Issue #20, opened 2017-03-06
Selecting a vulkan shader in GAPIC causes a crash


AWoloszyn:
As above, GAPIS crashes when this happens.

AWoloszyn:
Fixed by #31

PR #31, opened 2017-03-06
Cleaning up stuff to make the build sane

ianthehat:
This is all stuff to make the build more sane, and needed before a switch to bazel is viable.
It's all worthwhile even if we don't though.


Qining:
Confirmed this fixed the issue #20

ben-clayton:
Windows appears to work. I assume you've tested Linux. How about mac?

gapil/api.go
```diff
@@ -44,10 +44,15 @@ type ResolveResult struct {
 	Errs parse.ErrorList
 }
 
+type Loader interface {
```
ben-clayton: Docs please

AWoloszyn:
I don't think we should include the .vscode directory. 
This will clobber any custom settings/tasks that any individual user has in vscode.
(e.g. I have tasks that run ./do build etc, and this will clobber them)

ianthehat:
Yes, tested on mac
Yes, tested clean

.vscode : I will take this out for now while we discuss, but it seems to me this is the project configuration for vs code, and should be checked in.
it would be nice to have a library of tasks available to everyone, you can always use user settings to override the key bindings to the tasks you personally use a lot, and most of the other things you might put in here you can override with your user settings anyway
I do wish there was a way in vs code to do per project user settings, not found that yet.

gapil/api.go
ianthehat: done

AWoloszyn:
The main issue is that the workspace settings override the user settings. 

The big one for me is debug settings. The executable/arguments/etc that I use to start up a program make no sense to check in, and it would mean I will always have dirty files in my git repository. (making the flow of git add -u && git commit broken).

Adding the vscode directory, but keeping it in .gitignore may be a good compromise, since then there is a default that users can modify if they choose.

PR #34, opened 2017-03-06
robot: poll adb

mrkane27:


mrkane27:
Simplified it a bit.

Should be able to add new devices, but removing them doesn't really do much, new traces will fail.
Haven't yet figured out how to actually unregister the worker on device remove.

test/robot/trace/client.go
```diff
@@ -18,56 +18,109 @@ import (
...
-		return err
+type runner struct {
+	*client
+	device bind.Device
+}
+
+func (c *client) setRunner(serial string, r *runner) {
+	c.l.Lock()
+	defer c.l.Unlock()
+
```
ben-clayton: This is exciting. Shouldn't you be setting something here?

test/robot/trace/client.go
```diff
@@ -18,56 +18,109 @@ import (
 	"fmt"
 	"time"
```
ben-clayton: nit: kill the annoying blank line

test/robot/web/client/data.go
```diff
@@ -181,15 +181,20 @@ func newTask(entry map[string]interface{}, kind Item) *task {
...
-		t.status = grid.InProgress
-		t.result = grid.Unknown
-	case 2:
-		t.status = grid.Current
-		t.result = grid.Succeeded
-	case 3:
-		t.status = grid.Current
+	if st, ok := entry["status"].(float64); ok {
+		switch int(st) {
+		case 1:
```
ben-clayton: Don't we have an enum for these magic values somewhere?

mrkane27: we do but it's somewhere in proto code, and I think it would pull in a bunch of deps into Gopher js code.

test/robot/trace/client.go
mrkane27: removed

test/robot/web/client/data.go
ben-clayton: Fair enough

PR #37, opened 2017-03-06
Vulkan:Use memory tracker for coherent mem on Linux

Qining:
Test: tested with Vulkan sample applications.

### DIFF
--- a/gapii/cc/spy_base.cpp
+++ b/gapii/cc/spy_base.cpp
@@ -20,6 +20,10 @@
 #include "core/cc/log.h"
 #include "core/cc/thread.h"
 
+#if (TARGET_OS == GAPID_OS_LINUX) || (TARGET_OS == GAPID_OS_ANDROID)
+#include <core/memory_tracker/cc/memory_tracker.h>
+#endif // TARGET_OS
+
 #define __STDC_FORMAT_MACROS
 #include <inttypes.h>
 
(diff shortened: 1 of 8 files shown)


gapii/cc/spy_base.cpp
```diff
@@ -20,6 +20,10 @@
 #include "core/cc/log.h"
 #include "core/cc/thread.h"
 
+#if (TARGET_OS == GAPID_OS_LINUX) || (TARGET_OS == GAPID_OS_ANDROID)
```
ben-clayton: Have you considered:

```cpp
#if (TARGET_OS == GAPID_OS_LINUX) || (TARGET_OS == GAPID_OS_ANDROID)
#define USE_MEMORY_TRACKER 1
#else
#define USE_MEMORY_TRACKER 0
#endif
```

And using `USE_MEMORY_TRACKER` where you do this?

gapii/cc/spy_base.h
```diff
@@ -33,6 +33,10 @@
 #include "core/cc/coder/memory.h"
 #include "core/cc/coder/atom.h"
 
+#if (TARGET_OS == GAPID_OS_LINUX) || (TARGET_OS == GAPID_OS_ANDROID)
+#include <core/memory_tracker/cc/memory_tracker.h>
```
ben-clayton: If you were to add the `USE_MEMORY_TRACKER` define, I'd recommend it was defined in `memory_tracker.h`

ben-clayton: Actually scratch that. That's a stupid idea.

PR #41, opened 2017-03-07
Lower the shader version requirement for Gapic

Qining:
Lower the shader requirement because the shaders do not use any features
from vesion 150.


Qining:
Just tried to build on my laptop. However the laptop is running Ubuntu 14.04 and the Intel Graphics driver only supports: 1.10, 1.20, 1.30, 1.00 ES and 3.00 ES. So gapic will fail building.

Then I tried to change the version requirement to 130, and it works. So I guess 130 should be good enough here.

PR #42, opened 2017-03-07
GLSL version dependent on OpenGL version.

pmuetschard:


gapic/src/main/com/google/gapid/glviewer/ShaderSource.java
```diff
@@ -37,16 +42,18 @@ public static Shader loadShader(String name) {
   }
 
   public static Shader loadShader(URL resource) {
+    String version = (GL11.glGetInteger(GL30.GL_MAJOR_VERSION) == 3 &&
```
ben-clayton: The test on GL_MAJOR_VERSION here seems iffy. What are the possible values? Maybe add a versionAtLeast function?

pmuetschard: Done.

Qining:
Confirmed this works on my laptop (Ubuntu 1404) now. LGTM

ben-clayton:
Thanks

PR #52, opened 2017-03-08
Added QueryPool Suport to MEC.

AWoloszyn:
This adds Query-pool to Vulkan Mid-Execution capture.

It also updates the documents to show this.


gapis/gfxapi/vulkan/vulkan.api
```diff
@@ -3236,19 +3236,40 @@ cmd VkResult vkResetEvent(
   return ?
 }
 
+@override
+@custom
+@no_replay
+cmd void RecreateQueryPool(
```
ben-clayton: Nit: It might make sense from a cleanliness POV to move non-Vulkan APIs to their own API file.

AWoloszyn: I will probably do that as a separate CL. Move them all as its own operation.

PR #56, opened 2017-03-09
README for memory tracker

Qining:


core/memory_tracker/README.md
```diff
@@ -0,0 +1,64 @@
+# Track Coherent Memory
+
+## Problem and Solution
+Coherent memories are implicitly synchronized which means there is no *flush*
+or *invalid* command to explicitly *write* or *read* to/from those memories.
+Henceforth we cannot rely on intercepting the *flush* and *invalid* commands to
+track the modifications for coherent memory.
+
+The solution is to utilize segmentation fault signal handler to tell to which
```
AWoloszyn: This is only true on  Posix-based systems. We have to use other methods on other operating systems. I think it would make sense to describe this slightly differently, or at least a bit more explicitly.

Even something like:
The following explanation is for Posix-based systems. We intend to implement a similar system for Windows that uses SetUnhandledExceptionFilter.

core/memory_tracker/README.md
```diff
@@ -0,0 +1,64 @@
...
+memory page has been written and then read the data of that page to track the
+data in coherent memory.
+
+## Goal
+The goal is to implement a memory tracker that:
+* Tracks the memory pages that have been written to.
+* Transparent to the application under debugging.
+ * Assumes zero modification on the application under debugging
+ * For real segmentation faults, the system handler or, if any, the handler
```
AWoloszyn: For any memory that is not tracked, the previously installed handler should be called.

core/memory_tracker/README.md
```diff
@@ -0,0 +1,64 @@
...
+|OS        |OpenGL              |Vulkan              |
+|----------|--------------------|--------------------|
+|Windows   |:white_medium_square:|:white_medium_square:|
+|Linux     |:white_medium_square:|:white_check_mark:  |
+|Android   |:white_medium_square:|:white_check_mark:  |
+
+
+## Overall Workflow
+* Start Tracing.
+* Register the memory tracker's segfault handler.
```
AWoloszyn: We only register the segfault handler the first time someone calls "MapMemory" on a coherent buffer

core/memory_tracker/README.md
Qining: Fixed in b95e3717

core/memory_tracker/README.md
Qining: Fixed in b95e3717

core/memory_tracker/README.md
Qining: Fixed in b95e3717

PR #57, opened 2017-03-09
Use __func__ instead of __FILE__ for log messages

ianthehat:


ben-clayton:
If this is to fix the bazel build issue, can you put this CL into that branch instead?

dsrbecky:
Did this commit help by any chance: https://github.com/dsrbecky/gapid/commit/c266651604933e96e0d631ba2482c6c4ed541c6b

ianthehat:
It's not a bazel issue, I can reproduce it with a normal compiler command line, something is different about how cmake is running the compiler and I am not sure what, but I am fairly sure the code as it stood was not standards compliant.
I also thing it's a bad idea even if you can fix it, in a world where you are having problems with build times, using the very slow constexpr interpreter to do compile time string processing using recursive functions to produce information that you don't really care about anyway seems like a mistake...

ianthehat:
Tested with dsrbecky@c266651, still does not compile

dsrbecky:
Closing for now - it is not needed on the master branch.
Once we switch to bazel, we can revisit the issue, or I can try to fix the compile error.

PR #61, opened 2017-03-09
Adding hidden imports

ianthehat:
This are imports that generated files will use. These are not needed for
correctness, but when the generated source files are not in the source
tree, it makes a few of the tools behave a little better.

Also adding doc.go to packages that did not have one and moving the
package comments into them.


ben-clayton:
> it makes a few of the tools behave a little better.

I don't follow. Please can you explain what this is trying to solve?
I'd also like it if the imports had a comment next to them explaining why they exist.

ianthehat:
This is all the imports for the package that are *only* imported by the generated files.
It's not needed for bazel, but there are a couple of cases where tools behave better when the package dependancies are understood, and in the new bazel world the generated sources are not visible to those tools.

ianthehat:
Now with added comments

PR #62, opened 2017-03-09
Vulkan:Start Memory tracker only after tracing starts

Qining:


Qining:
Close as it is already merged with PR #63

PR #63, opened 2017-03-09
Vulkan:Fallback to read the whole coherent when tracing is not on Linux or Android

Qining:
Based on #56


gapis/gfxapi/vulkan/templates/vk_spy_helpers.cpp.tmpl
```diff
@@ -392,22 +396,21 @@ void VulkanSpy::trackMappedCoherentMemory(CallObserver*, uint64_t start, size_va
...
-#if (TARGET_OS == GAPID_OS_LINUX) || (TARGET_OS == GAPID_OS_ANDROID)
-    const size_val page_size = mMemoryTracker.page_size();
     auto &memory_object = this->DeviceMemories[memory];
-    // Get the valid mapped range
     const auto mapped_size = memory_object->mMappedSize;
     const auto mapped_location = (uint64_t)(memory_object->mMappedLocation);
-    const auto mapped_end = (uint64_t)(mapped_location + mapped_size);
-    const auto mapped_offset = (uint64_t)(memory_object->mMappedOffset);
     void *offset_addr = (void *)(offset_in_mapped + mapped_location);
+#if (TARGET_OS == GAPID_OS_LINUX) || (TARGET_OS == GAPID_OS_ANDROID)
```
AWoloszyn: I think it would make sense to move this check out.
In memory_tracker.h use

```
#if (TARGET_OS == GAPID_OS_LINUX || TARGET_OS == GAPID_OP_ANDROID)
#define COHERENT_TRACKING_ENABLED
... rest of what is in memory_tracker.h
#endif
```

Then use `COHERENT_TRACKING_ENABLED` elsewhere.

Qining: Fixed in 79b4a54

PR #78, opened 2017-03-13
Fix OpenGL viewport for HDPI.

pmuetschard:


pmuetschard:
Trying to kick the status checks...

PR #81, opened 2017-03-13
Restore the state after editing the capture.

### LINKED ISSUE #21: Client: Changing a field collapses all views
When changing a parameter, we should transparently change captures while keeping all the view states the same.
We don't want to have to re-find the context, re-expand the tree view, re-select the textures etc.

pmuetschard:
Closes #21.

### DIFF
--- a/gapic/src/main/com/google/gapid/views/ThumbnailScrubber.java
+++ b/gapic/src/main/com/google/gapid/views/ThumbnailScrubber.java
@@ -355,37 +355,56 @@ public class ThumbnailScrubber extends Composite
       Rectangle clip = gc.getClipping();
       Point size = (imageSize == null) ? new Point(THUMB_SIZE, THUMB_SIZE) : imageSize;
       int first = (int)((xOffset.longValueExact() + clip.x) / (size.x + 2 * MARGIN));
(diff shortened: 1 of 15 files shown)


gapic/src/main/com/google/gapid/views/ThumbnailScrubber.java
```diff
@@ -343,10 +373,12 @@ public void paint(BigInteger xOffset, BigInteger yOffset, GC gc) {
               imageSize =
                   new Point(Math.max(MIN_SIZE, bounds.width), Math.max(MIN_SIZE, bounds.height));
               updateSize.run();
+              selectAndScroll(selectedIndex);
```
ben-clayton: Are we changing state inside paint?

pmuetschard: We update the state / prepare the state of the next frame in the animation. I don't think it would make a difference if it was done outside of paint, as it, too, would have to run on the UI thread.

ben-clayton: It seems wrong for a function called paint() to be changing any sort of state.

pmuetschard: Is this what you had in mind?

PR #82, opened 2017-03-14
Show an error message in 'do' in case of bad checkout.

### LINKED ISSUE #55: Try harder to prevent people git cloning the project.
We've had a couple of people ask why the build fails.

Turns out they `git cloned` the project instead of using `go get`. `go get` forces the developer to checkout into `../github.com/google/gapid/` which is expected by `do`.

Let's:
* Make it more obvious in [BUILDING.md](https://github.com/google/gapid/blob/master/BUILDING.md)
* Try detecting a bad checkout and giving a sensible error message from `do`

pmuetschard:
Closes #55.


ben-clayton:
Awesome, thank you

PR #83, opened 2017-03-14
Mention environment variables and Homebrew alternatives for building

antiagainst:


BUILDING.md
```diff
@@ -65,22 +72,29 @@ The build output will be in the directory you specified with `do config`.
 
 ### Install Android SDK
 * Unzip the [Android SDK](https://dl.google.com/android/repository/tools_r25.2.3-macosx.zip) to a directory of your choosing.
+  * Instead, if you have Homebrew, you can install via `brew install android-sdk` and the default installation location would be `$HOMEBREW_PREFIX/opt/android-sdk`.
```
ben-clayton: `$HOMEBREW_PREFIX` does not appear to be a automatically defined environment variable. You *could* suggest:

```echo `brew config | grep HOMEBREW_PREFIX | awk '{print $2}'`/opt/android-sdk```

or something, but I'd probably just suggest something like:

  * Instead, if you have Homebrew, you can install via `brew install android-sdk` and the default installation location would be *HOMEBREW_PREFIX*`/opt/android-sdk`, where *HOMEBREW_PREFIX* can be found using `brew config`.

It's also worth mentioning that `do` does not currently support resolving env vars.
Alternatively to this, we could just make do search for the brew location to fill it in as a suggestion...

File: BUILDING.md [hunk 2]
```diff
@@ -65,22 +72,29 @@ The build output will be in the directory you specified with `do config`.
...
     * SDK Platform
   * Android SDK Build-tools 21.1.2
 
 ### Install Android NDK
 * Unzip the [Android NDK](https://dl.google.com/android/repository/android-ndk-r13b-darwin-x86_64.zip) to a directory of your choosing.
+  * Instead, if you have Homebrew, you can install via `brew install android-ndk` and the default installation location would be `$HOMEBREW_PREFIX/opt/android-ndk`.
 
-### Install [CMake](https://cmake.org/files/v3.7/cmake-3.7.1-Darwin-x86_64.dmg).
+### Install [CMake](https://cmake.org/files/v3.7/cmake-3.7.1-Darwin-x86_64.dmg)
+  * Instead, if you have Homebrew, you can install via `brew install cmake`.
```
ben-clayton: The version of the dependencies use also worries me a little. I've seen different behavior between moderately recent versions of CMake, hence the explicit links.

BUILDING.md
antiagainst: I was thinking about including the `brew config | grep ... | cut ...` command part earlier. But I feel it's too verbose. Just mentioning `brew config` sounds better.

BUILDING.md
antiagainst: Oh, really? That's weird. Do you know the reason? CMake 3.7.2 (from HomeBrew) seems working fine for me.

When I see that a project needs Softwrae x.y, I would expect Software > x.y should work and install a more recent one if exists, unless the project specifically says only x.y. And I assume it's not only me thinking like this. Should we be more specific in the instruction about the version then?

ben-clayton:
Thanks

PR #86, opened 2017-03-14
Be more verbose on replay if Vulkan returns errors.

AWoloszyn:


AWoloszyn:
Thanks for the feedback.
For the most part, Vulkan commands should not be emitting errors, but there are some cases where they may be expected (Allocating subsequently smaller chunks of memory until success for example). What might be a bit nicer, is passing the "expected return" down, for any methods that have non-remapped returns, then we can emit an error on replay if they don't match.

Does this make any sense?

ben-clayton:
> What might be a bit nicer, is passing the "expected return" down, for any methods that have non-remapped returns, then we can emit an error on replay if they don't match.

Instead of doing this, we expect to catch these issues when building the report. In gles we perform a `glGetError()` call after each command. That way we're separating out diagnostics from the performance critical replays. See [gles/find_issues.go](https://github.com/google/gapid/blob/b1733c6/gapis/gfxapi/gles/find_issues.go#L109-L148) on how we do this.

AWoloszyn:
Thanks for the information, I will abandon this CL, and take that approach.

ben-clayton:
Feel free to add this logic to the debug build however.

Issue #22, opened 2017-03-06
./do build will build all android ABI builds at the same time


AWoloszyn:
Since this is a ninja build it will run all sub-builds concurrently, which all run ninja. These each spawn their own concurrent children. This can cause an enormous amount of memory to be used, since 3x the number of compiles/links happen than is normal. In the worst case this will cause paging and eventual build failure on machines with insufficient ram/swap space.

ben-clayton:
I think this is unlikely to be fixed before we switch to bazel.

AWoloszyn:
That is fair. In order to prevent lower memory devices from failing however, would it make sense to maybe just expose a "concurrent-processes" flag so it can just run
ninja -j1 

Will be slower, but it should at least work on any machine.

dsrbecky:
Do you have ccache on your machine?  That should generally decrease the number of compiles that are needed in total.

AWoloszyn:
This was for a clean build on Windows.
On a machine with 4 cores, 8G ram + 8GB pagefile ,it was all filled short order when trying to compile/link 18 things at the same time. (The problem is mostly linking libinterceptor.so, which can take upwards of 2-3GB by itself).

ccache unfortunately won't help me with my clean-build case, or the linking issues.

dsrbecky:
ccache could still help with the clean-build case - it does the caching based on source code hashes.
I am mentioning just in case it happens to work as workaround, I know it is not a proper solution.

antiagainst:
I experienced the same problem on my MacBook Pro (15-inch, 2016) with 16GB memory. For fresh build, system loads is almost 200, memory is filled up, and GBs of swap files are used. Brings the machine to a kinda unusable state.

dsrbecky:
This is an old issue.  I have done a lot of improvements since then to the build system.  We split the generated files to fewer files (so fewer concurrent compiles).  I also tweaked the generated pb into "small code size" mode.  I moved a lot of code from headers to cpp.  I claim it is fixed, re-open if you see it again.

PR #87, opened 2017-03-14
Add -numjobs option to `do build`

antiagainst:
A workaround for #22.


antiagainst:
Some statistics on MacBook Pro (15-inch, 2016):

./do build -numjobs 1  7918.06s user 511.53s system 352% cpu 39:52.01 total

Issue #89, opened 2017-03-14
Client review: Add properties to texture view


pmuetschard:
Should include Width, Height, Format at the very least.

PR #90, opened 2017-03-14
Add Properties to Texture view.

pmuetschard:
For #89.

### DIFF
--- a/gapis/gfxapi/resource.go
+++ b/gapis/gfxapi/resource.go
@@ -25,8 +25,15 @@ type ResourceMap map[Resource]id.ID
 
 // Resource represents an asset in a capture.
 type Resource interface {
-	// ResourceName returns the UI name for the resource.
-	ResourceName() string
+	// ResourceHandle returns the UI identity for the resource.
+	// For GL this is the GLuint object name, for Vulkan the pointer.
+	ResourceHandle() string
+
+	// ResourceLabel returns the UI name for the resource.
+	ResourceLabel() string
+
+	// Order returns an integer used to sort the resources for presentation.
+	Order() uint64
 
 	// ResourceType returns the type of this resource.
 	ResourceType() ResourceType
--- a/gapis/service/service.proto
+++ b/gapis/service/service.proto
@@ -442,10 +442,14 @@ message ResourcesByType {
 message Resource {
   // The resource instance unique identifier.
   path.ID id = 1;
-  // The resource name.
-  string name = 2;
+  // The resource identifier used for display.
+  string handle = 2;
+  // The resource label.
+  string label = 3;
+  // An int used to sort the resources for presentation.
+  uint64 order = 4;
   // The list of command indices where the resource was used.
-  repeated uint64 accesses = 3;
+  repeated uint64 accesses = 5;
 }
 
 // Context represents a single rendering context in the capture.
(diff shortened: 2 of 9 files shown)


gapis/gfxapi/resource.go
```diff
@@ -25,8 +25,11 @@ type ResourceMap map[Resource]id.ID
 
 // Resource represents an asset in a capture.
 type Resource interface {
-	// ResourceName returns the UI name for the resource.
-	ResourceName() string
+	// ResourceName returns the UI identity for the resource.
```
ben-clayton: I'm really not sure about these changes. ResourceName was supposed to be the UI label. If you want a unique key for the resource, that's the path to the resource.
`ResourceName()` returns values that are not unique between resource types, nor resources of the same type. I don't see what purpose it serves.

pmuetschard: Here's my thinking:
1. Although the path is unique, there is no value in displaying it to the user.
2. We should display the GL texture/shader/program names. There are many places where we only show the GL names, so I don't think we can get away with not allowing the user to use the GL name to lookup a texture in the texture view, for example. Even if we change the linking, a user will inevitably find themselves in a situation where they want to find "texture 5". I know that the names are not unique, though for the "interesting" resources they typically are unique within a resource type.
3. Separating the label from the name and letting the UI do the formatting is preferable, 1st normal form and all.
4. I changed the name to be numeric to allow sorting. I don't think we'll run into an API that won't use a numeric identifier for resources. I'd rather address this if/when it happens, than making things difficult for us now. Here's a [Screenshot](https://screenshot.googleplex.com/pRHvitY3YXy) of the new UI. Clicking the ID column will correctly do a numeric sort).

Please let me know which of the above you disagree with.

ben-clayton: > 1. Although the path is unique, there is no value in displaying it to the user.

I didn't mean to suggest that we'd present the path. I thought you were using `ResourceName` as an internal key, which you certainly shouldn't do as it's not unique.

> 2. We should display the GL texture/shader/program names. 

We always were

> 3. Separating the label from the name and letting the UI do the formatting is preferable, 1st normal form and all.

Okay, I can buy that.

> I changed the name to be numeric to allow sorting. 

I see that it's useful to sort by numerical value instead of string, but Vulkan is dealing with pointers. Decimal representations of pointers just seem wrong.

I think I'm mostly just offended with the concept of `ResourceName` returning an integer. "Name" may be the GL terminology, but it isn't for Vulkan.

How about:

```go
// ResourceHandle returns the API handle for the resource.
// For GL this is the GLuint object name, for Vulkan the pointer.
ResourceHandle() string

// ResourceLabel returns the annotated name for the resource.
ResourceLabel() string

// Order returns an integer used to sort the resources for presentation.
Order() int
```

We then use `Order` to sort the resources in [`gapis/resolve/resources.go`](https://github.com/google/gapid/blob/b1733c6bd337e1121973106b3cdb16c550e815b6/gapis/resolve/resources.go)

pmuetschard: Done.

gapis/service/service.proto
```diff
@@ -442,8 +442,12 @@ message ResourcesByType {
 message Resource {
   // The resource instance unique identifier.
   path.ID id = 1;
-  // The resource name.
-  string name = 2;
+  // The resource identifier used for display.
+  string handle = 2;
+  // The resource label.
+  string label = 4;
```
ben-clayton: Superdooperubernit: While we're still pre-stable, I'm happy to break protos for the sake of a simple ordering.

pmuetschard: Done.

PR #106, opened 2017-03-16
Vulkan:MEC support for vkCmdCopyImageToBuffer

Qining:


gapii/cc/vulkan_inlines.inl
```diff
@@ -280,6 +280,23 @@ void inline CommandListRecreator<std::shared_ptr<RecreateCopyBufferToImageData>>
...
 
+template <>
+void inline CommandListRecreator<
+    std::shared_ptr<RecreateCopyImageToBufferData>>::
+operator()(VkCommandBuffer commandBuf, CallObserver* observer, VulkanSpy* spy,
+           const std::shared_ptr<RecreateCopyImageToBufferData>& t) {
+  if (!spy->Images.count(t->mSrcImage) || !spy->Buffers.count(t->mDstBuffer)) {
+    return;
+  }
+  std::vector<VkBufferImageCopy> regions;
```
AWoloszyn: We should probably consider reserving the space.
regions.reseve(t->mRegions.size())

Qining: Agreed, I remember we have multiple places have the same problem. Will file a CL later to resolve all of them.

PR #115, opened 2017-03-17
The dead code elimination transformer needs to be first.

pmuetschard:
Fixes the wireframe and wireframe overlay replay modes.


dsrbecky:
Ups, my bad.  Thanks.

I am happy with it as-is.  However, wouldn't the better solution be to add the wireframe transform after the dead-code one?  As a bonus point it would fix the TODO.

pmuetschard:
Fixed the TODO. The dead code elimination still needs to be pre-pended, because of the wireframe overlay.

PR #116, opened 2017-03-19
Postback test broken

ianthehat:
You are not allowed to pass a nil abi to Connect, also the result was ignored anyway and connect called again 2 lines later...


ben-clayton:
Good catch.

PR #119, opened 2017-03-20
Vulkan:Use std::vector::reserve() in recreating commands data

Qining:


gapii/cc/vulkan_inlines.inl
```diff
@@ -52,17 +52,20 @@ void inline CommandListRecreator<std::shared_ptr<RecreateCmdPipelineBarrierData>
     VkCommandBuffer commandBuf, CallObserver* observer, VulkanSpy* spy,
     const std::shared_ptr<RecreateCmdPipelineBarrierData>& t) {
     std::vector<VkMemoryBarrier> memory_barriers;
+    memory_barriers.reserve(t->mMemoryBarriers.size());
```
AWoloszyn: Good change! 

Another option, would have been to initialize the vector to the right size, and then set into it.
```
std::vector<VkMemoryBarrier> memory_barriers(t->mMemoryBarriers.size());
for (size_t ....) {
  memory_barriers[i] = t->mMemoryBarriers[i];
}
```

PR #120, opened 2017-03-20
Vulkan:Find color/depth attachments from LastUsedFramebuffer's RenderPass

### LINKED ISSUE #100: Depth input attachment image is considered as depth render target
In state.go, line 53, only the image usage bit is checked. Actually we should check the attachment type in the `LastUsedFramebuffer`.

Qining:
Fix #100

### DIFF
--- a/gapis/gfxapi/vulkan/state.go
+++ b/gapis/gfxapi/vulkan/state.go
@@ -21,57 +21,48 @@ import (
 )
 
 func (st *State) getFramebufferAttachmentInfo(attachment gfxapi.FramebufferAttachment) (w, h uint32, f VkFormat, attachmentIndex uint32, err error) {
-	if st.LastUsedFramebuffer != nil {
-		var index int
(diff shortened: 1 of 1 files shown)


gapis/gfxapi/vulkan/state.go
```diff
@@ -21,57 +21,48 @@ import (
...
-		case gfxapi.FramebufferAttachment_Color3:
-			index = 3
-		case gfxapi.FramebufferAttachment_Depth:
-			break
-		case gfxapi.FramebufferAttachment_Stencil:
-			fallthrough
-		default:
-			return 0, 0, VkFormat_VK_FORMAT_UNDEFINED, 0, fmt.Errorf("Framebuffer attachment %v currently unsupported", attachment)
-		}
+	return_error := func(format_str string, e ...interface{}) (w, h uint32, f VkFormat, attachmentIndex uint32, err error) {
```
ben-clayton: Nit: Don't use underscores in go variable names. govet will moan.

Qining: Fixed in af7520c, changed to `returnError`.

PR #121, opened 2017-03-20
Simplify logging interface & add RPC method

ben-clayton:
Instead of `log.Context`, now use `context.Context`.

To log something use:
```go
log.I(ctx, "cat says %v", "meow")
log.W(ctx, "dog says %v", "woof")
log.E(ctx, "fox says %v", "???")
```

Values are added to the context with:
```go
ctx := log.V{"key": value}.Bind(ctx)
```

GAPII and GAPIR log messages are parsed into the structured form so they can be filtered.

New streaming RPC `GetLogStream()` monitors all log messages from GAPIS and all subprocesses interleaved. Can be used by GAPIC to display all logs across all processes.


cmd/gapit/trace.go
```diff
@@ -99,13 +99,13 @@ func (verb *traceVerb) startLocalApp(ctx context.Context) error {
...
-	env.Set("LD_PRELOAD", libgapii.System())
-
-	boundPort, err := process.StartWithEnv(ctx, verb.Local.App.System(), env)
+	boundPort, err := process.Start(ctx, verb.Local.App.System(), process.StartOptions{
+		Env: shell.CloneEnv().
+			AddPathStart("VK_INSTANCE_LAYERS", "VkGraphicsSpy").
+			AddPathStart("VK_DEVICE_LAYERS", "VkGraphicsSpy").
+			AddPathStart("VK_LAYER_PATH", libVkLayerGraphicsSpy.Parent().System()).
+			Set("LD_PRELOAD", libgapii.System()),
+	})
```
pmuetschard: This change seems unrelated.

ben-clayton: It's related. There's new logic to handle stdout and stderr, which adds more options to the start function. Given the number of parameters we're passing in, I opted for a struct.

pmuetschard: Yes, I noticed that later. Sorry, I didn't realize it sends out comments right away. Is there no draft comment mode like Critique/Gerrit?

ben-clayton: Yes. Next to the single line comment button there should be a "Start Review" green button.

ben-clayton:
Now also contains GAPIC changes to display the merged logs across GAPIC, GAPIS and GAPIR.

gapic/src/main/com/google/gapid/views/LogView.java
```diff
@@ -53,12 +82,107 @@ public LogView(Composite parent) {
...
+        for (Log.SourceLocation location : message.getCallstackList()) {
+          TreeItem line = newItem(callstack, severity);
+          line.setText(Column.TEXT.index, String.format("%s:%d", location.getFile(), location.getLine()));
+        }
+      }
+    }
     dirty.set(false);
   }
 
+  private String formatTime(Timestamp time) {
```
pmuetschard: Can be static

gapic/src/main/com/google/gapid/views/LogView.java
```diff
@@ -53,12 +82,107 @@ public LogView(Composite parent) {
...
+      case Info:
+        return theme.infoBackground();
+      case Warning:
+        return theme.warningBackground();
+      case Error:
+        return theme.errorBackground();
+      case Fatal:
+        return theme.fatalBackground();
+    }
+    return theme.infoBackground();
```
pmuetschard: Please move this into a default case to make the behavior more explicit.

gapic/src/main/com/google/gapid/views/LogView.java
```diff
@@ -53,12 +82,107 @@ public LogView(Composite parent) {
...
+      case Info:
+        return theme.infoForeground();
+      case Warning:
+        return theme.warningForeground();
+      case Error:
+        return theme.errorForeground();
+      case Fatal:
+        return theme.fatalForeground();
+    }
+    return theme.infoForeground();
```
pmuetschard: Please move this into a default case to make the behavior more explicit.

gapic/src/main/com/google/gapid/views/LogView.java
```diff
@@ -53,12 +82,107 @@ public LogView(Composite parent) {
...
+    TreeItem item = new TreeItem(parent, 0);
+    item.setBackground(severityBackgroundColor(severity));
+    item.setForeground(severityForegroundColor(severity));
+    return item;
+  }
+
+  private Color severityBackgroundColor(Log.Severity severity) {
+    switch (severity) {
+      case Verbose:
+        return theme.verboseBackground();
```
pmuetschard: nit: since the case statements are so simple, I'd prefer them to be on one line:
```
      case Verbose: return theme.verboseBackground();
      case Debug: return theme.debugBackground();
      ...
```

gapic/src/main/com/google/gapid/util/Logging.java
```diff
@@ -15,54 +15,64 @@
  */
 package com.google.gapid.util;
 
+import com.google.gapid.proto.log.Log;
+import com.google.gapid.rpclib.schema.Method;
```
pmuetschard: not actually needed

ben-clayton:
I'm following up with a final CL that avoids the `tree.removeAll()`

gapic/src/main/com/google/gapid/util/Logging.java
```diff
@@ -15,54 +15,64 @@
...
-    private LogLevel(Level level, String gapisLevel, String gapirLevel) {
+    LogLevel(Level level, Log.Severity severity, String gapisLevel, String gapirLevel) {
       this.level = level;
+      this.severity = severity;
       this.gapisLevel = gapisLevel;
       this.gapirLevel = gapirLevel;
     }
+
+    static LogLevel fromLevel(Level level) {
+      for (LogLevel logLevel : LogLevel.values()) {
```
pmuetschard: logLevel shadows Logging.logLevel

gapic/src/main/com/google/gapid/util/Logging.java
```diff
@@ -15,54 +15,64 @@
...
+        if (logLevel.level == level) {
```
pmuetschard: This should be range based as it was before.
java.util.logging.Level is not an enum, so you also should not use '==', but .equals.

gapic/src/main/com/google/gapid/util/Logging.java
```diff
@@ -99,12 +109,20 @@ public static File getLogDir() {
...
+    return buffer.getMessages();
   }
 
   public static void setListener(Runnable listener) {
-    memoryHandler.listener = (listener == null) ? () -> { /* empty */ } : listener;
+    buffer.listener = (listener == null) ? () -> { /* empty */ } : listener;
+  }
+
+  /**
+   * Adds a {@link Log.Message} to the buffer, bypassing the Java {@link LogManager}.
```
pmuetschard: Unfortunately this needs to be ```{@link com.google.gapid.proto.log.Log.Message}```

gapic/src/main/com/google/gapid/util/Logging.java
```diff
@@ -165,44 +183,25 @@ private static void protosToString(LogRecord rec) {
...
-            return 'I';
-          } else {
-            return 'D';
-          }
-      }
+  private static final String JAVA_PREFIX = "java.";
+  private static final String GAPID_PREFIX = "com.google.gapid.";
+  private static final String GOOG_PREFIX = "com.google.";
+
+  private static String shorten(String className) {
```
pmuetschard: make this protected, as you are using it from multiple inner classes

gapic/src/main/com/google/gapid/util/Logging.java
```diff
@@ -225,18 +224,28 @@ public void close() {
   /**
    * {@link Handler} that keeps a number of messages in memory for later retrieval.
    */
-  private static class MemoryHandler extends Handler {
-    private static final int MAX_SIZE = 10 * 1024 * 1024;
-
+  private static class Buffer extends Handler {
     public Runnable listener;
-    private final StringBuilder buffer = new StringBuilder();
+    private final LinkedList<Log.Message> buffer = new LinkedList<>();
```
pmuetschard: buffer is shadowing Logging.buffer. I suggest renaming Logging.buffer to Logging.messageBuffer

gapic/src/main/com/google/gapid/server/GapitPkgInfoProcess.java
```diff
@@ -72,10 +79,35 @@ protected Exception prepare(ProcessBuilder pb) {
...
+    }
+  }
+
   @Override
   protected OutputHandler<PkgInfo.PackageList> createStdoutHandler() {
     return new BinaryHandler<PkgInfo.PackageList>(in -> {
-      byte[] data = ByteStreams.toByteArray(in);
+      SearchingInputStream is = new SearchingInputStream(in);
+      if (!is.search(PACKAGE_DATA_MARKER.getBytes())) {
+        throw new RuntimeException("The gapit command didn't produce the data marker.");
```
pmuetschard: Resource leak: 'is' is not closed at this location.

I'd suggest making the search method static and avoid the wrapping in another input stream. The SearchingInputStream class has no additional state, so it's not needed.

gapic/src/main/com/google/gapid/views/LogView.java
ben-clayton: Done.

gapic/src/main/com/google/gapid/views/LogView.java
ben-clayton: Done.

gapic/src/main/com/google/gapid/views/LogView.java
ben-clayton: Done.

gapic/src/main/com/google/gapid/util/Logging.java
ben-clayton: Done.

gapic/src/main/com/google/gapid/util/Logging.java
ben-clayton: Done.

gapic/src/main/com/google/gapid/util/Logging.java
ben-clayton: Done.

gapic/src/main/com/google/gapid/util/Logging.java
ben-clayton: I'm sure I've asked this before, but why is this an issue?
Done anyway.

gapic/src/main/com/google/gapid/server/GapitPkgInfoProcess.java
ben-clayton: > The SearchingInputStream class has no additional state, so it's not needed.

Well, it has the buffered reader state. I've wrapped it with a try-with-resources,

ben-clayton:
> FYI: i've checked in a .settings folder which enables a lot of warnings in Eclipse.

Thanks, but I use IntelliJ. :)

gapic/src/main/com/google/gapid/util/Logging.java
ben-clayton: Changed in later CL.

gapic/src/main/com/google/gapid/util/Logging.java
pmuetschard: The JVM has no concept (almost) of inner classes. To the JVM, an inner class and its container outer class are simply two different classes. You cannot call private methods of another class. The compiler adds synthetic methods to make it look like you can call private methods from inner/outer classes. These synthetic methods add unnecessary overhead and pollute stack traces. If you're going to allow access via these synthetic methods, why not give access directly?

gapic/src/main/com/google/gapid/util/Logging.java
ben-clayton: Is this Eclipse telling you this? IntelliJ seems fine with it as-is.

pmuetschard: Yes and no. Eclipse does give me the warning, but it comes straight from javadoc:
```src/main/com/google/gapid/util/Logging.java:145: warning - Tag @link: reference not found: Log.Message```

File: gapic/src/main/com/google/gapid/util/Logging.java [hunk 13]
```diff
@@ -165,44 +183,25 @@ private static void protosToString(LogRecord rec) {
...
+      return className.substring(GAPID_PREFIX.length());
+    } else if (className.startsWith(GOOG_PREFIX)) {
+      return "cg." + className.substring(GOOG_PREFIX.length());
     }
+    return className;
   }
 
   /**
-   * Similar to {@link java.util.logging.ConsoleHandler}, except that we log to standard output
+   * Similar to {@link java.util.logging.ConsoleHandler}, except that we logMessage to standard output
```
pmuetschard: Search replace?

gapic/src/main/com/google/gapid/util/Logging.java
```diff
@@ -77,6 +85,10 @@ static LogLevel fromLevel(Level level) {
     setLevel(Level.ALL);
   }};
 
+  private static final Listener NULL_LISTENER = (m) -> {};
```
pmuetschard: I have a warning on empty blocks, so please add a comment like ```/* do nothing */``` to explicitly state that it is indeed the intent.

gapic/src/main/com/google/gapid/util/Logging.java
```diff
@@ -222,30 +235,63 @@ public void close() {
...
 
   /**
-   * {@link Handler} that keeps a number of messages in memory for later retrieval.
+   * MessageIterator is an iterator for the {@link Buffer}.
+   * As the {@link Buffer} is a fixed size ring buffer, messages can be discarded if the buffer
+   * fills more quickly than it is consumed. This iterator has been implemented to read the next
+   * most recently available messages. Some messages may be skipped if the iterator lags behind
+   * new messages.
+   */
+  private static class MessageIterator implements Iterator<Log.Message> {
```
pmuetschard: This class breaks the Iterator contract. I would simply not implement Iterator.

Also, you are accessing private member variables of another class, which has the same problems as accessing private methods.

gapic/src/main/com/google/gapid/util/Logging.java
```diff
@@ -222,30 +235,63 @@ public void close() {
...
+      generation++;
     }
 
-    public synchronized void add(Log.Message message) {
-      buffer.addFirst(message);
-      while (buffer.size() > maxSize) {
-        buffer.removeLast();
-      }
-      listener.run();
+    public Iterator iterator() {
```
pmuetschard: missing <Log.Message> on return type.

gapic/src/main/com/google/gapid/util/Logging.java
```diff
@@ -109,12 +121,12 @@ public static File getLogDir() {
...
 
-  public static Log.Message[] getLogMessages() {
-    return buffer.getMessages();
+  public static Iterator<Log.Message> getMessageIterator() {
+    return buffer.iterator();
   }
 
-  public static void setListener(Runnable listener) {
-    buffer.listener = (listener == null) ? () -> { /* empty */ } : listener;
+  public static void setListener(Listener listener_) {
```
pmuetschard: Name shadowing is acceptable on setters, if you're shadowing the variable you're setting.

If you don't like to do that, please use new<Name> instead of <name>_.

gapic/src/main/com/google/gapid/util/Logging.java
```diff
@@ -123,6 +135,7 @@ public static void setListener(Runnable listener) {
    */
   public static void logMessage(Log.Message message) {
```
pmuetschard: This is now very much round-about: Buffer.publish calls Logging.logMessage, which calls Buffer.add and then fires the event.

Better to keep it they way it was:
logMessage and Buffer.publish call Buffer.add
Buffer.add fires the event

gapic/src/main/com/google/gapid/util/Logging.java
```diff
@@ -222,30 +235,63 @@ public void close() {
...
   /**
-   * {@link Handler} that keeps a number of messages in memory for later retrieval.
+   * MessageIterator is an iterator for the {@link Buffer}.
+   * As the {@link Buffer} is a fixed size ring buffer, messages can be discarded if the buffer
+   * fills more quickly than it is consumed. This iterator has been implemented to read the next
+   * most recently available messages. Some messages may be skipped if the iterator lags behind
+   * new messages.
+   */
+  private static class MessageIterator implements Iterator<Log.Message> {
+    private final Buffer buffer;
```
pmuetschard: Shadows Logging.buffer.

gapic/src/main/com/google/gapid/util/Logging.java
```diff
@@ -245,30 +254,34 @@ public void publish(LogRecord record) {
...
+
+      builder.addValues(Log.Value.newBuilder().setName("thread").setValue(Pods.pod(thread)));
+      builder.addValues(Log.Value.newBuilder().setName("class").setValue(Pods.pod(klass)));
+      builder.addValues(Log.Value.newBuilder().setName("method").setValue(Pods.pod(method)));
+
+      Throwable thrown = record.getThrown();
+      if (thrown != null) {
+        for (StackTraceElement el : thrown.getStackTrace()) {
+          builder.addCallstack(Log.SourceLocation.newBuilder()
+              .setFile(el.getFileName())
```
pmuetschard: el.getFileName can return null (protos don't like nulls).

gapic/src/main/com/google/gapid/server/GapisConnection.java
```diff
@@ -117,9 +141,44 @@ public boolean isConnected() {
...
+        this.channel = channel;
+        this.monitor = monitor;
+      }
+
+      @Override
+      public void run() {
+        Service.GetLogStreamRequest request = Service.GetLogStreamRequest.newBuilder().build();
+        Iterator<Log.Message> it = GapidGrpc.newBlockingStub(channel).getLogStream(request);
+        while (it.hasNext()) {
+          monitor.onLogMessage(it.next());
```
pmuetschard: This now causes an exception on clean shutdown. Before closing the connection to gapis, the logger thread should be stopped.

I think you'd be better off using a StreamObserver and the async case rather than the blocking case with an extra thread.

gapic/src/main/com/google/gapid/util/Logging.java
ben-clayton: Given that the class is static, does it?

pmuetschard: Yes, because so is Logging.buffer.

gapic/src/main/com/google/gapid/server/GapisConnection.java
ben-clayton: AFAICT, streamed RPCs aren't exposed in the FutureStub (if that's what you're referring to).

gapic/src/main/com/google/gapid/util/Logging.java
ben-clayton: Fair point.

gapic/src/main/com/google/gapid/util/Logging.java
```diff
@@ -15,54 +15,64 @@
...
-import java.util.logging.FileHandler;
-import java.util.logging.Formatter;
-import java.util.logging.Handler;
-import java.util.logging.Level;
-import java.util.logging.LogManager;
-import java.util.logging.LogRecord;
-import java.util.logging.Logger;
-import java.util.logging.StreamHandler;
+import java.util.LinkedList;
+import java.util.logging.*;
```
ben-clayton: Done. Another IntelliJ default that I continually battle.

gapic/src/main/com/google/gapid/util/Logging.java
ben-clayton: I'd say it's pretty obvious on this line. Done anyway.

gapic/src/main/com/google/gapid/util/Logging.java
```diff
@@ -15,54 +15,64 @@
...
 
 /**
  * Logging setup and utilities.
  */
 public class Logging {
   /**
-   * Possible log level flag values.
+   * Possible logMessage level flag values.
```
ben-clayton: Refactor gone crazy. Fixed.

gapic/src/main/com/google/gapid/util/Logging.java
ben-clayton: Done.

gapic/src/main/com/google/gapid/util/Logging.java
ben-clayton: Done.

gapic/src/main/com/google/gapid/util/Logging.java
```diff
@@ -15,54 +15,64 @@
...
+      }
+      return INFO;
+    }
   }
 
   public static final Flag<LogLevel> logLevel = Flags.value(
       "logLevel", LogLevel.INFO, "Logging level [OFF, ERROR, WARNING, INFO, DEBUG, ALL].");
   public static final Flag<String> logDir = Flags.value(
-      "logDir", System.getProperty("java.io.tmpdir"), "Directory for log files.");
+      "logDir", System.getProperty("java.io.tmpdir"), "Directory for logMessage files.");
```
ben-clayton: Done.

gapic/src/main/com/google/gapid/util/Logging.java
```diff
@@ -83,7 +93,7 @@ public static void init() {
 
     if (!logDir.get().isEmpty()) {
       try {
-        FileHandler fileHandler = new FileHandler(logDir.get() + File.separator + "gapic.log");
+        FileHandler fileHandler = new FileHandler(logDir.get() + File.separator + "gapic.logMessage");
```
ben-clayton: Done.

gapic/src/main/com/google/gapid/util/Logging.java
ben-clayton: Done.

gapic/src/main/com/google/gapid/util/Logging.java
ben-clayton: Done.

gapic/src/main/com/google/gapid/server/GapisConnection.java
ben-clayton: Now unbinds the `LogMonitor` on `close()`. Added warning for unexpected `StatusRuntimeException`s that are thrown by the thread.

gapic/src/main/com/google/gapid/util/Logging.java
ben-clayton: I've separated `Buffer` and `Handler`. The inner class was doing too much - hence the extraction of the event logic. Hopefully this is now cleaner and would be nice and simple to make the classes public if needed in the future.

core/log/severity.go
```diff
@@ -14,59 +14,66 @@
...
+	Error Severity = 4
+	// Fatal indicates a fatal error and the process should be terminated.
+	Fatal Severity = 5
+)
 
-// Alert is is shorthand for ctx.At(AlertLevel)
-func (ctx logContext) Alert() Logger {
-	return ctx.At(severity.Alert)
+func (s Severity) String() string {
+	switch s {
```
dsrbecky: Handle Verbose...

core/log/severity.go
```diff
@@ -14,59 +14,66 @@
...
+	}
+	return "?"
 }
 
-// Critical is shorthand for ctx.At(CriticalLevel)
-func (ctx logContext) Critical() Logger {
-	return ctx.At(severity.Critical)
+// Short returns the severity string with a single character.
+func (s Severity) Short() string {
+	switch s {
```
dsrbecky: Handle Verbose...

core/log/severity.go
ben-clayton: Done.

core/log/severity.go
ben-clayton: Done.

gapic/src/main/com/google/gapid/views/LogView.java
ben-clayton: Done.

dsrbecky:
+1

gapic/src/main/com/google/gapid/util/Logging.java
ben-clayton: Done.

gapic/src/main/com/google/gapid/server/GapisConnection.java
```diff
@@ -18,20 +18,36 @@
...
 import io.grpc.Metadata;
+import io.grpc.Status;
+import io.grpc.StatusRuntimeException;
 import io.grpc.okhttp.OkHttpChannelProvider;
 
 /**
  * A connection to a running Graphics API Server (GAPIS).
  */
 public abstract class GapisConnection implements Closeable {
+  private static final Logger LOG = Logger.getLogger(Client.class.getName());
```
pmuetschard: Client.class -> GapisConnection.class

gapic/src/main/com/google/gapid/server/GapitPkgInfoProcess.java
```diff
@@ -23,7 +23,8 @@
 import com.google.gapid.proto.pkginfo.PkgInfo;
 import com.google.gapid.proto.pkginfo.PkgInfo.PackageList;
 
-import java.io.File;
+import java.io.*;
```
pmuetschard: Another .* import.

I think you can tell intellij not to do this by setting the minimum number of classes needed before it collapses imports to some very high number.

gapic/src/main/com/google/gapid/util/Logging.java
```diff
@@ -41,40 +42,80 @@
...
       "logLevel", LogLevel.INFO, "Logging level [OFF, ERROR, WARNING, INFO, DEBUG, ALL].");
   public static final Flag<String> logDir = Flags.value(
       "logDir", System.getProperty("java.io.tmpdir"), "Directory for log files.");
 
-  private static final MemoryHandler memoryHandler = new MemoryHandler() {{
+  private static final int BUFFER_SIZE = 1000;
+
+  private static final RingBuffer buffer = new RingBuffer(BUFFER_SIZE);
+
+  private static final Handler handler = new Handler() {{
```
pmuetschard: You no longer need to hang on to this reference, so just create the handler in the init() method and add it to the root logger.

gapic/src/main/com/google/gapid/util/Logging.java
```diff
@@ -223,52 +255,98 @@ public void close() {
...
+    public MessageIterator iterator() {
+      return new MessageIterator(this);
+    }
+  }
+
+  /**
+   * {@link Handler} that handles {@link LogRecord}s converting them to
+   * {@link com.google.gapid.proto.log.Log.Message}s and then broadcasting them to listeners.
+   */
+  protected static class Handler extends java.util.logging.Handler {
```
pmuetschard: Maybe name this class ProtoHandler, or ProtoConverter?

gapic/src/main/com/google/gapid/views/LogView.java
```diff
@@ -15,48 +15,175 @@
...
+    for (int i = 0; i < MAX_NEW_ITEMS_PER_UPDATE; i++) {
+      Log.Message message = messageIterator.next();
+      if (message == null) {
+        // All messages consumed.
+        dirty.set(false);
+        return;
+      }
+      Log.Severity severity = message.getSeverity();
+      TreeItem item = newItem(tree, severity);
+      String[] lines = message.getText().split("(\n)|(\r\n)");
```
pmuetschard: split("\r?\n")

gapic/src/main/com/google/gapid/util/Logging.java
ben-clayton: They both sound like the Proto is the source when it's the target. Opted for `LogToMessageHandler`.

PR #123, opened 2017-03-20
Update the atom selection, if needed, when the context changes.

pmuetschard:
The current atom selection is already maintained when changing contexts.
However, if a context is selected that does not contain the current
selection, the selection needs to be updated for the new context.

### DIFF
--- a/gapic/src/main/com/google/gapid/models/ApiContext.java
+++ b/gapic/src/main/com/google/gapid/models/ApiContext.java
@@ -15,6 +15,8 @@
  */
 package com.google.gapid.models;
 
+import static com.google.gapid.util.Ranges.commands;
+import static com.google.gapid.util.Ranges.first;
 import static com.google.gapid.util.Ranges.last;
 import static java.util.Arrays.stream;
 
@@ -185,6 +187,11 @@ public class ApiContext extends CaptureDependentModel<ApiContext.FilteringContex
         public boolean contains(long index) {
           return true;
         }
+
+        @Override
+        public CommandRange findClosest(CommandRange range) {
+          return range;
+        }
       };
     }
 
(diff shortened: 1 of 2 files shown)


gapic/src/main/com/google/gapid/models/AtomStream.java
```diff
@@ -84,6 +86,17 @@ protected void fireLoadEvent() {
     }
   }
 
+  @Override
+  public void onContextSelected(FilteringContext ctx) {
+    if (selection != null && !ctx.contains(selection)) {
+      if (ctx.contains(last(selection))) {
+        selectAtoms(last(selection), 1, false);
+      } else {
+        selectAtoms(ctx.findClosest(selection), false);
```
ben-clayton: I assume there's guards against this selection triggering another selection?

gapic/src/main/com/google/gapid/models/ApiContext.java
```diff
@@ -204,6 +211,20 @@ public boolean contains(long index) {
...
     }
 
+    public CommandRange findClosest(CommandRange range) {
+      int index = Ranges.contains(context.getRangesList(), last(range));
+      if (index >= 0) {
+        return commands(last(range), 1);
+      } else if (index == -1) {
+        // The given range ends before any of our commands.
+        return (context.getRangesCount() == 0) ? null : commands(first(context.getRanges(0)), 1);
+      } else {
```
ben-clayton: Is the else block even reachable?

gapic/src/main/com/google/gapid/models/AtomStream.java
pmuetschard: It only changes the selection if it's currently inconsistent and if so switches it to a consistent state.

gapic/src/main/com/google/gapid/models/ApiContext.java
pmuetschard: Yes. It's the most commonly executed case. 
Index is positive if the context happens to contain the given range.
Index is -1 if the given range is before the context is created.
Index is < -1 if the given range is after the context is created, but the context does not contain the range.

ben-clayton: Ah, I quickly parsed the [`contains()`](https://github.com/pmuetschard/gapid/blob/98b422d06a9163cd65a06258012e92f6d6456fd6/gapic/src/main/com/google/gapid/util/Ranges.java#L94-L98) function and my brain freaked out with the double ternary in the lambda that's used as a parameter. Candidate for a function comment perhaps? :smile:

gapic/src/main/com/google/gapid/models/AtomStream.java
ben-clayton: Okay, I guess that's okay assuming the new selection couldn't change context again.

PR #131, opened 2017-03-23
Show thumbnails in the texture selection table.

pmuetschard:


pmuetschard:
ping?

gapis/service/path/path.proto
```diff
@@ -249,6 +249,6 @@ message Thumbnail {
     image.Format desired_format = 3;
 
     oneof object {
-        Command command = 4;
+        ResourceData resource = 4;
```
ben-clayton: Is this supposed to be here?

pmuetschard: Yes. It's even in its own commit. There is no atom that implements the Thumbnailer interface.

ben-clayton: Sorry, that's what I get for reviewing while being in a meeting (with you).

Issue #25, opened 2017-03-06
Vulkan assumes all replays should be done on Android devices.


AWoloszyn:
We should check the contents of the capture, and pick the device for replay appropriately.

ben-clayton:
Is this still an issue?

AWoloszyn:
This is now fixed.

PR #138, opened 2017-03-24
Filter the replay target based on memory layout.

AWoloszyn:
This allows us to replay Vulkan applications on Desktop if they were
captured on desktop without any user intervention.

### DIFF
--- a/gapis/replay/interfaces.go
+++ b/gapis/replay/interfaces.go
@@ -32,8 +32,9 @@ type Support interface {
 	CanReplayOnLocalAndroidDevice(context.Context) bool
 
 	// CanReplayOn returns true if the API can be replayed on the specified
-	// device.
-	CanReplayOn(context.Context, *device.Instance) bool
+	// device. The MemoryLayout is the memory layout that was in the capture
+	// file.
+	CanReplayOn(context.Context, *device.Instance, *device.MemoryLayout) bool
 }
 
 // QueryIssues is the interface implemented by types that can verify the replay
--- a/gapis/server/server.go
+++ b/gapis/server/server.go
@@ -29,6 +29,7 @@ import (
 	"github.com/google/gapid/core/app/benchmark"
 	"github.com/google/gapid/core/event/task"
 	"github.com/google/gapid/core/log"
+	"github.com/google/gapid/core/os/device"
 	"github.com/google/gapid/core/os/device/bind"
 	"github.com/google/gapid/framework/binary"
 	"github.com/google/gapid/framework/binary/registry"
@@ -125,9 +126,25 @@ func (s *server) LoadCapture(ctx context.Context, path string) (*path.Capture, e
 	return capture.Import(ctx, name, in)
 }
 
+// Returns all devices, sorted by Android first, and then Host
+func getSortedDevices(ctx context.Context) []bind.Device {
+	all := bind.GetRegistry(ctx).Devices()
+	androidDevices := make([]bind.Device, 0, len(all))
+	nonAndroidDevices := make([]bind.Device, 0, len(all))
+	for _, dev := range all {
+		instance := dev.Instance()
+		if instance.GetConfiguration().GetOS().GetKind() == device.Android {
+			androidDevices = append(androidDevices, dev)
+		} else {
+			nonAndroidDevices = append(nonAndroidDevices, dev)
+		}
+	}
+	return append(androidDevices, nonAndroidDevices...)
+}
+
 func (s *server) GetDevices(ctx context.Context) ([]*path.Device, error) {
 	s.deviceScanDone.Wait(ctx)
-	devices := bind.GetRegistry(ctx).Devices()
+	devices := getSortedDevices(ctx)
 	paths := make([]*path.Device, len(devices))
 	for i, d := range devices {
 		paths[i] = path.NewDevice(d.Instance().Id.ID())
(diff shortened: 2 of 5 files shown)


gapis/replay/interfaces.go
```diff
@@ -33,7 +33,7 @@ type Support interface {
 
 	// CanReplayOn returns true if the API can be replayed on the specified
 	// device.
```
ben-clayton: Please document what the `MemoryLayout` represents. I assume it is of the device that took the capture. If so, would it make more sense to pass that?

AWoloszyn:
This should be a good first-pass at fixing #25 
This will also be made better when once we have a more robust file header.

gapis/server/server.go
```diff
@@ -152,23 +168,47 @@ func (s *server) GetDevicesForReplay(ctx context.Context, p *path.Capture) ([]*p
 
 	all := bind.GetRegistry(ctx).Devices()
 	filtered := make([]bind.Device, 0, len(all))
-nextDevice:
+nextAndroidDevice:
```
ben-clayton: What's the benefit for splitting this into too passes?

ben-clayton: s/too/two

AWoloszyn: The main benefit is that it gives us a nice deterministic order to what devices will be returned if we ever run into the case where we have more than one MemoryLayout that match (Host/Android).

I.e. we will always have Android devices first, then Host. So you know that Android will be the first in the list in this case.

gapis/replay/interfaces.go
AWoloszyn: The MemoryLayout represents the MemoryLayout that was contained in the Architecture Atom at the start of the capture.

We cannot use the device object in the capture, because it is null at this point, for all traces. IIRC it gets filled in later when we go and actually start the replay, but at this point all we have is the MemoryLayout. This should all change once we get a header that contains this information.

AWoloszyn: Added some comments to clarify.

Qining:
lgtm

gapis/server/server.go
ben-clayton: Humm okay. The order of devices [is not documented](https://github.com/google/gapid/blob/master/gapis/service/service.go#L70) to have any particular order. We should probably change that. If we do, I'd prefer it if the `GetDevices` RPC kept the same order. 
Finally, can we try to kill the copy-paste code? Adding a for loop or extracting to a function would suffice.

AWoloszyn: Done

gapis/server/server.go
```diff
@@ -125,9 +126,22 @@ func (s *server) LoadCapture(ctx context.Context, path string) (*path.Capture, e
...
+	all := bind.GetRegistry(ctx).Devices()
+	filtered := make([]bind.Device, 0, len(all))
+	for _, device := range all {
+		if instance.GetConfiguration().GetOS().GetKind() == dev.Android {
+			filtered = append(device, filtered...)
+		} else {
+			filtered = append(filtered, device)
+		}
+	}
+}
```
ben-clayton: Does this compile? You don't appear to be returning anything...

gapis/server/server.go
```diff
@@ -29,6 +29,7 @@ import (
 	"github.com/google/gapid/core/app/benchmark"
 	"github.com/google/gapid/core/event/task"
 	"github.com/google/gapid/core/log"
+	dev "github.com/google/gapid/core/os/device"
```
ben-clayton: Why do you need the named import?

gapis/server/server.go
```diff
@@ -150,22 +180,22 @@ func (s *server) GetDevicesForReplay(ctx context.Context, p *path.Capture) ([]*p
...
 				"api":    fmt.Sprintf("%T", api),
 				"device": instance,
 			}.Bind(ctx)
-			if api.CanReplayOn(ctx, instance) {
+			if api.CanReplayOn(ctx, instance, layout) {
 				log.I(ctx, "Compatible")
 			} else {
 				log.I(ctx, "Incompatible")
-				continue nextDevice
+				continue nextAndroidDevice
```
ben-clayton:
```go
continue nextDevice
```

AWoloszyn: Fixed.

gapis/server/server.go
AWoloszyn: Yeah, sorry fixed that.

gapis/server/server.go
AWoloszyn: Without the named import, this conflicts with os/device/bind.
Unless I am missing something about how go works.
Basically it cannot find device.Android, but with the named import it can find dev.Android

ben-clayton: > Without the named import, this conflicts with os/device/bind.

Nope, I don't get how there's any collision here.
The only collision you could have is if there's another `device` declared at the usage scope.

ben-clayton: Please can you show me the compilation error when you strip the named import?

AWoloszyn: ../../gapis/server/server.go:32: imported and not used: "github.com/google/gapid/core/os/device"
../../gapis/server/server.go:135: device.Android undefined (type bind.Device has no field or method Android)

gapis/server/server.go
```diff
@@ -125,9 +126,24 @@ func (s *server) LoadCapture(ctx context.Context, path string) (*path.Capture, e
 	return capture.Import(ctx, name, in)
 }
 
+// Returns all devices, sorted by Android first, and then Host
+func getSortedDevices(ctx context.Context) []bind.Device {
+	all := bind.GetRegistry(ctx).Devices()
+	filtered := make([]bind.Device, 0, len(all))
```
ben-clayton: Uber nit: its sorted, not filtered.

gapis/server/server.go
```diff
@@ -125,9 +126,24 @@ func (s *server) LoadCapture(ctx context.Context, path string) (*path.Capture, e
...
+	for _, dev := range all {
+		instance := dev.Instance()
+		if instance.GetConfiguration().GetOS().GetKind() == device.Android {
+			filtered = append([]bind.Device{dev}, filtered...)
```
ben-clayton: Uber nit: While the number of devices is small, this is acceptable, but this does not scale particularly well. Would be nicer to have two lists you append to, then join at the end.

AWoloszyn: Will do

PR #151, opened 2017-03-27
Thread Mid-Execution capture through Gapit.

AWoloszyn:


gapii/cc/connection_header.h
```diff
@@ -48,6 +50,8 @@ class ConnectionHeader {
     uint32_t mVersion;                      // 2 or 3
     uint32_t mObserveFrameFrequency;        // non-zero == enabled. Version: 2+
     uint32_t mObserveDrawFrequency;         // non-zero == enabled. Version: 2+
+    uint32_t mStartFrame;                   // non-zero == Frame to start at. version 4+
+    uint32_t mNumFrames;                    // non-zero == Number of frames to capture. version 4+
```
Qining: Just wonder if we should put the `mStartFrame` and `mNumFrames` to the end of the header struct?

gapii/cc/spy.cpp
```diff
@@ -186,6 +189,11 @@ Spy::Spy()
                 (header.mFlags & ConnectionHeader::FLAG_DISABLE_PRECOMPILED_SHADERS) != 0;
         mRecordGLErrorState =
                 (header.mFlags & ConnectionHeader::FLAG_RECORD_ERROR_STATE) != 0;
+        // This will be over-written if we also set the header flags
+        mSuspendCaptureFrames = header.mStartFrame;
+        mCaptureFrames = header.mNumFrames;
+        mSuspendCaptureFrames.store((header.mFlags & ConnectionHeader::FLAG_DEFER_START)?
+            -1: mSuspendCaptureFrames.load());
```
Qining: Maybe make the `-1` a constant defined somewhere with an easy to understand name?

gapii/cc/connection_header.h
AWoloszyn: I wanted to leave the "numbers" with the other "numbers" in the header. In this case mFlags is a bit-field.

gapii/cc/spy.cpp
AWoloszyn: Done

PR #155, opened 2017-03-28
Added argument passing to gapit for local applications.

AWoloszyn:
This allows us to specify the local arguments when running a desktop
application.


cmd/gapit/trace.go
```diff
@@ -132,9 +132,14 @@ func (verb *traceVerb) startLocalApp(ctx context.Context) (func(), error) {
 			AddPathStart("VK_LAYER_PATH", libVkLayerGraphicsSpy.Parent().System())
 	}
 
+	r := regexp.MustCompile("'.+'|\".+\"|\\S+")
```
ben-clayton: Nit: Instead of `"'.+'|\".+\"|\\S+"` you can use backticks so you don't have to escape: `` `'.+'|".+"|\S+` ``

PR #159, opened 2017-03-28
Conversion for float-point values with arbitrary bit size

dsrbecky:
We will need 10-bit and 11-bit variants in particular.


core/math/f64/f64_test.go
```diff
@@ -0,0 +1,106 @@
...
+		{"1 10001 0000000000", -4.0},
+		{"1 10001 0100000000", -5.0},
+		{"1 11111 0000000000", math.Inf(-1)},
+	}
+
+	for _, c := range checks {
+		v, err := strconv.ParseUint(strings.Replace(c.f16, " ", "", -1), 2, 64)
+		if err != nil {
+			t.Errorf("%v\n", err)
+		}
```
ben-clayton: Replace:
```go
	if err != nil {
		t.Errorf("%v\n", err)
	}
```
With:
```go
	assert.To(t).For("err").ThatError(err).Succeeded()
```

core/math/f64/f64_test.go
```diff
@@ -0,0 +1,106 @@
...
+		expected, got := c.f64, f64.FromBits(v, 5, 10)
+		if math.Float64bits(expected) != math.Float64bits(got) {
```
ben-clayton:
```go
	assert.To(t).For("%v", c.f64).That(expected).Equals(got)
```

core/math/f64/f64_test.go
```diff
@@ -0,0 +1,106 @@
...
+	for _, c := range checks {
+		v, err := strconv.ParseUint(strings.Replace(c.f11, " ", "", -1), 2, 64)
+		if err != nil {
+			t.Errorf("%v\n", err)
+		}
+		expected, got := c.f64, f64.FromBits(v, 5, 6)
+		if math.Float64bits(expected) != math.Float64bits(got) {
+			t.Errorf("Testing %v: expected %v, got %v\n", c.f11, expected, got)
+		}
+	}
```
ben-clayton: See above.

dsrbecky:
Done.  I have also merged the two tests together to remove code duplication.  More importantly, I think it is useful to see the binary representations of different sizes next to each other.

Issue #165, opened 2017-03-29
Build is broken on OSX


ben-clayton:
[`core/cc/posix/thread.cpp:25:9`](https://github.com/google/gapid/blob/master/core/cc/posix/thread.cpp#L25): error: use of undeclared identifier 'malloc'

ben-clayton:
I have a fix - just waiting for the build to finish before creating the PR.

PR #167, opened 2017-03-29
Fix OSX build.

ben-clayton:
Fixes: #165

### DIFF
--- a/core/cc/posix/thread.cpp
+++ b/core/cc/posix/thread.cpp
@@ -17,6 +17,7 @@
 #include "core/cc/thread.h"
 
 #include <cstdint>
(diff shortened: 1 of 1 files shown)


core/cc/posix/thread.cpp
```diff
@@ -17,6 +17,7 @@
 #include "core/cc/thread.h"
 
 #include <cstdint>
+#include <cstdlib>
```
AWoloszyn: I cannot leave the comment on the line in question but at line 35 we should also have a free,.

ben-clayton: https://github.com/google/gapid/pull/168

PR #169, opened 2017-03-29
Pretty print debug information into translated GLSL files.

dsrbecky:


dsrbecky:
Where? ;-D

gapis/shadertools/cc/spv_manager.cpp
```diff
@@ -819,14 +836,19 @@ void SpvManager::appendDebugInstruction(std::vector<instruction_t>* debugs, Inst
 
       i.name = new char[str_name.length() + 1];
```
ben-clayton: What's freeing this allocation?

dsrbecky: deleteGlslCodeWithDebug in libmanager

ben-clayton: Fair enough.

PR #180, opened 2017-03-30
Use git submodules for Khronos repositories

dsrbecky:
This uses the current upstream version of the repositories.

I have pushed our local modifications of SPIRV-Tool upstream,
as this was necessary to even just build our shadertools.

There is still more work before we can fully switch to glslang.


.gitmodules
```diff
@@ -55,3 +55,21 @@
 [submodule "third_party/src/github.com/inconshreveable/mousetrap"]
 	path = third_party/src/github.com/inconshreveable/mousetrap
 	url = https://github.com/inconshreveable/mousetrap.git
+[submodule "third_party/glslang"]
```
AWoloszyn: Just a note, shaderc is only guaranteed to work against github.com/google/glslang. 
google/glslang is kept reasonably up-to-date with khronos/glslang, it has not diverged, but we are a few revisions back to make sure that it always works with ToT shaderc.

dsrbecky:
It can be generated... but I was not too keen to do the cmake dance if the code might be fairly short lived in the first place.

I am not quite sure why the code is not checked-in with the rest of SPIRV-Headers, I will have to ask the SPIRV guys. @AWoloszyn any thoughts?

AWoloszyn:
The generated headers are not a part of spirv-headers. They use the json files to create what they need, so if there were to be checked in they would be in SPIRV-Tools. However it is generally considered bad practice to check in generated files, so we generate them at build-time. Having generated files in your build-tree is a good way to accidentally get things out of date.

This is usually not a problem, because including SPIRV-Tools into a project using the normal CMake mechanisms (add_subdirectory(SPIRV-Tools)) will handle everything for you so long as you have python installed. I think the unique what that GAPID uses CMake may make this very difficult though.

dsrbecky:
I follow the general pros and cons regarding generated files.  My question is mostly about the SPIRV-Headers repository.  This repository consists mostly of checked in generated files in the first place - and that is one of its main reasons for existence.  So we are in situations where:
 1) Spriv opcode enum mapping is generated and checked in as part of SPIRV-Headers.
 2) Spriv opcode string mapping is generated and not check in as part of SPIRV-Tools.
Both generated using different systems and different means.  That seems like in inconsistency waiting to happen...

So given that SPIRV-Headers already generates C++ headers and checks those in, it would seem more consistent to me to generate the other data at the same time.  After all, doing things like converting opcode to string seems like a reasonable thing people might expect from SPIRV-Headers.

Correct me if I am wrong, but as far as I can tell the generated code in SPIRV-Tools does not actually contain any code that is specific to SPIRV-Tools - it just seems to fill the deficiencies of SPIRV-Headers, and other people would probably benefit from just adding that code to SPIRV-Headers instead.

dsrbecky:
...I would not necessarily suggest adding it the current form exactly as-is.

For example, both the Python and Lua headers already can map opcode to string and back due to the reflective nature of those languages.  It is up to discussion how best represent the reflection data in C.  Personally, I would just add big { { "OpCode", 42 }, ... } table as that is obvious and is sufficient for bi-directional mapping as well as enumeration.

I also realize that SPIRV-Tools is interested in the opcode grammar which is currently ignored by the SPIRV-Headers generator.

AWoloszyn:
So the basic answer is:
SPIRV-Headers is the official source of truth for SPIR-V. It is intended to be reasonably minimal, and complete.

SPIRV-Tools is a consumer of SPIRV-Headers, but has it's own interpretation of what data is needed by the headers. SPIRV-Tools changes and massages the data to contain exactly what it needs. If SPIRV-Tools needs more information in the headers, or changes to the organization of the headers, that change is limited to SPIRV-Tools. If this was pushed up into SPIRV-Headers, then the freedom to make changes that are beneficial to SPIRV-Tools are lost, since projects other than SPIRV-Tools can/do rely on the contents of SPIRV-Headers.

dsrbecky:
So the question is how much of this massaging is needed...  My general opinion would be that SPIRV-Headers should make most of the data from the .json file also available for the languages (in minimal and stable way).  I would go for simple C-table, which would mean that SPIRV-Tools might need to do a bit more work at start-up to access the data - say, iterate over the table, and build fast string to opcode lookup map.

My general feeling is that the extra data the SPIRV-Tools is extracting is very reasonable - it seems reasonably desirable to covert opcode to string for error messages, or to get number/type of operands to check correct usage, etc...   I am bluntly suggesting that SPIRV-Headers needs more information for most consumers :-)

AWoloszyn:
It may be possible, although any changes would have to be backwards compatible at this point, since a number of external projects already use spirv.hpp as is. It would be unfortunate to force every external project to update. 

As for what may be useful to external projects, the only data-points I have are SPIRV-Cross and SPIRV-LLVM both use spirv.hpp as-is, and SPIRV-Tools generates something more custom, and I am not sure if SPIRV-LLVM/SPIRV-Cross would actually benefit from a more complex spirv.hpp, I don't have a lot of insight into those projects. The purpose of the json file is so that projects can generate whatever format matches their needs, and avoid the need for run-time grossness.

dsrbecky:
Backwards compatible - I would probably include any additions in separate header in the first place.

I have browsed thought SPIRV-Cross a bit in the past - they index the individual words of the instruction.

Anyway, I will stop bike-shedding now :-)  It is a bit off-topic to the pull-request anyway.

pmuetschard:
This broke the shader tab for vulkan. It no longer shows the assembly for SPIRV shaders.

PR #195, opened 2017-03-31
Add usage hints and cancel OOV thumbnails for better replay performance.

pmuetschard:
Controls replay priorities and batching delays.
Far better handling for RPC cancellation.
Much cleaner GAPIS replay internals.
GAPIC cancels thumbnail requests as they scroll out of view.
GAPIC no longer batches requests.

### DIFF
--- a/gapic/src/main/com/google/gapid/util/Trees.java
+++ b/gapic/src/main/com/google/gapid/util/Trees.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright (C) 2017 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.gapid.util;
+
+import static java.util.Collections.emptySet;
+
+import com.google.common.collect.Sets;
+
+import org.eclipse.swt.graphics.Rectangle;
+import org.eclipse.swt.widgets.Tree;
+import org.eclipse.swt.widgets.TreeItem;
+
+import java.util.Set;
+
+/**
+ * Utilities for dealing with {@link Tree Trees} and {@link TreeItem TreeItems}.
+ */
+public class Trees {
+  private Trees() {
+  }
+
+  /**
+   * @return the {@link TreeItem TreeItems} that are currently visible in the tree.
+   */
+  public static Set<TreeItem> getVisibleItems(Tree tree) {
+    TreeItem top = tree.getTopItem();
+    if (top == null) {
+      // Work around bug where getTopItem() returns null when scrolling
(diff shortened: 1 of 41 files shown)


pmuetschard:
I've added two commits at the top to fix some issues I've found and address my nitpicky code review comments from #160.

gapic/src/main/com/google/gapid/util/Trees.java
```diff
@@ -15,112 +15,80 @@
...
   /**
-   * @return the parent {@link Tree} or {@link TreeItem} of the {@link TreeItem}.
+   * @return the next parent or {@code null} if the bottom has been reached.
    */
-  public static Object getParent(Object treeOrTreeItem) {
-    if (!(treeOrTreeItem instanceof TreeItem)) {
-      return null;
+  private static TreeItem getVisibleSiblings(TreeItem item, int treeBottom, Set<TreeItem> visible) {
+    TreeItem parent = item.getParentItem();
+    TreeItem[] siblings = (parent == null) ? item.getParent().getItems() : parent.getItems();
```
ben-clayton: Fold this function into the caller above to avoid repeating these lines for each item.

gapic/src/main/com/google/gapid/util/Trees.java
```diff
@@ -15,112 +15,80 @@
...
+        top = getVisibleSiblings(top, treeBottom, visible);
+      } while (top != null);
     }
-    return ((TreeItem)treeOrTreeItem).getItem(index);
+    return visible;
   }
 
   /**
-   * @return the index of the child {@link TreeItem} of the {@link Tree} or {@link TreeItem}.
+   * @return whether the bottom has been reached.
```
ben-clayton: Add a comment to say visible is filled with resulta

gapic/src/main/com/google/gapid/widgets/LoadableImage.java
```diff
@@ -105,13 +102,10 @@ protected void onUiThreadError(Void error) {
   }
 
   public LoadableImage unload() {
-    if (--loadCount != 0) {
+    if (--loadCount != 0 || state != State.LOADING) {
```
ben-clayton: The reason I didn't merge these if statements together is it's now less obvious that you're pre-decrementing. Given that swapping the order of these two conditionals can introduce bugs, I'd strongly recommend doing the decrement on a line above.

gapic/src/main/com/google/gapid/widgets/LoadableImage.java
```diff
@@ -105,13 +102,10 @@ protected void onUiThreadError(Void error) {
...
       return this;
     }
-    if (state == State.LOADING) {
-      future.cancel(true);
-    }
-    updateImage(null);
```
ben-clayton: Why not release the image? If we ever limit the size of the cache, you'd want to do this anyway, right?

gapic/src/main/com/google/gapid/widgets/LoadableImage.java
```diff
@@ -63,10 +63,7 @@ protected LoadableImage(Widget widget, Supplier<ListenableFuture<Object>> future
   }
 
   public LoadableImage load() {
-    if (loadCount++ != 0) {
-      return this;
-    }
-    if (state != State.NOT_STARTED) {
+    if (loadCount++ != 0 || state != State.NOT_STARTED) {
```
ben-clayton: Move increment to line above. See my comment below.

pmuetschard:
Squashed and rebased. Ready to submit, in my opinion.

PR #203, opened 2017-04-03
Added Dimension decoding/encoding to binaryxml.

AWoloszyn:


AWoloszyn:
dump_manifest throws up if you try to dump any app that has any dimension values.
eg: android:minWidth=100px

PR #206, opened 2017-04-04
Building with bazel

ianthehat:
Windows is still WIP, but linux and mac are solid.


ianthehat:
This diverged too much, abandoning

PR #210, opened 2017-04-05
Some optimizations for traces with 100s of shaders/textures.

pmuetschard:
- Don't starve async scheduled runnables, if repaint is slow
- Only update the shader dropdown if required to do so
- Only load texture thumbnails that are visible
- Batch updates to the texture table

### DIFF
--- a/gapis/resolve/resource_data.go
+++ b/gapis/resolve/resource_data.go
@@ -74,9 +74,10 @@ func buildResources(ctx context.Context, p *path.Command) (*ResolvedResources, e
 	for k, v := range resources {
 		res, err := v.ResourceData(ctx, state)
 		if err != nil {
-			return nil, fmt.Errorf("Could not get data for resource %v", k)
+			resourceData[k] = err;
+		} else {
+			resourceData[k] = res
 		}
-		resourceData[k] = res
 	}
 	return &ResolvedResources{idMap, resources, resourceData}, nil
 }
@@ -103,6 +104,9 @@ func (r *ResourceDataResolvable) Resolve(ctx context.Context) (interface{}, erro
 	}
 	id := r.Path.Id.ID()
 	if val, ok := res.resourceData[id]; ok {
+		if err, isErr := val.(error); isErr {
+			return nil, err
+		}
 		return val, nil
 	}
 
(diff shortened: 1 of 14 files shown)


gapic/src/main/com/google/gapid/widgets/VisibilityTrackingTreeViewer.java
```diff
@@ -70,6 +70,13 @@ private void updateVisibility() {
   }
 
   @Override
+  protected void inputChanged(Object input, Object oldInput) {
+    // Cause all items to receive a show event.
```
ben-clayton: all new visible items, right?

gapis/resolve/resource_data.go
```diff
@@ -74,9 +74,10 @@ func buildResources(ctx context.Context, p *path.Command) (*ResolvedResources, e
 	for k, v := range resources {
 		res, err := v.ResourceData(ctx, state)
 		if err != nil {
-			return nil, fmt.Errorf("Could not get data for resource %v", k)
+			resourceData[k] = err;
```
ben-clayton: [Okay, you are making me feel uneasy now.](https://github.com/google/gapid/pull/212#discussion_r109881143)
I swear I didn't see this before making those comments...

AWoloszyn:
This greatly helps for the enormous traces I have been looking at.

gapic/src/main/com/google/gapid/widgets/VisibilityTrackingTreeViewer.java
pmuetschard: Yes. Only the visible ones. Updated the comment.

gapis/resolve/resource_data.go
pmuetschard: Oh, and here I was worried that isErr is not :notes:idiomatic go:notes:

ben-clayton:
> Batch updates to the texture table

Where is this batching of which you speak?

pmuetschard:
> Where is this batching of which you speak?

https://github.com/google/gapid/pull/210/files#diff-a48d43384608915e3d14df3b796da55bR537

"intercepts" .refresh() calls, and bundles those that come in quick succession together as one.

ben-clayton:
Ack. Thanks.

PR #212, opened 2017-04-05
Gracefully handle unhandled texture types.

dsrbecky:
This ensures that one unhandled texture does not break all previews.

### DIFF
--- a/gapis/gfxapi/gfxapi.proto
+++ b/gapis/gfxapi/gfxapi.proto
@@ -50,6 +50,10 @@ enum ResourceType {
 	ShaderResource = 5;
 	// ProgramResource represents the Program resource type
 	ProgramResource = 6;
+	// UninitializedTextureResource represents texture resource type
+	// which has been generated but was never assigned dimensions.
+	// TODO: Clean up texture handling as per issue #237
+	UninitializedTextureResource = 7;
 }
 
 // FramebufferAttachment values indicate the type of frame buffer attachment.
--- a/gapis/gfxapi/gles/resources.go
+++ b/gapis/gfxapi/gles/resources.go
@@ -51,13 +51,16 @@ func (t *Texture) Order() uint64 {
 }
 
 // ResourceType returns the type of this resource.
-func (t *Texture) ResourceType() gfxapi.ResourceType {
+func (t *Texture) ResourceType(ctx context.Context) gfxapi.ResourceType {
 	switch t.Kind {
+	case GLenum_GL_NONE:
+		return gfxapi.ResourceType_UninitializedTextureResource
 	case GLenum_GL_TEXTURE_2D:
 		return gfxapi.ResourceType_Texture2DResource
 	case GLenum_GL_TEXTURE_CUBE_MAP:
 		return gfxapi.ResourceType_CubemapResource
 	default:
+		log.E(ctx, "Unknown texture resource type: %v", t.Kind)
 		return gfxapi.ResourceType_UnknownResource
 	}
 }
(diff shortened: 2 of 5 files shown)


gapis/gfxapi/gles/resources.go
```diff
@@ -108,7 +106,7 @@ func (t *Texture) ResourceData(ctx context.Context, s *gfxapi.State) (interface{
 		return &gfxapi.Cubemap{Levels: levels}, nil
 
 	default:
-		return nil, &service.ErrDataUnavailable{Reason: messages.ErrNoTextureData(t.ResourceHandle())}
+		return nil, fmt.Errorf("Unhandled texture type: %v", t.Kind)
```
ben-clayton: The old message handled the case that the name had been generated, but nothing had been bound yet. Please reintroduce the old error into a new `case 0:`

gapis/resolve/resource_data.go
```diff
@@ -74,7 +75,8 @@ func buildResources(ctx context.Context, p *path.Command) (*ResolvedResources, e
 	for k, v := range resources {
 		res, err := v.ResourceData(ctx, state)
 		if err != nil {
-			return nil, fmt.Errorf("Could not get data for resource %v", k)
+			log.E(ctx, "Could not get data for resource %v: %v", v.ResourceHandle(), err)
+			continue // Skip this resource, but sill return all the others.
```
ben-clayton: s/sill/still

With the above changed, we should actually pass the error down so it can be shown in the UI:

```go
	if err != nil {
		resourceData[k] = err
	} else {
		resourceData[k] = res
	}
```

and then at the end of `func (r *ResourceDataResolvable) Resolve(ctx context.Context)`:

```go
	id := r.Path.Id.ID()
	val, ok := res.resourceData[id]
	if !ok {
		return nil, fmt.Errorf("Cannot find resource with id: %v", id)
	}
	if err, isErr := val.(error); isErr {
		return nil, err
	} else {
 		return val, nil
	}
 }
```

dsrbecky:
I had to rewrite this after Pascal's recent CL - he (reasonably) does not request data for unknown resources now, so the old version still did not report anything.

gapis/gfxapi/gfxapi.proto
```diff
@@ -50,6 +50,9 @@ enum ResourceType {
 	ShaderResource = 5;
 	// ProgramResource represents the Program resource type
 	ProgramResource = 6;
+	// UninitializedTextureResource represents texture resource type
+	// which has been generated but does not have dimensions yet.
+	UninitializedTextureResource = 7;
```
ben-clayton: This isn't a new resource type. Resources exist for the whole trace. Only their data is tied to a single point in time.

dsrbecky: I am aware of that - however, we seem to tie the dimensions for the whole trace and the "the dimensions were not assigned" should be one of them.

ben-clayton: Dimensions could change. Perhaps not for GLES, but it seems feasible for other APIs.
It seems wrong to special case this. What's wrong with reporting (0,0) or (-1,-1) as the dimensions?

dsrbecky: I am also aware of that, but in that case the resource RPC is wrong in the first place...

ben-clayton: Add a TODO to address https://github.com/google/gapid/issues/237, and I'll approve.

dsrbecky: done

PR #218, opened 2017-04-05
BGR channel order support.

pmuetschard:


pmuetschard:
> Can you also add the INTEGER ones to sampleAsFloat function below?

Done.

> Did you find an use case for all of those in an app? 

The APK we've been looking at uses them...

PR #219, opened 2017-04-05
Minor Log View improvements

pmuetschard:


core/log/log_pb/log.proto
```diff
@@ -46,13 +46,24 @@ message Message {
 
     // The key-value pairs of extra data.
     repeated Value values = 8;
+
+    // The backtraces passed along with the log message (e.g. a Java exception).
+    repeated Backtrace backtrace = 9;
```
ben-clayton: I've been considering adding this for go chained errors too. We already call this "Cause", and it would be nice to keep this name for consistency's sake.

core/log/log_pb/log.proto
```diff
@@ -46,13 +46,24 @@ message Message {
...
+}
+
+message Backtrace {
```
ben-clayton: Cause

core/log/log_pb/log.proto
pmuetschard: Done.

core/log/log_pb/log.proto
pmuetschard: Done.

PR #220, opened 2017-04-05
Ensure that framebuffer references valid texture.

dsrbecky:


dsrbecky:
This is still valid check though...

PR #225, opened 2017-04-06
When creating symlinks, try to use relative paths.

AWoloszyn:
When we create symlinks, if we use relative paths, then if they are in a
directory that unmounted/remounted somewhere else, the symlinks will
continue to work. If relative paths are not possible, then fall back
to the original absolute paths.

### DIFF
--- a/core/os/file/file.go
+++ b/core/os/file/file.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"io"
 	"os"
+	"path/filepath"
 
 	"github.com/google/gapid/core/log"
 )
@@ -42,6 +43,9 @@ func RemoveAll(f Path) error {
 
 // Symlink creates a new symbolic-link at link to target.
 func Symlink(link, target Path) error {
+	if relPath, err := filepath.Rel(link.Parent().value, target.value); err == nil {
+		return os.Symlink(relPath, link.value)
+	}
 	return os.Symlink(target.value, link.value)
 }
 


core/os/file/file.go
```diff
@@ -42,7 +43,11 @@ func RemoveAll(f Path) error {
 
 // Symlink creates a new symbolic-link at link to target.
 func Symlink(link, target Path) error {
-	return os.Symlink(target.value, link.value)
+	if rel_path, err := filepath.Rel(link.value+"/..", target.value); err == nil {
```
ben-clayton: Golint: don't use underscores in variable names. 

Instead of "/..", does `Parent()` not work? Do either support files instead or directories?

core/os/file/file.go
```diff
@@ -42,7 +43,11 @@ func RemoveAll(f Path) error {
...
+		return os.Symlink(rel_path, link.value)
+	} else {
```
ben-clayton: Golint: don't use else if true block returns.

core/os/file/file.go
AWoloszyn: I will double-check but this should support both. (I will also test with Parent())

AWoloszyn: I have validated this, this works for both directories and files.

core/os/file/file.go
AWoloszyn: Done

core/os/file/file.go
```diff
@@ -42,6 +43,10 @@ func RemoveAll(f Path) error {
 
 // Symlink creates a new symbolic-link at link to target.
 func Symlink(link, target Path) error {
+
```
ben-clayton: nit: blank newline.

PR #240, opened 2017-04-10
GAPII: Generate protopack files

ben-clayton:
This PR moves us over to using the proto-pack file format instead of the old codergen serialization format.
GAPIS supports both, so this CL should have no impact. I will, however, be killing the codergen support over the coming days.

Note that I'm not using the `ScratchAllocator` or proto `Arena`s for the protobuf generation. This means we are doing one hell of a lot of temporary heap allocations. I plan to fix this, but not right now as I'm not convinced that this is a noticable performance hit anyway.

Next up is header section describing the device that was took the trace.


ben-clayton:
\*ping\*

gapii/cc/pack_encoder.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+namespace {
+
+const uint64_t specialSection = 0;
+const char magic[] = "protopack";
+
+// PackEncoderImpl implements the PackEncoder interface.
+class PackEncoderImpl : public gapii::PackEncoder {
+public:
+    PackEncoderImpl(std::shared_ptr<core::StreamWriter> output);
```
AWoloszyn: NIT: I won't hold up the review on this, but I have a strong preference to avoid shared_ptr when there is a clear owner for an object.

In this case, I think either PackEncoderImpl should own StreamWriter, (in which case it would be a std::unique_ptr<>) that we can pass in to move ownership, or the owner of PackEncoder should own it (so a raw-pointer passed in).

In practice for GAPII, I think it makes sense for spy.cpp to own the connection, and then pass out handles (pointers to it) to any objects that need to access. This would that neither connection, nor the encoder itself, would have to be shared pointers.

ben-clayton: I tried moving over to `std::unique_ptr` but the connection is actually shared (it's implemented by `ConnectionStream` that also implements `core::StreamReader`, which is held by other code).
I acknowledge that `unique_ptr` would be preferable here, and I'll see if I can tidy it up later.

gapii/cc/pack_encoder.cpp
```diff
@@ -0,0 +1,144 @@
...
+}
+
+void PackEncoderImpl::message(const Message* msg) {
+    auto desc = msg->GetDescriptor();
+
+    auto insert = mIds.insert(std::make_pair(desc, mIds.size() + 1));
+    if (insert.second) {
+        writeType(desc);
+    }
+	writeSection(insert.first->second, "", msg);
```
ben-clayton: Done

gapis/gfxapi/templates/api_spy.cpp.tmpl
```diff
@@ -190,7 +171,7 @@
     {{$synthetic := GetAnnotation $ "synthetic"}}
 
     {{Template "C++.ReturnType" $}} {{$spyname}}::{{$name}}({{Macro "C++.CallParameters" $ | Strings "CallObserver* observer" | JoinWith ", "}}) {
-      GAPID_DEBUG({{Template "C++.PrintfCommandCall" $}});
+      GAPID_WARNING({{Template "C++.PrintfCommandCall" $}});
```
ben-clayton: Nope, I didn't. Not sure how this crept in.

ben-clayton:
I've also added a small fix for nullptr strings that made protoc cry.
Tested on gles and vulkan.

PR #242, opened 2017-04-10
Added most of the Vulkan texture formats.

AWoloszyn:
Still missing a couple of compressed formats, and the single
ERGB format.


AWoloszyn:
Done! Not sure why I wasn't notified about your comment though.

PR #243, opened 2017-04-10
Instead of a Yes/No replay prioritize the replay devices.

AWoloszyn:
This lets us handle the case where we have multi-api traces.
We can still replay on Device, even though in the general case
we want to replay GLES on Host.

### DIFF
--- a/gapis/server/server.go
+++ b/gapis/server/server.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime/pprof"
+	"sort"
 
 	"github.com/google/gapid/core/app/auth"
 	"github.com/google/gapid/core/app/benchmark"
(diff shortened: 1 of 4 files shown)


gapis/server/server.go
```diff
@@ -152,6 +153,30 @@ func (s *server) GetDevices(ctx context.Context) ([]*path.Device, error) {
...
+func (p prioritizedDevices) Len() int {
+	return len(p)
+}
+
+func (p prioritizedDevices) Less(i, j int) bool {
+	return p[i].priority < p[j].priority
+}
+
+func (p prioritizedDevices) Swap(i, j int) {
+	tempDevice := p[i].device
```
ben-clayton:
```go
func (p prioritizedDevices) Swap(i, j int) { p[i], p[j] = p[j], p[i] }
```

gapis/server/server.go
```diff
@@ -184,29 +209,38 @@ func (s *server) GetDevicesForReplay(ctx context.Context, p *path.Capture) ([]*p
 	}
 
 	all := getSortedDevices(ctx)
-	filtered := make([]bind.Device, 0, len(all))
-nextDevice:
+	filtered := make([]prioritizedDevice, 0, len(all))
```
ben-clayton: nit: More common to use the 2 parameter flavor:

```go
filtered := make([]prioritizedDevice, len(all))
```

ben-clayton: Ignore me, it's the EOD.

PR #249, opened 2017-04-11
Succeed SPIRV Disassembly on all success codes.

AWoloszyn:
This allows other success codes to still return their SPIR-V
from the disassembler.


gapis/shadertools/cc/libmanager.cpp
```diff
@@ -285,7 +285,7 @@ const char* getDisassembleText(uint32_t* spirv_binary, size_t length) {
     (SPV_BINARY_TO_TEXT_OPTION_FRIENDLY_NAMES |
      SPV_BINARY_TO_TEXT_OPTION_INDENT));
 
-  if (result != SPV_SUCCESS) {
+  if (result < 0) {
```
AWoloszyn: Any >0 return code will give you SPIR-V output.

Issue #152, opened 2017-03-27
GAPID Packaging


pmuetschard:
# Package Layout

## Windows
The installer allows the user to install the package anywhere, suggesting `%PROGRAMFILES%\gapid`. The installer will install GAPID into the selected folder with the following structure. Along with an installer, we provide a `.zip` file that can be extracted anywhere. Both will follow the same structure:
```
<install root or gapid inside .zip>\
├─ jre\...
├─ lib\
│  ├─ GraphicsSpyLayer.json
│  ├─ gapic.jar
│  ├─ libgapii.dll
│  ├─ libVkLayer_VirtualSwapchain.dll
│  └─ VirtualSwapchainLayer.json
├─ strings\en-us.stb
├─ build.properties
├─ gapid-arm64-v8a.apk
├─ gapid-armeabi-v7a.apk
├─ gapid-x86.apk
├─ gapid.bat (shell script to launch the client)
├─ gapir.exe
├─ gapis.exe
├─ gapit.exe
├─ libgcc_s_seh-1.dll
├─ libstdc++-6.dll
└─ libwinpthread-1.dll
```

## MacOS
The built package consists of a `.dmg` disk image containing the `GAPID.app` application and link to `/Applications` to make "installing" easy. The `GAPID.app` can be run from anywhere and follows the following structure:
```
GAPID.app/
└─ Contents/
   ├─ MacOS/
   │  └─ <see below>
   ├─ Resources/GAPID.icns
   └─ Info.plist
```
Along with the `.dmg` disk image, we also provide a `.zip` file with the same content's as inside the `Contents/MacOS/` folder of the `.app` package:
```
<Content/MacOS inside .app or gapid inside .zip>/
├─ jre/...
├─ lib/
│  ├─ gapic.jar
│  └─ libgapii.dylib
├─ strings/en-us.stb
├─ build.properties
├─ gapid-arm64-v8a.apk
├─ gapid-armeabi-v7a.apk
├─ gapid-x86.apk
├─ gapid (shell script to launch the client)
├─ gapir
├─ gapis
└─ gapit
```
When running the `.app` application, the `gapid` script is invoked. When using the `.zip` file, the only way to run the application is via the terminal by running the `gapid` script or the executables.

## Linux
[...]

PR #250, opened 2017-04-11
New Package Layout

pmuetschard:
See #152


core/app/layout/layout.go
```diff
@@ -29,10 +30,15 @@ const (
 	ErrUnknownABI             = fault.Const("Unknown device ABI: %+v")
 )
 
+// LibraryType enumerates the possible GAPID dynamic libraries.
+type LibraryType int;
+const (
+	LibGraphicsSpy LibraryType = iota
```
ben-clayton: This PR needs gofmting. I would encourage you to try out VSCode with the go plugin - it's an awesome go editing environment and also automatically formats as you save.

pmuetschard: Done.

pmuetschard:
I've run stuff through gofmt and fixed the linter squiggles from VS code.
I've uploaded a change to make tracing find the MSYS dlls, which has the side effect of fixing windows tracing from the command line without the ENV setup of 'do run'.

AWoloszyn:
Just tried this out for Vulkan on Linux / Android. Everything is working :+1:

PR #256, opened 2017-04-12
Split GL objects to shared and non-shared.

dsrbecky:
Separate the ID->object maps based on whether the object
can be shared between GL contexts or not.

Other then the change in gles.api, this is just a renaming commit.


dsrbecky:
I think I will push this in before the shared-context PR

PR #260, opened 2017-04-12
Don't crash on incomplete texture cubes.

pmuetschard:


gapis/gfxapi/texture.go
```diff
@@ -108,7 +112,9 @@ func (t *Cubemap) ConvertTo(ctx context.Context, f *image.Format) (interface{},
 		out.Levels[i] = &CubemapLevel{}
 		dst, src := out.Levels[i].faces(), m.faces()
 		for j := range src {
-			if obj, err := src[j].ConvertTo(ctx, f); err == nil {
+			if src[j] == nil {
+				dst[j] = nil
```
ben-clayton: I'm still confused by this code. Isn't `dst` always going to be an array of 6 nil `image.Info2D` pointers?
In which case, would it not be more sensible to have:

```go
dst, src := [6]*image.Info2D{}, m.faces()
for j, srcFace := range src {
	if srcFace == nil {
		continue
	}
	cnvFace, err := srcFace.ConvertTo(ctx, f); 
	if err != nil {
		return nil, err
	}
	dst[j] = cnvFace
}
```

pmuetschard: Done.

PR #261, opened 2017-04-12
Binding a framebuffer causes its attachments to be read.

pmuetschard:
This fixes a trace I came across where the texture size is modified after it has been attached:

```
glBindFramebuffer(1)
glBindTexture(1)
glTexImage2D(1x1, null)
glFramebufferTexture2D(1, 1)
glTexImage2D(1024x1024, null)
```

### DIFF
--- a/gapis/gfxapi/gles/dependency_graph.go
+++ b/gapis/gfxapi/gles/dependency_graph.go
@@ -289,13 +289,16 @@ func (g *DependencyGraph) getBehaviour(ctx context.Context, s *gfxapi.State, id 
 				fb := c.Objects.Framebuffers[c.BoundDrawFramebuffer]
 				if (a.Mask & GLbitfield_GL_COLOR_BUFFER_BIT) != 0 {
 					for _, att := range fb.ColorAttachments {
+						b.read(g, getAttachmentSize(g, c, att))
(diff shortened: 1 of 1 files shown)


dsrbecky:
Hmmm.... glBindFramebuffer should not be reading any data.
I feel like this is just masking the real problem.
I think Clears are Draws should be reading the size maybe.
If you send me the capture I can take a look....

pmuetschard:
Draw calls already have a read dependency onto the fb attachements. The problem with waiting until a draw call is that the reads there will only propagate to the the write by the glClear call. I've modified the change so that glClear will read the fb size instead.

gapis/gfxapi/gles/dependency_graph.go
```diff
@@ -418,3 +421,20 @@ func getAttachmentData(g *DependencyGraph, c *Context, att FramebufferAttachment
 	}
 	return
 }
+
+func getAttachmentSize(g *DependencyGraph, c *Context, att FramebufferAttachment) (key stateKey) {
```
dsrbecky: Nit: I would prefer getAttachmentDataAndSize instead of copy-pasting the method (in the same way it is done with textures).

PR #263, opened 2017-04-12
Texture thumbnail fixes.

pmuetschard:
- Fixes a bug in the resize code for uncompressed formats
- If luminance channel is requested of a format that doesn't have a luminance channel, but has a single (other) color channel, use that channel as luminance.

### DIFF
--- a/core/stream/convert.go
+++ b/core/stream/convert.go
@@ -115,7 +115,7 @@ func resolveImplicitMappings(count int, mappings []mapping, srcFmt *Format, srcD
 			}
 		case Channel_Luminance:
 			if c := srcFmt.GetSingleColorComponent(); c != nil {
-				// A format with a signle color channel is equivalent to a luninance format.
+				// A format with a single color channel is equivalent to a luninance format.
 				m.src = buf{srcData, c, srcFmt.BitOffsets()[c], uint32(srcFmt.Stride()) * 8}
 			}
 			// TODO: RGB->Luminance conversion (#276)
--- a/core/stream/convert.go
+++ b/core/stream/convert.go
@@ -114,10 +114,11 @@ func resolveImplicitMappings(count int, mappings []mapping, srcFmt *Format, srcD
 				m.src = buf0
 			}
 		case Channel_Luminance:
-			if c := srcFmt.HasSingleColorComponent(); c != nil {
+			if c := srcFmt.GetSingleColorComponent(); c != nil {
+				// A format with a signle color channel is equivalent to a luninance format.
 				m.src = buf{srcData, c, srcFmt.BitOffsets()[c], uint32(srcFmt.Stride()) * 8}
 			}
-			// TODO: RGB->Luminance conversion
+			// TODO: RGB->Luminance conversion (#276)
 		}
 	}
 	return mappings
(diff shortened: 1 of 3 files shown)


core/stream/format.go
```diff
@@ -129,6 +129,21 @@ func (f *Format) HasColorComponent() bool {
 	return false
 }
 
+// HasSingleColorComponent returns a component if the format contains a single
+// color component, otherwise nil. See ColorChannels for channels considered colors.
+func (f *Format) HasSingleColorComponent() *Component {
```
ben-clayton: This signature confuses me. I would expect it to return a bool, iff there's a single color component. Instead it returns the first color component it finds.

core/stream/convert.go
```diff
@@ -113,6 +113,11 @@ func resolveImplicitMappings(count int, mappings []mapping, srcFmt *Format, srcD
 			} else if srcFmt.HasColorComponent() {
 				m.src = buf0
 			}
+		case Channel_Luminance:
+			if c := srcFmt.HasSingleColorComponent(); c != nil {
+				m.src = buf{srcData, c, srcFmt.BitOffsets()[c], uint32(srcFmt.Stride()) * 8}
+			}
+			// TODO: RGB->Luminance conversion
```
ben-clayton: Make it clearer that we're just reusing the first color component we find for luminance. A link to a github issue would also be good.

core/stream/format.go
pmuetschard: Done.

core/stream/convert.go
pmuetschard: Done.

core/stream/format.go
```diff
@@ -129,6 +129,21 @@ func (f *Format) HasColorComponent() bool {
 	return false
 }
 
+// GetSingleColorComponent returns a component if the format contains a single
+// color component, otherwise nil. See ColorChannels for channels considered colors.
+func (f *Format) GetSingleColorComponent() *Component {
```
ben-clayton: Can we call this `GetFirstColorComponent`?

ben-clayton: Then the comment becomes:

```
// GetFirstColorComponent returns the first color component of the format, or nil if the
// format has no colors. See ColorChannels for channels considered colors.
```

core/stream/convert.go
```diff
@@ -113,6 +113,12 @@ func resolveImplicitMappings(count int, mappings []mapping, srcFmt *Format, srcD
 			} else if srcFmt.HasColorComponent() {
 				m.src = buf0
 			}
+		case Channel_Luminance:
+			if c := srcFmt.GetSingleColorComponent(); c != nil {
+				// A format with a signle color channel is equivalent to a luninance format.
```
ben-clayton: s/signle/single

core/stream/format.go
pmuetschard: That is not what it does, though, nor what I want it to do. I want it to check whether the format represents a single color channel, not a format that has at least one color channel.

ben-clayton: Sorry, I obviously didn't actually read the code.

PR #269, opened 2017-04-13
Link CGO products relative to the source directory

fluxlalonde:
This change allows tools outside of the GAPID repository to use GAPID Go code.


fluxlalonde:
Augh!  I need to abandon this - it assumes a build location for the gapid project that's not real!

PR #274, opened 2017-04-18
Allow Go version above the requirement to be used

1ace:
Currently, the code reject any version that isn't `1.8`:
```
Requires Go version '1.8', got '1.8.1'
```

This string comparison isn't perfect (it will be wrong again when go `1.10` comes out), but at least it fixes the issue until someone write a correct version comparison test *(I don't know what the proper way of doing this in go is; looking it up, it looks like [that lib](https://github.com/blang/semver) might be good, but I'm not going to make the decision of suggesting we import that lib into this project)*.

### DIFF
--- a/cmd/do/do.go
+++ b/cmd/do/do.go
@@ -125,7 +125,7 @@ func init() {
 		hostExeExt = ".exe"
 	}
 
-	if got := strings.TrimLeft(runtime.Version(), "go"); got != requiredGoVersion {
+	if got := strings.TrimLeft(runtime.Version(), "go"); got < requiredGoVersion {
 		fmt.Fprintf(os.Stderr, "Requires Go version '%s', got '%s'", requiredGoVersion, got)
 		os.Exit(1)
 	}


AWoloszyn:
Adding @ben-clayton to the review since he handles more of the Go build.

One note however if that we rely on gopherjs which will throw an error when compiling with a golang version whose major version does not match.

```go
// +build !go1.9
// +build go1.8
```

This will cause a much more difficult to diagnose issue, which IRRC is why we have the version check there in the first place. It might be better to have the test be >= 1.8 && < 1.9.

ben-clayton:
I'm not particularly fond of using a string less-than for a version check. We could parse the version with [regexp](https://golang.org/pkg/regexp/), or using a simple `Sscanf`:
```go
	major, minor, point := 0, 0, 0
	fmt.Sscanf(runtime.Version(), "go%d.%d.%d", &major, &minor, &point)
```

> One note however if that we rely on gopherjs which will throw an error when compiling with a golang version whose major version does not match.

As much as I love the concept of GopherJS, I'm leaning towards dropping it, as it pulls in a load of random dependencies, and we only use it to produce some JS that could be written by hand.

ben-clayton:
Oh and @1ace - thank you for taking the time to check out the project and contribute!

ben-clayton:
Made a slightly cleaner version of this change here: https://github.com/google/gapid/pull/367

ben-clayton:
Thank you for the suggestion @1ace!

1ace:
Happy to see it fixed, and even happier to see a better fix than my plain string comparison :)

PR #278, opened 2017-04-18
Vulkan:Add TRANSFER_SRC bit to recreated images and STORE_OP_STORE to recreated renderpasses

Qining:
This fixes replay for multisampled attachments with MEC

### DIFF
--- a/gapis/gfxapi/vulkan/replay.go
+++ b/gapis/gfxapi/vulkan/replay.go
@@ -128,6 +128,40 @@ func (t *makeAttachementReadable) Transform(ctx context.Context, id atom.ID, a a
 			out.MutateAndWrite(ctx, id, newAtom)
 			return
 		}
+	} else if recreateImage, ok := a.(*RecreateImage); ok {
+		pinfo := recreateImage.PCreateInfo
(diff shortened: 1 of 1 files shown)


gapis/gfxapi/vulkan/replay.go
```diff
@@ -128,6 +128,40 @@ func (t *makeAttachementReadable) Transform(ctx context.Context, id atom.ID, a a
 			out.MutateAndWrite(ctx, id, newAtom)
 			return
 		}
+	} else if recreateImage, ok := a.(*RecreateImage); ok {
```
ben-clayton: Are you aware of type switch statements?

```go
switch a := a.(type) {
	case *VkCreateImage:
		// a will be of type *VkCreateImage
	case *RecreateImage:
		// a will be of type *RecreateImage
	...
}
```

Qining: Yes. I think it is better to keep the form unchanged here? I can change to type switch if you prefer that way.

ben-clayton: By all means, keep as is, as this isn't part of this CL - but it would be nice if this was cleaned up.

PR #282, opened 2017-04-19
Fixed shadertools.

AWoloszyn:
My last fix was wrong. It assumed that the auto return was of type
spv_result as it had previously been compared with SPV_SUCCESS. These
were both incorrect, as tools.Disassemble returns a bool.


ben-clayton:
Thank you

PR #287, opened 2017-04-19
Handle glDebugMessageCallback

dsrbecky:
We need to ignore in compat as it will crash gapir (at random point in the driver if/when it tries to use it)

On the other hand it is really useful, so wire it to so gapir reports messages from driver.

### DIFF
--- a/gapir/cc/linux/gles_renderer.cpp
+++ b/gapir/cc/linux/gles_renderer.cpp
@@ -178,6 +178,21 @@ void GlesRendererImpl::createPbuffer(int width, int height) {
     mPbuffer = glXCreatePbuffer(mDisplay, mFBConfig, pbufferAttribs);
 }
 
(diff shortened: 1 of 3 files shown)


gapir/cc/linux/gles_renderer.cpp
```diff
@@ -287,6 +302,13 @@ void GlesRendererImpl::bind() {
             mNeedsResolve = false;
             mApi.resolve();
         }
+
+        if (mApi.mFunctionStubs.glDebugMessageCallback != nullptr) {
+            mApi.mFunctionStubs.glDebugMessageCallback(reinterpret_cast<void*>(&DebugCallback), nullptr);
```
ben-clayton: Add a TODO and github issue to disable this for profiling, please.

dsrbecky: Done - #289

PR #292, opened 2017-04-20
Replace most atom.NoID uses with derived ID.

dsrbecky:
Derived ID has a high bit set to indicate it is generated atom.

The derived ID has no semantic value and it is just for debugging.


dsrbecky:
Done

PR #293, opened 2017-04-20
Added a mode to gapit to only trace a single API.

AWoloszyn:
This will allow us to filter out traced commands during trace.
Specifically this means we don't have to care about GLES during a
vulkan trace.

### DIFF
--- a/cmd/gapit/trace.go
+++ b/cmd/gapit/trace.go
@@ -49,12 +49,19 @@ func init() {
 	})
 }
 
+// These are hard-coded and need to be kept in sync with the api_index
+// in the *.api files.
+const coreAPI = uint32(1 << 0)
+const glesAPI = uint32(1 << 1)
+const vulkanAPI = uint32(1 << 2)
+
 func (verb *traceVerb) Run(ctx context.Context, flags flag.FlagSet) error {
 	options := client.Options{
 		ObserveFrameFrequency: uint32(verb.Observe.Frames),
 		ObserveDrawFrequency:  uint32(verb.Observe.Draws),
 		StartFrame:            uint32(verb.Start.At.Frame),
 		FramesToCapture:       uint32(verb.Capture.Frames),
+		APIs:                  uint32(0xFFFFFFFF),
 		APK:                   verb.APK,
 	}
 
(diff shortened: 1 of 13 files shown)


gapii/cc/call_observer.cpp
```diff
@@ -112,7 +113,7 @@ void CallObserver::invoke() {
 
 void CallObserver::encodeAndDeleteCommand(::google::protobuf::Message* cmd) {
     observePending();
-    auto encoder = mSpyPtr->getEncoder();
+    auto encoder = mSpyPtr->getEncoder(mApi);
```
Qining: When we call `encodeAndDeletCommand`, what is the value of `mApi`? I think the value is `0` as in spy.cpp, the observer is constructed with `0` value. But I think the value needs to be `1` as that is the bit mask value for Core API?

gapii/cc/spy.cpp
```diff
@@ -205,13 +205,14 @@ Spy::Spy()
...
-
+    set_valid_apis(header.mAPIs);
+    GAPID_ERROR("APIS %08x", header.mAPIs);
     GAPID_INFO("GAPII connection established. Settings:");
     GAPID_INFO("Observe framebuffer every %d frames", mObserveFrameFrequency);
     GAPID_INFO("Observe framebuffer every %d draws", mObserveDrawFrequency);
     GAPID_INFO("Disable precompiled shaders: %s", mDisablePrecompiledShaders ? "true" : "false");
 
-    CallObserver observer(this);
+    CallObserver observer(this, 0);
```
Qining: Is it fine that the `mApi` value of the observer is `0`? I think we need it to be `1` as we built an Architecture atom below?

cmd/gapit/trace.go
```diff
@@ -49,12 +49,17 @@ func init() {
 	})
 }
 
+const coreAPI = uint32(1 << 0)
```
ben-clayton: Please add a comment saying these hard-coded constants need to be kept in-sync with the `api_index` in the API files.

gapii/cc/call_observer.cpp
AWoloszyn: The Call Observer is re-created for every command. This is up to date with each command that may be run. (Vulkan/Gles/Core all set it when they create the call observer).

gapii/cc/spy.cpp
AWoloszyn: Good catch.

cmd/gapit/trace.go
AWoloszyn: Done

Qining:
looks good for me

PR #299, opened 2017-04-21
core/data/pack: Handle bad data without exploding.

ben-clayton:


core/data/pack/types.go
```diff
@@ -86,7 +86,11 @@ func (t *Types) AddMessage(msg proto.Message) (Type, bool) {
...
 // It uses the proto type registry to look up the name.
 func (t *Types) AddName(name string) (Type, bool) {
-	typ := proto.MessageType(name).Elem()
+	typ := proto.MessageType(name)
+	if typ == nil {
+		return Type{}, false
+	}
+	typ = typ.Elem()
 	msg := reflect.New(typ).Interface().(proto.Message)
 	return t.add(msg, name, typ)
```
dsrbecky: Shouldn't this return true now?

ben-clayton: It does. `add` has and always returns a `(Type, bool)` pair.

PR #301, opened 2017-04-21
Fix the format problem for Vulkan Depth Attachment

Qining:
This bring back the depth buffer for Vulkan

In short, when we copy the depth data from a depth_stencil format image,
we need to extract only the depth part, so we need a mapping from
depth_stencil format to the corresponding pure depth format.

But when we create the staging image, we still need to use the
depth_stencil format, as vkCmdBlitImage requires the src image and dst
image to be generally the same (if either of srcImage or dstImage was
created with a depth/stencil format, the other must also have been
created with EXACTLY the same format.

Also, when the source image is a depth/stencil image, the filter of
vkCmdBlitImage must be VK_FILTER_NEAREST


core/stream/fmts/ds.go
```diff
@@ -29,6 +29,18 @@ var (
 		}},
 	}
 
+	DS_NU16S8 = &stream.Format{
```
AWoloszyn: Can you add the new formats to fmts_test.go

Qining: Added in 28a34ec

Qining: Fixed in f326a5a

PR #304, opened 2017-04-25
Add core/data/protoconv: A registry of proto <-> go-struct converters.

ben-clayton:


core/data/protoconv/protoconv.go
```diff
@@ -0,0 +1,175 @@
...
+		panic("toObject must be a function")
+	}
+	if err := checkToFunc(toP); err != nil {
+		panic(fmt.Errorf(`toProto does not match signature:
+  func(context.Context, O) (P, error)
+Got:   %v
+Error: %v`, printFunc(toP), err))
+	}
+	if err := checkToFunc(toO); err != nil {
+		panic(fmt.Errorf(`toProto does not match signature:
```
AWoloszyn: toObject does not match siganture:

ben-clayton: Good catch. Done.

PR #309, opened 2017-04-26
Move code from {api}_types.h to {api}_types.cpp

dsrbecky:
This reduces uncached clean build time by about 25%.

The header is frequently referenced so make it smaller.
Constructors are moved to (newly created) cpp file.
Reference to the protobuf header is also moved to cpp.


gapis/gfxapi/templates/api_types.h.tmpl
```diff
@@ -47,6 +43,18 @@
   #include <stdint.h>
   #include <string.h>
 ¶
+  // Forward declarations
+  namespace google {
+    namespace protobuf {
```
ben-clayton: nit: no indent for namespaces.

PR #323, opened 2017-04-27
Add .gfxtrace capture header.

ben-clayton:
This CL breaks backwards compatibility with the old file formats, but cleans up lots of code.


ben-clayton:
Not ready yet.

PR #327, opened 2017-04-28
Vulkan:Fix crash caused by accessing NULL handle in Vulkan DCE

Qining:
Check if the given Vulkan Handle exist in the state before getting the object of the handle. Sometime using VK_NULL_HANDLE (value 0) is valid, but no object should be obtained from that null handle.

### DIFF
--- a/gapis/gfxapi/vulkan/dependency_graph.go
+++ b/gapis/gfxapi/vulkan/dependency_graph.go
@@ -334,12 +334,16 @@ func (g *DependencyGraph) getBehaviour(ctx context.Context, s *gfxapi.State, id 
 
 	// Helper function that gets the overlapped memory bindings for a given image
 	getOverlappedBindingsForImage := func(image VkImage) []*vulkanDeviceMemoryBinding {
-		if GetState(s).Images.Get(image).IsSwapchainImage {
+		if !GetState(s).Images.Contains(image) {
+			log.E(ctx, "Error Image: %v: does not exist in state", image)
+		}
+		imageObj := GetState(s).Images.Get(image)
+		if imageObj.IsSwapchainImage {
 			return []*vulkanDeviceMemoryBinding{}
-		} else if GetState(s).Images.Get(image).BoundMemory != nil {
-			boundMemory := GetState(s).Images.Get(image).BoundMemory.VulkanHandle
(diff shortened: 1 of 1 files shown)


gapis/gfxapi/vulkan/dependency_graph.go
```diff
@@ -334,12 +334,16 @@ func (g *DependencyGraph) getBehaviour(ctx context.Context, s *gfxapi.State, id
 
 	// Helper function that gets the overlapped memory bindings for a given image
 	getOverlappedBindingsForImage := func(image VkImage) []*vulkanDeviceMemoryBinding {
-		if GetState(s).Images.Get(image).IsSwapchainImage {
+		if !GetState(s).Images.Contains(image) {
+			log.E(ctx, "Error Image: %v: does not exist in state", image)
```
ben-clayton: Should this return?

gapis/gfxapi/vulkan/dependency_graph.go
```diff
@@ -349,10 +353,14 @@ func (g *DependencyGraph) getBehaviour(ctx context.Context, s *gfxapi.State, id
 
 	// Helper function that gets the overlapped memory bindings for a given buffer
 	getOverlappedBindingsForBuffer := func(buffer VkBuffer) []*vulkanDeviceMemoryBinding {
-		if GetState(s).Buffers.Get(buffer).Memory != nil {
-			boundMemory := GetState(s).Buffers.Get(buffer).Memory.VulkanHandle
-			offset := uint64(GetState(s).Buffers.Get(buffer).MemoryOffset)
-			size := uint64(uint64(GetState(s).Buffers.Get(buffer).Info.Size))
+		if !GetState(s).Buffers.Contains(buffer) {
+			log.E(ctx, "Error Buffer: %v: does not exist in state", buffer)
```
ben-clayton: Likewise, after erroring, is there any point in continuing the function?

gapis/gfxapi/vulkan/dependency_graph.go
Qining: just added in fa86381

gapis/gfxapi/vulkan/dependency_graph.go
Qining: Just added in fa86381

PR #328, opened 2017-04-28
Enable Geometry View for Vulkan

Qining:
So far we don't have a reliable way to get the semantics info of the vertex buffer data.  The semantics is guessed with the order of attributes and the number of elements for each vertex.

Formats are added to map Vulkan RGBA format to gfxapi.Format with channel XYZW.


Qining:
The vertex buffer streams are assigned with names in the following format (assume the vertex is bound to binding 0, and the attribute has location number 1:
`binding=0,location=1`
Note there is no blank space after ','.

gapis/gfxapi/vulkan/draw_call_mesh.go
```diff
@@ -0,0 +1,412 @@
...
+			vertexCount, firstVertex, binding, attribute)
+		if err != nil {
+			return nil, err
+		}
+		translatedFormat, err := translateVertexFormat(attribute.Format)
+		if err != nil {
+			// TODO(qining): This is an error, should emit error message here
+			continue
+		}
+		name := fmt.Sprintf("binding:%v,location:%v", binding.Binding, attribute.Location)
```
Qining: Here we put the binding and location info to the name field of the vertex stream.

pmuetschard: What is the most common way it is formatted in the shader? E.g. you pointed out that there is no space after the comma. Is that typical of how people write it? I'd argue that the format in the name should be what is conventionally used in shaders. The UI is just going to show the name as it and will not (nor should it) try to parse the name.

FYI, in the code you are using ':' as the separator, while in your comment you used '='.

AWoloszyn: In the shader it will look something like
`layout (binding=0, location=0)` however like C code this is entirely up to the programmer.
```glsl
layout(binding = 0, location=1)
layout ( binding=0, location   =   1)
layout ( location = 0, binding = 1 ) 
```
whatever the programmer prefers, so we should probably just pick a convention that would work, personally I prefer
```glsl
layout(binding=0, location=0) 
```

Qining: Oops, I think I should change to '='. As for the blank space, this is just a problem about what we want the user to see. Currently the name is not used in the server side for anything.

Qining:
Changed vertex stream naming to `binding=0, location=1`

AWoloszyn:
I might also note a TODO: where we can disassemble the shader in the future to pull out the Debug name if the shader still has debug info.

Qining:
TODO added in 89a479d

gapis/gfxapi/vulkan/vulkan.api
```diff
@@ -2766,6 +2766,7 @@ cmd void vkGetDeviceQueue(
   pQueue[0] = id
 }
 
+@DrawCall
```
pmuetschard: Is every vkQueueSubmit guaranteed to contain a draw call? Isn't it also possible to have multiple draw calls in a command buffer? If either is true, please add a TODO (somewhere) to indicate the current limitations.

With the proto changes we're currently making, it might be helpful to give Ben some info here: https://github.com/ben-clayton/gapid/issues/17

Qining: Agree, TODO added in 91d2528.

Issue #303, opened 2017-04-25
Possible to use gapii for arm linux target (non android) ?


eternity74:
Hi.
I'd like to use this tool our env (non android, arm linux gles application)
I thought that I build libgapii.so and use LD_PRELOAD=libgapii.so for the gles application then use  gapit with local_port from ubuntu PC (local_port will forward to arm linux target device via ssh) to get trace. Does it make sense ?
If it makes sense, could anyone give a hint to me for building only gapii for embebed arm linux device ?
I am trying to build gapii under our env but it is hard to setup for me.
Thanks.

ben-clayton:
Hello,

Yes, it should be possible, but it isn't something we have been testing so far. All of the symbols should be exposed in the generated `gles_exports.cpp` via the [`EXPORT` macro](https://github.com/google/gapid/blob/master/core/cc/target.h#L28).

The bit you're missing is the build steps to generate a `libgapii.so` for vanilla linux arm.

Android is handled a bit differently. It uses its own toolchain, and relies on [a different mechanism](https://github.com/google/gapid/blob/master/gapii/cc/android/installer.cpp) instead of `LD_PRELOAD` in order to hook into the calls.

We could certainly add the rules to the CMake build system to build an ARM set of libraries, however we are planning on switching to a different (and more reliable) build system soon. @ianthehat may be able to provide an ETA for this.
That said, if you wanted to brave the CMake rules, we'd happily accept a PR to generate this flavor of `libgapii.so` so long as it can be disabled with a flag.

If you're happy to wait, we can consider this a feature request, and we'll add support sometime after we've switched build systems.

Many thanks.

eternity74:
Hi ben-clayton,
Thanks for your help.
I think there might be some conflict between x86 linux (GL) and  arm linux (GLES). I am not sure, I can make it work for both flavors. I will try to make libgapii.so with arm linux toolchain (GLES) by myself anyway.
If I have successful result, I will propose it to be accepted.
Also I'm interesting in new build system. Could you give some more information about it and ETA ?
Thanks a lot.

ben-clayton:
> If I have successful result, I will propose it to be accepted.
Awesome, thank you.

The new build system is [bazel](https://bazel.build/). We're working to iron out some issues with Windows (the bazel team are still fixing some windows specific issues).
There are a number of benefits to bazel over what we currently have:
* It uses its own internal file system to keep track of dependencies. This removes the possibility of broken dependencies that cause flakey builds.
[...]

PR #331, opened 2017-05-02
added gapii target for arm linux (#303)

eternity74:
We can build libgapii.so for arm linux target with this patch.
To build libgapii.so, please follow below steps
 1. set proper path for sysroot and compiler in cmake/Toolchain/ArmLinux.cmake
    (see ARMLINUX_SYSROOT and ARMLINUX_COMPILER_BIN_PATH)
 2. ./do config
    'true' at "Build additional gapii for armlinx [Default: false]"

libgapii.so will be created at release/bin/armlinux


eternity74:
I signed it!

cmd/do/env.go
```diff
@@ -57,8 +57,8 @@ func env(cfg Config) *shell.Env {
 
 	path = append(path,
 		cfg.bin().System(),
-		cfg.JavaHome.Join("bin").System(),
 	)
+	path = append([]string{cfg.JavaHome.Join("bin").System()}, path...)
```
ben-clayton: What was the reason for this change?

eternity74: I have multiple java version on build system and java 1.7 was used basically.
And I set java 1.8 installation path during ./do config but during the build, java 1.7 was used bacause of /usr/bin/java which is version 1.7
If you prefer keep original code, I can revert this change.
Thanks.

ben-clayton: > If you prefer keep original code, I can revert this change.

Yes please, it may have unexpected side-effects in our build.

eternity74: I revert this change.
Thanks.

cmd/do/config.go
```diff
@@ -63,6 +63,7 @@ type Config struct {
 	NinjaPath      file.Path `desc:"Path to ninja executable"`
 	PythonPath     file.Path `desc:"Path to python executable"`
 	MSYS2Path      file.Path `desc:"Path to msys2 root" os:"windows"`
+	ArmLinuxGapii  bool      `desc:"Build additional gapii for armlinx"`
```
ben-clayton: nit: s/armlinx/armlinux

eternity74: done. I uploaded change.

AWoloszyn:
I rebased and double-checked this. Everything still seems to be working. Merging it.

PR #333, opened 2017-05-02
Fixes for broken shader-editing in Vulkan.

AWoloszyn:
This fixes shader editing in Vulkan. There were 3 issues fixed.
1) libmanager.cpp was checking result (bool) != SPV_SUCCESS (0), which
always returned the wrong thing.
2) The memory layout was not set when re-encoding the new
VkCreateShaderModule paramters, so this only worked on ARM32.
3) We missed RecreateShaderModule in mid-execution capture. Now we can
modify shaders in RecreateShaderModule.

### DIFF
--- a/gapis/gfxapi/vulkan/resources.go
+++ b/gapis/gfxapi/vulkan/resources.go
@@ -621,6 +621,9 @@ func (shader *ShaderModuleObject) SetResourceData(ctx context.Context, at *path.
 		if a, ok := list.Atoms[i].(*VkCreateShaderModule); ok {
 			edits(uint64(i), a.Replace(ctx, data))
 			return nil
+		} else if a, ok := list.Atoms[i].(*RecreateShaderModule); ok {
+			edits(uint64(i), a.Replace(ctx, data))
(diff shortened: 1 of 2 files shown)


gapis/gfxapi/gles/resources.go
```diff
@@ -203,17 +204,28 @@ func (shader *Shader) SetResourceData(ctx context.Context, at *path.Command,
 	for resource.Accesses[index] > at.Index && index >= 0 {
 		index--
 	}
+
+	to_mutate := len(list.Atoms)
+	if to_mutate > 5 {
+		to_mutate = 5
+	}
+	// Hack until the architecture is part of the header
```
ben-clayton: I'll have the header CL up for review later today. Do you want to wait for this to be in so you don't have to add more hacks?

AWoloszyn: I will wait and then rebase this on top of your change. I was unsure when that CL would make it up, and I would rather not make you rebase on top of these changes either.

PR #335, opened 2017-05-02
Add protopack capture header

ben-clayton:
This CL breaks backwards compatibility with the old file formats.


gapii/cc/spy.cpp
```diff
@@ -73,102 +107,41 @@ const uint32_t kStartMidExecutionCapture =  0xdeadbeef;
...
-        }
-        return result;
-    case 4:
-        if (x != '\n' && x != '\r') {
-            val.push_back((char) x);
-        } else {
-            if (val.back() == ']') {
-              val.pop_back();
+Spy* Spy::get() {
+    bool init;
```
AWoloszyn: I am not sure about the status of support in the NDK, but I think std::call_once might help simplify this.

ben-clayton: Ack. Created #336 to track.

PR #337, opened 2017-05-02
Vulkan:Assignment to LastDrawInfo should only happen in command callbacks

Qining:
Also moved the CmdBoundBuffers and related structs around.

### DIFF
--- a/gapis/gfxapi/vulkan/vulkan.api
+++ b/gapis/gfxapi/vulkan/vulkan.api
@@ -5510,7 +5510,6 @@ sub void doCmdBindDescriptorSets(CmdBindDescriptorSets bind) {
       readCoherentMemoryInBuffer(v.Buffer, v.Offset, v.Range)
     }
   }
-  clearLastDrawInfoDescriptorSets()
   for _, binding, set in bind.DescriptorSets {
       LastDrawInfo.DescriptorSets[binding] = set
   }
--- a/gapis/gfxapi/vulkan/vulkan.api
+++ b/gapis/gfxapi/vulkan/vulkan.api
@@ -5496,6 +5496,26 @@ cmd void vkCmdSetStencilReference(
     u32                reference) {
 }
 
+@internal
+class CmdBindDescriptorSets {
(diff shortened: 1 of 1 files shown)


gapis/gfxapi/vulkan/vulkan.api
```diff
@@ -5496,6 +5496,26 @@ cmd void vkCmdSetStencilReference(
...
+}
+
+sub void doCmdBindDescriptorSets(CmdBindDescriptorSets bind) {
+  for _ , _ , v in bind.BoundBuffers {
+    if (v.Buffer != null) {
+      v.Buffer.LastBoundQueue = LastBoundQueue
+      readCoherentMemoryInBuffer(v.Buffer, v.Offset, v.Range)
+    }
+  }
+  clearLastDrawInfoDescriptorSets()
```
AWoloszyn: This is not quite right.  From the spec:
`
vkCmdBindDescriptorSets causes the sets numbered [firstSet.. firstSet+descriptorSetCount-1] to use the bindings stored in pDescriptorSets[0..descriptorSetCount-1] for subsequent rendering commands (either compute or graphics, according to the pipelineBindPoint). Any bindings that were previously applied via these sets are no longer valid.
`
It does not unbind any previously bound descriptors. In fact if you have a bound descriptor, a future pipeline may read from it

```

Similarly, when binding a pipeline, the pipeline can correctly access any previously bound descriptor sets which were bound with compatible pipeline layouts, as long as all lower numbered sets were also bound with compatible layouts.

Layout compatibility means that descriptor sets can be bound to a command buffer for use by any pipeline created with a compatible pipeline layout, and without having bound a particular pipeline first. It also means that descriptor sets can remain valid across a pipeline change, and the same resources will be accessible to the newly bound pipeline.
```

Qining: Removed the call to ClearDescriptorSets in 6957177.

The clear function is kept for future 'correct' logic of clearing bound descriptor sets.

PR #343, opened 2017-05-03
Don't include alpha when computing the color range for tone mapping.

pmuetschard:


gapic/src/main/com/google/gapid/image/ArrayImageBuffer.java
```diff
@@ -367,16 +367,31 @@ private FloatPixelInfo(float min, float max) {
       this.max = max;
     }
 
-    public static PixelInfo compute(FloatBuffer buffer) {
+    public static PixelInfo compute(FloatBuffer buffer, boolean hasAlpha) {
```
ben-clayton: Maybe add a comment saying that you expect 4 components if there's alpha?

pmuetschard: Renamed hasAlpha to isRGBA to be more explicit.

PR #344, opened 2017-05-05
Enable gapit dump information about the ABI used for the trace

Qining:


cmd/gapit/flags.go
```diff
@@ -115,12 +115,13 @@ type (
...
-		Observations   bool `help:"if true then the bytes of every observation are also dumped."`
-		Raw            bool `help:"if true then the value of constants, instead of their names, will be dumped."`
-		ShowDeviceInfo bool `help:"if true then show originating device information."`
+		Gapis            GapisFlags
+		Gapir            GapirFlags
+		Extras           bool `help:"if true then extras are also dumped."`
+		Observations     bool `help:"if true then the bytes of every observation are also dumped."`
+		Raw              bool `help:"if true then the value of constants, instead of their names, will be dumped."`
+		ShowDeviceInfo   bool `help:"if true then show originating device information."`
+		ShowTraceABIInfo bool `help:"if true then show information of the ABI used for the trace."`
```
ben-clayton: Übernit: Drop `Trace` from `ShowTraceABIInfo`.

I'm trying to keep consistency between 'trace' and 'capture'. __capture__ is the result of a __trace__ action. __trace__ is always the verb. In this case, it's already a long parameter to type, so I just suggest dropping the `Trace` altogether.

Qining: Fixed in 43665bb

PR #345, opened 2017-05-08
Switch off some of the OS checks for the more generic check.

AWoloszyn:
This lets us turn on/off Coherent memory tracking more easily.


Qining:
looks good to me

PR #352, opened 2017-05-12
Merge 'protos' branch back into master

ben-clayton:
This is an enormous PR that brings in the work Pascal and I have been working on for the last month.

The main work of this PR is to kill codergen and use protobufs for everything. This involved redesigning many of the RPC APIs.

While I'd like to extract out little pieces, so much as changed that extracting tiny bits is infeasible.

There are still some rough edges. Given the small user base, I think this is Good Enough™ to go back into master so we're all working on the same code.

Apologies for the size of the review.


AWoloszyn:
The only issue I see is that thumbnails are broken for Vulkan. In the client, all thumbnails are (!), but the framebuffers are generated correctly.

cmd/gapit/common.go
```diff
@@ -54,9 +67,9 @@ func getGapis(ctx context.Context, gapisFlags GapisFlags, gapirFlags GapirFlags)
 		return nil, log.Err(ctx, err, "Failed to connect to the GAPIS server")
 	}
 
-	if h := log.GetHandler(ctx); h != nil {
-		go client.GetLogStream(ctx, h)
-	}
+	// if h := log.GetHandler(ctx); h != nil {
```
ben-clayton: Fixed.

dsrbecky:
Are you tempted to extract out the non-proto related parts?
I am not saying to properly split it out to bite-sized chunks, just extract out the CLs which can be cleanly pushed first?

ben-clayton:
> Are you tempted to extract out the non-proto related parts?

At this point, no, not really. Even CLs that look unrelated likely have merge conflicts. When I merge this PR I'll be sure to rebase instead of squash. So what you're suggesting is simply reordering the stack of changes. Given that Pascal and I have been working in this branch for the last few weeks, I'd bet that last minute rebases are just going to introduce more bugs than taking what we have now.

gapis/gfxapi/gles/glsl/preprocessor/preprocessorImpl.go
```diff
@@ -314,7 +314,7 @@ func (p *preprocessorImpl) processDefine(args []TokenInfo) {
 	macro := args[0]
 
 	if _, ok := p.macros[macro.Token.String()]; ok {
-		p.err.Errorf("Macro '%s' already defined.", args[0].Token)
+		delete(p.macros, macro.Token.String())
```
dsrbecky: I think redefiniton is error only if the macro changes.  This code will die anyway.

ben-clayton:
Thanks all.

gapis/service/service.proto
```diff
@@ -67,32 +67,45 @@ message ServerInfo {
...
-message Hierarchies { repeated Hierarchy list = 1; }
+message Events { repeated Event list = 1; }
 message StringTableInfos { repeated stringtable.Info list = 1; }
-
-message Object {
-  bytes data = 1;
-}
+message Threads { repeated path.Thread list = 1; }
 
 message Value {
```
ben-clayton: Both of your suggestions also are overloaded - `Any` is used in paths, `Response` isn't great as `Value` is also used by `Set()`.
My justification for the name is when you consider the package each of these are declared in. `pod.Value` is a plain-old-data value, `box.Value` is a boxed value, `service.Value` is any value that is used by the service package. I guess there's a question about whether we could fold `box.Value` and `service.Value` together. I'll think about how easy that would be.

On a side note, this wasn't actually renamed / introduced in these CLs.

gapis/service/path/path.proto
```diff
@@ -167,7 +264,7 @@ message ImageInfo {
 // MapIndex is a path to a value held inside a map.
 message MapIndex {
```
ben-clayton: > Could this be just:

No, because the `map` is a path to a server-side object, and your suggestion uses two client-side values.

dsrbecky: message Index {
path.Any value = 1;
box.Value key = 2;
}
?

gapis/service/box/box.proto
```diff
@@ -0,0 +1,115 @@
...
+
+message Value {
+  // value_id is used to describe cyclic values. The identifier is scoped to
+  // this nested value declaration; it is not global.
+  uint32 value_id = 1;
+  oneof val {
+    // If true then this value is a back-reference to a previously described
+    // value with the same id.
+    bool back_reference = 2;
+    pod.Value pod = 3;
```
ben-clayton: POD is used by the database.

gapis/gfxapi/gles/glsl/preprocessor/preprocessorImpl.go
ben-clayton: ack

Issue #370, opened 2017-05-16
Need an API to resolve state paths to state node paths.


pmuetschard:
Follow results for state elements return a state path (see below). The client needs an API to convert that into a StateTreeNode path similar to the API that maps Command paths to CommandTreeNode paths.

Example of a path to a state node as returned by follow:

```
map_index {
  box { pod { sint64: 60 } }
  field {
    name: "Programs"
    field {
      name: "SharedObjects"
      map_index {
        box { pod { uint64: 3545856288 } }
        field {
          name: "Contexts"
          state {
            after { capture { id { data: "..." } }, index: 19206 }
 } } } } } }
```

PR #374, opened 2017-05-16
Path following in the command tree.

pmuetschard:
State paths do not yet select the appropriate node in the state tree, yet. Waiting on an API to do that. See #370.

### DIFF
--- a/gapic/src/main/com/google/gapid/views/AtomTree.java
+++ b/gapic/src/main/com/google/gapid/views/AtomTree.java
@@ -16,10 +16,12 @@
 package com.google.gapid.views;
 
 import static com.google.gapid.image.Images.noAlpha;
+import static com.google.gapid.models.Follower.nullPrefetcher;
 import static com.google.gapid.models.Thumbnails.THUMB_SIZE;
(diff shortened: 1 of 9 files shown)


gapic/src/main/com/google/gapid/views/AtomTree.java
```diff
@@ -206,47 +212,67 @@ public void widgetSelected(SelectionEvent e) {
...
+      }
+
       private void updateHover(int x, int y) {
         TreeItem item = tree.getItem(new Point(x, y));
-        if (item != null && (item.getData() instanceof AtomStream.Node) &&
-            item.getImage() != null && item.getImageBounds(0).contains(x, y)) {
-          hover(item);
-        } else {
-          hover(null);
+        // When hovering over the far left of deep items, getItem returns null. Let's check a few
```
ben-clayton: WTF?

pmuetschard: Yeah... and it's not actually SWT's fault either. On OSX it also appears to be inconsistent :(

PR #375, opened 2017-05-16
Windows build fixes for the ASTC 3rd party library.

pmuetschard:
This seems pretty hacky, but it makes the build work. I tested that the resulting gapis doesn't crash, but I couldn't test that ASTC stuff still works.


pmuetschard:
test\test-core-image.exe passes, so this seems.... safe. meh.

PR #378, opened 2017-05-17
Continuous and presubmit build configs.

pmuetschard:


pmuetschard:
Presubmits now build :)

PR #379, opened 2017-05-18
Replace codegen encode / decode logic with reflection.

ben-clayton:
* Removes a whole load of complicated and delicate template logic with a single, readable implementation.
* Cuts the generated go codegen down by about a 1/3rd.
* Unifies the existing code with the `memory.Writer` logic (used by `atom.Data`).
* Knocks about 7 seconds off a clean build on my '16 MBP.


gapis/memory/decoder.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/google/gapid/core/os/device"
+)
+
+// Decoder provides methods to read primitives from a binary.Reader, respecting
+// a given MemoryLayout.
+// Decoder will automatically handle alignment and types sizes.
+type Decoder struct {
+	r binary.Reader
+	m *device.MemoryLayout
+	o uint64
```
dsrbecky: So this field just shadows the offset of the Reader right?  I always find it a bit sad to see copied state.  I worry they might go out of sync.  Oh well...

ben-clayton: Fair point. I'll look to see if I can remove this in a followup CL.

PR #413, opened 2017-05-20
Fix string decode

ben-clayton:


dsrbecky:
What is going on here?  That looks very messy.
Can &array[1] be unaligned?  That seems crazy.

If you are just trying to keep backwards compatibility for broken captures, just force the char aligment to 1 on load, of hard code it to 1 in the encoder/decoder for now?

ben-clayton:
> Can &array[1] be unaligned?

This is what a trace from Pascal suggested, but you're right, this makes no sense.
@pmuetschard - can you try taking the trace again and seeing if the issue persists?

dsrbecky:
My first guess when reading the PR was that the trace files contained the incorrect alignment for char due to some bug, or that backwards compatibility added the alignments from the first commit of this PR.  Either way we should tack down the underlying cause.

ben-clayton:
Agreed, closing for now.

dsrbecky:
Although the fist commit is probably still needed right?  (Make a separate PR for that?)

ben-clayton:
Way ahead of you. https://github.com/google/gapid/pull/419

PR #419, opened 2017-05-20
Fix alignment of char in device.Little32/64

ben-clayton:
Also tidy the `memory.Encoder` / `Decoder` to use the memory variable length types.


ben-clayton:
This is the good bits from https://github.com/google/gapid/pull/413

PR #423, opened 2017-05-23
Fix out of bound problem in glTexSubImage2D

ek9852:
If we have an x-offset and running on last row,
we will get out of bound, since it wrongly slice a whole
stride.


dsrbecky:
Thank you!

PR #427, opened 2017-05-24
Vulkan:Add support of VkEvent, resolve hanging on host blocking commands

Qining:
Resolve the hanging on commands like: vkQueueWaitIdle caused by vkCmdWaitEvents.

Add support to VkEvent related commands.

### DIFF
--- a/gapis/gfxapi/templates/api_spy.cpp.tmpl
+++ b/gapis/gfxapi/templates/api_spy.cpp.tmpl
@@ -197,6 +197,9 @@
 
         observer->invoke();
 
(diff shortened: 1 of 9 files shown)


gapis/gfxapi/templates/api_spy.cpp.tmpl
```diff
@@ -197,6 +197,9 @@
 
         observer->invoke();
 
+        {{if (GetAnnotation $ "mayBlockTrace")}}
```
Qining: changed to "blocking" in 3544215

PR #432, opened 2017-05-25
Search/Filter for activity chooser and other fixes

### LINKED ISSUE #181: The Select Activity for capture dialog needs search
There is a lot of applications available on rooted phone.
It would be nice to be able to search thought the list.

pmuetschard:
Closes #181

### DIFF
--- a/gapic/src/main/com/google/gapid/views/ActivityPickerDialog.java
+++ b/gapic/src/main/com/google/gapid/views/ActivityPickerDialog.java
@@ -16,35 +16,44 @@
 package com.google.gapid.views;
 
 import static com.google.gapid.widgets.Widgets.createComposite;
-import static com.google.gapid.widgets.Widgets.createTree;
+import static com.google.gapid.widgets.Widgets.createTreeForViewer;
 import static org.eclipse.jface.dialogs.IDialogConstants.OK_ID;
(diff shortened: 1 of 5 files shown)


gapic/src/main/com/google/gapid/views/ActivityPickerDialog.java
```diff
@@ -171,67 +263,111 @@ protected void showError(String message) {
...
+    }
+
+    @Override
+    public StyledString getStyledText(Object element) {
+      if (element instanceof PkgInfo.Package) {
+        return new StyledString(((PkgInfo.Package)element).getName());
+      } else if (element instanceof PkgInfo.Activity) {
+        PkgInfo.Activity a = (PkgInfo.Activity)element;
+        return new StyledString(
+            a.getName(), Action.findLaunchAction(a) != null ? theme.labelStyler() : null);
```
ben-clayton: Does `theme.labelStyler()` make the text bold? If so, this is really not clear.

pmuetschard: It styles it as a label, which does mean bold. The name of the styler is not the best, it probably should be called em(phasized)Styler.

PR #436, opened 2017-05-25
Adding support for shell am instrument

dsrbecky:


dsrbecky:
Still not done....
ApplicationLoaders.getClassLoader breakpoint never seems to be hit....

Issue #401, opened 2017-05-19
Better GAPIC startup and handling of crashing GAPIS


pmuetschard:
Currently, the client starts GAPIS at startup, before showing anything to the user (i.e. before even creating the main window). For people that don't use the command line, this results int multiple seconds where "nothing is happening".

GAPIC also becomes completely unusable if GAPIS crashes, instead it should give the user the option to try and restart GAPIS and try again, or restart GAPIS and "close" the current trace.

GAPIC needs a UI to allow the user to connect to a remote GAPIS, this can currently only be done via command line flags.

ben-clayton:
Dropping priority, as I'm not sure that restarting GAPIS is going to help much - if it crashed once, it's likely to crash immediately again.

Issue #275, opened 2017-04-18
Make DataUnavailableException a "soft error"


dsrbecky:
I frequently get a lot of messages like this in the stdout with full callstack:
E20170418-143055880[grpc-default-executor-8][util.UiCallback.onFinish] error in UiCallback.onRpcThread
E20170418-143055880[grpc-default-executor-8][util.UiCallback.onFinish] com.google.gapid.server.Client$DataUnavailableException: No texture data has been associated with texture Texture<49> at this point in the trace.
E20170418-143055880[grpc-default-executor-8][util.UiCallback.onFinish] 	at com.google.gapid.server.Client.throwIfError(Client.java:212)
E20170418-143055880[grpc-default-executor-8][util.UiCallback.onFinish] 	at com.google.gapid.server.Client.lambda$get$1(Client.java:87)
E20170418-143055880[grpc-default-executor-8][util.UiCallback.onFinish] 	at com.google.common.util.concurrent.AbstractTransformFuture$AsyncTransformFuture.doTransform(AbstractTransformFuture.java:211)

There is generally one for each texture without explicitly loaded data.
This can happen either if the the data was not loaded yet, or if the texture is used as render target.
Either way, I do not feel like this is an error that should be logged - it is expected behavior.

pmuetschard:
See the discussion on service.proto here: https://github.com/ben-clayton/gapid/pull/19

pmuetschard:
oops.

dsrbecky:
I do not see the discussion... :-/

pmuetschard:
Look harder?? :) 

In Ben's first comment (the one with the big red X), click on "show outdated" next to service.proto (the first file)

dsrbecky:
Ops.... I was clicking on the filename which is of no use...

I do not feel strongly about any of the discussed solutions.  As long as the log does not get spammed with errors which are not really errors, I am happy.

As a side note, I personally categorize log messages as Error for cases where unexpected behavior has happened.  Error in the log or UI for me means something has gone wrong - either a bug, or hardware failure.  There should be no errors under normal usage.  Anything that is expected under normal use gets logged as warning.

To extend that though, should rename "soft errors" into warnings?  Eg. WarnDataUnavailable.  Just a though.

pmuetschard:
Everything specific to this bug is done. There are separate bugs open (i.e. #401, #403, and #407) to deal with error handling.

PR #437, opened 2017-05-26
Beginning soft error work.

pmuetschard:
This should take care of a most of the visible things for #275 , but error handling still needs a bit of love, but is dependent on #401.

### DIFF
--- a/gapic/src/main/com/google/gapid/widgets/LoadableImage.java
+++ b/gapic/src/main/com/google/gapid/widgets/LoadableImage.java
@@ -17,6 +17,7 @@ package com.google.gapid.widgets;
 
 import static java.util.logging.Level.FINE;
 
+import com.google.common.base.Preconditions;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.gapid.image.Images;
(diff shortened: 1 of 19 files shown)


gapic/src/main/com/google/gapid/widgets/LoadableImage.java
```diff
@@ -235,6 +176,82 @@ public default void onLoadingStart() { /* empty */ }
...
+  /**
+   * Builder for {@link LoadableImage}. If built using a future, it is assumed
+   * to already be loading, while if built with a supplier, the {@link #load()}
+   * method needs to be invoked to start the loading process.
+   */
+  public static class Builder {
+    private final LoadingIndicator loading;
+    private Supplier<ListenableFuture<Object>> futureSupplier;
+    private boolean small;
+    private boolean shouldLoad;
```
ben-clayton: `shouldLoad` is a bit magic. Perhaps some documentation explaining what it is and why would help someone like me understand why it exists.

pmuetschard: Added a comment to make things clearer.

PR #440, opened 2017-05-26
Implement 3D textures and texture arrays

dsrbecky:


gapis/gfxapi/gles/glsl_compat.go
```diff
@@ -210,6 +215,25 @@ func (t *glslTransformState) apply(child, parent interface{}) interface{} {
 				t.Type = ast.TSampler2D
 			}
 		}
+		if t.renameViewID {
```
ben-clayton: Seems like a good candidate for a test. Especially if we intend to switch parsers.

gapis/gfxapi/gles/resources.go
```diff
@@ -101,17 +102,17 @@ func (t *Texture) ResourceData(ctx context.Context, s *gfxapi.State) (interface{
 					Data:   image.NewID(face.Data.ResourceID(ctx, s)),
 				}
 				switch j {
```
ben-clayton: Nit:

```go
switch j + GLenum_GL_TEXTURE_CUBE_MAP_POSITIVE_X {
    case GLenum_GL_TEXTURE_CUBE_MAP_NEGATIVE_X:
    ...
```
Is more readable and less error prone.

PR #441, opened 2017-05-26
Vulkan video

AWoloszyn:


gapii/cc/spy.cpp
```diff
@@ -108,6 +108,9 @@ const uint32_t kStartMidExecutionCapture =  0xdeadbeef;
 const int32_t  kSuspendIndefinitely = -1;
 
 const uint8_t kCoreAPI = 0;
```
ben-clayton: Not too happy with these constants needing to match those in the API file.
At the very least please can we have a comment?

AWoloszyn: Agreed. I will add a comment right now.
I haven't figure out a good way to handle this. We could maybe generate some relatively empty header files that we can include without having to deal with weird include cycles.

AWoloszyn: Actually come to think of it, I might be able to emit something into vulkan_spy.h / gles_spy.h and handle it that way. I will give it a shot.

gapis/gfxapi/templates/api_spy.h.tmpl
```diff
@@ -62,6 +62,8 @@
 ¶
     {{ForEach (AllCommands $) "DeclareCommand" | JoinWith "\n"}}
 ¶
+    const static uint32_t kApiIndex = {{$.Index}};
```
ben-clayton: Awesome, thank you!

Issue #439, opened 2017-05-26
Gapit dump with observations data and replay is not working for Vulkan


Qining:
The problem is all the dispatchable handles in Vulkan are read/written as Uint64.

In memory/read.go line 48.
When the type is a dispatchable Vulkan handle, for example, VkQueue, the TypeOf VkQueue is not equal to reflect.TypeOf(Size(0)). This means all the dispatchable handles are processed in the default branch below and are considered as Uint64.

The problem is in the commit: 9e4444f97fe8011404e554878d05fd8aefaf3875

ben-clayton:
Sorry for the bug. The simplest fix would be to annotate the memory types with a private dummy function:

```go
func (Size) isSize() {}
```

And then test for this private function via an interface:

```
// IsSize returns true if v is a Size or alias to a Size.
func IsSize(v interface{}) {
    _, ok := v.(type interface{ isSize() })
    return ok
}
```

`IsSize` can just be inlined where ever we check for `tySize`

ben-clayton:
Updated the comment above once I remembered that type aliases don't adopt the aliasee's functions.

PR #442, opened 2017-05-26
Make SizeTy interface to group all size_t types

Qining:
This fixes the incorrect parsing of Vulkan dispatchable handles, which
is caused by not considering those dispatchable handles as size_t type
values.

Fix issue #439 .

### DIFF
--- a/gapis/gfxapi/templates/api.go.tmpl
+++ b/gapis/gfxapi/templates/api.go.tmpl
@@ -125,11 +125,25 @@
   {{else if IsVoid     $u}}
   {{else}}
     {{$to := Macro "Go.Type" $.To}}
-
     type {{$.Name}} {{$to}}
-    {{if (GetAnnotation $ "dispatchHandle")}}
-      // Dummy function to make dispatchable handle implement SizeTy interface
+    {{if IsSize $u}}
+      // Dummy function to make {{$.Name}} implement SizeTy interface
       func ({{$.Name}}) IsMemorySize() {}
+
+    {{end}}
+    {{if IsChar $u}}
+      // Dummy function to make {{$.Name}} implement CharTy interface
+      func ({{$.Name}}) IsChar() {}
+
+    {{end}}
+    {{if IsInt $u}}
+      // Dummy function to make {{$.Name}} implement IntTy interface
+      func ({{$.Name}}) IsInt() {}
+
+    {{end}}
+    {{if IsUint $u}}
+      // Dummy function to make {{$.Name}} implement UintTy interface
+      func ({{$.Name}}) IsUint() {}
 
     {{end}}
   {{end}}
--- a/gapis/gfxapi/templates/api.go.tmpl
+++ b/gapis/gfxapi/templates/api.go.tmpl
@@ -127,6 +127,11 @@
     {{$to := Macro "Go.Type" $.To}}
 
     type {{$.Name}} {{$to}}
+    {{if (GetAnnotation $ "dispatchHandle")}}
+      // Dummy function to make dispatchable handle implement SizeTy interface
+      func ({{$.Name}}) IsMemorySize() {}
+
+    {{end}}
   {{end}}
 
   {{template "LabelsForLabeledType" $}}
--- a/gapis/memory/alignof_sizeof.go
+++ b/gapis/memory/alignof_sizeof.go
@@ -29,7 +29,7 @@ func AlignOf(t reflect.Type, m *device.MemoryLayout) uint64 {
 		return uint64(m.GetPointer().GetAlignment())
 	case t == tyInt, t == tyUint:
 		return uint64(m.GetInteger().GetAlignment())
-	case t == tySize:
+	case t.Implements(reflect.TypeOf((*SizeTy)(nil)).Elem()):
 		return uint64(m.GetSize().GetAlignment())
 	default:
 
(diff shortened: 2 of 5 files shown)


gapis/memory/alignof_sizeof.go
```diff
@@ -29,7 +29,7 @@ func AlignOf(t reflect.Type, m *device.MemoryLayout) uint64 {
 		return uint64(m.GetPointer().GetAlignment())
 	case t == tyInt, t == tyUint:
 		return uint64(m.GetInteger().GetAlignment())
-	case t == tySize:
+	case t.Implements(reflect.TypeOf((*SizeTy)(nil)).Elem()):
```
ben-clayton: How about adding a `tySizeTy` next to `tySize` in types.go?

gapis/memory/alignof_sizeof.go
```diff
@@ -29,7 +29,7 @@ func AlignOf(t reflect.Type, m *device.MemoryLayout) uint64 {
 		return uint64(m.GetPointer().GetAlignment())
 	case t == tyInt, t == tyUint:
```
ben-clayton: Aren't you going to hit exactly the same problem with `Int`, `Uint` and `Char`?
Please can you fix all the types together?

Qining: Interfaces added in 89556d0.

gapis/memory/alignof_sizeof.go
Qining: In 89556d0, the tySize is replaced with `reflect.TypeOf((*SizeTy)(nil)).Elem()`, and all the `==` type comparisons are changed to something like `t.Implements(CharTy):`.

Any new types defined with `type xxx ϟmem.Char` will needs to implements `IsChar()`. However, I have not found any of such types in the code.

gapis/memory/types.go
```diff
@@ -18,21 +18,45 @@ import "reflect"
 
 var (
 	tyPointer = reflect.TypeOf((*Pointer)(nil)).Elem()
-	tyChar    = reflect.TypeOf(Char(0))
-	tyInt     = reflect.TypeOf(Int(0))
-	tyUint    = reflect.TypeOf(Uint(0))
-	tySize    = reflect.TypeOf(Size(0))
+	tyChar    = reflect.TypeOf((*CharTy)(nil)).Elem()
```
ben-clayton: These should really be called `tyXXXTy` to keep with the naming convention we've used elsewhere.

gapis/memory/types.go
```diff
@@ -18,21 +18,45 @@ import "reflect"
...
+	IsUint()
+}
+
+// Dummy function to make Uint implement UintTy interface
+func (Uint) IsUint() {}
+
 // Char is the possibly signed but maybe unsigned C/C++ char.
 type Char uint8
 
+// UintTy is the interface implemented by types that should be treated as char type.
```
ben-clayton: `// CharTy` ...

gapis/gfxapi/templates/api.go.tmpl
```diff
@@ -127,6 +127,11 @@
     {{$to := Macro "Go.Type" $.To}}
 
     type {{$.Name}} {{$to}}
+    {{if (GetAnnotation $ "dispatchHandle")}}
```
ben-clayton: Wouldn't it be better to add the dummy method `IsMemoryXXX` if the pseudonym's underlying type is a `int`, `size`, `char`, etc?

gapis/memory/types.go
```diff
@@ -18,24 +18,45 @@ import "reflect"
...
+	IsUint()
+}
+
+// Dummy function to make Uint implement UintTy interface
+func (Uint) IsUint() {}
 
 // Char is the possibly signed but maybe unsigned C/C++ char.
 type Char uint8
 
+// UintTy is the interface implemented by types that should be treated as char type.
```
ben-clayton: `// CharTy is the interface`...

gapis/gfxapi/templates/api.go.tmpl
Qining: But I remember so far we don't have types defined with int, uint, and char. Types in Vulkan are defined as uint32_t, int32_t according to Spec.

gapis/memory/types.go
Qining: OK, I will added them back. But I think `t == tyChar` case has already been covered by `t.Implements(tyCharTy)`.

Qining: Sorry, I misunderstood. I will rename them, but won't add the original type checking back.

gapis/gfxapi/templates/api.go.tmpl
Qining: I see, so I should have a template to check whether the underlying type is int, uint, and char. I will do that.

Qining: Added in 58a1e33

gapis/memory/types.go
Qining: fixed

gapis/memory/types.go
Qining: fixed

gapis/memory/types.go
Qining: changed back in 58a1e33

PR #443, opened 2017-05-29
Vulkan:Missing pieces of VkEvent Support

Qining:
* MEC support of VkEvent
* Add read observation for buffers used in descriptors via buffer view
* Add vkCmdSet/ResetEvent and vkCmdWaitEvents in dependency graph

based on #442


AWoloszyn:
I want to wait until the sizeTy changes are in so I can test this, then I will approve if all is good.

Issue #237, opened 2017-04-10
Clean up gfxapi.Resource protos


ben-clayton:
Knowing what comes back from a `Get` on a `path.Resource` is full of undocumented knowledge.
By restructuring the protos a little, this can be much cleaner and intuitive:

```proto

// ResourceData represents the resource state at a single point in a capture
message ResourceData {
	oneof data {
		Texture texture = 1;
		Shader shader = 2;
		Program program = 3;
	}
}

// Texture represents a texture resource.
message Texture {
	oneof type {
		Texture1D texture_1d = 1;
		Texture2D texture_2d = 2;
		Texture3D texture_3d = 3;
		Cubemap cubemap = 4;
	}
}

```

PR #455, opened 2017-05-30
Clean up resource.proto

ben-clayton:
Implement the changes as proposed by #237.

Fixes: #237

### DIFF
--- a/gapis/gfxapi/gfxapi.proto
+++ b/gapis/gfxapi/gfxapi.proto
@@ -39,22 +39,12 @@ enum DrawPrimitive {
 enum ResourceType {
 	// UnknownResource represents an unknown resource type
 	UnknownResource = 0;
-	// Texture1DResource represents the Texture1D resource type
-	Texture1DResource = 1;
-	// Texture2DResource represents the Texture2D resource type
-	Texture2DResource = 2;
-	// Texture3DResource represents the Texture2D resource type
-	Texture3DResource = 3;
-	// CubemapResource represents the Cubemap resource type
-	CubemapResource = 4;
+	// Texture represents the Texture resource type
+	TextureResource = 1;
 	// ShaderResource represents the Shader resource type
-	ShaderResource = 5;
+	ShaderResource = 2;
 	// ProgramResource represents the Program resource type
-	ProgramResource = 6;
-	// UninitializedTextureResource represents texture resource type
-	// which has been generated but was never assigned dimensions.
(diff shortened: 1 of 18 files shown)


gapis/gfxapi/gfxapi.proto
```diff
@@ -102,6 +92,26 @@ enum UniformType {
...
+	oneof data {
+		Texture texture = 1;
+		Shader shader = 2;
+		Program program = 3;
+	}
+}
+
+// Texture represents a texture resource.
+message Texture {
+	oneof type {
```
dsrbecky: All of the texture types are essentially identical in most respects.
Would you consider changing the "type" to an just an enum value,
and storing the rest in array of layers...
1D, 2D has always just one layer,
3D, 2D_array has arbitrary number of layers,
Cubemap and Cubemap_array have 6*N layers.

Layers->Levels->Data seems to be model used by the Vulkan API, and I have independently switched the GLSL API to follow the same model.
We do not necessarily have to switch in this PR though.

ben-clayton: > 3D, 2D_array has arbitrary number of layers,

Are you proposing that a 3D texture is just N x `image.Info2D`? If so, I think that's really clunky for the UI to use - especially if we need to actually recreate the 3D texture for a rotatable 3d view.

My personal opinion is that the server interface should be explicit - I steer away from structures that have fields (depth, layers) that only make sense when some other field (type) equals some magic value. This does result in more code, but should be clearer with less chance of misinterpretation.

> I have independently switched the GLSL API to follow the same model.

I'm more than happy for the internal model to be compact, my argument is just for the public interface.

This is purely my opinion though, so I'm happy to open this up for a vote.

dsrbecky: > Are you proposing that a 3D texture is just N x image.Info2D?
Yes.  As far as I am concerted 3D texture and 2D texture array are interchangeable, including the UI visualization.

> My personal opinion is that the server interface should be explicit
I am in favor of explict, however what we are doing here is to create special type for the case when array length is 1.  I would prefer less copy-pasted code.

> This is purely my opinion though, so I'm happy to open this up for a vote.
Fair enough, @pmuetschard @AWoloszyn ... opinions?

pmuetschard: I would vote for separate structures, so it's easier to extend them. For example, it would be helpful to communicate to the client whether a cube texture is considered complete by GL. Having separate structures keeps things like that cleaner.

PR #466, opened 2017-05-31
A few trivial multiview compat additions

dsrbecky:


gapis/gfxapi/gles/compat.go
```diff
@@ -1038,6 +1038,14 @@ func compat(ctx context.Context, device *device.Instance) (transform.Transformer
 				return
 			}
 
+		case *GlFramebufferTextureMultiviewOVR:
+			{
+				// TODO: Support properly.
```
ben-clayton: Care to elaborate on 'properly'?

dsrbecky: I am not completely sure how 'properly' will look yet.

PR #490, opened 2017-06-01
State Tree Linking

### LINKED ISSUE #484: Follow ObjectIDs in state view
I assume the ability to follow object IDs was lost during proto change?

pmuetschard:
Closes #484

### DIFF
--- a/gapis/resolve/state_tree.go
+++ b/gapis/resolve/state_tree.go
@@ -219,6 +219,10 @@ func shallowPathsEqual(a, b path.Node) bool {
 		if b, ok := b.(*path.ArrayIndex); ok {
 			return a.Index == b.Index
 		}
+	case *path.Slice:
+		if b, ok := b.(*path.Slice); ok {
+			return a.Start == b.Start && a.End == b.End
+		}
 	}
 	return false
 }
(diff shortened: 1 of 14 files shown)


gapis/resolve/state_tree.go
```diff
@@ -242,7 +242,7 @@ func (n *stn) buildChildren(ctx context.Context, tree *stateTree) {
 				children = append(children, &stn{
 					name:           fmt.Sprintf("[%d - %d]", n.subgroupOffset+s, n.subgroupOffset+e-1),
 					value:          reflect.ValueOf(slice.ISlice(s, e, tree.state.MemoryLayout)),
-					path:           n.path,
+					path:           path.NewArrayIndex(n.subgroupOffset+s, n.path),
```
ben-clayton: Isn't this assigning the path to the first element in the slice? Shouldn't it be a subslice of the slice?

Also, don't the tests need updating?

gapis/resolve/state_tree_test.go
```diff
@@ -491,7 +491,7 @@ func TestStateTreeNode(t *testing.T) {
 			&service.StateTreeNode{
 				NumChildren:    10,
 				Name:           "[10 - 19]",
-				ValuePath:      rootPath.Field("ReferenceB").Field("Array").Path(),
+				ValuePath:      rootPath.Field("ReferenceB").Field("Array").Slice(10, 19).Path(),
```
ben-clayton: Now that these paths are unique, you can probably kill the `flags    int` field from the test structure. The only reason this existed was that there was a many-to-1 mapping of value path -> STN. Now I believe they're all 1:1.

gapis/resolve/state_tree.go
pmuetschard: Done and done. <sub>I'm so <sub>ashamed...</sub></sub>

gapis/resolve/state_tree_test.go
pmuetschard: Done.

Issue #113, opened 2017-03-17
Add the ability to remove commands from the trace.


AWoloszyn:
Would be nice to be able to be able to examine the trace with some commands removed in order to determine their effect.

PR #492, opened 2017-06-02
Bunch'o'fixes for #479

### LINKED ISSUE #479: gapis panic
gapis panic
Target: Pixel XL, Qualcomm chipset
Reproducible : yes, when select command 317440. Not sure panic triggered by background thread or by user click action.
Gfxtrace: https://drive.google.com/open?id=0B0S0_4EgiFmcTWdvLTlJMmdKcEU

Log:
panic: runtime error: invalid memory address or nil pointer dereference
  panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x4970df1]

goroutine 180020 [running]:
github.com/google/gapid/gapis/gfxapi/gles.subApplyDynamicContextState(0x6c8e8e0, 0xc52746e900, 0x6c91880, 0xc4207aca80, 0x0, 0xc480eae8a0, 0xc4acd0ff40, 0x0, 0xc456fd3c00, 0xc420b6a030, ...)
	/Users/ggl/go/src/github.com/google/gapid/gapis/gfxapi/gles/mutate.go:5641 +0x181
github.com/google/gapid/gapis/gfxapi/gles.(*EglMakeCurrent).mutate(0xc4207aca80, 0x6c8e8e0, 0xc52746e900, 0xc480eae8a0, 0x0, 0x2fed2c7f90f72f97, 0xc4a64d1c60)
	/Users/ggl/go/src/github.com/google/gapid/gapis/gfxapi/gles/mutate.go:18016 +0x3f0
github.com/google/gapid/gapis/gfxapi/gles.(*EglMakeCurrent).Mutate(0xc4207aca80, 0x6c8e8e0, 0xc52746e900, 0xc480eae8a0, 0x0, 0x0, 0x0)
	/Users/ggl/go/src/github.com/google/gapid/gapis/gfxapi/gles/custom_replay.go:269 +0x14d
github.com/google/gapid/gapis/resolve.Memory(0x6c8e8e0, 0xc52746e900, 0xc52746e4e0, 0xc4fc49ae68, 0x4010943, 0x597bae0)
	/Users/ggl/go/src/github.com/google/gapid/gapis/resolve/memory.go:44 +0x198
github.com/google/gapid/gapis/resolve.ResolveInternal(0x6c8e8e0, 0xc52746e8a0, 0x6c8f420, 0xc52746e4e0, 0x585ee40, 0xc452376618, 0x6c8f420, 0xc52746e4e0)
	/Users/ggl/go/src/github.com/google/gapid/gapis/resolve/resolve.go:250 +0x9b0
github.com/google/gapid/gapis/resolve.ResolveService(0x6c8e8e0, 0xc52746e8a0, 0x6c8f420, 0xc52746e4e0, 0xc52746e8a0, 0x59dce60, 0x1, 0x6c64760)
	/Users/ggl/go/src/github.com/google/gapid/gapis/resolve/resolve.go:203 +0x4d
github.com/google/gapid/gapis/resolve.(*GetResolvable).Resolve(0xc452376630, 0x6c8e8e0, 0xc52746e8a0, 0x6c64760, 0xc452376630, 0x1, 0x6c692e0)
	/Users/ggl/go/src/github.com/google/gapid/gapis/resolve/get.go:35 +0xa9
github.com/google/gapid/gapis/database.(*record).resolve(0xc52746e840, 0x6c8e820, 0xc4acd0fe40, 0xc4e613fc00, 0xc52746e480)
	/Users/ggl/go/src/github.com/google/gapid/gapis/database/memory.go:72 +0x8a
github.com/google/gapid/gapis/database.(*memory).resolveLocked.func1(0xc52746e840, 0xc4acd0fe80, 0xc420371f00)
	/Users/ggl/go/src/github.com/google/gapid/gapis/database/memory.go:144 +0x45
created by github.com/google/gapid/gapis/database.(*memory).resolveLocked
	/Users/ggl/go/src/github.com/google/gapid/gapis/database/memory.go:151 +0x4ff

ben-clayton:
Fixes #479.

### DIFF
--- a/core/data/deep/copy.go
+++ b/core/data/deep/copy.go
@@ -21,6 +21,16 @@ import (
 	"github.com/google/gapid/core/data"
 )
 
+// Clone makes a deep copy of v.
+func Clone(v interface{}) (interface{}, error) {
+	s := reflect.ValueOf(v)
+	d := reflect.New(s.Type())
+	if err := reflectCopy(d.Elem(), s, "val", map[reflect.Value]reflect.Value{}); err != nil {
+		return nil, err
+	}
+	return d.Elem().Interface(), nil
+}
+
 // Copy recursively copies all fields, map and slice elements from the value
 // src to the pointer dst.
 func Copy(dst, src interface{}) error {
--- a/gapis/gfxapi/gles/api/extras.api
+++ b/gapis/gfxapi/gles/api/extras.api
@@ -77,15 +77,3 @@ sub void ApplyProgramInfoExtra(ProgramId programId, ref!ProgramInfo info) {
     }
   }
 }
-
-// DeviceInfo is an atom extra used to describe information about the device producing the trace.
-@internal @serialize @unused
-class DeviceInfo {
-  string ProductManufacturer
-  string ProductModel
-  string BuildDescription
-  string BuildVersionRelease
-  string BuildVersionSdk
-  string Debuggable
-  string AbiList
-}
--- a/gapis/gfxapi/gles/compat.go
+++ b/gapis/gfxapi/gles/compat.go
@@ -20,6 +20,7 @@ import (
 	"reflect"
 	"strings"
 
+	"github.com/google/gapid/core/data/deep"
 	"github.com/google/gapid/core/log"
 	"github.com/google/gapid/core/math/interval"
 	"github.com/google/gapid/core/math/u64"
@@ -271,16 +272,17 @@ func compat(ctx context.Context, device *device.Instance) (transform.Transformer
 			// TODO(dsrbecky): This might make some atoms valid for replay which were invalid on trace.
 			scs := FindStaticContextState(a.Extras())
 			if scs != nil && !version.IsES && scs.Constants.MajorVersion < 3 {
-				clone := *a
-				for i, e := range clone.extras.All() {
+				clone, err := deep.Clone(a)
+				if err != nil {
+					panic(err)
+				}
+				a = clone.(*EglMakeCurrent)
+				for _, e := range a.extras.All() {
 					if cs, ok := e.(*StaticContextState); ok {
-						scs := *cs
-						scs.Constants.MajorVersion = 3
-						scs.Constants.MinorVersion = 0
-						clone.extras[i] = &scs
+						cs.Constants.MajorVersion = 3
+						cs.Constants.MinorVersion = 0
 					}
 				}
-				a = &clone
 			}
 
 			// Mutate to set the context, Version and Extensions strings.
@@ -893,6 +895,11 @@ func compat(ctx context.Context, device *device.Instance) (transform.Transformer
 			// depending on their ability to eliminate dead code. In particular,
 			// dead code in pixel shader can allow code removal in the vertex shader.
 			// As these should have no side-effects, just drop them.
+			return
+
+		case *GlGetProgramInterfaceiv:
+			// Introduced as core in 4.3, but macOS caps out at 4.1.
+			// As this should have no side-effects, just drop them.
 			return
 
 		case *GlLabelObjectEXT,
--- a/gapis/gfxapi/gles/extras.go
+++ b/gapis/gfxapi/gles/extras.go
@@ -17,6 +17,7 @@ package gles
 import (
 	"context"
 
+	"github.com/google/gapid/core/data/deep"
 	"github.com/google/gapid/gapis/atom"
 	"github.com/google/gapid/gapis/atom/atom_pb"
 	"github.com/google/gapid/gapis/gfxapi/gles/gles_pb"
@@ -63,7 +64,11 @@ func FindErrorState(extras *atom.Extras) *ErrorState {
 func FindStaticContextState(extras *atom.Extras) *StaticContextState {
 	for _, e := range extras.All() {
 		if cs, ok := e.(*StaticContextState); ok {
-			return cs
+			clone, err := deep.Clone(cs)
+			if err != nil {
+				panic(err)
+			}
+			return clone.(*StaticContextState)
 		}
 	}
 	return nil
@@ -74,18 +79,11 @@ func FindStaticContextState(extras *atom.Extras) *StaticContextState {
 func FindDynamicContextState(extras *atom.Extras) *DynamicContextState {
 	for _, e := range extras.All() {
 		if cs, ok := e.(*DynamicContextState); ok {
-			return cs
-		}
-	}
-	return nil
-}
-
-// FindDeviceInfo searches for the DeviceInfo in the extras,
-// returning the DeviceInfo if found, otherwise nil.
-func FindDeviceInfo(extras *atom.Extras) *DeviceInfo {
-	for _, e := range extras.All() {
-		if di, ok := e.(*DeviceInfo); ok {
-			return di
+			clone, err := deep.Clone(cs)
+			if err != nil {
+				panic(err)
+			}
+			return clone.(*DynamicContextState)
 		}
 	}
 	return nil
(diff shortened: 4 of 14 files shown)


gapis/gfxapi/gles/gles.api
```diff
@@ -515,34 +515,47 @@ sub void ApplyStaticContextState(ref!Context ctx, ref!StaticContextState staticS
...
-    stencil_buffer.Height = dynamicState.BackbufferHeight
-    stencil_buffer.InternalFormat = dynamicState.BackbufferStencilFmt
-    ctx.SharedObjects.Renderbuffers[stencil_id] = stencil_buffer
-
-    if dynamicState.ResetViewportScissor {
-      ctx.FragmentOperations.Scissor.Box.Width = dynamicState.BackbufferWidth
-      ctx.FragmentOperations.Scissor.Box.Height = dynamicState.BackbufferHeight
-      ctx.Rasterization.Viewport.Width = dynamicState.BackbufferWidth
-      ctx.Rasterization.Viewport.Height = dynamicState.BackbufferHeight
+    if 0 in ctx.Objects.Framebuffers {
```
dsrbecky: I am not happy about this - I do want to write the whole API file based on the assumption the data might get inconsistent at any point.  This is just one case, but there is still hundreds left all over the API file, and all over the compat layers, etc... 

If the state might get corrupt, we need to fix the underlying cause - which is context cancellation in this case.  I still maintain my original position that context cancellation is generally bad idea - it is just a trigger that will make your code start failing at random point under hard to reproduce conditions.  If you kick of a job, the job should finish.  It is fine for very expensive jobs to have high level check if the context has been cancelled - so the top most loop of replay could bail out early.  But having leaf sub systems fail and then trying to deal with the fallout is just asking for trouble.

Having said that, we had this discussion many time before, so I am happy with you keeping the cancellation policy as it is, as long as you are the one who deals who plays whack-a-mole with the resulting problems :-)

ben-clayton: > am not happy about this - I do want to write the whole API file based on the assumption the data might get inconsistent at any point.

The code generated from the API cannot crash, full-stop.

I understand that you don't want to add seat-belts everywhere through the API files, but NPEs are not acceptable for _any_ sequence of calls (typical or unusual). Longer term we should be fuzzing the API to prevent panics from occurring, ever.

For this particular case, it would be usual for the framebuffer attachments to be missing, but it is possible once we implement #113, as you just have to delete the first `eglMakeCurrent` which holds the `StaticContextState`. I'm not saying this is the only solution to the problem, but it solves the crashes I encountered (I also wrote a simple stress test of 1000's of concurrent query-wait-cancels to ensure there were no other hidden panics). 

Most of the work I have done with the static analysis was exactly to address this - if we can accurately determine what all the possible entry states are for any command, then we can avoid adding unnecessary seat-belts everywhere.

I've considered attempting to automatically handle things like NPEs, turning them into errors, but each time I've considered it, I've come to the conclusion it's a bad idea. Either the NPE is a failure on our part (which we should fix), or is an error case that we should be reporting with a sensible human readable message.

>  If you kick of a job, the job should finish.

What I failed to add in this PR description are the performance improvements these changes give. Quickly stepping though the commands stacked up walks of state mutation, which could completely stall getting the most recent results. Actually cancelling atom-walks gives noticeable improvements to the responsiveness in the UI.

For that reason, I disagree.

> fine for very expensive jobs to have high level check if the context has been cancelled 

I don't really want to start this argument again, but I'll repeat what I said before - the things that are checking cancellation state are "expensive jobs". The ones that caused these issues were all to do with resolvables - which by nature **are** expensive (otherwise they wouldn't be resolvables in the first place) - and they just happen to be called from `atom.Mutate()`. The obvious problem is that mutate logic depends on the results of these possibly cancelled jobs, and so we have to have some basic handling of context creation in the mutator.

I get that this solution isn't perfect - the cancellation will only be noticed at the end of the `Mutate()` call, and so part of the logic within the `Mutate()` can be using invalid data if the cancellation happens early. The idiomatic way to deal with this is to return and check `errors` everywhere, but the template code generation makes this extremely difficult and I'm worried about runtime / compile time impact. This was the best compromise I could come up with for now. If we finally migrate to the LLVM based tech, then we could probably use exceptions to handle these sorts of cases (including `abort`), which would terminate execution as soon as the cancellation is raised.

dsrbecky: I agree that any sequence of commands should not cause panic, but you have to convince me that is the case here.  With eglMakeCurrent missing this would just result in no context being bound, which would make other commands gracefully stop.

I agree that with quick scan thought the commands, you might want to cancel the unneeded ones.  Let's say you have 1000 requests pending and currently processing the first one, but only needing the last one.  I say you should finish the first one, skip the 998 which have not even started yet, and then do the last one.  If you really want to try hard, I say you can check for cancellation in the top-most loop so you bail out once then next atom is fully processed.  I fail to see why it is important to cancel immediately in the most nested function and then deal with the error propagation and state inconsistency problems.  It sounds to me like yet another really complicated solution for really simple problem.

ben-clayton: Putting the offline discussion here so that it is public:

> With `eglMakeCurrent` missing this would just result in no context being bound, which would make other commands gracefully stop.

Okay, so the failure probably isn't with a preceding `eglMakeCurrent`, but more likely `glGenFramebuffers` reading zeros (due to cancellation). As we get back zeros, we're overwriting the 0'th framebuffer with a newly initialized one, corrupting the state. As discussed, I'll try adding an assert that the read value is not 0, and run my tests again. If this is the case, I'm happy to substitute the first CL with this fix.

> I say you should finish the first one, skip the 998 which have not even started yet, and then do the last one.

So we're now running all state mutations on a single executor? You know there's about 15 state mutations run concurrently whenever a command is clicked? Even if we wanted to go down this path (which I would suspect would make things much slower), adding a queued executor sounds like a few weeks work.

> If you really want to try hard, I say you can check for cancellation in the top-most loop so you bail out once then next atom is fully processed.

That's where the code does exit the loop. I have not changed things so that the `Mutate()` call will return any earlier than it used to. The template changes do two things:

1. Panics if we get an error back that isn't `nil` or `context.Canceled`. It completely ignores `context.Canceled` if it sees it.
2. Returns `ctx.Err()` instead of `nil` at the end of the function. This is just so the callee of `Mutate` doesn't have to check both the `error` returned by `Mutate` and `ctx.Err()` to know whether it should stop.

In terms of state inconsistency - there is a small window in which we can get unexpected data. If the context is cancelled between the `Mutate` start and end, you'll get undefined data on memory reads **for that command only**. After that the `Mutate` function will return `context.Canceled` and the callee should abort the loop. In this situation the state should not be used as the request was cancelled - stop whatever your doing and return.

> I fail to see why it is important to cancel immediately in the most nested function and then deal with the error propagation and state inconsistency problems.

As I've tried to explain, `context.Context` is designed so that cancellation propagates down to all child `Context`s. There is no way to stop a child context being cancelled when the parent is cancelled, unless you build an entirely new context, and copy over the content. This is infeasibly slow, and will lose any data added to the context by non-gapid packages (say gRPC).
`Mutate()` calls into other packages. Some of these packages respect cancellation if they are blocking or just take a long time. Things like database reads respect cancellation, and return the `context.Canceled` error when the callee context says it should stop.

I understand that handling cancellation mid-execution is a PITA, and not something we want to do, so instead of building "yet another really complicated solution", I tried doing the simplest solution - ignore the cancellation inside the `Mutate`, tell the callee if a cancellation occurred.

Please, if you have a better / less-complicated solution, share it.

dsrbecky: glGenFramebuffers returning 0's is a bit less scary since we want to handle invalid traces anway, and it means I do not have check the whole code-base on where else I have (correctly) assumed the default framebuffer has attachments (and all other similar assumptions).

Nonetheless, I would still argue that hardening of the API against invalid input should be done by tests and fuzzers, not by race condition on user machines with valid traces.

> now running all state mutations on a single executor
Well, I phrased it in single-threaded, but the logic applies to multi-threaded.  I assume we have a semaphore somewhere to limit the number of concurrent tasks to the number of CPUs, right?

ben-clayton: So, the good news is that I believe `glGenFramebuffers` is the culprit. I'll change the initial CL. I'll also add a PR for my stress-test gapit verb.

> Nonetheless, I would still argue that hardening of the API against invalid input should be done by tests and fuzzers, not by race condition on user machines with valid traces.

Of course it should. I never said that.

ben-clayton:
Removed the first CL and rebased. Create #498 to track the API issue discussed.

PR #493, opened 2017-06-02
Move packages from core-> test/robot

ben-clayton:
The general rule for core is that it is code used by multiple components.
This PR moves those packages only used by robot from `core` to `test/robot`.


ben-clayton:
> Does that mean your are officially giving up on the new parser?

No. I'm purely enforcing the layout policy we agreed upon.

If get time to migrate everything to lingo, then it can move back to core. Until then, I want to discourage people using it as that'll simply fragment the code even more.

PR #499, opened 2017-06-02
Vulkan:vkGetEventStatus on replay side

Qining:
On replay side, the vkGetEventStatus can proceed only when the return
code matches with the return code in trace

### DIFF
--- a/gapis/gfxapi/templates/vulkan_gfx_api_extras.tmpl
+++ b/gapis/gfxapi/templates/vulkan_gfx_api_extras.tmpl
@@ -377,6 +377,32 @@ bool Vulkan::replayGetFenceStatus(Stack* stack, bool pushReturn) {
     }
 }
 ¶
+bool Vulkan::replayGetEventStatus(Stack* stack, bool pushReturn) {
+    auto expected = gapir::Vulkan::VkResult(stack->pop<uint32_t>());
+    auto event = stack->pop<uint64_t>();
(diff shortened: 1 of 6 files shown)


gapis/gfxapi/templates/vulkan_gfx_api_extras.tmpl
```diff
@@ -377,6 +377,32 @@ bool Vulkan::replayGetFenceStatus(Stack* stack, bool pushReturn) {
...
+    auto event = stack->pop<uint64_t>();
+    auto device = static_cast<size_val>(stack->pop<size_val>());
+    if (stack->isValid()) {
+        GAPID_INFO("vkGetEventStatus(%" PRIsize ", %" PRIu64 ")", device, event);
+        if (mVkDeviceFunctionStubs.find(device) != mVkDeviceFunctionStubs.end() &&
+        mVkDeviceFunctionStubs[device].vkGetEventStatus) {
+            VkResult return_value;
+            do {
+                return_value = mVkDeviceFunctionStubs[device].vkGetEventStatus(device, event);
+            } while (return_value != expected);
```
AWoloszyn: You don't want this.

If in the trace, the event has not yet been set, but in the replay it has already been set, we will loop forever.

Qining: Fixed in d122b8c.

A new parameter 'wait' is added to determine whether use the busy loop.

PR #503, opened 2017-06-05
Fixes to get robot (just about) going.

ben-clayton:


test/robot/replay/client.go
```diff
@@ -71,25 +71,41 @@ func doReplay(ctx context.Context, action string, in *Input, store *stash.Client
...
 	}
-	if err := store.GetFile(ctx, in.Gapit, gapit); err != nil {
-		return nil, err
-	}
-	if err := store.GetFile(ctx, in.Gapis, gapis); err != nil {
-		return nil, err
-	}
-	if err := store.GetFile(ctx, in.Gapir, gapir); err != nil {
+
+	if err := store.GetFile(ctx, in.Trace, tracefile); err != nil {
```
baldwinn860: does in.Trace not already get stored in tracefile from the for loop above?

ben-clayton: That was totally intentional and purely there to test you. Fixed.

PR #515, opened 2017-06-06
GAPIS heartbeat

ben-clayton:
Matches the heartbeat logic between gapis <-> gapir.

Kills gapis if nobody wants to talk to it anymore.

### DIFF
--- a/gapic/src/main/com/google/gapid/server/GapidClientGrpc.java
+++ b/gapic/src/main/com/google/gapid/server/GapidClientGrpc.java
@@ -15,12 +15,14 @@
  */
 package com.google.gapid.server;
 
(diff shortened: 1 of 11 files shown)


gapic/src/main/com/google/gapid/server/GapidClientGrpc.java
```diff
@@ -29,6 +31,7 @@
  * A {@link GapidClient} based on a gRPC service.
  */
 public class GapidClientGrpc implements GapidClient {
+  private static final PingRequest PING_REQUEST = Service.PingRequest.newBuilder().build();
```
pmuetschard: PingRequest.getDefaultInstance() does the same, so no need to create a static constant.

PR #518, opened 2017-06-06
Replace minRequiredVersion with @if

dsrbecky:


gapis/gfxapi/gles/gles.api
```diff
@@ -243,6 +243,146 @@ enum DeclaringApi {
...
+  bool GL_NV_gpu_shader5                               = true
+  bool GL_EXT_polygon_offset_clamp                     = true
+  bool GL_NV_viewport_swizzle                          = true
+  bool GL_EXT_window_rectangles                        = true
+  bool GL_OES_EGL_image_external                       = true
+  bool EGL_KHR_image_base                              = true
+  bool EGL_ANDROID_create_native_client_buffer         = true
+  bool EGL_ANDROID_image_native_buffer                 = true
+}
+AvailableApis Api
```
ben-clayton: API is so overloaded, I'm slightly concerned with adding yet another definition to the codebase.
Part of this is GL version, part of it is extension. Can they not be split into two categories?

dsrbecky: done

PR #532, opened 2017-06-07
Add StartOfFrame annotation

Qining:
The new StartOfFrame is exactly the same as the original EndOfFrame
annotation (as shown in resolve/events.go)
For GLES, atoms like eglSwapBuffer will be annotated with StartOfFrame,
as the framebuffer image should be read before mutating such atoms. This
CL should not change any behavior on GLES side.

For Vulkan, vkQueuePresentKHR will be annotated with the new EndOfFrame
annotation. So when grabbing framebuffer images, vkQueuePresentKHR will
be mutated (but not called on replay side), so the server will grab the
image which is to be presented, not the image just rendered.

### DIFF
--- a/cmd/perf/bench.go
+++ b/cmd/perf/bench.go
@@ -313,7 +313,7 @@ func getAtomIndicesAndSampleGrabber(bench *Benchmark, session *session) (err err
 	switch bench.Input.BenchmarkType {
 	case "frames":
 		arr, grab = arr.filter(func(idx int) bool {
-			return session.atoms[idx].AtomFlags().IsBeginOfFrame()
+			return session.atoms[idx].AtomFlags().IsEndOfFrame()
 		}), getFrame
 	case "state":
 		grab = getStateAfter
@@ -347,7 +347,7 @@ func (s *session) grabSamples() error {
 	}
 	currentFrame := 0
 	for index, atom := range s.atoms {
-		if atom.AtomFlags().IsBeginOfFrame() {
+		if atom.AtomFlags().IsEndOfFrame() {
 			currentFrame++
 		}
 		if interesting[index] {
--- a/cmd/perf/bench.go
+++ b/cmd/perf/bench.go
@@ -313,7 +313,7 @@ func getAtomIndicesAndSampleGrabber(bench *Benchmark, session *session) (err err
 	switch bench.Input.BenchmarkType {
 	case "frames":
 		arr, grab = arr.filter(func(idx int) bool {
-			return session.atoms[idx].AtomFlags().IsEndOfFrame()
+			return session.atoms[idx].AtomFlags().IsBeginOfFrame()
 		}), getFrame
 	case "state":
 		grab = getStateAfter
@@ -347,7 +347,7 @@ func (s *session) grabSamples() error {
 	}
 	currentFrame := 0
 	for index, atom := range s.atoms {
-		if atom.AtomFlags().IsEndOfFrame() {
+		if atom.AtomFlags().IsBeginOfFrame() {
 			currentFrame++
 		}
 		if interesting[index] {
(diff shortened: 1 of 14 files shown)


Qining:
One thing I don't like about this approach is that, BeginOfFrame and EndOfFrame must not exist in one API at the same time. Probably we should note this somewhere, or handle such cases in our code?

gapii/cc/spy.cpp
```diff
@@ -352,6 +352,36 @@ void Spy::onPostDrawCall(uint8_t api) {
...
+
+void Spy::onPostBeginOfFrame(CallObserver* observer) {
+    if (!is_suspended() && mCaptureFrames >= 1) {
+        mCaptureFrames -= 1;
+        if (mCaptureFrames == 0) {
+            mConnection->close();
+            set_suspended(true);
+        }
+    }
+    if (mSuspendCaptureFrames.load() > 0) {
```
Qining: Currently we don't have Mid Execution Capture supported on GLES side. And this onPostBeginOfFrame is only used for GLES. So probably we should not have a function body like this here, which could be misleading and probably wrong?

AWoloszyn: It is a hard call, if we have multiple APIs in the same capture, should we start vulkan MEC here?

cmd/perf/bench.go
```diff
@@ -313,7 +313,7 @@ func getAtomIndicesAndSampleGrabber(bench *Benchmark, session *session) (err err
 	switch bench.Input.BenchmarkType {
 	case "frames":
 		arr, grab = arr.filter(func(idx int) bool {
-			return session.atoms[idx].AtomFlags().IsEndOfFrame()
```
ben-clayton: These should probably be `IsEndOfFrame`

gapic/src/main/com/google/gapid/models/AtomStream.java
```diff
@@ -203,10 +203,10 @@ public int getStartOfFrame(long index) {
   /**
    * @return the index of the last command of the frame that contains the given command.
    *
-  public int getEndOfFrame(long index) {
```
ben-clayton: Nope

cmd/perf/bench.go
Qining: changed back to EndOfFrame.

gapic/src/main/com/google/gapid/models/AtomStream.java
Qining: Changed back to EndOfFrame in 6f5f49c

Qining:
@ben-clayton BeginOfFrame has been changed to StartOfFrame. The incorrect changes in the java code and bench.go have been reverted.

Just curious, since GLES does not have any atoms with EndOfFrame annotation, can it cause problems in bench.go?

gapic/src/main/com/google/gapid/models/AtomStream.java
```diff
@@ -193,7 +193,7 @@ public Atom getAtom(long index) {
   public int getStartOfFrame(long index) {
     Atom[] atoms = getData().getAtoms();
     for (int i = (int)index; i > 0; i--) {
-      if (atoms[i - 1].isEndOfFrame()) {
+      if (atoms[i - 1].isStartOfFrame()) {
```
ben-clayton: Please revert this too.

Qining: Done.

ben-clayton:
> Just curious, since GLES does not have any atoms with EndOfFrame annotation, can it cause problems in bench.go?

bench.go should not be using atoms directly - it should be ported over to the service and use events.
The whole of perfz needs a major overhaul.

PR #533, opened 2017-06-07
Vulkan:Postpone the messaging of framebuffer observation atom

Qining:
Depends on #532 

Gather the framebuffer observation atom data prior to the EndOfFrame atom:
vkQueuePresentKHR, but do not encode and message the atom. The atom will
be messaged after the vkQueuePresentKHR, so that on the replay side, let
the read_framebuffer.go grab the image to be presented first, then we
can compare the framebuffer images from trace and replay sides for the same
vkQueuePresentKHR atom.

### DIFF
--- a/gapii/cc/spy.cpp
+++ b/gapii/cc/spy.cpp
@@ -151,7 +151,8 @@ Spy::Spy()
   , mObserveFrameFrequency(0)
   , mObserveDrawFrequency(0)
   , mDisablePrecompiledShaders(false)
-  , mRecordGLErrorState(false) {
+  , mRecordGLErrorState(false)
+  , mPendingFramebufferObservation(nullptr) {
 
 
 #if TARGET_OS == GAPID_OS_ANDROID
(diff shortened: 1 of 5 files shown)


gapii/cc/spy.h
```diff
@@ -16,94 +16,113 @@
...
-    int mNumDraws;
-    int mNumDrawsPerFrame;
-    int mObserveFrameFrequency;
-    int mObserveDrawFrequency;
-    bool mDisablePrecompiledShaders;
-    bool mRecordGLErrorState;
-
-    std::unordered_map<ContextID, GLenum_Error> mFakeGlError;
-    std::unique_ptr<core::AsyncJob> mDeferStartJob;
+ public:
```
Qining: clang-format adds more changes then necessary.

gapii/cc/spy.cpp
```diff
@@ -459,13 +501,19 @@ void Spy::observeFramebuffer(uint8_t api) {
...
-        atom_pb::FramebufferObservation observation;
-        observation.set_originalwidth(w);
-        observation.set_originalheight(h);
-        observation.set_datawidth(downsampledW);
-        observation.set_dataheight(downsampledH);
-        observation.set_data(downsampledData.data(), downsampledData.size());
-        mEncoder->message(&observation);
+        // atom_pb::FramebufferObservation observation;
+        auto observation = std::unique_ptr<atom_pb::FramebufferObservation>(
+            new atom_pb::FramebufferObservation());
```
AWoloszyn: Do we not have a make_unique anywhere? (Should we add a make_unique)

gapis/resolve/events.go
```diff
@@ -59,18 +59,30 @@ func Events(ctx context.Context, p *path.Events) (*service.Events, error) {
...
+		if p.FirstInFrame && (f.IsBeginOfFrame() || i == 0) {
 			events = append(events, &service.Event{
 				Kind:    service.EventKind_FirstInFrame,
 				Command: p.Commands.Capture.Command(uint64(i)),
 			})
 		}
+		if p.FirstInFrame && f.IsEndOfFrame() {
+			events = append(events, &service.Event{
+				Kind:    service.EventKind_FirstInFrame,
+				Command: p.Commands.Capture.Command(uint64(i) + 1),
```
AWoloszyn: This may go out of bounds

gapii/cc/spy.cpp
Qining: Hmm, we really don't have make_unique. Should we add one? In this CL or a separate one? @ben-clayton

gapis/resolve/events.go
Qining: Oops. Fixed in fa66850

PR #535, opened 2017-06-08
Track bound GL objects by reference rather than by ID

dsrbecky:
This simplifies the code and will help us track lifetime of objects correctly.


dsrbecky:
Part of #544

PR #537, opened 2017-06-08
gapic: Restore the expansion state in the state tree.

### LINKED ISSUE #505: Clicking on new commands collapses all state.

pmuetschard:
Fixes #505

### DIFF
--- a/gapic/src/main/com/google/gapid/util/Paths.java
+++ b/gapic/src/main/com/google/gapid/util/Paths.java
@@ -127,6 +127,10 @@ public class Paths {
     return Path.Any.newBuilder().setStateTreeNode(node).build();
   }
 
+  public static Path.Any any(Path.State state) {
+    return Path.Any.newBuilder().setState(state).build();
+  }
+
   public static Path.Any commandTree(Path.Capture capture, FilteringContext context) {
     return Path.Any.newBuilder()
         .setCommandTree(context.commandTree(Path.CommandTree.newBuilder())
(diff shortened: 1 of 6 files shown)


gapic/src/main/com/google/gapid/views/StateView.java
```diff
@@ -330,6 +348,67 @@ public void onStateSelected(Path.Any path) {
...
+      @Override
+      protected TreePath[] onRpcThread(Result<List<TreePath>> result)
+          throws RpcException, ExecutionException {
+        List<TreePath> list = result.get();
+        return list.toArray(new TreePath[list.size()]);
+      }
+
+      @Override
+      protected void onUiThread(TreePath[] treePaths) {
+        setExpanded(treePaths, paths, retry);
```
ben-clayton: What's with retry?

gapic/src/main/com/google/gapid/util/Paths.java
```diff
@@ -400,6 +413,64 @@ public static int compare(Path.Command a, Path.Command b) {
     }
   }
 
+  public static Path.Any reparent(Path.Any path, Path.State newState) {
```
ben-clayton: There must be a cleaner / less fragile way of getting / setting parents. Unfortunately I don't know what that is right now.

gapic/src/main/com/google/gapid/views/StateView.java
pmuetschard: I think it's due to the dynamic loading and laziness of our tree model that the tree viewer sometimes gives up and doesn't actually expand a path. I detect that by asking if the path is expanded right after expanding it and if it's not expanded, schedule a retry. The retry count is just there to avoid an infinite loop if things go sideways. It does have the additional benefit that if, after playing with Tree (A) and selecting an atom and loading Tree (B), you select another atom quickly - loading Tree (C) - before Tree (B) had a chance to update its expansion state, it will eventually restore Tree (A)'s expansion state, rather than Tree (B)'s empty one.

PR #544, opened 2017-06-08
Refactor GL object lifetime tracking

### LINKED ISSUE #18: gles.api doesn't correctly distinguish between reserved and active names.
From [glIsTexture](https://www.khronos.org/opengles/sdk/docs/man32/html/glIsTexture.xhtml):
_"A name returned by glGenTextures, but not yet associated with a texture by calling glBindTexture, is not the name of a texture."_

We do not represent this in the API file.

Similar (Tri-State): [glIsQuery](https://www.khronos.org/opengles/sdk/docs/man32/html/glIsQuery.xhtml), [glIsRenderbuffer](https://www.khronos.org/opengles/sdk/docs/man32/html/glIsRenderbuffer.xhtml), [glIsVertexArray](https://www.khronos.org/opengles/sdk/docs/man32/html/glIsVertexArray.xhtml)

Different (Tri-State):
[glIsProgram](https://www.khronos.org/opengles/sdk/docs/man32/html/glIsProgram.xhtml):
"A program object marked for deletion with glDeleteProgram but still in use as part of current rendering state is still considered a program object and glIsProgram will return GL_TRUE"

[glIsShader](https://www.khronos.org/opengles/sdk/docs/man32/html/glIsShader.xhtml):
"A shader object marked for deletion with glDeleteShader but still attached to a program object is still considered a shader object and glIsShader will return GL_TRUE."

Two-State: [glIsSampler](https://www.khronos.org/opengles/sdk/docs/man32/html/glIsSampler.xhtml)

dsrbecky:
Fixes #18

### DIFF
--- a/gapis/gfxapi/gles/api/asynchronous_queries.api
+++ b/gapis/gfxapi/gles/api/asynchronous_queries.api
@@ -40,7 +40,16 @@ cmd void glBeginQuery(GLenum target, QueryId query) {
     }
   }
   ctx := GetContext()
-  ctx.ActiveQueries[target] = query
+  ctx.ActiveQueries[target] = GetOrCreateQuery(query)
+}
+
+sub ref!Query GetOrCreateQuery(QueryId id) {
+  ctx := GetContext()
+  if (id != 0) && (ctx.Objects.Queries[id] == null) {
+    ctx.Objects.Queries[id] = new!Query(ID: id)
+    ctx.Objects.GeneratedNames.Queries[id] = true
+  }
+  return ctx.Objects.Queries[id]
 }
 
 @if(Version.GLES30)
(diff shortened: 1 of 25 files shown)


gapis/gfxapi/gles/api/asynchronous_queries.api
```diff
@@ -85,13 +99,17 @@ cmd void glEndQuery(GLenum target) {
...
+sub void GenQueries(GLsizei count, QueryId* queries) {
+  if count < 0 { glErrorInvalidValue() }
   q := queries[0:count]
   ctx := GetContext()
   for i in (0 .. count) {
     id := as!QueryId(?)
-    // TODO: This should only reserve the name, but not create the object.
-    ctx.Objects.Queries[id] = new!Query(ID: id)
+    assert(id != 0)
+    ctx.Objects.GeneratedNames.Queries[id] = true
```
dsrbecky: All of the glGen* methods are pretty much identical, but I have not managed to figure out a nice way to avoid the copying pasting in the API language.
  cmd void glGenQueries(GLsizei count, QueryId* ids) {
    GenNames(ctx.Objects.GeneratedNames.Queries, count, ids)
  }
Maps are value objects, so this would not work, would it.  And there is no way to take a pointer in our language.  I miss C++

dsrbecky:
Does this mean you are also happy with #535

dsrbecky:
I added 2 more commits.

ben-clayton:
As discussed, I'd recommend the following pattern:

```
cmd ProgramId glCreateProgram() {
    ctx := GetContext()
    id := ?
    ctx.Objects.Shared.Programs[id] = new!Program(ID: id, RefCount: 1)
    return id
}

cmd void glDeleteProgram(ProgramId program) {
    if program != 0 {
        if !p.DeleteStatus {
            p.DeleteStatus = true
            AdjustProgramRefCount(p, -1)
        }
    }
}

sub void AdjustProgramRefCount(ref!Program program, s32 delta) {
    assert(s.RefCount >= 1)
    program.RefCount += delta
    if (program.RefCount == 0) {
        // delete
        ctx.Objects.Shared.Programs[program.ID] = null
    }
}
```

But I'm not going to force this on you if you strongly oppose.

dsrbecky:
I renamed the RefCount fields and decremented by 1 everywhere.... is that better?

PR #545, opened 2017-06-08
gapir: Add --wait-for-debugger flag.

ben-clayton:
If you have a crashing replay, this enables you to:

```bash
./do run gapit report --gapir-args="--wait-for-debugger" <trace>
```

Then in another console:
```bash
lldb -n gapir
```

For posix / linux I haven't found a reliable way to detect debugger attachment, so you need to set `gIsDebuggerAttached` to true to continue.

### DIFF
--- a/cmd/gapir/cc/main.cpp
+++ b/cmd/gapir/cc/main.cpp
@@ -23,6 +23,7 @@
 #include "gapir/cc/server_listener.h"
 
 #include "core/cc/connection.h"
+#include "core/cc/debugger.h"
 #include "core/cc/log.h"
 #include "core/cc/socket_connection.h"
 #include "core/cc/supported_abis.h"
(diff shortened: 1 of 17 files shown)


Qining:
I found this line might need some attention:
https://github.com/google/gapid/blob/master/core/cc/socket_connection.cpp#L193
Waiting for debugger to attach may cause a timeout?

BTW. The problem of gapir failure on my mahcine is found:
It seems that the clang 3.8 on my machine makes a mistake in compiling the code at "gles_renderer.cpp: line 189". The disassembled code shows there is a jump instruction jumps to a `ud2` instruction, which is undefined. This triggers a SIGILL signal and then gapir exit.

Switched to gcc 4.8.4, and the problem is fixed, there is no more `ud2` instruction in the generated code.

ben-clayton:
> I found this line might need some attention

I doubt [that line](https://github.com/google/gapid/blob/master/core/cc/socket_connection.cpp#L193) is the issue as the wait-for-debugger logic happens before that line is ever executed. There is a possible timeout on the gapis side however - I could look to extend it if the gapir-flag is present, but I didn't have any issues with the default when using this.

Qining:
LGTM. Wonder if others are happy with this.

ben-clayton:
I'd rather try to find a way that kills the global, or start a markdown document that describes this and the `--wait-for-debugger` flag.

PR #555, opened 2017-06-11
Merge the dead code elimination and dependency graph code for gles and vulkan

Qining:


Qining:
@ben-clayton @dsrbecky @AWoloszyn 
Dead code elimination transform is moved to: gapis/atom/transform
Dependency graph boilerplate code is moved to: gapis/resolve/dependencygraph
API specific dependency graph code is located in gapis/gfxapi/gles/dependency_graph_behaviour_provider.go and gapis/gfxapi/vulkan/dependency_graph_behaviour_provider.go

The test of dead_code_elimination in gapis/gfxapi/gles/dead_code_elimination_test.go is divided into two parts. First API agnostic liveness tree test, is moved to gapis/atom/transform. Another GLES test stays in gapis/gfxapi/gles/

Two interfaces are added:
1) HasDependencyGraphBehaviourProvider
For APIs that implement this interface, behaviour provider will be requested when building the dependency graph.
If an API does not have dependency graph behaviour provider, all atoms of that API will be labelled: KeepAlive.

2) BehaviourProvider
Provider should: For a given Atom, determine the AtomBehaviour of it and returns the AtomBehaviour. The function: GetBehaviourForAtom() is called when resolving the dependency graph.

Design decision:
Add a new Behaviour Provider interface, because building dependency graph for Vulkan does need some extra book keeping.

Not building the whole dependency graph in the provider, because the atom list may contain atoms from different APIs, so each atom should be processed by its corresponding API. This is why the provider just process one atom, instead of the whole atom list.

Some parts not so good, or can be improved:
GetBehaviourForAtom() does need the dependency graph to be passed in. Ideally, the provider should not need access to the graph at all. This is because the AtomBehaviour describes how 'stateKey' affects the 'stateAddress'. The translation between 'StateKey' and 'StateAddress' is owned by the dependency graph. One way to avoid the graph to be passed into provider is postpone the translation. However I'm not sure if it should be done in this CL.

gapis/resolve/dependencygraph/dependency_graph.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+}
+
+func (b *AtomBehaviour) Write(g *DependencyGraph, state StateKey) {
+	if state != nil {
+		b.Writes = append(b.Writes, g.addressMap.addressOf(state))
+	}
+}
+
+type HasDependencyGraphBehaviourProvider interface {
```
dsrbecky: I am not sure if the "Has" prefix is very go-like.
Personally, it seems fine to me.  @ben-clayton ?

gapis/resolve/dependencygraph/dependency_graph.go
```diff
@@ -0,0 +1,173 @@
...
+
+	s := c.NewState()
+	t0 := dependencyGraphBuildCounter.Start()
+	for i, a := range g.Atoms {
+		api := a.API()
+		if _, ok := behaviourProviders[api]; !ok {
+			if bp, ok := api.(HasDependencyGraphBehaviourProvider); ok {
+				behaviourProviders[api] = bp.GetDependencyGraphBehaviourProvider(ctx)
+			} else {
+				// API does not provides dependency information, always keep atoms for
```
dsrbecky: nit: "provide"

gapis/resolve/dependencygraph/dependency_graph.go
ben-clayton: No, it's not go like. Please drop the `Has`.

Qining:
Rebased on top of #613

gapis/config/config.go
```diff
@@ -20,7 +20,7 @@ const (
 	DebugReplay                = false
 	DebugReplayBuilder         = false
 	DisableDeadCodeElimination = false
-	DebugDeadCodeElimination   = false
+	DebugDeadCodeElimination   = true
```
Qining: Sorry, this is my fault. Fixed.

gapis/resolve/dependencygraph/dependency_graph.go
Qining: Dropped in 5aea7b6

gapis/resolve/dependencygraph/dependency_graph.go
Qining: Fixed in 5aea7b6

Qining:
@dsrbecky @ben-clayton @AWoloszyn 
efde581 is the commit rebased on top of current master with synchronization cl and updated gles dep graph. PTAL.

dsrbecky:
LGTM

Qining:
@dsrbecky 
Replay of GLES fixed at: https://github.com/Qining/gapid/blob/8202971d612951896754485e7a9f03c2b9888eab/gapis/resolve/dependencygraph/dependency_graph.go#L174

The problem is that, GLES atoms may depends on the side effects of Core API atoms. So just keep all the Core API atoms alive is not enough, we needs to mutate them.

PR #599, opened 2017-06-13
Misc bug fixes

ben-clayton:


cmd/do/gapic.go
```diff
@@ -147,14 +147,14 @@ func (e *gapicEnv) build(ctx context.Context, options BuildOptions) {
...
-	jargs := make([]string, 0)
+	jargs := []string{}
 	env := env(e.cfg)
 	switch e.platform {
 	case "linux":
 		env.Set("SWT_GTK3", "0").
 			Set("LIBOVERLAY_SCROLLBAR", "0")
 	case "osx":
-		jargs = append(args, "-XstartOnFirstThread")
+		jargs = []string{"-XstartOnFirstThread"}
```
pmuetschard: I would leave this as an append instead of overriding the jargs value.

ben-clayton: Done

PR #612, opened 2017-06-15
Allow gapit video to output individual commands.

AWoloszyn:
This is in contrast to replaying only at a frame level.

### DIFF
--- a/cmd/gapit/flags.go
+++ b/cmd/gapit/flags.go
@@ -110,11 +110,12 @@ type (
 			Width  int `help:"maximum video width"`
 			Height int `help:"maximum video height"`
 		}
-		Type   VideoType `help:"type of output to produce"`
-		Text   string    `help:"summary prefix (use '║' for aligned columns, '¶' for new line)"`
-		Frames struct {
+		Type     VideoType `help:"type of output to produce"`
+		Text     string    `help:"summary prefix (use '║' for aligned columns, '¶' for new line)"`
+		Commands bool      `help:"Treat every command as its own frame"`
+		Frames   struct {
 			Start int `help:"frame to start capture from"`
-			End   int `help:"frame to end capture on: -1 for last frame"`
+			Count int `help:"number of frames after Start to capture: -1 for all frames"`
 		}
 	}
 	DumpShadersFlags struct {
--- a/cmd/gapit/video.go
+++ b/cmd/gapit/video.go
@@ -57,7 +57,7 @@ func init() {
 	verb.Max.Width = 1920
 	verb.Max.Height = 1280
 	verb.FPS = 5
-	verb.Frames.End = allTheWay
+	verb.Frames.Count = allTheWay
 	app.AddVerb(&app.Verb{
 		Name:      "video",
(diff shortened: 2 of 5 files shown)


cmd/gapit/flags.go
```diff
@@ -110,11 +110,12 @@ type (
...
-		Type   VideoType `help:"type of output to produce"`
-		Text   string    `help:"summary prefix (use '║' for aligned columns, '¶' for new line)"`
-		Frames struct {
+		Type     VideoType `help:"type of output to produce"`
+		Text     string    `help:"summary prefix (use '║' for aligned columns, '¶' for new line)"`
+		Commands bool      `help:"Treat every command as its own frame"`
+		Frames   struct {
 			Start int `help:"frame to start capture from"`
-			End   int `help:"frame to end capture on: -1 for last frame"`
+			End   int `help:"number of frames after Start to capture: -1 for last frame"`
```
ben-clayton: Should this be renamed `Count`?

cmd/gapit/video.go
```diff
@@ -115,6 +123,7 @@ func (verb *videoVerb) regularVideoSource(
 				errors[i] = err
 				atomic.AddUint32(&errorCount, 1)
 			}
+			log.W(ctx, "Got frame: %d", i)
```
ben-clayton: Please either drop the severity or remove.

cmd/gapit/video.go
```diff
@@ -75,18 +75,26 @@ func (verb *videoVerb) regularVideoSource(
...
-	eofEvents, err := getEvents(ctx, client, &path.Events{
+	requestEvents := path.Events{
 		Commands:    capture.Commands(),
 		LastInFrame: true,
-	})
+	}
+
+	if verb.Commands {
+		requestEvents.LastInFrame = false
+		requestEvents.AllCommands = true
```
ben-clayton: What's with `AllCommands`?

gapis/service/path/path.proto
```diff
@@ -152,6 +152,7 @@ message Events {
     bool push_user_markers = 8;
     bool pop_user_markers = 9;
     bool framebuffer_observations = 10;
+    bool all_commands = 11;
```
ben-clayton: I don't understand the purpose of this event. You already have `service.Capture.num_commands` - what more does this tell you?

cmd/gapit/flags.go
AWoloszyn: Yes it really should be. I will rename

gapis/service/path/path.proto
AWoloszyn: This purpose of this is so that anything (code / interface etc) that relies on events, does not have to split their implementation into 2. 1 to handle events, and 1 to handle "everything".

Eg: Video.go:83
```go
	if verb.Commands {
		requestEvents.LastInFrame = false
		requestEvents.AllCommands = true
	}
```

Is the only change on the client that is necessary to handle either of these cases, rather than having to create a secondary loop that works on Commands instead of returned Events.

AWoloszyn:
This went in as a part of #613

PR #613, opened 2017-06-15
Handle blocking synchronization events in replay.

AWoloszyn:
We now analyze the synchronization events in the Vulkan replay. If we would end up in a case where the replay would block, re-write the command-buffers to avoid this case.

This also prevents these specific cases from being batched with incompatible cases.

### DIFF
--- a/gapis/gfxapi/api.go
+++ b/gapis/gfxapi/api.go
@@ -15,11 +15,13 @@
 package gfxapi
 
 import (
+	"context"
 	"fmt"
 
 	"github.com/google/gapid/core/data/id"
 	"github.com/google/gapid/core/image"
 	"github.com/google/gapid/gapil/constset"
+	"github.com/google/gapid/gapis/service/path"
 )
 
 // API is the common interface to a graphics programming api.
@@ -41,6 +43,10 @@ type API interface {
 
 	// Context returns the active context for the given state.
 	Context(*State) Context
+
+	// ResolveSynchroniation fills the given SynchronizationData with the
+	// set of blocking operations between commands in the given capture for this API.
+	ResolveSynchronization(ctx context.Context, d *SynchronizationData, c *path.Capture) error
 }
 
 // ID is an API identifier
--- a/gapis/gfxapi/core/core.go
+++ b/gapis/gfxapi/core/core.go
@@ -15,9 +15,14 @@
 package core
 
 import (
+	"context"
+
 	"github.com/google/gapid/core/image"
 	"github.com/google/gapid/gapis/gfxapi"
+	"github.com/google/gapid/gapis/service/path"
 )
+
+type CustomState struct{}
 
 // GetFramebufferAttachmentInfo returns the width, height and format of the specified framebuffer attachment.
 func (api) GetFramebufferAttachmentInfo(state *gfxapi.State, attachment gfxapi.FramebufferAttachment) (width, height uint32, format *image.Format, err error) {
@@ -26,5 +31,9 @@ func (api) GetFramebufferAttachmentInfo(state *gfxapi.State, attachment gfxapi.F
 
 // Context returns the active context for the given state.
 func (api) Context(*gfxapi.State) gfxapi.Context {
+	return nil
+}
+
+func (api) ResolveSynchronization(ctx context.Context, d *gfxapi.SynchronizationData, c *path.Capture) error {
 	return nil
 }
--- a/gapis/gfxapi/synchronization_data.go
+++ b/gapis/gfxapi/synchronization_data.go
@@ -1,0 +1,117 @@
+// Copyright (C) 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package gfxapi
+
+import "sort"
+
+// An index, typically an atom ID that defines the location of one side
+// synchronization dependency.
+type SynchronizationIndex uint64
+type SynchronizationIndices []SynchronizationIndex
+
+// The index of a subcommand within a command
+type SubcommandIndex []uint64
+
+// ExecutionRanges contains the information about a blocked command.
+// LastIndex is the final subcommand that exists within this command.
+// Ranges defines which future command will unblock the command in question, and
+// which subcommandis the last that will be run at that point.
+type ExecutionRanges struct {
+	LastIndex SubcommandIndex
+	Ranges    map[SynchronizationIndex]SubcommandIndex
+}
+
+// SynchronizationData contains a map of synchronization pairs.
+// The SynchronizationIndex is the command that will be blocked from
+// completion, and what subcommands will be made available by future commands.
+type SynchronizationData struct {
+	CommandRanges map[SynchronizationIndex]ExecutionRanges
+}
+
+// NewSynchronizationData creates a new clean SynchronizationData object
+func NewSynchronizationData() *SynchronizationData {
+	s := new(SynchronizationData)
+	s.CommandRanges = make(map[SynchronizationIndex]ExecutionRanges)
+	return s
+}
+
+// LessThan returns true if s comes before s2.
+func (s SubcommandIndex) LessThan(s2 SubcommandIndex) bool {
+	for i := range s {
+		if i > len(s2)-1 {
+			// This case is a bit weird, but
+			// {0} > {0, 1}, since {0} represents
+			// the ALL commands under 0.
+			return true
+		}
+		if s[i] < s2[i] {
+			return true
+		}
+		if s[i] > s2[i] {
+			return false
+		}
+	}
+	return false
+}
+
+// Decrement returns the subcommand that preceded this subcommand.
+// Decrement will decrement its way UP subcommand chains.
+// Eg: {0, 1}.Decrement() == {0, 0}
+//     {1, 0}.Decrement() == {0}
+//     {0}.Decrement() == {}
+func (s *SubcommandIndex) Decrement() {
+	for len(*s) > 0 {
+		if (*s)[len(*s)-1] > 0 {
+			(*s)[len(*s)-1]--
+			return
+		}
+		*s = (*s)[:len(*s)-1]
+	}
+}
+
+// Len returns the length of subcommand indices
+func (s SynchronizationIndices) Len() int {
+	return len(s)
+}
+
+// Swap swaps the 2 subcommands in the given slice
+func (s SynchronizationIndices) Swap(i, j int) {
+	s[i], s[j] = s[j], s[i]
+}
+
+// Less returns true if s[i] < s[j]
+func (s SynchronizationIndices) Less(i, j int) bool {
+	return s[i] < s[j]
+}
+
+// SortedKeys returns the keys of 's' in sorted order
+func (s SynchronizationData) SortedKeys() SynchronizationIndices {
+	v := make(SynchronizationIndices, 0, len(s.CommandRanges))
+	for k, _ := range s.CommandRanges {
+		v = append(v, k)
+	}
+	sort.Sort(v)
+	return v
+}
+
+// SortedKeys returns the keys of 'e' in sorted order
+func (e ExecutionRanges) SortedKeys() SynchronizationIndices {
+	v := make(SynchronizationIndices, 0, len(e.Ranges))
+	for k, _ := range e.Ranges {
+		v = append(v, k)
+	}
+	sort.Sort(v)
+	return v
+}
--- a/gapis/gfxapi/templates/api.go.tmpl
+++ b/gapis/gfxapi/templates/api.go.tmpl
@@ -1081,6 +1081,7 @@
     {{range $g := $.Globals}}
       {{Title $g.Name}} {{Template "Go.Type" $g}}
     {{end}}
+    CustomState `nobox:"true"`
   }
 
   func (g *State) Init() {
--- a/gapis/gfxapi/vulkan/command_buffer_rebuilder.go
+++ b/gapis/gfxapi/vulkan/command_buffer_rebuilder.go
@@ -1,0 +1,299 @@
+// Copyright (C) 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package vulkan
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/google/gapid/gapis/atom"
+	"github.com/google/gapid/gapis/gfxapi"
+	"github.com/google/gapid/gapis/memory"
+)
+
+func rebuildCmdBeginRenderPass(
+	ctx context.Context,
+	commandBuffer VkCommandBuffer,
+	s *gfxapi.State,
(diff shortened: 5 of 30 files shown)


gapis/gfxapi/api.go
```diff
@@ -41,6 +43,8 @@ type API interface {
 
 	// Context returns the active context for the given state.
 	Context(*State) Context
+
+	ResolveSynchronization(ctx context.Context, d *SynchronizationData, c *path.Capture) error
```
ben-clayton: godocs please

gapis/gfxapi/core/core.go
```diff
@@ -15,8 +15,12 @@
 package core
 
 import (
+	"context"
+
 	"github.com/google/gapid/core/image"
 	"github.com/google/gapid/gapis/gfxapi"
+
```
ben-clayton: nit: remove blank newline

gapis/gfxapi/synchronization_data.go
```diff
@@ -0,0 +1,88 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package gfxapi
+
+import "sort"
+
```
ben-clayton: Please document all public symbols.

gapis/gfxapi/synchronization_data.go
```diff
@@ -0,0 +1,88 @@
...
+			break
+		}
+		if s[i] < s2[i] {
+			return true
+		}
+	}
+	return false
+}
+
+func (s *SubcommandIndex) Decrement() {
```
ben-clayton: Please can we have some basic tests on this? Sort of thing that can hide bugs which takes days to trace.

gapis/gfxapi/vulkan/command_buffer_rebuilder.go
```diff
@@ -0,0 +1,300 @@
...
+
+package vulkan
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/google/gapid/gapis/atom"
+	"github.com/google/gapid/gapis/gfxapi"
+
```
ben-clayton: nit: please remove blank line.

gapis/resolve/synchronization_data.go
```diff
@@ -0,0 +1,40 @@
...
+package resolve
+
+import (
+	"context"
+
+	"github.com/google/gapid/gapis/capture"
+	"github.com/google/gapid/gapis/gfxapi"
+)
+
+// Resolve builds a ResolvedResources object for all of the resources
```
ben-clayton: Please update comment

gapis/gfxapi/vulkan/vulkan_terminator.go
```diff
@@ -0,0 +1,378 @@
...
+		}
+	}
+
+	return lrp, subpass
+}
+
+// RebuildCommandBuffer takes the commands from commandBuffer up to, and
+// including idx. It then appends any recreate* arguments to the end
+// of the command buffer.
+func RebuildCommandBuffer(ctx context.Context,
```
ben-clayton: Need to be public?

gapis/gfxapi/vulkan/vulkan_terminator.go
```diff
@@ -0,0 +1,378 @@
...
+		NewVkEndCommandBuffer(commandBufferId, VkResult_VK_SUCCESS))
+	cleanup = append(cleanup, func() {
+		allocateData.Free()
+		commandBufferData.Free()
+		beginInfoData.Free()
+	})
+	return VkCommandBuffer(commandBufferId), x, cleanup
+}
+
+func CutCommandBuffer(ctx context.Context, id atom.ID,
```
ben-clayton: Need to be public? Either way, please document.

gapis/resolve/synchronization_data.go
```diff
@@ -0,0 +1,40 @@
...
+// at the path r.After
+func (r *SynchronizationResolvable) Resolve(ctx context.Context) (interface{}, error) {
```
ben-clayton: If this is just used inside the vulkan package, then you can move the resolvable into there.
If you can do that, do you still need `ResolveSynchronization` on the API interface?

gapis/gfxapi/vulkan/vulkan.go
```diff
@@ -70,3 +73,47 @@ func (api) Mesh(ctx context.Context, o interface{}, p *path.Mesh) (*gfxapi.Mesh,
...
+
+	s.HandleSubcommand = func(a interface{}) {
+		rootIdx := gfxapi.SynchronizationIndex(i)
+		if k, ok := submissionMap[s.CurrentSubmission]; ok {
+			rootIdx = gfxapi.SynchronizationIndex(k)
+		} else {
+			submissionMap[s.CurrentSubmission] = i
+		}
+
+		if rng, ok := d.CommandRanges[rootIdx]; ok {
```
Qining: So keys of `d.CommandRanges` will always be the indices of vkQueueSubmit?

gapis/gfxapi/vulkan/vulkan_terminator.go
```diff
@@ -0,0 +1,378 @@
...
+			subpass = 0
+		case *RecreateCmdNextSubpassData:
+			subpass += 1
+		case *RecreateCmdEndRenderPassData:
+			lrp = nil
+			subpass = 0
+		}
+	}
+
+	walkCommands(c, queue.PendingCommands, f)
```
Qining: Why do we walk through the pending commands first, then the not pending ones? It seems that all the results from walking through the pending commands will be overwritten by the following walk through of the not-pending commands.

gapis/gfxapi/vulkan/vulkan_terminator.go
```diff
@@ -0,0 +1,378 @@
...
+			}
+		}
+	}
+}
+
+// resolveCurrentRenderPass walks all of the current and pending commands
+// to determine what renderpass we are in after the idx'th subcommand
+func resolveCurrentRenderPass(ctx context.Context, s *gfxapi.State, submit *VkQueueSubmit,
+	idx gfxapi.SubcommandIndex, lrp *RenderPassObject, subpass uint32) (*RenderPassObject, uint32) {
+	if len(idx) == 0 {
```
Qining: When can the length be zero?

gapis/gfxapi/vulkan/vulkan_terminator.go
AWoloszyn: Will remove.

gapis/gfxapi/vulkan/vulkan_terminator.go
AWoloszyn: Will remove.

gapis/resolve/synchronization_data.go
AWoloszyn: Just to leave a trace of our offline discussion here:
This  can be used to define synchronization between theads of execution. We can also use this data-structure when we want to start to show this information in the UI.

gapis/gfxapi/vulkan/vulkan_terminator.go
```diff
@@ -0,0 +1,378 @@
...
+		case *RecreateCmdEndRenderPassData:
+			lrp = nil
+			subpass = 0
+		}
+	}
+
+	walkCommands(c, queue.PendingCommands, f)
+	submitInfo := submit.PSubmits.Slice(uint64(0), uint64(submit.SubmitCount), l)
+	additional := 0
+	if len(idx) == 1 {
```
Qining: I think a cleaner way to express this "check length -> determine which level of loop to have extra 1 round -> return if done" is:
1) Just walk through the submit infos, command buffers, commands according to the idx value. For each level, if we do not hit the end of idx array, walk through the level.
2) Add a switch case statement, for different length of idx, add one more loop at corresponding level
3) Return lrp and subpass.

gapis/gfxapi/vulkan/command_buffer_rebuilder.go
```diff
@@ -0,0 +1,300 @@
...
+	"github.com/google/gapid/gapis/memory"
+)
+
+func recreateCmdBeginRenderPass(
```
Qining: Maybe we can call it rebuildCmdBeginRenderPass, so be consistent with rebuildCommandBuffer, and differentiate from MEC support?

gapis/gfxapi/vulkan/vulkan_terminator.go
```diff
@@ -0,0 +1,378 @@
...
+	submitCopy := NewVkQueueSubmit(submit.Queue, submit.SubmitCount, submit.PSubmits,
+		submit.Fence, submit.Result)
+	submitCopy.Extras().Add(a.Extras().All()...)
+
+	newCommandBuffers := make([]VkCommandBuffer, 1)
+	lastSubmit := uint64(0)
+	lastCommandBuffer := uint64(0)
+	if !skipAll {
+		lastSubmit = idx[0]
+		lastCommandBuffer = idx[1]
```
Qining: Is it possible to have a case where `len(idx) == 1`, like we cut precisely at boundary of submit infos?

gapis/gfxapi/vulkan/vulkan.api
```diff
@@ -3379,6 +3384,7 @@ cmd VkResult vkSetEvent(
     // If all pending events are signaled, so all removed from the pending
     // event list in the queue object, we should roll out the pending commands
     if len(q.PendingEvents) == 0 {
+      LastBoundQueue = Queues[queue]
```
Qining: Thanks to find this bug!

gapis/gfxapi/vulkan/vulkan_terminator.go
```diff
@@ -0,0 +1,378 @@
...
+		VkCommandBufferUsageFlags(VkCommandBufferUsageFlagBits_VK_COMMAND_BUFFER_USAGE_ONE_TIME_SUBMIT_BIT),
+		NewVkCommandBufferInheritanceInfoᶜᵖ(memory.Nullptr),
+	}
+
+	beginInfoData := atom.Must(atom.AllocData(ctx, s, beginInfo))
+	x = append(x,
+		NewVkBeginCommandBuffer(commandBufferId, beginInfoData.Ptr(), VkResult_VK_SUCCESS).AddRead(beginInfoData.Data()))
+
+	// DestroyResourcesAtEndOfFrame will handle this actually removing the
+	// command buffer. We have no way to handle WHEN this will be done
```
Qining: Stale comments?

gapis/gfxapi/vulkan/vulkan_terminator.go
```diff
@@ -0,0 +1,378 @@
...
+	cleanup := make([]func(), 0)
+
+	writeIdx := uint64(0)
+	if len(idx) > 0 {
+		writeIdx = idx[0]
+	}
+	additional := 0
+	if len(idx) == 1 {
+		additional = 1
+	}
```
Qining: Hmm, this 'using writeIdx to determine if we are in a command buffer, and need an +1 when copying the original commands' is quite confusing to read. We probably needs a better way.

And as a new command buffer is always needed, we can creates that new command buffer first and that creation does not need `writeIdx` and `additional`, then put this 'copying the original commands, if any' logic in one place closer to each other.

gapis/gfxapi/vulkan/vulkan_terminator.go
```diff
@@ -0,0 +1,378 @@
...
+	lrp, lsp = resolveCurrentRenderPass(ctx, s, submit, idx, lrp, lsp)
+
+	extraCommands := make([]interface{}, 0)
+	if lrp != nil {
+		lastPass := uint32(len(lrp.SubpassDescriptions))
+		for i := 0; uint32(i) < lastPass-lsp-1; i++ {
+			extraCommands = append(extraCommands, RecreateCmdNextSubpass{})
+		}
+		extraCommands = append(extraCommands, RecreateCmdEndRenderPassData{})
```
Qining: I guess the line 311 is not right, it should be RecreateCmdNextSubpassData

gapis/gfxapi/vulkan/vulkan_terminator.go
Qining: I agree creating a new command buffer in all cases. But if `len(idx) == 1`, isn't `idx[1]` will crash?

gapis/gfxapi/vulkan/vulkan_terminator.go
Qining: Probably we can add some comments about that, walking through the pending commands only matters when we stop at a subcommand.

Qining:
LGTM, @ben-clayton

gapis/gfxapi/synchronization_data.go
```diff
@@ -0,0 +1,117 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package gfxapi
+
+import "sort"
+
+// An index, typically an atom ID that defines the location of one side
```
ben-clayton: `typically an atom ID` when is it not?

gapis/gfxapi/synchronization_data.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// NewSynchronizationData creates a new clean SynchronizationData object
+func NewSynchronizationData() *SynchronizationData {
+	s := new(SynchronizationData)
+	s.CommandRanges = make(map[SynchronizationIndex]ExecutionRanges)
+	return s
+}
+
+// LessThan returns true if "s" comes before "s2".
```
ben-clayton: Übernit: no quotes when referencing parameters.

gapis/gfxapi/synchronization_data.go
```diff
@@ -0,0 +1,117 @@
...
+		if s[i] > s2[i] {
+			return false
+		}
+	}
+	return false
+}
+
+// Decrement returns the subcommand that preceded this subcommand.
+// Decrement will decrement its way UP subcommand chains.
+// Eg: {0, 1}.Decrement() == {0, 0}
```
ben-clayton: I don't expect you to change this, but this is a prime candidate for one of go's testable examples. [They're pretty neat.](https://blog.golang.org/examples)

gapis/gfxapi/templates/api.go.tmpl
```diff
@@ -1081,6 +1081,9 @@
     {{range $g := $.Globals}}
       {{Title $g.Name}} {{Template "Go.Type" $g}}
     {{end}}
+    SubcommandIndex gfxapi.SubcommandIndex `nobox:"true"`
```
ben-clayton: I'm sure I started a comment on this, but I can't find it, so I don't know what your response was.

Instead of polluting all the APIs with mostly Vulkan specific fields, how about we add a single `customState` field / type that can be populated with whatever each API wants?

gapis/gfxapi/synchronization_data.go
ben-clayton: And if it is always an atom ID, can't we just use that?

AWoloszyn: In this case it is absolutely an atom ID.
using atom.ID will however cause a circular reference in the imports :(

gapis/gfxapi/templates/api.go.tmpl
AWoloszyn: That sounds good, I don't remember reading such a comment, (although I may have just missed it). That is a lot nicer than what I have been doing, so I will do that.

gapis/gfxapi/synchronization_data.go
AWoloszyn: Cool, I will Keep that in mind for the future. I didn't even know those were a thing.

gapis/gfxapi/synchronization_data.go
AWoloszyn: Done

AWoloszyn:
I rebased on top of latest #612

gapis/gfxapi/templates/api.go.tmpl
```diff
@@ -1081,6 +1081,7 @@
     {{range $g := $.Globals}}
       {{Title $g.Name}} {{Template "Go.Type" $g}}
     {{end}}
+    CustomStateData CustomState `nobox:"true"`
```
ben-clayton: You could make this an anonymous field if it helps usage.

AWoloszyn: Done, thanks for the tip. I keep on forgetting those even exist in Go

AWoloszyn:
Squashed into 3 commits.

PR #622, opened 2017-06-19
gapic: Implement a shared future.

pmuetschard:
A shared future wraps a future with a ref count. Canceling a shared future will decrement the count and only cancel the delegate if the count reaches 0.

### DIFF
--- a/gapic/src/main/com/google/gapid/rpc/SharedFuture.java
+++ b/gapic/src/main/com/google/gapid/rpc/SharedFuture.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright (C) 2017 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


gapic/src/main/com/google/gapid/rpc/SharedFuture.java
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  public ListenableFuture<V> share() {
+    refCount.incrementAndGet();
+    return this;
+  }
+
+  @Override
+  public boolean cancel(boolean mayInterruptIfRunning) {
+    if (refCount.decrementAndGet() == 0) {
```
ben-clayton: Handle [multiple calls to `cancel`](http://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Future.html?is-external=true#cancel-boolean-).

gapic/src/main/com/google/gapid/rpc/SharedFuture.java
```diff
@@ -0,0 +1,80 @@
...
+  public SharedFuture(ListenableFuture<V> delegate) {
+    this.delegate = delegate;
+  }
+
+  public static <V> SharedFuture<V> shared(ListenableFuture<V> future) {
+    return new SharedFuture<V>(future);
+  }
+
+  public ListenableFuture<V> share() {
+    refCount.incrementAndGet();
```
ben-clayton: assert that the count was >= 1 prior to incrementing?

ben-clayton:
What's happening with this PR?

PR #626, opened 2017-06-19
Initialized Versions and Extensions

### LINKED ISSUE #564: State.Version shows true for everything?
```
GLES10: true
GLES20: true
GLES30: true
GLES31: true
GLES32: true
```

What's that mean? Candidate for hiding.

### LINKED ISSUE #609: Set ctx.Version and ctx.Extension values

dsrbecky:
Closes #609
Closes #564

### DIFF
--- a/gapis/gfxapi/gles/gles.api
+++ b/gapis/gfxapi/gles/gles.api
@@ -98,7 +98,18 @@ class Versions {
   bool GLES31                                          = true
   bool GLES32                                          = true
 }
+// TODO: This should be thread local variable
 Versions Version
+
+sub void InitVersions(ref!Context ctx) {
+  Version = Versions(
+    GLES10: VersionGreaterOrEqual(ctx, 1, 0),
+    GLES20: VersionGreaterOrEqual(ctx, 2, 0),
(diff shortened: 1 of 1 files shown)


dsrbecky:
Yeah, I am not exactly happy with it as-is either.

As a background, I wanted to have the @if syntax, and those globals felt out of that as something the gives decent @if expressions, is syntactically valid as-is, and has easy to understand semantic meaning.  I am more concerned with having some way to expresses version checks (what he had before did not work) than the implementation details.

As the comments suggest, my vision for future was to make those user hidden and thread local.  I am willing to spin it different ways though.

We can go a bit towards #if and make the expressions a bit more special so we could have say @if(GLES30 || GL_EXT_FOO) without any global variables per se.

I am still not entirely sold on the property idea.  In principle I like them, but in practice I need to see more value that we can get out of them to justify the cost.

We could also add back macros in a way: "define Version Contexts[CurrentThread].Info.SupportedVersions"  - which is basically like properties, except it is just AST level syntax sugar, so lesser cost?  Overall, I would like move the language to more C++ looking, but that is another debate.

Overall, I do not see this is most pressing issue at the moment.  If anything we should hide those from UI.

PR #630, opened 2017-06-20
gapic: Separate flags for gapis and gapir log level.

pmuetschard:


dsrbecky:
So is --log-level now effectively ignored?

Intuitively, I would expect --gapis-log-level and --gapir-log-level to default to --log-level if they have not been specified.

pmuetschard:
> So is --log-level now effectively ignored?

--log-level is the logging level for the UI

> Intuitively, I would expect --gapis-log-level and --gapir-log-level to default to --log-level if they have not been specified.

Done.

pmuetschard:
...ping...

PR #647, opened 2017-06-27
Robot: `./do upload` zips local build, Artifacts based on build host.

baldwinn860:
`./do upload` now produces a zip file of the local build when the expected file does not exist by using a new verb `./do robot upload package <filename>`

Artifacts are now based on the build host rather than merged from multiple hosts with possible overriding android toolsets.

Layouts now support OS X `.dylib` files and finding Gapis.

### DIFF
--- a/cmd/robot/build.go
+++ b/cmd/robot/build.go
@@ -166,19 +166,30 @@ func (v *buildUploadVerb) process(ctx context.Context, id string) error {
 	return nil
 }
 
-func zipFile(zip *zip.Writer, zipVirtualPath string, filePath file.Path) error {
+func zipFile(zipWriter *zip.Writer, zipVirtualPath string, filePath file.Path) error {
 	fileReader, err := os.Open(filePath.String())
 	if err != nil {
 		return err
 	}
 	defer fileReader.Close()
 
-	zipWriter, err := zip.Create(zipVirtualPath)
+	fileHeader, err := fileReader.Stat()
 	if err != nil {
 		return err
 	}
 
-	_, err = io.Copy(zipWriter, fileReader)
+	zipHeader, err := zip.FileInfoHeader(fileHeader)
+	if err != nil {
+		return err
+	}
+	zipHeader.Name = zipVirtualPath
+
+	zipFile, err := zipWriter.CreateHeader(zipHeader)
+	if err != nil {
+		return err
+	}
+
+	_, err = io.Copy(zipFile, fileReader)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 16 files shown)


test/robot/build/build.proto
```diff
@@ -63,32 +63,42 @@ message Information {
...
-  // The stash id of the gapid.apk
-  string gapid_apk = 7;
+  // The tools provided from a particular host.
+  HostToolSet host = 2;
+  // The tools provided for each supported android abi.
+  repeated AndroidToolSet android = 3;
+}
+
+// HostToolSet is the information for the tools extracted from a package for a particular host device.
+message HostToolSet {
```
ben-clayton: I'm guessing that we'll want to include a vulkan interceptor for host. @AWoloszyn - is this right?
If so, this can be added in a later PR.

test/robot/build/build.proto
```diff
@@ -63,32 +63,42 @@ message Information {
...
+  string virtual_swap_chain_lib = 4;
   // VirtualSwapChainLib is the stash id of the vulkan virtual-swap-chain loader
   // json file.
-  string virtual_swap_chain_json = 9;
+  string virtual_swap_chain_json = 5;
+}
+
+// AndroidToolSet is the information for the tools extracted from a package for a particular android ABI.
+message AndroidToolSet {
+  // The type of host system this tool set runs on.
```
ben-clayton: Is this right? Shouldn't it be `// The Android ABI this tool set runs on.` ?

test/robot/build/package.go
```diff
@@ -150,20 +150,29 @@ func (pkg *Package) mergeTool(ctx context.Context, tool *ToolSet) {
...
 			}
-			if tool.Gapis != "" {
-				t.Gapis = tool.Gapis
-			}
-			if tool.Gapit != "" {
-				t.Gapit = tool.Gapit
+			for _, android := range tool.Android {
+				if android.GapidApk != "" {
+					merged := false
+					for _, a := range t.Android {
```
ben-clayton: Maybe add a `func (*ToolSet) FindAndroidToolSet(*device.ABI) *AndroidToolSet` helper method that returns `nil` if the ABI cannot be found?

test/robot/build/package.go
```diff
@@ -173,11 +182,25 @@ func (pkg *Package) mergeTool(ctx context.Context, tool *ToolSet) {
 }
 
 // GetTools will return the toolset that match the abi, if there is one.
```
ben-clayton: `// GetHostTools will return the host toolset that matches the abi, if there is one.`

Would have been caught with VSCode and the go plugin.

test/robot/build/package.go
```diff
@@ -173,11 +182,25 @@ func (pkg *Package) mergeTool(ctx context.Context, tool *ToolSet) {
...
+func (p *Package) GetHostTools(abi *device.ABI) *ToolSet {
 	for _, t := range p.Tool {
 		if t.Abi.SameAs(abi) {
 			return t
 		}
 	}
 	return nil
 }
+
+func (p *Package) GetTargetTools(hostAbi *device.ABI, targetAbi *device.ABI) *AndroidToolSet {
```
ben-clayton: `// GetTargetTools will return the target toolset that matches the host and target abi, if there is one.`

Would have been caught with VSCode and the go plugin.

test/robot/build/package.go
```diff
@@ -173,11 +182,25 @@ func (pkg *Package) mergeTool(ctx context.Context, tool *ToolSet) {
...
+	hostTools := p.GetHostTools(hostAbi)
+	if hostTools == nil {
+		return nil
+	}
+
+	for _, a := range hostTools.Android {
```
ben-clayton: You could use the proposed `ToolSet.FindAndroidToolSet()` helper here too.

cmd/robot/build.go
```diff
@@ -166,19 +166,30 @@ func (v *buildUploadVerb) process(ctx context.Context, id string) error {
 	return nil
 }
 
-func zipFile(zip *zip.Writer, zipVirtualPath string, filePath file.Path) error {
+func zipFile(zipWriter *zip.Writer, zipVirtualPath string, filePath file.Path) error {
```
ben-clayton: For better or for worse, [go prefers short, often single letter variable names](https://talks.golang.org/2014/names.slide#7). `w` would be completely acceptable for `zipWriter` here. Most of the code in this PR uses names that are far more verbose than we'd normally have, but I'm not going to ask you to change it. Just a FYI for future reference.

test/robot/trace/client.go
```diff
@@ -175,7 +175,7 @@ func doTrace(ctx context.Context, action string, in *Input, store *stash.Client,
 	output, err := cmd.Call(ctx)
 	output = fmt.Sprintf("%s\n\n%s", cmd, output)
 	if err != nil {
-		return nil, log.Errf(ctx, err, "gapit call failed %v", output);
+		return nil, log.Errf(ctx, err, "gapit call failed %v", output)
```
ben-clayton: Would have been automatically fixed on save with VSCode and the go plugin.

PR #665, opened 2017-06-29
Add two missing draw calls

dsrbecky:
To stop API verification complaining


dsrbecky:
Yeap - it is glDrawArrays with 'count' automatically set

Issue #651, opened 2017-06-28
"do run gapic" failed


zhaijialong:
I built it on a windows10 pc. After successful build , I got the message when runing "do run gapic". 

lots of thanks for your help in advance.

```
E:\gpaid\src\src\github.com\google\gapid>do run gapic
[0/1] Install the project...
-- Install configuration: "Debug"
-- Up-to-date: e:/gapid/debug/pkg/./gapir.exe
-- Up-to-date: e:/gapid/debug/pkg/./gapis.exe
-- Up-to-date: e:/gapid/debug/pkg/./libwinpthread-1.dll
-- Up-to-date: e:/gapid/debug/pkg/./libstdc++-6.dll
-- Up-to-date: e:/gapid/debug/pkg/./libgcc_s_seh-1.dll
-- Up-to-date: e:/gapid/debug/pkg/./gapit.exe
-- Up-to-date: e:/gapid/debug/pkg/./libwinpthread-1.dll
-- Up-to-date: e:/gapid/debug/pkg/./libstdc++-6.dll
-- Up-to-date: e:/gapid/debug/pkg/./libgcc_s_seh-1.dll
-- Up-to-date: e:/gapid/debug/pkg/./lib/VirtualSwapchainLayer.json
-- Up-to-date: e:/gapid/debug/pkg/./lib/libVkLayer_VirtualSwapchain.dll
-- Up-to-date: e:/gapid/debug/pkg/./lib/libgapii.dll
-- Up-to-date: e:/gapid/debug/pkg/./lib/GraphicsSpyLayer.json
-- Up-to-date: e:/gapid/debug/pkg/./strings/en-us.stb
-- Up-to-date: e:/gapid/debug/pkg/./gapid-aarch64.apk
-- Up-to-date: e:/gapid/debug/pkg/./gapid-armeabi.apk
-- Up-to-date: e:/gapid/debug/pkg/./gapid-x86.apk
-- Up-to-date: e:/gapid/debug/pkg/./build.properties
I20170628-134659461[main][server.GapiPaths.checkForTools] Looking for GAPID in e:\gapid\debug\pkg -> e:\gapid\debug\pkg\gapis.exe
[...]

PR #669, opened 2017-06-29
deviceinfo: Fix character encoding for GetComputerName.

ben-clayton:
Should be UTF8.

Fixes: #651

### DIFF
--- a/core/os/device/deviceinfo/cc/windows/query.cpp
+++ b/core/os/device/deviceinfo/cc/windows/query.cpp
@@ -150,6 +150,7 @@ bool createContext(void* platform_data) {
     if (!GetComputerNameW(host_wide, &size)) {
         snprintf(gContext.mError, sizeof(gContext.mError),
                  "Couldn't get host name: %d", GetLastError());
+        return false;
     }
     WideCharToMultiByte(
         CP_UTF8,                    // CodePage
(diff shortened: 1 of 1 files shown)


core/os/device/deviceinfo/cc/windows/query.cpp
```diff
@@ -146,7 +146,21 @@ bool createContext(void* platform_data) {
     gContext.mNumCores = sysInfo.dwNumberOfProcessors;
 
     DWORD size = MAX_COMPUTERNAME_LENGTH + 1;
-    GetComputerNameA(gContext.mHostName, &size);
+    WCHAR host_wide[MAX_COMPUTERNAME_LENGTH + 1];
+    if (!GetComputerNameW(host_wide, &size)) {
+        snprintf(gContext.mError, sizeof(gContext.mError),
+                 "Couldn't get host name: %d", GetLastError());
+    }
```
AWoloszyn: I think we want an early-return here. If we host_wide is empty, or garbage, I would be worried about what WideCharToMultiByte would return.

ben-clayton: Agreed, that's a real mistake. Thanks for spotting.

ben-clayton:
Done

PR #672, opened 2017-06-29
Kokoro mac: Update path to CMake executable.

ben-clayton:
Hopefully fixes the mac buildbot.


ben-clayton:
Mac buildbot has got past the point it previously died...

PR #673, opened 2017-06-29
Record thread name on context creation

dsrbecky:


gapii/cc/spy.cpp
```diff
@@ -262,7 +263,15 @@ std::shared_ptr<StaticContextState> GlesSpy::GetEGLStaticContextState(CallObserv
 
     Constants constants;
     getContextConstants(constants);
-    std::shared_ptr<StaticContextState> out(new StaticContextState(constants));
+
+    std::string threadName;
+#if TARGET_OS == GAPID_OS_ANDROID
+    char buffer[256] = { 0 };
```
ben-clayton: Ubernit: 256 seems a little excessive, [this documentation](http://man7.org/linux/man-pages/man2/prctl.2.html) claims the string is no longer than 15 characters + null-terminator

dsrbecky: I noticed, but the size is not passed down to the getter function.
I am thinking better safe then sorry.

PR #677, opened 2017-06-30
Framebuffers

AWoloszyn:

PR #678, opened 2017-06-30
Screenshots

AWoloszyn:
This sits on top of #677 
Only 0583f1b759fb90478537330a20f5dd54224e56c7 is new

### DIFF
--- a/cmd/gapit/screenshot.go
+++ b/cmd/gapit/screenshot.go
@@ -1,0 +1,133 @@
+// Copyright (C) 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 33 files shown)


cmd/gapit/screenshot.go
```diff
@@ -0,0 +1,133 @@
...
+		"format": ii.Format,
+	}.Bind(ctx)
+	if ii.Width == 0 || ii.Height == 0 {
+		return nil, log.Err(ctx, nil, "Framebuffer has zero dimensions")
+	}
+	data, err = img.Convert(data, w, h, 1, ii.Format, img.RGBA_U8_NORM)
+	if err != nil {
+		return nil, log.Err(ctx, err, "Failed to convert frame to RGBA")
+	}
+	stride := w * 4
```
Qining: nit: Is there a better way to get the `4` instead of hard coding here?

ben-clayton: We're hard coding the conversion to RGBA 3 lines above, so I think 4 is justified to be hard-coded too.

Qining: OK, I was thinking if there is a way to get the `4` from the format. Anyway. LGTM.

PR #681, opened 2017-07-03
Move constructor functions to CommandBuilder

ben-clayton:
This CL should have no side-effects, as it creates an empty CommandBuilder struct that holds all the same constructor functions.

This is in preparation of adding threads to each atom and killing the core package.


gapis/gfxapi/vulkan/vulkan_terminator.go
```diff
@@ -295,14 +296,14 @@ func rebuildCommandBuffer(ctx context.Context,
 // index it would remain valid. This means closing any open
 // RenderPasses.
 func cutCommandBuffer(ctx context.Context, id atom.ID,
-	a atom.Atom, idx sync.SubcommandIndex, out transform.Writer) {
-	submit := a.(*VkQueueSubmit)
```
ben-clayton: @AWoloszyn - a bit of a cheeky unrelated change here, but this just seemed odd.

AWoloszyn: Thanks, looks good to me

PR #686, opened 2017-07-04
gapil: Add support for $Thread - the current thread.

ben-clayton:
This adds language support for $ globals, and registers the $Thread global. This isn't declared or used anywhere yet.


dsrbecky:
Why the dollar prefix?  Our build-in functions do not have magic names.

ben-clayton:
Why not? I think it makes it crystal clear that it's special and hasn't been declared elsewhere.

> Our build-in functions do not have magic names.

True, maybe they should.

PR #687, opened 2017-07-04
More thread-identifier plumbing

ben-clayton:
Still doesn't actually do anything, or is used by anything.


ben-clayton:
> I am suddenly worried if all the CommandBuilders have a thread set

That's exactly the reason I added `CommandBuffer{Thread: 0}` even though the change wasn't necessary. Searching the codebase for `CommandBuffer{}` is easy.

PR #688, opened 2017-07-04
Un-revert and fix "gapis/replay/executor: Optimize the gapir <-> gapis io."

ben-clayton:
The issue was that both gapir and gapis could fill their TPC/IP buffers,
waiting for each other to consume data. gapir was blocked sending resource
ids, and gapis was blocked sending the actual data.

### DIFF
--- a/gapis/replay/executor/executor.go
+++ b/gapis/replay/executor/executor.go
@@ -16,11 +16,13 @@
 package executor
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"fmt"
 	"io"
 
+	"github.com/google/gapid/core/data/binary"
 	"github.com/google/gapid/core/data/endian"
 	"github.com/google/gapid/core/data/id"
 	"github.com/google/gapid/core/log"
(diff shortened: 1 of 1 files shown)


gapis/replay/executor/executor.go
```diff
@@ -113,105 +113,111 @@ func (r executor) execute(ctx context.Context) error {
...
-		data := obj.([]byte)
-		n, err := r.connection.Write(data)
+		obj, err := db.Resolve(ctx, rID)
 		if err != nil {
-			return log.Errf(ctx, err, "Failed to send resource with id: %v", rid)
+			return log.Errf(ctx, err, "Failed to resolve resource with id: %v", rID)
 		}
 
-		totalReturnedSize += uint64(n)
+		n, _ := buf.Write(obj.([]byte))
```
dsrbecky: Another memcpy of the data?  How about just storing the objs and then writing them later?

ben-clayton: This was intentional as I wanted to reduce the number of syscalls on the sending side to a minimum. I did however add the `bufio.NewWriter(connection)` line after this, which should reduce the calls significantly (still more than once by just building a single buffer).
Done anyway.

PR #694, opened 2017-07-05
Update Vulkan README.md and add support for vkResetDescriptorPool

Qining:
Added forward declaration for internal classes, so that on CPP side we won't have undefined-type error.


gapis/gfxapi/templates/api_types.h.tmpl
```diff
@@ -64,6 +64,9 @@
   {{range $p := $.Pseudonyms}}
     {{Template "DeclareType" $p}}
   {{end}}
+  {{range $c := $.Classes}}
+    {{if GetAnnotation $c "internal"}} class {{$c.Name}}; {{end}}
+  {{end}}
```
Qining: @ben-clayton @AWoloszyn Not sure if this is acceptable, but this does solve the undefined-type problem when compiling CPP

gapis/gfxapi/vulkan/vulkan.api
```diff
@@ -9132,6 +9142,7 @@ enum RecordingState {
   @unused VkDescriptorPoolCreateFlags     Flags
   @unused u32                             MaxSets
   @unused map!(u32, VkDescriptorPoolSize) Sizes
+  @unused map!(VkDescriptorSet, ref!DescriptorSetObject) DescriptorSets
```
AWoloszyn: If you just make this a map of u32->VkDescriptorSet you will avoid the circular dependency. We probably don't need the refs to the Objects here.

Qining: Right, but this means we will need to traverse through all the descriptorsets in a pool when we want to free a specific descriptorset.

Qining:
@AWoloszyn Updated.

DescriptorPoolObject now has a map from VkDescriptorSet -> Ref!DescriptorSetObject
DescriptorSetObject now does not have Ref!DescriptorPoolObject, but a VkDescriptorPool.

gapis/gfxapi/vulkan/vulkan.api
```diff
@@ -4607,6 +4607,11 @@ cmd VkResult vkResetDescriptorPool(
     VkDevice                   device,
     VkDescriptorPool           descriptorPool,
     VkDescriptorPoolResetFlags flags) {
+  sets := DescriptorPools[descriptorPool].DescriptorSets
```
Qining: If I understand it correctly, this should create a copy, so we won't erasing items while iterating them.

AWoloszyn: I think so, although I would look at the generated golang code.
I have a feeling that in golang we don't actually copy the map, but reference it.

Qining: Interesting.
In CPP side, it is a copy, so no problem.
In Golang side, it is not a copy, however, it is **safe** to remove item while iterating!
https://stackoverflow.com/questions/23229975/is-it-safe-to-remove-selected-keys-from-golang-map-within-a-range-loop

ben-clayton: Can you make a note of the behaviors you're relying on?
If I ever get around to finishing the LLVM compiler, I'm going to have to be careful to either correct this code or keep the same behaviors.

Qining: OK, I will add comments there and rebase on top of current master.

Qining:
@ben-clayton @AWoloszyn Rebase DONE and updated. All comments resolved.

PR #701, opened 2017-07-06
gapis: Make sure host is first in GetDevices()

ben-clayton:
While Android may have preference for replay, the host should be considered the primary device.


dsrbecky:
Let's merge this?

PR #703, opened 2017-07-06
Fix glLinkProgram

dsrbecky:


dsrbecky:
PATL - I made it into an error if the location is not -1 at the same time.

It was gl_InstanceID

PR #710, opened 2017-07-07
Code Shufflin'

ben-clayton:
First stage in merging the `atom` and `gfxapi` packages.
Most of the key-players in `atom` have been moved.


gapis/gfxapi/test/gfxapi_test_import/imported.go
```diff
@@ -1,49 +0,0 @@
-// Copyright (C) 2017 Google Inc.
```
dsrbecky: This file seems pretty empty, so assume it was dropped on purpose during the move?

gapis/atom/list.go
```diff
@@ -1,59 +0,0 @@
...
-package atom
-
-import (
-	"context"
-
-	"github.com/google/gapid/gapis/api"
-)
-
-// List is a list of atoms.
-type List struct {
```
dsrbecky: Ohhh, that is kind of pity to see it go. I might add it back as part of GC optimization in distant future.
(provided it proves worthwhile, and provided the commands still live in Go memory by then)

dsrbecky:
Is it really for review?  The PR keeps changing under my feet.

gapis/atom/list.go
dsrbecky: Ohhh, that is kind of pity to see it go. I might add it back as part of GC optimization in distant future.
(provided it proves worthwhile, and provided the commands still live in Go memory by then)

ben-clayton: Why do you want it? It made use awkward, and hid the fact we were making whole-capture copies of the entire command stream everywhere.

gapis/gfxapi/test/gfxapi_test_import/imported.go
ben-clayton: Will delete in a followup CL.

ben-clayton: Sorry, just properly read the comment. Yeah, it was left over crap from the named imports and wasn't referenced.

PR #711, opened 2017-07-07
Check GLSL code during find-issues

dsrbecky:
Finish the TODO.


dsrbecky:
So, the code works, determines the code is not valid.
But gapit report seems to be ignoring all but the first issue, any idea why?

PR #713, opened 2017-07-07
Use GL_ARB_separate_shader_objects in compat GLSL

dsrbecky:
Try to enable the extension if it is available

### DIFF
--- a/gapis/shadertools/cc/libmanager.cpp
+++ b/gapis/shadertools/cc/libmanager.cpp
@@ -228,6 +228,15 @@ code_with_debug_info_t* convertGlsl(const char* input, size_t length, const opti
 
   std::string source = spirv2glsl(std::move(spirv_new));
 
+  // Add support for layout(location) as needed by separate shader objects
+  // TODO: Postprocessing source code like this is fairly hacky. Find a better way.
+  auto extensionPos = source.find("#ifdef GL_ARB_shading_language_420pack");
+  if (extensionPos != std::string::npos) {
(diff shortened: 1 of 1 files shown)


gapis/shadertools/cc/libmanager.cpp
```diff
@@ -228,6 +228,15 @@ code_with_debug_info_t* convertGlsl(const char* input, size_t length, const opti
 
   std::string source = spirv2glsl(std::move(spirv_new));
 
+  // Add support for layout(location) as needed by separate shader objects
+  // TODO: Postprocessing source code like this is fairly hacky. Find a better way.
+  auto extensionPos = source.find("#ifdef GL_ARB_shading_language_420pack");
```
ben-clayton: This is a bit gross. Maybe we should use a regex to at least handle more than 1 whitespace between the #ifdef and the extension name?

ben-clayton:
Did you mean to close this?

ben-clayton:
Oh, you emailed me. Shame.

dsrbecky:
Well, glslang check still warns us the the code we produce is invalid.
After a bit of spec reading, I think it might be right.
Although it does seem to fix it on my computer.... I think.... "gapit report" is moody today.
I would be still curious to know if this PR fixes it for you.

ben-clayton:
> I would be still curious to know if this PR fixes it for you.

Afraid not, I still get the same issues appearing.

dsrbecky:
Thought so, pity.
I will see if just removing the locations works.
Does gapit report show the shader compilation issues for you?

PR #716, opened 2017-07-10
Remove all layout(location = ...) qualifiers in compat.

dsrbecky:


gapis/shadertools/cc/spv_manager.cpp
```diff
@@ -112,6 +112,18 @@ void SpvManager::renameViewIndex() {
   }
 }
 
+// Remove all layout(location = ...) qualifiers.
+// For most use cases we should be binding/remapping locations regardless whether they
+// are assigned explicit location or compiler generated one.
+// TODO: Handle separate shader objects - we just hope the game uses consistent block names for now.
+bool SpvManager::removeLayoutLocations() {
```
AWoloszyn: It might be worth renaming any variables that have layout information in a deterministic way. This way you don't have to worry about block name consistency.

layout(location = 1) float foo;
can become
float _f_loc1;

gapis/shadertools/cc/spv_manager.cpp
```diff
@@ -112,6 +112,18 @@ void SpvManager::renameViewIndex() {
...
+  for (auto& it : module->annotations()) {
+    if (it.opcode() == SpvOpDecorate && it.GetSingleWordOperand(1) == SpvDecorationLocation) {
+      it.ToNop();
+    }
+  }
+}
```
ben-clayton: Missing a return statement.

gapis/shadertools/cc/spv_manager.cpp
dsrbecky: I started that, but actually finding if the variable is a block or not is non-trivial, and we do not want to rename non-blocks (for now, I wouldn't entirety rule it out)

The question, would such renaming scheme work 100% of the time?

gapis/shadertools/cc/spv_manager.cpp
dsrbecky: done (removed the return type)

PR #717, opened 2017-07-10
Add compat for GlTexBufferEXT

dsrbecky:


gapis/api/gles/compat.go
```diff
@@ -562,6 +562,13 @@ func compat(ctx context.Context, device *device.Instance) (transform.Transformer
 				log.E(ctx, "Error decompressing texture: %v", err)
 			}
 
+		case *GlTexBufferEXT:
+			if !version.IsES { // Strip suffix on desktop.
```
ben-clayton: Maybe `if version.AtLeastGL(3, 1) {` to be explicit? (I still don't think we've finalized the target GL version, but it would be nicer/safer to include it)

dsrbecky: I pretty sure whatever we settle on, it will be >=3.2 (the first core version)

ben-clayton: Still, there's no harm in having it, right?

dsrbecky: You know that the main reason that the recent APK was crashing was proliferation of boolean flags though compat, right?

dsrbecky: I have added it, but I worried about this exponentiation grow of different code paths we take

gapis/api/gles/compat.go
```diff
@@ -562,6 +562,13 @@ func compat(ctx context.Context, device *device.Instance) (transform.Transformer
 				log.E(ctx, "Error decompressing texture: %v", err)
 			}
 
+		case *GlTexBufferEXT:
+			if !version.IsES && version.AtLeast(3, 1) { // Strip suffix on desktop.
```
ben-clayton: `version.AtLeastGL(3, 1)` is short hand for this test.

gapis/api/gles/compat.go
ben-clayton: > exponentiation grow of different code paths

This doesn't add any more code paths than the original CL.

gapis/api/gles/compat.go
dsrbecky: Of course it does, in order to know how compat behaves, you need to know the exact version and list of all supported extensions on the target machine.
We might one day wonder why this does not work for someone until we think to ask if he is using 3.0 by any chance.

dsrbecky: Changed to short hand

ben-clayton: The reason I ask to have it, is that it's easier to see a single untransformed command that errors with "Attempted to unsupported function XXX", than try and trace through compat to figure out what things could have produced the call in the first place.
By annotating these transformations with the target GL version, it's clear what we're expecting from a replay device. At the moment, I can't just look at the code and figure out what the maximum GL version we need for all compat functions to work. Longer term, I'd hope that we'd correctly handle the GLES and GL paths, and also know ahead of time when said device won't replay - that's what [`GetReplayPriority`](https://github.com/google/gapid/blob/73a534ddb2aff14c1302f55bb296599ef154cb0e/gapis/replay/interfaces.go#L28-L34) should really be returning.

dsrbecky: Well, we had this discussion before - flexibility vs simplicity.

You are going NxM approach where ideally, every trace can be replayed on any target, but every user can essentially have unique replay stream depending on his version and extensions.

I prefer more of 1:1 approach, where given trace always produces the same compat replay stream for every user, but we might potentially limit some users with old GL versions.

In this case, my approach with some old GL version will yield "Unsupported function glTexBuffer" with related atom number.  Not ideal, but not hard to understand either.  I would be in favor to add something long lines of "requiresGL(3,1)" which would log error, and record the fact that this trace requires GL3.1 and why.  We might even do the compat translation before choosing target device, so we can pick the replay device intelligently.

With has has been submitted, the compat will produce unmodified stream for this case.  Which might or might not work.  Worst case scenario, the driver supports the extension, but the extension is different to the non-extension and things behave differently.  No error is reported, but other things break.  There is no indication that GL3.1 is needed either.  The user is now executing code path that we have never tested, and that can not reproduce locally.  If we get the version and extension list, and stare at compat.go long enough, we might figure it out.  More likely we will just start closing bugs with "Works for me"

PR #724, opened 2017-07-11
Use wglCreateContextAttribsARB to create the context of the highest supported profile

ben-clayton:
This doesn't seem necessary as the standard wglCreateContext() function seems to create a everything-compatible context. By being explicit we're being more in keeping with Linux, and we have a place to put context creation flags if we need them.

### DIFF
--- a/gapir/cc/windows/gles_renderer.cpp
+++ b/gapir/cc/windows/gles_renderer.cpp
@@ -18,6 +18,7 @@
 #include "gapir/cc/gles_renderer.h"
 
 #include "core/cc/gl/formats.h"
+#include "core/cc/gl/versions.h"
 #include "core/cc/log.h"
 
 #include <windows.h>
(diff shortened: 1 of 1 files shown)


gapir/cc/windows/gles_renderer.cpp
```diff
@@ -188,13 +189,59 @@ void GlesRendererImpl::setBackbuffer(Backbuffer backbuffer) {
...
+            wglMakeCurrent(mDeviceContext, nullptr);
+            wglDeleteContext(temp_context);
+        }
     }
 
-    if (mSharedContext != nullptr) {
-        wglShareLists(mSharedContext, mRenderingContext);
+    if (wglCreateContextAttribsARB != nullptr) {
+        for (auto gl_version : core::gl::sVersionSearchOrder) {
+            std::vector<int> attribs;
```
dsrbecky: FYI, You can just use the initializer { .... ,... } syntax even for vector.  I think.

ben-clayton: Ack. If we did add support for `KHR_context_flush_control`, we'd need it conditional based on support.

dsrbecky: True

PR #725, opened 2017-07-11
MSVC fixes

ben-clayton:
Requires building a project by hand, but you use this to debug and profile.


dsrbecky:
Requires building by hand?
The build server does not use msvc, right?

ben-clayton:
Yes, as the PR description says, you need to import all the source files manually to build at the moment.

PR #735, opened 2017-07-13
Fix Robot on Windows

baldwinn860:
A lot of subtle changes to filepaths and environment settings to get robot to work properly on windows. I'm sure this is not the end of these kinds of bugs. But here is a quick look at the grid after testing some apks on my machine (apk names redacted). 
image

### DIFF
--- a/cmd/embed/main.go
+++ b/cmd/embed/main.go
@@ -35,6 +35,7 @@ var (
 	pkg    string
 	output string
 	root   string
+	web    bool
 )
 
 const header = `
@@ -51,6 +52,7 @@ func main() {
 	flag.StringVar(&pkg, "package", "", "the package to use, defaults to the dir name")
 	flag.StringVar(&output, "out", "embed.go", "the file to generate")
 	flag.StringVar(&root, "root", "", "The root path for embedding")
+	flag.BoolVar(&web, "web", false, "If true, replace back slash separators in filepaths with forward slashes")
 	app.ShortHelp = "embed: A tool to embed files into go source."
 	app.Run(run)
 }
@@ -124,6 +126,9 @@ func run(ctx context.Context) error {
 			entry.filename = filename
 		}
 		entry.name = nameReplacer.Replace(filename)
+		if web {
+			entry.filename = strings.Replace(entry.filename, "\\", "/", -1)
+		}
 		entry.contents, err = ioutil.ReadFile(entry.path)
 		if err != nil {
 			return err
--- a/core/app/layout/layout.go
+++ b/core/app/layout/layout.go
@@ -140,7 +140,7 @@ func init() {
 	}
 }
 
-// DirToBinABI returns the ABI name that corresponds to the binLayoutDir given
+// DirToBinABI returns the ABI name that corresponds to the given binLayout Directory
 func DirToBinABI(dir string) string {
 	return dirToBinABI[dir]
 }
--- a/core/app/layout/layout.go
+++ b/core/app/layout/layout.go
@@ -56,9 +56,16 @@ type FileLayout interface {
 	Json(ctx context.Context, lib LibraryType) (file.Path, error)
 }
 
+func withExecutablePlatformSuffix(exe string) string {
+	if runtime.GOOS == "windows" {
+		return exe + ".exe"
+	}
+	return exe
+}
+
 var libTypeToName = map[LibraryType]string{
-	LibGraphicsSpy:      withPlatformSuffix("libgapii"),
-	LibVirtualSwapChain: withPlatformSuffix("libVkLayer_VirtualSwapchain"),
+	LibGraphicsSpy:      withLibraryPlatformSuffix("libgapii"),
+	LibVirtualSwapChain: withLibraryPlatformSuffix("libVkLayer_VirtualSwapchain"),
 }
 
 var libTypeToJson = map[LibraryType]string{
@@ -66,7 +73,7 @@ var libTypeToJson = map[LibraryType]string{
 	LibVirtualSwapChain: "VirtualSwapchainLayer.json",
 }
 
-func withPlatformSuffix(lib string) string {
+func withLibraryPlatformSuffix(lib string) string {
 	switch runtime.GOOS {
 	case "windows":
 		return lib + ".dll"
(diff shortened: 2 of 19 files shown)


test/robot/record/shelf.go
```diff
@@ -44,11 +45,18 @@ func NewShelf(ctx context.Context, shelfURL string) (Shelf, error) {
...
 	case "", "file":
 		if location.Host != "" {
-			return nil, log.Err(ctx, nil, "Host not supported for file shelves")
+			return nil, log.Errf(ctx, nil, "Host not supported for file shelves")
 		}
 		if location.Path == "" {
 			return nil, log.Err(ctx, nil, "Path must be specified for file shelves")
 		}
+		if runtime.GOOS == "windows" {
+			if l := len(location.Path); l > 2 && location.Path[0] == '/' && location.Path[2] == ':' {
```
ben-clayton: If this is really needed, you can use [strings.TrimPrefix](https://golang.org/pkg/strings/#TrimPrefix)

cmd/embed/main.go
```diff
@@ -51,6 +52,7 @@ func main() {
 	flag.StringVar(&pkg, "package", "", "the package to use, defaults to the dir name")
 	flag.StringVar(&output, "out", "embed.go", "the file to generate")
 	flag.StringVar(&root, "root", "", "The root path for embedding")
+	flag.BoolVar(&web, "web", false, "If true, replace back slash separators in filepaths with forward slashes")
```
ben-clayton: Would it not be simpler and easier to have embedded paths always use a forward slash and fix up [`loadReferenceImage`](https://github.com/google/gapid/blob/ded000ea2f5ba94cb92f944c69e91da84af23676/test/integration/replay/gles/image.go#L59) to use `path.Join` instead of `filepath.Join`?

core/app/layout/layout.go
```diff
@@ -124,6 +131,14 @@ var binABIToDir = map[string]string{
 	"x86":         "android-x86",
 }
 
+var DirToBinABI = func() map[string]string {
```
ben-clayton: It's more idiomatic to declare a var and initialize it with an `init()` function.
That said, why not offer it as a function that takes a dir parameter and returns the bin ABI?
Either way it a new public symbol - please document it.

test/robot/monitor/monitor.go
```diff
@@ -107,7 +108,9 @@ func Run(ctx context.Context, managers Managers, owner DataOwner, update func(ct
 			data.Gen.Update()
 			// Run the update
 			if update != nil {
-				update(ctx, &managers, data)
+				if err := update(ctx, &managers, data); len(err) != 0 {
```
ben-clayton: nit: `if errs :=`

test/robot/record/shelf.go
```diff
@@ -44,11 +45,18 @@ func NewShelf(ctx context.Context, shelfURL string) (Shelf, error) {
 	switch location.Scheme {
 	case "", "file":
 		if location.Host != "" {
-			return nil, log.Err(ctx, nil, "Host not supported for file shelves")
+			return nil, log.Errf(ctx, nil, "Host not supported for file shelves")
```
ben-clayton: Revert?

test/robot/stash/factory.go
```diff
@@ -49,6 +50,16 @@ func Dial(ctx context.Context, address string) (*Client, error) {
 			location.Scheme = "file"
 		}
 	}
+	if location.Scheme == "file" {
+		if runtime.GOOS == "windows" {
+			if l := len(location.Path); l > 2 && location.Path[0] == '/' && location.Path[2] == ':' {
+				// windows file urls have an extra slash before the volume label that needs to be remove
+				// see https://github.com/golang/go/commit/844b625ebcc7101e09fb87828a0e71db942a2416d
```
baldwinn860: accidentally moved the 'd' from "removed" on the previous line, which broke the link.

cmd/embed/main.go
baldwinn860: My thinking was that embed should keep the path separator of the host system in most cases since they represent the actual offsets of the files they come from on those systems. It's just embedded web pages that want to be able to look up by those paths from a web url. The other option I considered was having the static file webserver "fixup" the incoming url's path separators to whatever the host used (like the fix for the root index), which would probably be the simplest and easiest solution of all. Let me know which option sounds best.

core/app/layout/layout.go
```diff
@@ -131,13 +131,19 @@ var binABIToDir = map[string]string{
...
+	dirToBinABI = make(map[string]string)
 	for abi, dir := range binABIToDir {
-		result[dir] = abi
+		dirToBinABI[dir] = abi
 	}
-	return result
-}()
+}
+
+// DirToBinABI returns the ABI name that corresponds to the binLayoutDir given
```
ben-clayton: given...

core/os/file/path.go
```diff
@@ -82,7 +83,7 @@ func (p Path) System() string { return p.value }
 
 // URL returns the full absolute path using the / separator prefixed by the URL
 // file scheme.
-func (p Path) URL() string { return "file:///" + filepath.ToSlash(p.value) }
+func (p Path) URL() (*url.URL, error) { return url.Parse("file:///" + filepath.ToSlash(p.value)) }
```
ben-clayton: In this particular case, if an error is raised, it is almost certainly an issue with the `file` package, not the user of the package. So, I'd say it is acceptable to drop the `error` return value and internally panic if `url.Parse` returns an error.

test/robot/record/shelf.go
```diff
@@ -19,6 +19,8 @@ import (
 	"net/url"
 	"runtime"
```
ben-clayton: nit: kill the newline.

test/robot/record/shelf.go
```diff
@@ -36,29 +38,23 @@ type Shelf interface {
...
 		}
-		if runtime.GOOS == "windows" {
-			if l := len(location.Path); l > 2 && location.Path[0] == '/' && location.Path[2] == ':' {
-				// windows file urls have an extra slash before the volume label that needs to be removed
-				// see https://github.com/golang/go/commit/844b625ebcc7101e09fb87828a0e71db942a2416
-				location.Path = location.Path[1:]
-			}
+		if runtime.GOOS == "windows" && strings.IndexByte(shelfURL.Path, ':') == 2 {
+			// windows file urls have an extra slash before the volume label that needs to be removed
+			// see https://github.com/golang/go/commit/844b625ebcc7101e09fb87828a0e71db942a2416
```
ben-clayton: nit: [Might be better to link to this comment](https://github.com/golang/go/issues/6027#issuecomment-66083310).

test/robot/stash/factory.go
```diff
@@ -19,6 +19,8 @@ import (
 	"net/url"
 	"runtime"
```
ben-clayton: nit: remove blank newline.

PR #738, opened 2017-07-13
Shared contexts also need to share X-display

dsrbecky:
This concludes a long search for a cause of a driver crash.

Kudos to Ben for figuring this out!

### DIFF
--- a/gapir/cc/linux/gles_renderer.cpp
+++ b/gapir/cc/linux/gles_renderer.cpp
@@ -107,12 +107,17 @@ private:
     bool mQueriedExtensions;
 
     Display *mDisplay;
+    bool mOwnsDisplay; // True if we created mDisplay
     GLXContext mContext;
     GLXContext mSharedContext;
     GLXPbuffer mPbuffer;
     GLXFBConfig mFBConfig;
 };
 
+// NB: We keep a reference the shared GL context, so "parent" context
+//     must stay alive at least for the duration of this context.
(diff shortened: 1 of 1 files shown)


gapir/cc/linux/gles_renderer.cpp
```diff
@@ -121,9 +122,17 @@ GlesRendererImpl::GlesRendererImpl(GlesRendererImpl* shared_context)
...
         , mPbuffer(0) {
 
-    mDisplay = XOpenDisplay(nullptr);
-    if (mDisplay == nullptr) {
-        GAPID_FATAL("Unable to to open X display");
+    if (shared_context != nullptr) {
+        // Ensure that shared contexts also share X-display.
+        // Drivers are know to misbehave/crash without this.
+        mDisplay = shared_context->mDisplay;
+        mOwnsDisplay = false;
```
ben-clayton: So are shared contexts have an exclusive single owner? I thought they were shared - and would require refcounts?

dsrbecky: You are right in terms of GL/EGL, but we always create extra "root" context just for the purpose of context sharing (which has never any GL commands issued to it).  This reason for it is simpler debugging, and to deal with the fact sometimes recreate contexts just to resize buffers (maybe we should stop doing that).

ben-clayton: Maybe put that in a comment?

PR #743, opened 2017-07-14
Switch NDK compiler from gcc to clang

ben-clayton:
GCC is deprecated in the NDK.
Switch over to using the `android.toolchain.cmake` toolchain in the NDK instead of trying to roll our own.
The `--strip-all` linker flag reduces the final apks by about half.


ben-clayton:
GCC is deprecated in the NDK.
The `--strip-all` linker flag reduces the final apks by about half.

dsrbecky:
This seems to bump required cmake version to 3.6.0

ben-clayton:
What do you mean bump? The [BUILDING.md](https://github.com/google/gapid/blob/master/BUILDING.md) page asks you to download CMake 3.7.1. I believe it always has.

dsrbecky:
Time for me to update cmake :-)

Have you tried this on your windows box?

ben-clayton:
Yes, I wrote it on there.

core/os/device/deviceinfo/cc/android/query.cpp
```diff
@@ -103,6 +103,22 @@ void abiByName(const std::string name, device::ABI* abi) {
         memory_layout->set_endian(device::LittleEndian);
         abi->set_allocated_memorylayout(memory_layout);
         abi->set_architecture(device::X86);
+    } else if (name == "x86_64") {
```
dsrbecky: Is this intentionally included?

ben-clayton: Yes, the old CMake rules mixed up x86 and x86_64. I've untangled them.

dsrbecky:
It seems to make a clean build half a minute faster.  Yay!

On the other hand, I am bit confused as I tried to compare gcc vs clang sleep in the past and I did not observe any difference.

dsrbecky:
Does the new build still produce optimized code? (to make sure I am comparing apples to apples)

ben-clayton:
> On the other hand, I am bit confused as I tried to compare gcc vs clang sleep in the past and I did not observe any difference.

Might be the newer compilers. Might be that aapt has half the number of bytes to compress. Who knows.

> Does the new build still produce optimized code?

Yes
```cmake
# Debug and release flags.
list(APPEND ANDROID_COMPILER_FLAGS_DEBUG
	-O0)
if(ANDROID_ABI MATCHES "^armeabi")
	list(APPEND ANDROID_COMPILER_FLAGS_RELEASE
		-Os)
else()
	list(APPEND ANDROID_COMPILER_FLAGS_RELEASE
		-O2)
endif()
```

dsrbecky:
It builds on both my linux machines.
The windows build bot is not happy though for some reason.
specific_gfx_api.h.tmpl at <Write $filename>: error calling Write: open gles_gfx_api.h The requested operation cannot be performed on a file with a user-mapped section open.

ben-clayton:
Yeah, disturbingly it looks like `apic` is trying to run on the Android cmake builds. I'm guessing this has always been this way, but something about clang is stricter about file locks.
Will respin the build do see if it's intermittent.

dsrbecky:
I already tried to re-run the build several times - so fair it always fails.

ben-clayton:
The same issue each time?

dsrbecky:
Yes

ben-clayton:
Let's see if this helps...

ben-clayton:
w00t

ben-clayton:
Thanks for checking so thoroughly!

PR #744, opened 2017-07-17
Improve kokoro build speed

dsrbecky:
Kokoro will rsync the whole workspace back after build.
This is redundant for us so just delete most files at the end.

This saves about one third of the Windows build time.


dsrbecky:
PTAL

ben-clayton:
@pmuetschard ?

PR #745, opened 2017-07-17
gapir/cc/windows: Switch to using pbuffers.

ben-clayton:
Creating and adjusting windows for each renderer does not work well for multiple threads. Attempting to change the size of a window on a thread that wasn't the one that created it requires a message loop.

Instead implement the renderer similarly to linux, using offscreen pbuffers.


ben-clayton:
Merged as part of #749

PR #756, opened 2017-07-19
Add framebuffer sanity checks.

dsrbecky:


gapis/api/gles/api/framebuffer.api
```diff
@@ -457,6 +457,9 @@ cmd void glDeleteFramebuffers(GLsizei count, const FramebufferId* framebuffers)
   for i in (0 .. count) {
     id := f[i]
     if id != 0 {
+      // TODO: Create warning instead of error, and filer out
+      //       invalid ID in compat so replay does not see them.
+      assert(id in ctx.Objects.Framebuffers)
```
ben-clayton: Please can we make this an `abort()` as discussed?

dsrbecky:
PTAL, I made it a GL error instead of assert.  I does not follow the spec (added TODO for that), but it is good quick-fix.

PR #758, opened 2017-07-19
cmake: Add package compilation

dsrbecky:
This doubles the speed of incremental builds.

The rules used packages only to track dependencies (what needs
to be compiled) but never compiled them (which is works, since
go would compile whatever is needed when building tests/cmds).

Explicitly build and install packages from cmake now.
This should allow us to remove the big lock which
made go compilation essentially single-threaded.

Also make test dependencies more precise to make them parallel.


dsrbecky:
PTAL - build servers seem to be happy.
It was because of the dependencies for generated files indeed, thanks!

PR #760, opened 2017-07-19
Some changes to help make building with MSVC easier.

AWoloszyn:


AWoloszyn:
I would still like to get @ben-clayton  to look at this.

PR #762, opened 2017-07-19
Added the ability to trace if STDOUT is consumed.

AWoloszyn:


Qining:
LGTM

cmd/gapit/trace.go
```diff
@@ -111,7 +111,7 @@ func (verb *traceVerb) startLocalApp(ctx context.Context) (func(), error) {
 	// VK_DEVICE_LAYERS and LD_PRELOAD set to correct values to load the spy
 	// layer.
 	env := shell.CloneEnv()
-	cleanup, err := loader.SetupTrace(ctx, env)
+	cleanup, tempfile, err := loader.SetupTrace(ctx, env)
```
ben-clayton: portfile?

core/os/process/client.go
```diff
@@ -89,8 +131,10 @@ func Start(ctx context.Context, name string, opts StartOptions) (int, error) {
 	stdout := &portWatcher{
 		portChan: portChan,
 		stdout:   opts.Stdout,
+		portFile: opts.PortFile,
 	}
 
+	go stdout.waitForFile(ctx)
```
ben-clayton: Not the end of the world as gapit is the only thing that calls this, and it exits after tracing, but you could be orphaning go routines here. This could be solved with `task.WithCancel`, and defer-closing the context. That way when a port is found, the other method will give up. Just a nit though.

PR #774, opened 2017-07-24
Add dependency tracking for other clear variants

dsrbecky:


dsrbecky:
...another hunch related to the depth-fighting

ben-clayton:
> ...another hunch related to the depth-fighting

I thought the default was to always keep them alive?

Any noticeable performance gains?

dsrbecky:
They would be kept-alive by default, but the things the clears depend on might not be kept alive, which could possibly cause issues even for the clear.

PR #777, opened 2017-07-24
Increase default idle timeout

dsrbecky:


gapis/server/grpc.go
```diff
@@ -87,24 +87,25 @@ func (s *grpcServer) stopIfIdle(ctx context.Context, server *grpc.Server, keepAl
 	// no activity from the client in a contiguous N chunks of time.
 	// This avoids killing the server if the machine is suspended (where the
 	// client cannot send hearbeats, and the system clock jumps forward).
-	const maxMissCount = 3
-	idleTimeout /= maxMissCount
-	misses := 0
+	const waitTime = time.Second
+	var idleTime time.Duration = 0
```
ben-clayton: `= 0` is unnecessary and go vet will moan about it.

gapis/server/grpc.go
```diff
@@ -87,24 +87,25 @@ func (s *grpcServer) stopIfIdle(ctx context.Context, server *grpc.Server, keepAl
...
 			return
-		case <-time.After(idleTimeout):
-			misses++
-			if misses >= maxMissCount {
-				log.W(ctx, fmt.Sprintf("Stopping GAPIS server as it has been idle for more than %v (--idle-timeout)", idleTimeout))
-				time.Sleep(time.Second * 3) // Wait a little in the hope this message makes its way to the client(s).
+		case <-time.After(waitTime):
+			idleTime += waitTime
+			if idleTime >= idleTimeout {
+				log.E(ctx, fmt.Sprintf("Stopping GAPIS server as it has not received heartbeat (--idle-timeout %v)", idleTimeout))
```
ben-clayton: There's no such thing as a heartbeat in terms of the RPC interface. Any RPC calls will reset the timeout. How about:

`Stopping GAPIS as it has not received a request in %v`

gapis/server/grpc.go
```diff
@@ -87,24 +87,25 @@ func (s *grpcServer) stopIfIdle(ctx context.Context, server *grpc.Server, keepAl
 	// no activity from the client in a contiguous N chunks of time.
 	// This avoids killing the server if the machine is suspended (where the
 	// client cannot send hearbeats, and the system clock jumps forward).
-	const maxMissCount = 3
-	idleTimeout /= maxMissCount
-	misses := 0
+	const waitTime = time.Second
```
ben-clayton: This value should really be a fraction of `idleTimeout`.

gapis/server/grpc.go
```diff
@@ -87,24 +87,25 @@ func (s *grpcServer) stopIfIdle(ctx context.Context, server *grpc.Server, keepAl
...
-		case <-time.After(idleTimeout):
-			misses++
-			if misses >= maxMissCount {
-				log.W(ctx, fmt.Sprintf("Stopping GAPIS server as it has been idle for more than %v (--idle-timeout)", idleTimeout))
-				time.Sleep(time.Second * 3) // Wait a little in the hope this message makes its way to the client(s).
+		case <-time.After(waitTime):
+			idleTime += waitTime
+			if idleTime >= idleTimeout {
+				log.E(ctx, fmt.Sprintf("Stopping GAPIS server as it has not received heartbeat (--idle-timeout %v)", idleTimeout))
+				time.Sleep(time.Second * 4) // Wait a little in the hope this message makes its way to the client(s).
```
ben-clayton: Was 3 seconds really not enough?

gapis/server/grpc.go
```diff
@@ -87,24 +87,25 @@ func (s *grpcServer) stopIfIdle(ctx context.Context, server *grpc.Server, keepAl
...
 				return
+			} else if idleTime >= 4*time.Second {
+				log.W(ctx, fmt.Sprintf("GAPIS server has not received heartbeat for %v (--idle-timeout %v)", idleTime, idleTimeout))
```
ben-clayton: `GAPIS has not received a request in %v` ?

gapis/server/grpc.go
dsrbecky: You expect to defining factional second timeouts?

gapis/server/grpc.go
dsrbecky: Heartbeat seems clear.
The above message makes it sounds like we keep killing and restarting gapis.

gapis/server/grpc.go
dsrbecky: I just wanted to round up to power of two :-)
Nonetheless, better safe then sorry, gapis has a habit of dropping log messages, and it makes debugging tricky.

gapis/server/grpc.go
ben-clayton: Not necessarily, but there's also no point waking a process every second if the timeout is 1 day.

gapis/server/grpc.go
ben-clayton: It seems clear to you because you are looking at this code. It's less clear to someone who's using the public interface. I don't see how it makes any suggestion of restarting?

gapis/server/grpc.go
dsrbecky: There is also no harm, more importantly - it is not expected - so it is good to notify the user that things seem unresponsive.  And give heads up that it will be killed.

gapis/server/grpc.go
dsrbecky: As a user, "gapis has not received a request" seems perfectly reasonable if they have not clicked on anything.
"there was no heartbeat" communicates better that something died or become unresponsive.

gapis/server/grpc.go
AWoloszyn: I have seen timeouts for very large Vulkan traces.

gapis/server/grpc.go
AWoloszyn: To me a heartbeat is a specific message. Since we don't send specific heartbeats, maybe something else like "Stopping GAPIS because the client has not been communicating for %v"

dsrbecky: Isn't the regular 'ping' message that we regularly send precisely this specific message that you mean?

gapis/server/grpc.go
AWoloszyn: The main argument for keeping this a multiple of the timeout, is that if your timeout is long, you won't be constantly keeping your CPU from sleeping (or forcing your thread to get scheduled) and hurting battery-life / performance for unrelated processes.

Might be reasonable to say min(60s, maxMissCount/3) that way you will still get periodic reminders if you have a huge timeout, but won't wake the CPU often.

gapis/server/grpc.go
ben-clayton: If it were called heartbeat, sure, but it isn't.
We've wasted way more time arguing over this than it would have taken to have made the changes...

dsrbecky:
PTAL

gapis/server/grpc.go
```diff
@@ -87,24 +87,25 @@ func (s *grpcServer) stopIfIdle(ctx context.Context, server *grpc.Server, keepAl
...
 		case <-task.ShouldStop(ctx):
 			return
-		case <-time.After(idleTimeout):
-			misses++
-			if misses >= maxMissCount {
-				log.W(ctx, fmt.Sprintf("Stopping GAPIS server as it has been idle for more than %v (--idle-timeout)", idleTimeout))
+		case <-time.After(waitTime):
+			idleTime += waitTime
+			if idleTime >= idleTimeout {
+				log.E(ctx, fmt.Sprintf("Stopping GAPIS server: No communication with the client for %v (--idle-timeout %v)", idleTime, idleTimeout))
```
ben-clayton: Übernit (and completely my bad): GAPIS server is stuttering - the S stands for server.

PR #780, opened 2017-07-24
Fork the context for command mutation.

ben-clayton:
Removes the need to handle cancellation in the state mutators.

By creating a function that deals with general state mutation loops, we can put the ugliness of context-forking in a single place.

Related to the discussion in the thread of #479.

### DIFF
--- a/gapis/api/CMakeFiles.cmake
+++ b/gapis/api/CMakeFiles.cmake
@@ -25,6 +25,7 @@ set(files
     cmd_errors.go
     cmd_extras.go
     cmd_flags.go
+    cmd_foreach.go
     cmd_id_group_test.go
     cmd_id_group.go
     cmd_id_range.go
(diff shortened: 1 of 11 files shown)


dsrbecky:
I have been arguing for something like this for months, so I am obviously in favor :-)

I would have preferred if instead of re-creating the context, we would just tightly control which places can error-out because of cancellation.  Ideally just this ForeachCmd would be the termination point, and everything else would keep executing even in cancelled context.   However, I will settle for this PR as mid-ground.

There is ForeachCmd missing in dead code elimination, which means this PR does not work for GLES.

ben-clayton:
> There is ForeachCmd missing in dead code elimination

Link please.

dsrbecky:
https://github.com/google/gapid/blob/master/gapis/api/transform/dead_code_elimination.go#L80

ben-clayton:
This is only called from replay though, right? Which is in a `ForeachCmd` loop.

ben-clayton:
Okay, updated to include the DCE's flush.

PR #781, opened 2017-07-24
UI for tracing local apps

### LINKED ISSUE #428: UI Dialog for tracing Local Apps
It would be good to add a UI dialog for tracing Local applications.

This is something that works for Vulkan, and would help put us on parity with other applications.
The invocation is:
```
gapit trace -local-app "path/to/exe" -local-args "arguments to pass to exe" 
```
It would also be helpful to have an option to set the Working Directory, as many applications expect specific working directories.

pmuetschard:
Fixes #428

### DIFF
--- a/gapic/src/main/com/google/gapid/models/Settings.java
+++ b/gapic/src/main/com/google/gapid/models/Settings.java
@@ -68,6 +68,7 @@ public class Settings {
   public boolean traceDisablePcs = false;
   public String traceExecutable = "";
   public String traceArgs = "";
+  public String traceCwd = "";
   public boolean traceMidExecution = false;
   public int traceFrameCount = 0;
   public boolean skipWelcomeScreen = false;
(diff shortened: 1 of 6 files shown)


AWoloszyn:
One note is that the Mid-Execution works for both Desktop and Android. (It would be nice to have on both).

Some Desktop applications have a requirement for Working Directory (i.e. will fail if run from the wrong working-directory), so that would be good as well.

Lastly there is another argument for gapit trace that might be nice for everyone which is -capture-frames=N which will stop tracing after N frames. It is MORE useful with mid-execution so you don't have to be quick and trace/stop if you only want very few frames.

pmuetschard:
> One note is that the Mid-Execution works for both Desktop and Android. (It would be nice to have on both).

Done.

> Some Desktop applications have a requirement for Working Directory (i.e. will fail if run from the wrong working-directory), so that would be good as well.

I didn't see an option for gapit to specify the CWD. What flag am I to pass?

> Lastly there is another argument for gapit trace that might be nice for everyone which is -capture-frames=N which will stop tracing after N frames. It is MORE useful with mid-execution so you don't have to be quick and trace/stop if you only want very few frames.

Done.

AWoloszyn:
Awesome!
As for CWD, there is no flag to gapit, We should be able to just spawn the gapit process from that CWD. 

But looking at the code we already set the working directory for gapit with
```java
// Use the base directory as the working directory for the server.
pb.directory(GapiPaths.base());
```
I can add a command-line argument to gapit to take a working directory argument.

AWoloszyn:
#801 has a change that adds `-local-workingdir ` as an argument to GAPIT.

PR #807, opened 2017-07-26
Add SubCmdRoots for empty command-lists.

AWoloszyn:
This removes weird gaps in the UI that were causing indexing issues.

### DIFF
--- a/gapis/api/cmd_id_group.go
+++ b/gapis/api/cmd_id_group.go
@@ -365,7 +365,7 @@ func (g *CmdIDGroup) AddRoot(rootidx []uint64) *SubCmdRoot {
 		return g.Spans[i].(*SubCmdRoot)
 	} else {
 		if c != 1 {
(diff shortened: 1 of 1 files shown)


gapis/api/cmd_id_group.go
```diff
@@ -365,7 +365,7 @@ func (g *CmdIDGroup) AddRoot(rootidx []uint64) *SubCmdRoot {
 		return g.Spans[i].(*SubCmdRoot)
 	} else {
 		if c != 1 {
-			panic("This should not happen, a single command cannot span 2 groups")
+			panic("This should not happen, a single command cannot span %v groups", c)
```
ben-clayton: This does not compile:

`gapid/gapis/api/cmd_id_group.go:368: too many arguments to panic: panic("This should not happen, a single command cannot span %v groups", c)`

AWoloszyn:
Thanks for catching that. Fixed now.

PR #809, opened 2017-07-26
Fix object view in robot webserver and uploads.

baldwinn860:
When the format of build ToolSets changed it broke a part of the data
gathering functionality in the webserver. There was already a TODO to
change it to use the proto directly rather than using the json and
assuming the format doesn't change so I took it as mine for the future.

The master server was only listening from localhost for uploading
subjects and artifacts. Now it accepts from any valid connection, so I
can upload from another computer.

### DIFF
--- a/cmd/robot/master.go
+++ b/cmd/robot/master.go
@@ -88,7 +88,7 @@ func (v *masterVerb) Run(ctx context.Context, flags flag.FlagSet) error {
 	}
 	tempDir := file.Abs(tempName)
 	restart := false
-	serverAddress := fmt.Sprintf("localhost:%v", v.Port)
+	serverAddress := fmt.Sprintf(":%v", v.Port)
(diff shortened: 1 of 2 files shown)


cmd/robot/master.go
```diff
@@ -88,7 +88,7 @@ func (v *masterVerb) Run(ctx context.Context, flags flag.FlagSet) error {
 	}
 	tempDir := file.Abs(tempName)
 	restart := false
-	serverAddress := fmt.Sprintf("localhost:%v", v.Port)
+	serverAddress := fmt.Sprintf(":%v", v.Port)
```
ben-clayton: This change seems unrelated?

baldwinn860: This fixes remote uploads to master. It could be split to another CL if needed, but it is mentioned in the CL description.

PR #816, opened 2017-07-27
Just a minor fix to avoid crash

Qining:


gapis/api/vulkan/draw_call_mesh.go
```diff
@@ -293,7 +293,7 @@ func getVerticesData(ctx context.Context, s *api.State, thread uint64,
 	fullSize := uint64(vertexCount-1)*stride + perVertexSize
 
 	offset := uint64(attribute.Offset) + (uint64(firstVertex) * stride)
-	if offset > vertexSlice.count || offset+fullSize > vertexSlice.count {
+	if offset >= vertexSlice.count || offset+fullSize > vertexSlice.count {
```
ben-clayton: @AWoloszyn, @Qining:
Given that `offset` and `fullSize` are both unsigned, do you need the LHS of the `||`.
Secondly, don't you have exactly the same off-by-one bug on the RHS?

Qining: The `=` in the LHS should only matter if the `fullSize` is `0`. I don't think it can be `0`, but adding `=` just make it more correct. (Although I did observed a crash once which escaped the original guard, and this quite odd to me too)

I think the range to be used for slice is [offset, offset + fullSize). So I think RHS shouldn't have '='.

ben-clayton: Okay, makes sense.

PR #820, opened 2017-07-27
Win installer fixes

pmuetschard:
- Fix MSI installer to actually respect user picked location.
- Make the MSI installer create a start menu shortcut.


AWoloszyn:
Looks right to me, we should probably get this in.

PR #824, opened 2017-07-28
gapidapk: Update to latest gradle and android tools build.

ben-clayton:


ben-clayton:
Not merging, as this doesn't work with Kokoro yet.

PR #836, opened 2017-07-31
Fix resource destruction

dsrbecky:


gapis/api/gles/replay.go
```diff
@@ -348,10 +348,12 @@ func (t *destroyResourcesAtEOS) Flush(ctx context.Context, out transform.Writer)
 			tmp := s.AllocDataOrPanic(ctx, queries)
 			cmds = append(cmds, cb.GlDeleteQueries(GLsizei(len(queries)), tmp.Ptr()).AddRead(tmp.Data()))
 		}
+
+		cmds = append(cmds, cb.EglMakeCurrent(memory.Nullptr, memory.Nullptr, memory.Nullptr, memory.Nullptr, 1))
```
ben-clayton: Please comment that this is needed inside the loop as potentially shared objects are deleted and removed from the maps on the `MutateAndWrite` line. This wasn't obvious to me (hence the bug) and I wouldn't want myself or anyone else to break this again.

dsrbecky: done

PR #838, opened 2017-07-31
Make it possible to get log messages from trace driver

dsrbecky:


gapii/cc/spy.cpp
```diff
@@ -238,10 +247,61 @@ EGLBoolean Spy::eglInitialize(CallObserver* observer, EGLDisplay dpy, EGLint* ma
...
+    std::map<EGLint, EGLint> attribs;
+    while(attrib_list != nullptr && *attrib_list != EGL_NONE) {
+        EGLint key = *(attrib_list++);
+        EGLint val = *(attrib_list++);
+        attribs[key] = val;
+    }
+
+    // Modify attrib list
+    if (mRecordGLErrorState) {
+        attribs[EGL_CONTEXT_FLAGS_KHR] |= EGL_CONTEXT_OPENGL_DEBUG_BIT_KHR;
```
ben-clayton: Have you checked what happens if you set this flag when the driver doesn't support the extension?
WGL can error on context creation in this case - I don't know if this is the same for EGL.

dsrbecky: The extension and the debug bit are unrelated.
(other than the fact context without the debug bit may opt to generate no log callbacks at all)

PR #842, opened 2017-07-31
A bunch of summer cleanup + javadoc for gapic

pmuetschard:
- Fix a bunch of warnings.
- Add a jdoc verb to do to generate gapic's javadoc.
- Use PingRequest.getDefaultInstance()
- Refactor the thread coloring in the tree view.
- Simplify CommandFilter's toString.
- Be more consistent with inner class imports.
- MemoryProtos.java -> Memory.java
- Fix some proto import warnings.
- Make some of the code generation quiet for a less noisy build.
- Max line length.


cmd/do/do.go
```diff
@@ -321,6 +327,12 @@ func (verb *goVerb) Run(ctx context.Context, flags flag.FlagSet) error {
 	return nil
 }
 
+func (verb *jdocVerb) Run(ctx context.Context, flags flag.FlagSet) error {
```
ben-clayton: It might also be interesting to see if there's any free javadoc hosting sites like [godoc](https://godoc.org/github.com/google/gapid). I found [javadoc.io](http://www.javadoc.io/), but that requires the code to be in a Central Maven Repository.

gapic/src/main/com/google/gapid/util/Colors.java
```diff
@@ -42,6 +44,18 @@ public static RGB rgb(double r, double g, double b) {
     return new RGB((int)(r * 255), (int)(g * 255), (int)(b * 255));
   }
 
+  public static RGB getRandomColor(int index) {
+    float hue = (float)Math.toDegrees(index * GOLDEN_ANGLE) % 360;
+    return new RGB(hue, 1f, 1f);
```
ben-clayton: I actually spent quite a lot of time trying to get colors that were non-offensive.
The code I wrote is not a simple HSV to RGB function. Unless there's a good justification, I'd prefer it if you kept the original code.

gapic/src/main/com/google/gapid/util/Colors.java
```diff
@@ -42,6 +44,18 @@ public static RGB rgb(double r, double g, double b) {
     return new RGB((int)(r * 255), (int)(g * 255), (int)(b * 255));
   }
 
+  public static RGB getRandomColor(int index) {
```
ben-clayton: The colors were not supposed to be random. The code attempts to generate complementary colors for close indices, while also avoiding duplicates as the number of indices grow.

gapic/src/main/com/google/gapid/util/Colors.java
pmuetschard: Made a small adjustment to the color generator.

The only justification is that it's simpler and easier to understand. What does normalizing a color mean (especially without weighting the channels differently)?

https://screenshot.googleplex.com/oc8HLpgFAre - As a comparison of the first 10 colors generated by each algorithm.

gapic/src/main/com/google/gapid/util/Colors.java
pmuetschard: I mean "random" as in pseudo-random. I'm all ears for a better name.

cmd/do/do.go
pmuetschard: I haven't found anything good that would do it automatically. Probably because invoking javadoc is not as straight forward: it needs the .class files and the full classpath.

We could have a javadoc branch, where we check it in for each release, but I'm not sure it's worth it, because we're not really providing a library, where people will want to ready the javadoc.

gapic/src/main/com/google/gapid/util/Colors.java
ben-clayton: www.shadertoy.com
```glsl
const float GOLDEN_ANGLE = 2.39996322972865332f;
const float COLOR_INTENSITY = 0.2;
const float PI = 3.141592654;
vec3 hsv2rgb(vec3 c) {
    vec4 K = vec4(1.0, 2.0 / 3.0, 1.0 / 3.0, 3.0);
    vec3 p = abs(fract(c.xxx + K.xyz) * 6.0 - K.www);
    return c.z * mix(K.xxx, clamp(p - K.xxx, 0.0, 1.0), c.y);
}
void mainImage(out vec4 fragColor, in vec2 fragCoord) {
	ivec2 cell = ivec2(fragCoord.xy / vec2(64, 64));
	vec2 uv = vec2(cell);
	float angle = uv.x * GOLDEN_ANGLE;
	vec3 rgb;
	if (fragCoord.y > (iResolution.y/2.0)) {
		rgb = sin(angle + PI*vec3(0.0, 1.0, 2.0)/3.0);
		rgb += 1.0;
		rgb = normalize(rgb);
	} else {
		float hue = angle / (2.0*PI);
		vec3 hsv = vec3(hue, 1.0, 0.8);
		rgb = hsv2rgb(hsv);
	}
	fragColor = vec4(mix(vec3(1.0), rgb, COLOR_INTENSITY), 1.0);
}
```

Top: original, Bottom: new.

Personally, I feel that the top has a more neutral, pastel palette.

ben-clayton: Change the HSV params to `(hue, 0.8, 0.8)` and they're much closer. Maybe I'm now just mentally biased now, but I still prefer the original palette. However, I'd probably fail a blind test. :wink:

pmuetschard: I've adjusted both values now. I can't tell the difference between old and new.

gapic/src/main/com/google/gapid/util/Colors.java
ben-clayton: Meh, fine.

cmd/do/do.go
ben-clayton: Ack.

PR #844, opened 2017-08-01
Fix windows renderer

ben-clayton:


core/cc/linux/debugger.cpp
```diff
@@ -20,6 +20,10 @@ static volatile bool gIsDebuggerAttached = false;
 
 namespace core {
 
+void Debugger::waitForAttach() {
+    while (!isAttached()) {}
+}
+
 bool Debugger::isAttached() {
```
dsrbecky: Nit: I would go one step further and inline the isAttached

ben-clayton:
Grrrrr. This fixes the two linked bugs, but does hit the `ERROR_INVALID_PIXEL_FORMAT` issue. Back to the drawing board...

gapic/src/main/com/google/gapid/server/GapisProcess.java
```diff
@@ -88,9 +93,14 @@ protected Exception prepare(ProcessBuilder pb) {
       args.add(new File(logDir, "gapis.log").getAbsolutePath());
       args.add("-log-level");
       args.add(logLevel.get().gapisLevel);
+
+      String gapir = "--log " + new File(logDir, "gapir.log").getAbsolutePath() +
+          " --log-level " + logLevel.get().gapirLevel;
+      if (gapirArgs.isSpecified()) {
```
dsrbecky: So --log-level is not passed to gapir unless there are also user args?

ben-clayton: no?

core/cc/linux/debugger.cpp
ben-clayton: Why?

gapic/src/main/com/google/gapid/server/GapisProcess.java
dsrbecky: Ups, misread it

core/cc/linux/debugger.cpp
dsrbecky: It seems a bit redundant.
But the real reason is that if it is removed, all that is needed is to 1) attach debugger 2) force return from function.
With the extra function and without symbols, that attach flow no longer works.

ben-clayton:
Now works.

gapic/src/main/com/google/gapid/Main.java
```diff
@@ -172,14 +172,16 @@ public void dispose() {
   private static final Flag<?>[] ALL_FLAGS = {
     Flags.help,
     Flags.version,
-    GapiPaths.gapidPath,
+    Follower.logFollowRequests,
```
pmuetschard: Please don't re-order the flag. They were ordered by importance/relevance to most users, as they are shown in this same order for --help.

gapic/src/main/com/google/gapid/server/GapisProcess.java
```diff
@@ -82,15 +87,25 @@ protected Exception prepare(ProcessBuilder pb) {
...
     File logDir = Logging.getLogDir();
     if (logDir != null) {
       args.add("-log-file");
       args.add(new File(logDir, "gapis.log").getAbsolutePath());
       args.add("-log-level");
       args.add(logLevel.get().gapisLevel);
+
+      gapirFlags = "--log " + new File(logDir, "gapir.log").getAbsolutePath() +
+          " --log-level " + logLevel.get().gapirLevel;
+      if (gapirArgs.isSpecified()) {
```
pmuetschard: Use ```if (!gapirArgs.get().isEmpty())``` to honor the default value (in case anyone ever changes it).

gapic/src/main/com/google/gapid/server/GapisProcess.java
```diff
@@ -113,6 +128,10 @@ protected Exception prepare(ProcessBuilder pb) {
       args.add(adb);
     }
 
+    if (gapisArgs.isSpecified()) {
```
pmuetschard: Use ```if (!gapisArgs.get().isEmpty())```.

gapic/src/main/com/google/gapid/server/GapisProcess.java
```diff
@@ -113,6 +128,10 @@ protected Exception prepare(ProcessBuilder pb) {
...
+      args.add(gapisArgs.get());
```
pmuetschard: The help text says "argument**s**", however this passes the string as a single argument.

ben-clayton: It wasn't clear to me how you have a multi-arg flag. Is it possible?

gapic/src/main/com/google/gapid/Main.java
ben-clayton: Done. I do hate flag lists that aren't alphabetically ordered though (not that I would expect declaration order to affect presentation).

gapic/src/main/com/google/gapid/server/GapisProcess.java
ben-clayton: Done

gapic/src/main/com/google/gapid/server/GapisProcess.java
ben-clayton: Done

gapic/src/main/com/google/gapid/server/GapisProcess.java
pmuetschard: Looks like you got yourself into a bit of a pickle... however, since we've already implemented a "split a string into flags list" function in ```core/text```, I'd suggest passing the string to gapis and having gapis use that function to split it apart. Assuming, of course, you can then pass the result on to the flag parsing code...

Otherwise, I do have code somewhere, that can collect multiple instances of a flag into a list. This would still limit the flag's syntax to the ```--name=value``` form:
```./do run gapic --gapis-arg '--foo=bar' --gapis-arg '--baz=meow'```
Otherwise you'd have to use this really clumsy form:
```./do run gapic --gapis-arg --foo --gapis-arg bar --gapis-arg --baz --gapis-arg meow```
Let me know if you want me to add in the string-list flag type.

Another ugly alternative is to re-implement ```core/text```'s splitting function in Java. Unfortunately, guava's ```Splitter``` does not yet support "splitting with escapes".

PR #845, opened 2017-08-01
gapis/gles: Workarounds for EGL_EXT_buffer_age

ben-clayton:
The Android framework heavily uses `eglSwapBuffersWithDamageKHR` to update just the portions of the compositor that got touched.


ben-clayton:
Okay, the description I gave is misleading - I guess at the framebuffer level, it *could* be invalidating, but from an end-user perspective, marking everything outside of that partial update with undefined is just not helpful.

Most people would expect and more importantly want to see the final composition, not the partial updates.

I've updated the comment.

dsrbecky:
Can you share a trace with the problem you are trying to solve?
eglSwapBuffersWithDamageKHR are eglSwapBuffers should be equivalent.  Otherwise the app is wrong and we should be showing the pattern.

Issue #773, opened 2017-07-24
Gapis timeout


dsrbecky:
My gapis keeping exiting on a big trace:
W 00:17:00.031 gapis Stopping GAPIS server as it has been idle for more than 3.333333333s (--idle-timeout)
I am not stopping any process, and running just at info log-level.

Also, should we it an error?  It took me a while to notice it all the other log messages.

It also seems that gapis will explicitly kill gapir process on timeout.  Can we just let gapir die on its own?  I mistook the abrupt death for driver crash.

ben-clayton:
"3.333333333s" is a bug, it's actually 10 seconds.

I've seen this as well, but I'm not sure what the correct solution is.
It seems odd / wrong that GAPIC hasn't sent a ping to GAPIS within 10 seconds when it [should be pinging at 1 second intervals](https://github.com/google/gapid/blob/master/gapic/src/main/com/google/gapid/server/GapisProcess.java#L57-L58).

I guess it's possible that with the workload of opening a large trace the heartbeat thread is just not scheduled. If so, the only short-term solution I can think of is to bump the timeouts. @pmuetschard proposed scanning for process IDs, but that involves per-platform code.

> Also, should we it an error?

Sure.

> It also seems that gapis will explicitly kill gapir process on timeout. Can we just let gapir die on its own?

When gapis times out, the main function returns, the main context is cancelled so that things clean up, and [any outstanding processes are terminated so we don't orphan them](https://github.com/google/gapid/blob/master/core/os/shell/local.go#L47-L57). I'd argue that killing child processes is the correct thing to do when we exit - but maybe we should be using an interrupt instead of forcefully closing the child process. Unfortunately that isn't supported on windows.

dsrbecky:
Increased the timeout in #777 
It still dies after a minute though

AWoloszyn:
FYI, I got the 1M timeout on Windows for a particularly large trace. Things had not hung yet though.

PR #848, opened 2017-08-01
gapis: Fix random gapis timeouts.

ben-clayton:
I believe RPC limits the number of concurrent RPCs, and once a bunch of long-running calls consume all the slots, the pings don't get through.

Count the number of concurrent RPCs, and consider this when deciding to timeout.

Fixes: #773

### DIFF
--- a/gapis/server/grpc.go
+++ b/gapis/server/grpc.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"sync/atomic"
 	"time"
 
 	"github.com/google/gapid/core/app/auth"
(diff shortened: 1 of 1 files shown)


ben-clayton:
This is an urgent fix. Pulling magic-pixie TBR powers.

gapis/server/grpc.go
```diff
@@ -68,21 +62,40 @@ func NewWithListener(ctx context.Context, l net.Listener, cfg Config, srvChan ch
...
+	select {
+	case s.keepAlive <- struct{}{}:
+	default:
+	}
+	return func() {
+		select {
+		case s.keepAlive <- struct{}{}:
+		default:
+		}
+		atomic.AddUint32(&s.inFlightRPCs, ^uint32(0))
```
pmuetschard: Why use this xor "magic", when there is a signed version of the atomics? Are you really expecting 2147483648+ concurrent requests?

ben-clayton: The xor magic is documented on the go function. It's the expected usage.

PR #856, opened 2017-08-02
Allow saving geometry data as OBJ.

pmuetschard:


pmuetschard:
> Rather unexpected?!

Writing my GVR test app, I wound up wishing I had this, as I was decoding from a compressed format. I had some time yesterday that was not enough to start one of the bigger things...

PR #857, opened 2017-08-02
Show the current GL context in the state view.

### LINKED ISSUE #708: We need a way to determine current context in state view
Switch to $Thread removed the value from state view, which makes it close very hard to find the right context.

pmuetschard:
Fixes #708

### DIFF
--- a/gapis/api/gles/gles.api
+++ b/gapis/api/gles/gles.api
@@ -620,6 +620,7 @@ class Context {
 // Globals
 ContextID              NextContextID
 map!(u64, ref!Context) Contexts
+ref!Context            CurrentContext
 
 sub ref!Context GetContext() {
   context := Contexts[$Thread]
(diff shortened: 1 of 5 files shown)


ben-clayton:
I'm slightly on the fence about this change.
I wanted to tackle this by adding basic support for Properties (getters / setters) in gapil, and as a first step, just write the getter in go by hand so that the `state_tree` picked it up.
While this works as a quick-fix, there are two concerns:
1. You're adding more work into every command, everywhere. Hopefully go is smart enough to inline and eliminate the `SetThread` call if it has no body.
2. You're adding an externally assigned variable into the API, which the static analysis will not be able to understand. While this field is not used in the API, this isn't an issue, but it's going to end up getting used if it stays.

Unless David has any major objections, I'll approve this a stop-gap solution.

dsrbecky:
Short, simple and efficient.  I am all in :-)

PR #859, opened 2017-08-03
Add automatic checking for updates

ben-clayton:


gapic/src/main/com/google/gapid/MainWindow.java
```diff
@@ -91,6 +92,7 @@
   protected final Map<MainTab.Type, Action> viewTabs = Maps.newHashMap();
   protected final Set<MainTab.Type> hiddenTabs = Sets.newHashSet();
   protected Action editCopy;
+  private StatusBar statusBar;
```
pmuetschard: JFace already has a status bar widget - simply add "addStatusLine()" in the MainWindow constructor.

However, I've not added the status line yet, because if it's empty most of the time, which it looks like it would be in this change, it looks strange.

gapic/src/main/com/google/gapid/views/StatusBar.java
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  /**
+   * Updates the notification to the given text. Can be safely called on a non-UI thread.
+   *
+   * @param text the notification text.
+   * @param onClick the optional notifiction click handler.
+   */
+  public void setNotification(String text, Runnable onClick) {
+    scheduleIfNotDisposed(this, () -> {
```
pmuetschard: Callers should make sure to only call this on the UI thread.

gapic/src/main/com/google/gapid/views/StatusBar.java
```diff
@@ -0,0 +1,73 @@
...
+      notification.setText(text);
+      notification.setForeground(onClick != null ?
```
pmuetschard: Use the ```Link``` widget to do this automatically.

gapic/src/main/com/google/gapid/MainWindow.java
```diff
@@ -59,6 +60,12 @@
 import com.google.gapid.widgets.TabArea.TabInfo;
 import com.google.gapid.widgets.Widgets;
 
+import java.awt.Desktop;
```
pmuetschard: Let's not mix UI frameworks... use Program.launch(URL) to launch the browser.

gapic/src/main/com/google/gapid/models/Settings.java
```diff
@@ -224,10 +233,27 @@ private static int getInt(Properties properties, String name, int dflt) {
...
+    } catch (NumberFormatException e) {
+      return dflt;
+    }
+  }
+
   private static boolean getBoolean(Properties properties, String name) {
     return "true".equalsIgnoreCase(properties.getProperty(name, ""));
   }
 
+  private static boolean getBoolean(Properties properties, String name, boolean dflt) {
```
pmuetschard: This is unnecessary. If you want a boolean that is true by default, simply invert the meaning. I.e. rename autoCheckForUpdates to noAutoUpdates.

gapic/src/main/com/google/gapid/util/UpdateWatcher.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.Date;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Utility class for checking for new releases of GAPID.
+ */
+public class UpdateWatcher {
+  private static final int CHECK_INTERVAL_HOURS = 8;
+  private static final long CHECK_INTERVAL_MS = CHECK_INTERVAL_HOURS * 60 * 60 * 1000;
```
pmuetschard: ```TimeUnit.HOURS.toMillis(8)```

gapic/src/main/com/google/gapid/util/UpdateWatcher.java
```diff
@@ -0,0 +1,80 @@
...
+    } else {
+      scheduleCheck();
+    }
+  }
+
+  private void scheduleCheck() {
+    long now = new Date().getTime();
+    long timeSinceLastUpdateMS = now - settings.lastCheckForUpdates;
+    long delay = Math.max(CHECK_INTERVAL_MS - timeSinceLastUpdateMS, 0);
+    Scheduler.EXECUTOR.schedule(this::doCheck, delay, TimeUnit.MILLISECONDS);
```
pmuetschard: Since it's unlikely for anyone to have the application open for that long, I think we should just check once a day on startup.

gapic/src/main/com/google/gapid/util/UpdateWatcher.java
```diff
@@ -0,0 +1,80 @@
...
+        Release release = future.get();
+        if (release != null) {
+          settings.updateAvailable = true;
+          listener.onNewReleaseAvailable(release);
+        }
+      } catch (InterruptedException | ExecutionException e) {
+        /* never mind */
+      }
+    }
+    settings.lastCheckForUpdates = new Date().getTime();
```
pmuetschard: ```System.currentTimeMillis()```

gapic/src/main/com/google/gapid/util/UpdateWatcher.java
```diff
@@ -0,0 +1,80 @@
...
+    this.listener = listener;
+    if (settings.updateAvailable) {
+      Scheduler.EXECUTOR.schedule(this::doCheck, 0, TimeUnit.MILLISECONDS);
+    } else {
+      scheduleCheck();
+    }
+  }
+
+  private void scheduleCheck() {
+    long now = new Date().getTime();
```
pmuetschard: ```System.currentTimeMillis()```

gapic/src/main/com/google/gapid/MainWindow.java
ben-clayton: I can't really find much by way of documentation for JFace's status bar. The little I can find seems like it's rather restrictive.

>  However, I've not added the status line yet, because if it's empty most of the time, which it looks like it would be in this change, it looks strange.

It looks stranger at the moment than it needs to - the bar appears to have padding that I have no idea how to remove. If you are really opposed to having a status bar, where would you suggest we put the notification?

gapic/src/main/com/google/gapid/views/StatusBar.java
ben-clayton: For something that is a status bar, where statuses nearly always come from asynchronous tasks, this seems like it makes usage more awkward for no obvious gain. What are the down-sides to making this thread safe?

gapic/src/main/com/google/gapid/models/Settings.java
ben-clayton: Seems like a bit of a long-winded solution for the sake of a `true` default. I'd buy this argument more if other types didn't have non-default-initialized values. `noAutoUpdates` reads like it would apply the update automatically, so it would probably end up named as `noAutoCheckForUpdates`. Then you'd end up with a number of conditionals being a double-negative: `if (!noAutoCheckForUpdates)`, which I've seen cause countless bugs in my career.

gapic/src/main/com/google/gapid/util/UpdateWatcher.java
ben-clayton: I don't think it's that uncommon to leave an application open on your desktop, go home, come in the next day to pick up from where you left off. Having a new version notification is a nice thing to be greeted with.

gapic/src/main/com/google/gapid/MainWindow.java
ben-clayton: Done

gapic/src/main/com/google/gapid/views/StatusBar.java
ben-clayton: Done

gapic/src/main/com/google/gapid/util/UpdateWatcher.java
ben-clayton: Done

gapic/src/main/com/google/gapid/util/UpdateWatcher.java
ben-clayton: Done

gapic/src/main/com/google/gapid/util/UpdateWatcher.java
ben-clayton: Done

gapic/src/main/com/google/gapid/MainWindow.java
pmuetschard: > seems like it's rather restrictive.

Not really, but I'm OK with punting on doing the research on it. A cursory look suggests it uses a `ContributionManager` and `ContributionItem`s... Either way, it doesn't look like it does all that much stuff to render, so may not even be worth using.

gapic/src/main/com/google/gapid/MainWindow.java
pmuetschard: You're still using the Desktop class.

gapic/src/main/com/google/gapid/MainWindow.java
```diff
@@ -84,13 +89,15 @@
  * The main {@link ApplicationWindow} containing all of the UI components.
  */
 public class MainWindow extends ApplicationWindow {
+  private static final Logger LOG = Logger.getLogger(MainWindow.class.getName());
```
pmuetschard: nit: please add an empty line between class and instance members.

gapic/src/main/com/google/gapid/views/StatusBar.java
pmuetschard: The downside is the cognitive load. Up to this point, every widget and view class can assume that its entrie interface is only accessed from the UI thread. Thus, the thread transitions are not spread out everywhere, instead, whatever makes the RPC call, does the transition back to the UI thread in its callback. You should use RPC.listen and UICallback to do this nicely for you.

gapic/src/main/com/google/gapid/MainWindow.java
pmuetschard: Actually, remove it since it's not used.

gapic/src/main/com/google/gapid/util/UpdateWatcher.java
```diff
@@ -0,0 +1,79 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.gapid.util;
+
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.gapid.models.Settings;
+import com.google.gapid.proto.service.Service.Release;
+import com.google.gapid.server.Client;
+import java.util.Date;
```
pmuetschard: unused import

gapic/src/main/com/google/gapid/views/StatusBar.java
```diff
@@ -0,0 +1,66 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.gapid.views;
+
+import static com.google.gapid.widgets.Widgets.scheduleIfNotDisposed;
+
+import com.google.gapid.widgets.Theme;
```
pmuetschard: unused import.

gapic/src/main/com/google/gapid/MainWindow.java
ben-clayton: Really done.

gapic/src/main/com/google/gapid/views/StatusBar.java
ben-clayton: Moved the `scheduleIfNotDisposed` call to the caller. `RPC.listen` doesn't seem to fit well here.

gapic/src/main/com/google/gapid/MainWindow.java
ben-clayton: Done

gapic/src/main/com/google/gapid/util/UpdateWatcher.java
ben-clayton: Done

gapic/src/main/com/google/gapid/views/StatusBar.java
ben-clayton: Done

PR #871, opened 2017-08-04
Fix a crash due to interface conversion

Qining:
Not quite sure if this is a correct fix, but otherwise gapis crashes for some cases as:
```
panic: interface conversion: api.Span is *api.SubCmdRoot, not
*api.CmdIDRange [recovered]
```


Qining:
Close as this should be fixed with the support of rebuilding vkCmdExecuteCommands.

PR #875, opened 2017-08-07
Fix crash in deviceinfo query.

baldwinn860:
query::getDeviceInstance could return null, but was never checked in
get_device_instance before dereferencing, causing a crash when it
failed.

Test case is calling host.getHostDevice while accessing a headless
instance, which causes glXCreateNewContext to fail from within
createContext.


baldwinn860:
Note this still doesn't let us run on a headless instance (or one without gl contexts), but now it no longer crashes and instead gives us the actual error.

Issue #877, opened 2017-08-08
Server failed to encode response proto: oneof field has nil value


ben-clayton:
`server.CheckForUpdates()` is returning `nil, nil`. Proto expects a oneof (value or error).

PR #879, opened 2017-08-08
gapis/server: Don't return (nil, nil) from CheckForUpdates.

ben-clayton:
Fixes: #877


gapis/server/server.go
```diff
@@ -124,6 +125,12 @@ func (s *server) CheckForUpdates(ctx context.Context, includePrereleases bool) (
 			mostRecentVersion = version
 		}
 	}
+	if mostRecent == nil {
```
pmuetschard: There's another `return nil, nil` on line 106... why wouldn't GH let me comment there... sigh.

ben-clayton: Done.

Issue #880, opened 2017-08-08
Vulkan: Crash when resolve command tree node for subcommand


Qining:
if there are vkCmdExecuteCommands in the subcommand list of a vkQueueSubmit, the resolving process may crash. Note that this is irrelevant with the missing implementation of subcommand rebuilding support of vkCmdExecuteCommand, as no replay request needed to trigger this crash.

Qining:
Fixed in #881

PR #881, opened 2017-08-08
Fix a crash due to grouping in SubCmdRoot

Qining:
Otherwise gapis crash as:
```
panic: interface conversion: api.Span is *api.SubCmdRoot, not
*api.CmdIDRange [recovered]
```

Fix issue #880 

Generally, for SubCmdRoot, Subcommand should be added first (this only expand the single range of the subcommands in SubCmdRoot) and nested SubCmdRoots are added later (to divides the single range into multiple pieces).

Also, for SubCmdRoot, the Bounds() function should returns the inner-most index range. For a SubCmdRoot of vkQueueSubmit, the length of its ID is just 1, and we add this SubCmdRoot by intersecting with the top-level CmdIDGroup (the root), using the inner-most index (as there is only one) is correct.

For subcommand in a vkQueueSubmit, its SubCmdRoot (e.g. vkCmdExecuteCommands) will be intersected with the corresponding nesting sub group of vkQueueSubmit. We still should be using the inner-most index of the subcommand as the return of Bounds(). It is not right to use the outer-most index (which is the index of the vkQueueSubmit command) to locate the subcommand in the nesting sub group of vkQueueSubmit.

### DIFF
--- a/gapis/api/cmd_id_group.go
+++ b/gapis/api/cmd_id_group.go
@@ -126,10 +126,12 @@ func (r *CmdIDRange) item(i uint64) SpanItem {
 func (r *CmdIDRange) itemIndex(i CmdID) uint64    { return uint64(i - r.Start) }
 func (r *CmdIDRange) split(i uint64) (Span, Span) { return r.Split(i) }
 
-func (c *SubCmdRoot) Bounds() CmdIDRange          { return CmdIDRange{CmdID(c.Id[0]), CmdID(c.Id[0] + 1)} }
+func (c *SubCmdRoot) Bounds() CmdIDRange {
+	return CmdIDRange{CmdID(c.Id[len(c.Id)-1]), CmdID(c.Id[len(c.Id)-1] + 1)}
+}
 func (c *SubCmdRoot) itemCount() uint64           { return 1 }
 func (c *SubCmdRoot) item(uint64) SpanItem        { return *c }
-func (c *SubCmdRoot) itemIndex(i CmdID) uint64    { return 0 }
+func (c *SubCmdRoot) itemIndex(i CmdID) uint64    { return uint64(i) }
 func (c *SubCmdRoot) split(i uint64) (Span, Span) { return c, nil }
 
 func (g *CmdIDGroup) Bounds() CmdIDRange          { return g.Range }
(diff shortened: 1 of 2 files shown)


gapis/resolve/command_tree.go
```diff
@@ -319,6 +321,8 @@ func (r *CommandTreeResolvable) Resolve(ctx context.Context) (interface{}, error
 
 	for k, v := range snc.SubcommandGroups {
 		r := out.root.AddRoot([]uint64{uint64(k)})
+		// subcommands are added before nested SubCmdRoots.
+		sort.SliceStable(v, func(i, j int) bool { return len(v[i]) < len(v[j]) })
```
Qining: I would say it is still just a quick-fix. The order should not matter for a more reasonable logic

gapis/api/cmd_id_group.go
```diff
@@ -372,8 +374,20 @@ func (g *CmdIDGroup) AddRoot(rootidx []uint64) *SubCmdRoot {
 		switch first := g.Spans[s].(type) {
 		case *CmdIDGroup:
 			return first.AddRoot(rootidx)
+		case *CmdIDRange:
+			firstHalf := &CmdIDRange{first.Start, first.End}
```
ben-clayton: Why assign the end here then immediately replace it the line below?

Qining: oops, I was just trying to use copy-constructor to build a CmdIDRange. I will fix this.

Qining: Fixed

gapis/api/cmd_id_group.go
```diff
@@ -372,8 +374,18 @@ func (g *CmdIDGroup) AddRoot(rootidx []uint64) *SubCmdRoot {
 		switch first := g.Spans[s].(type) {
 		case *CmdIDGroup:
 			return first.AddRoot(rootidx)
+		case *CmdIDRange:
+			firstHalf := &CmdIDRange{first.Start, CmdID(rootidx[len(rootidx)-1]}
```
ben-clayton: You're missing a `)`. Please always check the code builds before submitting.

PR #884, opened 2017-08-09
Improve command tree grouping

dsrbecky:


gapis/api/cmd_id_group_test.go
```diff
@@ -448,7 +448,7 @@ func TestAddAtomsWithSplitting(t *testing.T) {
 	ctx := log.Testing(t)
 	got := buildTestGroup(700)
 
-	got.AddAtoms(func(CmdID) bool { return true }, 45)
+	got.AddAtoms(func(CmdID) bool { return true }, 45, 0)
```
ben-clayton: Can we please have a test for `maxNeighbours`?

gapis/resolve/command_tree.go
```diff
@@ -340,21 +355,8 @@ func (r *CommandTreeResolvable) Resolve(ctx context.Context) (interface{}, error
...
-		Filter:       p.Filter,
-		DrawCalls:    p.GroupByDrawCall,
-		FirstInFrame: true,
-		LastInFrame:  true,
-	})
-	if err != nil {
-		return log.Errf(ctx, err, "Couldn't get events")
-	}
-
+func addDrawEvents(ctx context.Context, events *service.Events, p *path.CommandTree, t *commandTree, last api.CmdID) {
```
ben-clayton: What's the fix for the "silently ignored error" of which you speak?

gapis/resolve/command_tree.go
```diff
@@ -372,20 +372,31 @@ func addDrawEvents(ctx context.Context, events *service.Events, p *path.CommandT
 }
 
 func addFrameEvents(ctx context.Context, events *service.Events, p *path.CommandTree, t *commandTree, last api.CmdID) {
-	frameCount, frameStart, frameEnd := 0, api.CmdID(0), api.CmdID(0)
+	frameCount, frameStart := 0, api.CmdID(0)
 	for _, e := range events.List {
 		i := api.CmdID(e.Command.Indices[0])
 		switch e.Kind {
```
ben-clayton: I'm not sure about removing the use of `FirstInFrame`. Don't we want to use this information if it's provided by the API? For example `eglSwapBuffers` is now annotated with `@StartOfFrame`, so we're actually deriving our start-of-frame here from the end-of-frame event, which is actually derived from the `@StartOfFrame` annotation. Ewww.

gapis/api/cmd_id_group.go
```diff
@@ -33,9 +33,10 @@ import (
 // or an command range that is within this group's span but outside of any
 // sub-group.
 type CmdIDGroup struct {
-	Name  string     // Name of this group.
-	Range CmdIDRange // The range of commands this group (and items) represents.
-	Spans Spans      // All sub-groups and sub-ranges of this group.
+	Name      string     // Name of this group.
+	Range     CmdIDRange // The range of commands this group (and items) represents.
+	Spans     Spans      // All sub-groups and sub-ranges of this group.
+	Thumbnail CmdID      // The command which shall be used to make thumbnail for this group
```
ben-clayton: I really don't think this information belongs here. The types in the `gapis/api` package are supposed to be core and generic data structures relating to APIs, Commands and basic resource types. Thumbnail is a very specific usage.

It's unfortunate that the `resolve.commandTree` structure doesn't have a separate node type for its representation of the command tree. 

Given that you're just doing a binary search on the events list to get the thumbnail for the group, you could just do this on demand in [`resolve.CommandTreeNode()`](https://github.com/google/gapid/blob/f0fccfd6c035b4feff3124ed19dffb6b251de2b4/gapis/resolve/command_tree.go#L78-L119)

gapis/resolve/command_tree.go
dsrbecky: return log.Errf(ctx, err, "Couldn't get events") was always ignored before

gapis/api/cmd_id_group.go
dsrbecky: Done

gapis/resolve/command_tree.go
dsrbecky: Done

gapis/api/cmd_id_group_test.go
dsrbecky: Done

dsrbecky:
PTAL

dsrbecky:
PS: I added the missing local changes to the last commit.

PR #886, opened 2017-08-09
Vulkan: Subcommand support for secondary command buffers and commands

Qining:
Supports cutting in vkCmdExecuteCommands and RecreateCmdExecuteCommands.

Sits on top of #881

### DIFF
--- a/gapis/api/vulkan/vulkan_terminator.go
+++ b/gapis/api/vulkan/vulkan_terminator.go
@@ -239,11 +239,9 @@ func rebuildCommandBuffer(ctx context.Context,
 		numCommandsToCopy += 1
 	case 2:
 		// Ends at a secondary command buffer
-		log.W(ctx, "Cut at secondary command buffer: %v", idx)
 		numSecondaryCmdBuffersToCopy = idx[1] + 1
 	case 3:
 		// Ends at a secondary command, copies including idx
-		log.W(ctx, "Cut at secondary command: %v", idx)
 		numSecondaryCmdBuffersToCopy = idx[1]
 		numSecondaryCommandsToCopy = idx[2] + 1
 	}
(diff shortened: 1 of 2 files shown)


gapis/api/vulkan/vulkan_terminator.go
```diff
@@ -219,73 +219,84 @@ func rebuildCommandBuffer(ctx context.Context,
...
-	}
-
-	for i := 0; i < int(commandsToCopy); i++ {
+	switch len(idx) {
+	case 1:
+		// Only primary commands, copies including idx
+		numCommandsToCopy += 1
+	case 2:
+		// Ends at a secondary command buffer
+		log.W(ctx, "Cut at secondary command buffer: %v", idx)
```
AWoloszyn: This is not actually a warning, we should not log.

gapis/api/vulkan/vulkan_terminator.go
```diff
@@ -219,73 +219,84 @@ func rebuildCommandBuffer(ctx context.Context,
...
+		numSecondaryCmdBuffersToCopy = idx[1] + 1
+	case 3:
+		// Ends at a secondary command, copies including idx
+		log.W(ctx, "Cut at secondary command: %v", idx)
```
AWoloszyn: Same here.

gapis/api/vulkan/vulkan_terminator.go
Qining: fixed in 20d358d0839a5ddb2416f1a474b78f9d7dc18c3f

gapis/api/vulkan/vulkan_terminator.go
Qining: fixed in 20d358d0839a5ddb2416f1a474b78f9d7dc18c3f

PR #890, opened 2017-08-10
Add specific GL error message for negative value.

dsrbecky:


gapis/messages/en-us.stb.md
```diff
@@ -27,6 +27,10 @@ The slice {{from_value}}:{{to_value}} for {{from_variable}}:{{to_variable}} is o
 
 Invalid value.
 
+# ERR_INVALID_VALUE_NON_NEGATIVE
```
ben-clayton: Nit: Double negative. How about `ERR_INVALID_NEGATIVE_VALUE` ?

PR #892, opened 2017-08-10
Robot: Fix error handling in trace client

baldwinn860:
When traces failed we would get no logs, because of the way the error codes were being handled, and then crash, because of the way errors from doTrace were handled. This fixes those.

### DIFF
--- a/test/robot/replay/replay.proto
+++ b/test/robot/replay/replay.proto
@@ -44,6 +44,8 @@ message Output {
   string log = 1;
   // Video is the movie of the replay frame capture output.
   string video = 2;
+  // CallError is the error string returned by the call to gapit.
+  string call_error = 3;
 }
 
 // Action holds the information about an execution of a task.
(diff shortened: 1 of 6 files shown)


test/robot/trace/client.go
```diff
@@ -189,7 +187,7 @@ func doTrace(ctx context.Context, action string, in *Input, store *stash.Client,
 		return nil, err
 	}
 	outputObj.Trace = traceID
-	return outputObj, nil
+	return outputObj, callErr
```
ben-clayton: As a general rule-of-thumb, you shouldn't return a value and an error. If you have an output object and want to indicate that something went wrong (but still got meaningful values), I'd suggest putting an error field in the output.

baldwinn860: Done.

test/robot/replay/replay.proto
```diff
@@ -44,6 +44,8 @@ message Output {
   string log = 1;
   // Video is the movie of the replay frame capture output.
   string video = 2;
+  // CallErr is the error string returned by the call to gapit.
```
ben-clayton: Nit: `CallError`, here and other protos.

PR #900, opened 2017-08-11
Fix tracing on multiple devices simultaneously.

baldwinn860:
The mixture of a deferred mutex unlock and a channel receive call made
it so the actions list would wait for one trace task to finish before
starting the next one. Moving the mutex fixed the issue.

### DIFF
--- a/test/robot/job/worker/action.go
+++ b/test/robot/job/worker/action.go
@@ -118,6 +118,8 @@ func (a *Actions) do(ctx context.Context, w *Worker, input Input) (string, error
 
 // update an action, and return the merged action.
 func (a *Actions) update(ctx context.Context, action Action) error {
+	a.mu.Lock()
+	defer a.mu.Unlock()
 	return a.ledger.Add(ctx, action)
 }
 
--- a/test/robot/job/worker/action.go
+++ b/test/robot/job/worker/action.go
@@ -94,11 +94,18 @@ func (a *Actions) apply(ctx context.Context, item Action) error {
 }
 
 func (a *Actions) do(ctx context.Context, w *Worker, input Input) (string, error) {
-	a.mu.Lock()
-	defer a.mu.Unlock()
-	action := proto.Clone(a.nullAction).(Action)
-	action.Init(id.Unique().String(), input, w.Info)
-	if err := a.ledger.Add(ctx, action); err != nil {
+	action, err := func() (Action, error) {
+		a.mu.Lock()
+		defer a.mu.Unlock()
+		action := proto.Clone(a.nullAction).(Action)
+		action.Init(id.Unique().String(), input, w.Info)
+		err := a.ledger.Add(ctx, action)
+		if err != nil {
+			return nil, err
+		}
+		return action, nil
+	}()
+	if err != nil {
 		return "", err
 	}
 	task := proto.Clone(a.nullTask).(Task)


test/robot/job/worker/action.go
```diff
@@ -95,12 +95,13 @@ func (a *Actions) apply(ctx context.Context, item Action) error {
 
 func (a *Actions) do(ctx context.Context, w *Worker, input Input) (string, error) {
 	a.mu.Lock()
-	defer a.mu.Unlock()
 	action := proto.Clone(a.nullAction).(Action)
 	action.Init(id.Unique().String(), input, w.Info)
 	if err := a.ledger.Add(ctx, action); err != nil {
 		return "", err
 	}
+	// unlock here so other do calls can be invoked after adding to ledger
```
ben-clayton: What happens if the `Add` call errors and returns before the mutex unlock?

baldwinn860: thanks for the catch, fixed.

pmuetschard: To avoid any problems down the line, it's probably worth creating a small private function that does the clone, init and add while holding the lock, so you can still use the defer to unlock. Using the defer is the preferred pattern, as it ensure that it will *always* unlock - even if another developer later on adds an early return, e.g.

baldwinn860: Ok, I was considering something like that, instead making a local function that would handle the defer. Would that be ok since this function (clone, init, add) doesn't need to be accessible anywhere else? or is the separate private function still preferred? I really like defer, but the fact that it is function oriented rather than scope oriented is still a gotcha for me.

baldwinn860: ok the latest version has a local function to handle defer correctly. If you prefer the separate private function let me know.

ben-clayton:
What exactly is this mutex supposed to be guarding against?

baldwinn860:
Ah, I realized a problem, ledger.Add will trigger the apply event handler, which requires the mutex to be locked, locking it in do keeps it locked for the local add, but when the worker is then started clients will call Update on the action, which calls ledger.Add again, I think this is the reason for the mutex to be held for the whole function. Adding a mutex lock to Actions.update should bring parity and still let multiple actions happen in parallel.

ben-clayton:
Okay, so `a.ledger.Add()` calls `apply()` via the `ledger.Watch(ctx, apply)` and `apply()` `"should be called with the mutation lock already held."`.

ben-clayton:
From the original PR description, what channel did you find blocking when the mutex was held?

baldwinn860:
in `worker.send` there is a blocking channel on the result of the task.

ben-clayton:
This code is making my head hurt. @ianthehat - if you're lurking, can you offer a suggestion here?

baldwinn860:
Here is the flow from my understanding:
 - Task managers call `worker.Manager.Do`
 - which calls `Actions.do`
 - which locks the mutex to call `ledger.Add`
 - which triggers `Actions.apply` which requires the lock to be held.

Prior to this change:
 - the lock is still held while `Actions.do` calls `Workers.send`
 - which starts a task and waits to receive the result
 - the task calls `worker.Manager.Update`
 - which calls `Actions.update`
 - which calls `ledger.Add` without locking the mutex because the parent `Actions.do` still holds it
 - `Actions.apply` is called with the mutex still locked and effectively single threaded
 - All other calls to `Actions.do` from other goroutines block because the mutex is still held

After this change:
 - the lock is released after `ledger.Add` finishes
 - `Workers.send` is called, starts the task and still waits to receive the result
 - the task calls `worker.Manager.Update` which calls `Actions.update`
 - which now locks the mutex before calling `ledger.Add` thus fulfilling `Actions.apply`'s contract
 - Other `Actions.do` can be spun up now that the mutex is not locked to the task completing.

PR #803, opened 2017-07-26
Revision of DCE for Vulkan

Qining:
This version of DCE drops more commands and save more transferring data.

Sits on top of #881 and #901 

TODO:
[x] Remove debugging comments and clean the code
[x] Blurbs for the exported type and methods
[x] More tests for the DCE transformation and the simple trie data sturcture
[x] Solve the problem of synchronization primitives
  Solved by keeping all the synchronization primitives alive.
[x] Handle EndRenderPass
  Solved by marking the corresponding EndRenderPass alive if a
  BeginRenderPass is found alive. However, no new backpropation will be
  initiated from the alive EndRenderPass

PR #901, opened 2017-08-14
A map-based trie for data indexed by SubCmdIdx

Qining:
Vulkan's new DCE #803 needs a data structure indexed by SubCmdIdx.

### DIFF
--- a/gapis/api/subcmd_idx_trie.go
+++ b/gapis/api/subcmd_idx_trie.go
@@ -1,0 +1,109 @@
+// Copyright (C) 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package api
+
+// SubCmdIdxTrie is a map-based trie using SubCmdIdx for indexing the data
+// stored inside.
+type SubCmdIdxTrie struct {
+	hasValue  bool
+	value     interface{}
+	nextLevel map[uint64]*SubCmdIdxTrie
+}
+
+// Constructs a new SubCmdIdxTrie instance and returns a pointer to the instance.
+func NewSubCmdIdxTrie() *SubCmdIdxTrie {
+	return &SubCmdIdxTrie{
+		hasValue:  false,
+		nextLevel: map[uint64]*SubCmdIdxTrie{},
+	}
+}
+
+// Tries to get the value stored in the trie by the given SubCmdIdx. If a value
+// is stored in the trie for the given SubCmdIdx, returns the value and true,
+// otherwise returns nil and false.
+func (trie *SubCmdIdxTrie) GetValue(indices SubCmdIdx) (interface{}, bool) {
+	t := trie
+	for _, k := range indices {
+		_, ok := t.nextLevel[k]
+		if ok {
+			t = t.nextLevel[k]
+		} else {
+			return nil, false
+		}
+	}
+	if t.hasValue {
+		return t.value, true
+	}
+	return nil, false
+}
+
+// Sets a value to the trie with the given SubCmdIdx as index
+func (trie *SubCmdIdxTrie) SetValue(indices SubCmdIdx, v interface{}) {
+	t := trie
+	for _, k := range indices {
+		_, ok := t.nextLevel[k]
+		if ok {
+		} else {
+			t.nextLevel[k] = NewSubCmdIdxTrie()
+		}
+		t = t.nextLevel[k]
+	}
+	t.value = v
+	t.hasValue = true
+}
+
+// Tries to remove a value indexed by the given SubCmdIdx in the trie. If a
+// value is found, removes it and returns true. If a value with that SubCmdIdx
+// is not found, returns false. Removing the value from a parent node won't
+// remove the child nodes and their value, i.e.: Removing with [0,1] won't
+// erase the value and the node like: [0, 1, 2].
+func (trie *SubCmdIdxTrie) RemoveValue(indices SubCmdIdx) bool {
+	t := trie
+	for _, k := range indices {
+		_, ok := t.nextLevel[k]
+		if ok {
+			t = t.nextLevel[k]
+		} else {
+			return false
+		}
+	}
+	t.hasValue = false
+	return true
+}
+
+// Tries to remove a trie node indexed by the given SubCmdIdx. If a node is
+// is found, removes it can all its child nodes, and returns true. Otherwise,
+// returns false.
+func (trie *SubCmdIdxTrie) RemoveNode(indices SubCmdIdx) bool {
+	if len(indices) == 0 {
+		return false
+	}
+	t := trie
+	for i, k := range indices {
+		if i == len(indices)-1 {
+			if _, ok := t.nextLevel[k]; ok {
+				delete(t.nextLevel, k)
+				return true
+			}
+		}
+		_, ok := t.nextLevel[k]
+		if ok {
+			t = t.nextLevel[k]
+		} else {
+			return false
+		}
+	}
+	return false
+}
--- a/gapis/api/subcmd_idx_trie.go
+++ b/gapis/api/subcmd_idx_trie.go
@@ -17,93 +17,72 @@ package api
 // SubCmdIdxTrie is a map-based trie using SubCmdIdx for indexing the data
 // stored inside.
 type SubCmdIdxTrie struct {
-	hasValue  bool
 	value     interface{}
 	nextLevel map[uint64]*SubCmdIdxTrie
 }
 
-// Constructs a new SubCmdIdxTrie instance and returns a pointer to the instance.
-func NewSubCmdIdxTrie() *SubCmdIdxTrie {
-	return &SubCmdIdxTrie{
-		hasValue:  false,
-		nextLevel: map[uint64]*SubCmdIdxTrie{},
-	}
-}
-
-// Tries to get the value stored in the trie by the given SubCmdIdx. If a value
-// is stored in the trie for the given SubCmdIdx, returns the value and true,
-// otherwise returns nil and false.
-func (trie *SubCmdIdxTrie) GetValue(indices SubCmdIdx) (interface{}, bool) {
-	t := trie
-	for _, k := range indices {
-		_, ok := t.nextLevel[k]
-		if ok {
-			t = t.nextLevel[k]
-		} else {
-			return nil, false
-		}
-	}
-	if t.hasValue {
-		return t.value, true
-	}
-	return nil, false
-}
-
-// Sets a value to the trie with the given SubCmdIdx as index
-func (trie *SubCmdIdxTrie) SetValue(indices SubCmdIdx, v interface{}) {
-	t := trie
-	for _, k := range indices {
-		_, ok := t.nextLevel[k]
-		if ok {
-		} else {
-			t.nextLevel[k] = NewSubCmdIdxTrie()
-		}
-		t = t.nextLevel[k]
-	}
-	t.value = v
-	t.hasValue = true
-}
-
-// Tries to remove a value indexed by the given SubCmdIdx in the trie. If a
-// value is found, removes it and returns true. If a value with that SubCmdIdx
-// is not found, returns false. Removing the value from a parent node won't
-// remove the child nodes and their value, i.e.: Removing with [0,1] won't
-// erase the value and the node like: [0, 1, 2].
-func (trie *SubCmdIdxTrie) RemoveValue(indices SubCmdIdx) bool {
-	t := trie
-	for _, k := range indices {
-		_, ok := t.nextLevel[k]
-		if ok {
-			t = t.nextLevel[k]
-		} else {
-			return false
-		}
-	}
-	t.hasValue = false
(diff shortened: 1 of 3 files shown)


gapis/api/subcmd_idx_trie.go
```diff
@@ -0,0 +1,109 @@
...
+		return t.value, true
+	}
+	return nil, false
+}
+
+// Sets a value to the trie with the given SubCmdIdx as index
+func (trie *SubCmdIdxTrie) SetValue(indices SubCmdIdx, v interface{}) {
+	t := trie
+	for _, k := range indices {
+		_, ok := t.nextLevel[k]
```
ben-clayton: What I said above is especially true here.

gapis/api/subcmd_idx_trie.go
```diff
@@ -0,0 +1,109 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package api
+
+// SubCmdIdxTrie is a map-based trie using SubCmdIdx for indexing the data
+// stored inside.
+type SubCmdIdxTrie struct {
+	hasValue  bool
```
ben-clayton: Why do you need `hasValue` and `value`? Are you expecting to have `nil` as a valid value? If not, is there any purpose in the `bool` return value on `GetValue`?

gapis/api/subcmd_idx_trie.go
```diff
@@ -0,0 +1,109 @@
...
+	return &SubCmdIdxTrie{
+		hasValue:  false,
+		nextLevel: map[uint64]*SubCmdIdxTrie{},
+	}
+}
+
+// Tries to get the value stored in the trie by the given SubCmdIdx. If a value
+// is stored in the trie for the given SubCmdIdx, returns the value and true,
+// otherwise returns nil and false.
+func (trie *SubCmdIdxTrie) GetValue(indices SubCmdIdx) (interface{}, bool) {
```
ben-clayton: Nit: [Drop `Get` from `GetValue`](https://golang.org/doc/effective_go.html#Getters).

gapis/api/subcmd_idx_trie.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+	t.hasValue = false
+	return true
+}
+
+// Tries to remove a trie node indexed by the given SubCmdIdx. If a node is
+// is found, removes it can all its child nodes, and returns true. Otherwise,
+// returns false.
+func (trie *SubCmdIdxTrie) RemoveNode(indices SubCmdIdx) bool {
+	if len(indices) == 0 {
```
ben-clayton: Pointless test - if the `len` were 0, you'd return with the same value.

gapis/api/subcmd_idx_trie.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// Tries to remove a value indexed by the given SubCmdIdx in the trie. If a
+// value is found, removes it and returns true. If a value with that SubCmdIdx
+// is not found, returns false. Removing the value from a parent node won't
+// remove the child nodes and their value, i.e.: Removing with [0,1] won't
+// erase the value and the node like: [0, 1, 2].
+func (trie *SubCmdIdxTrie) RemoveValue(indices SubCmdIdx) bool {
+	t := trie
+	for _, k := range indices {
```
ben-clayton: I can't help but feel like you're doing the same traversal logic 3 times here, with slightly different logic once you resolve the node with the given index. How about:

```go
func (t *SubCmdIdxTrie) index(indices SubCmdIdx) *SubCmdIdxTrie {
	var ok bool
	for _, k := range indices {
		if t, ok = t.nextLevel[k]; !ok {
			return nil
		}
	}
	return t
}

func (t *SubCmdIdxTrie) Value(indices SubCmdIdx) interface{} {
	if n := t.index(indices); n != nil {
		return n.value
	}
	return nil
}

func (t *SubCmdIdxTrie) SetValue(indices SubCmdIdx, v interface{}) {
	if n := t.index(indices); n != nil {
		n.value = v
	}
}

func (t *SubCmdIdxTrie) RemoveValue(indices SubCmdIdx) bool {
	if n := t.index(indices); n != nil {
		t.value = nil
		return true
	}
	return false
}
```

gapis/api/subcmd_idx_trie.go
```diff
@@ -0,0 +1,109 @@
...
+		}
+	}
+	t.hasValue = false
+	return true
+}
+
+// Tries to remove a trie node indexed by the given SubCmdIdx. If a node is
+// is found, removes it can all its child nodes, and returns true. Otherwise,
+// returns false.
+func (trie *SubCmdIdxTrie) RemoveNode(indices SubCmdIdx) bool {
```
ben-clayton: Continuing with the theme of `nil` == no value, you could simplify the interface for this type by handing `SubCmdIdxTrie.SetValue(i, nil)` as a `RemoveValue(i)`, and automatically remove nodes that have no value or children.

Qining: SetValue(i, nil) and RemoveValue(i) are a little bit different. SetValue(i, nil) will create nodes if necessary, but RemoveValue(i) will never create nodes.

Automatically removing nodes is a little bit overkill. This potentially needs to traverse from the bottom to the top of a trie, but right, we can have it that way.

gapis/api/subcmd_idx_trie.go
Qining: The reason I didn't make it this way is that SetValue() will create nodes if necessary, but Value() and Remove() do not modify the Trie and just return failure in such cases. Sure we can have a bool type to control or pass in a callback, but I think either way are not good. These are just three tiny loops, keep it straightforward might be better. But anyway, we can for sure add a 'bool' to the argument list of index() to solve it.

Qining:
Will have traverse() function in the next commit, and traverse() function will have two callbacks for pre-order and post-order traversal. This will enable both
1) Create new nodes during traversing
2) Remove empty nodes during traversing

gapis/api/subcmd_idx_trie.go
ben-clayton: Annoyingly my code snippet is stale - I did update this to account for the creation of nodes, but I have no idea where that comment went. :(

ben-clayton:
```go
func (t *SubCmdIdxTrie) index(indices SubCmdIdx, buildMissing bool) *SubCmdIdxTrie {
	var ok bool
	for _, k := range indices {
		if t, ok = t.nextLevel[k]; !ok {
			if !buildMissing {
				return nil
			}
			next = &SubCmdIdxTrie{}
			t.nextLevel[k], t = next, next
		}
	}
	return t
}

func (t *SubCmdIdxTrie) Value(indices SubCmdIdx) interface{} {
	if n := t.index(indices, false); n != nil {
		return n.value
	}
	return nil
}

func (t *SubCmdIdxTrie) SetValue(indices SubCmdIdx, v interface{}) {
	if n := t.index(indices, true); n != nil {
		n.value = v
	}
}

func (t *SubCmdIdxTrie) RemoveValue(indices SubCmdIdx) bool {
	if n := t.index(indices, false); n != nil {
		t.value = nil
		return true
	}
	return false
}
```

gapis/api/subcmd_idx_trie.go
ben-clayton: I just scanned #803 to usages of `RemoveNode` and I can't find any. Is it even used?

gapis/api/subcmd_idx_trie.go
Qining: `hasValue` removed in 12fca8db3ff814a230a9a9fe976420a7e914c4f2

gapis/api/subcmd_idx_trie.go
```diff
@@ -0,0 +1,109 @@
...
+
+// SubCmdIdxTrie is a map-based trie using SubCmdIdx for indexing the data
+// stored inside.
+type SubCmdIdxTrie struct {
+	hasValue  bool
+	value     interface{}
+	nextLevel map[uint64]*SubCmdIdxTrie
+}
+
+// Constructs a new SubCmdIdxTrie instance and returns a pointer to the instance.
```
Qining: Godocs updated in 12fca8db3ff814a230a9a9fe976420a7e914c4f2

gapis/api/subcmd_idx_trie.go
```diff
@@ -0,0 +1,109 @@
...
+func NewSubCmdIdxTrie() *SubCmdIdxTrie {
+	return &SubCmdIdxTrie{
+		hasValue:  false,
```
Qining: The initializer is removed in 12fca8db3ff814a230a9a9fe976420a7e914c4f2

gapis/api/subcmd_idx_trie.go
```diff
@@ -0,0 +1,109 @@
...
+		nextLevel: map[uint64]*SubCmdIdxTrie{},
```
Qining: Fixed in 12fca8db3ff814a230a9a9fe976420a7e914c4f2

gapis/api/subcmd_idx_trie.go
Qining: Changed to `t` in 12fca8db3ff814a230a9a9fe976420a7e914c4f2

Qining: Dropped in 12fca8db3ff814a230a9a9fe976420a7e914c4f2

gapis/api/subcmd_idx_trie.go
Qining: hmm, I've updated the `index()` function in 12fca8db3ff814a230a9a9fe976420a7e914c4f2

Now it takes three callback functions. First one for handling the target trie node, second and third are for pre-indexing-order and post-indexing-order callbacks.

So now, we handle can all the requirements for SetValue, RemoveValue

gapis/api/subcmd_idx_trie.go
Qining: No it is not used anywhere, it was added just to make the structure looks more complete. `RemoveNode` has been removed in 12fca8db3ff814a230a9a9fe976420a7e914c4f2.

Now, `RemoveValue()` will take care of deleting the empty nodes.

Qining:
12fca8db3ff814a230a9a9fe976420a7e914c4f2 addes three callbacks function to be called along indexing the `SubCmdIdx` in the `trie`.

All comments should have been resolved.

gapis/api/subcmd_idx_trie.go
Qining: removed in 12fca8db3ff814a230a9a9fe976420a7e914c4f2

gapis/api/subcmd_idx_trie.go
```diff
@@ -0,0 +1,88 @@
...
+			}
+		}, nil)
+}
+
+// RemoveValue tries to remove a value indexed by the given SubCmdIdx in the
+// trie. If a value is found, removes it and returns true. If a value with that
+// SubCmdIdx is not found, returns false.
+func (t *SubCmdIdxTrie) RemoveValue(indices SubCmdIdx) bool {
+	hadValue := false
+	r := t.index(indices, func(t *SubCmdIdxTrie) {
```
ben-clayton: Due to the closures, captured value, order of calls, I find this exceptionally hard to read and follow.
Here's an alternative implementation that passes the tests:

```go

// SubCmdIdxTrie is a map-based trie using SubCmdIdx for indexing the data
// stored inside.
type SubCmdIdxTrie struct {
	value    interface{}
	children map[uint64]*SubCmdIdxTrie
}

// Value returnes the value stored in the trie indexed by the given SubCmdIdx.
// if no value is found by the given SubCmdIdx, returns nil.
func (t *SubCmdIdxTrie) Value(indices SubCmdIdx) interface{} {
	for t != nil && len(indices) > 0 && len(t.children) > 0 {
		t, indices = t.children[indices[0]], indices[1:]
	}
	if t == nil || len(indices) > 0 {
		return nil // invalid index
	}
	return t.value
}

// SetValue sets a value to the trie with the given SubCmdIdx as index.
func (t *SubCmdIdxTrie) SetValue(indices SubCmdIdx, v interface{}) {
	if v == nil { // nil value behaves the same as no value
		t.RemoveValue(indices)
		return
	}
	for _, i := range indices {
		next, ok := t.children[i]
		if !ok { // child does not exist - create it
			if t.children == nil {
				t.children = map[uint64]*SubCmdIdxTrie{}
			}
			next = &SubCmdIdxTrie{}
			t.children[i] = next
		}
		t = next
	}
	t.value = v
}

// RemoveValue tries to remove a value indexed by the given SubCmdIdx in the
// trie. If a value is found, removes it and returns true. If a value with that
// SubCmdIdx is not found, returns false.
func (t *SubCmdIdxTrie) RemoveValue(indices SubCmdIdx) bool {
	n := len(indices)
	stack := make([]*SubCmdIdxTrie, n+1)
	stack[0] = t
	for i, j := range indices { // build a stack of nodes starting with t
		t = t.children[j]
		if t == nil {
			return false // invalid index
		}
		stack[i+1] = t
	}
	if t.value == nil {
		return false // cannot remove a value it doesn't have
	}
	t.value = nil             // clear the value
	for i := n; i >= 0; i-- { // remove nodes that have no value or children
		if t.value != nil || len(t.children) > 0 {
			break // node cannot be deleted
		}
		delete(stack[i].children, indices[i-1])
		t = stack[i]
	}
	return true
}
```

This implementation is fractionally faster too:

Old:
```bash
$ go test ./gapis/api/ --bench=.                                                                                                                          
BenchmarkSubCmdIdxTrie-48         100000             11770 ns/op 
```

New:
```bash
$ go test ./gapis/api/ --bench=.                                                                                                                                                    
BenchmarkSubCmdIdxTrie-48         200000             10194 ns/op
```

Qining:
@ben-clayton Updated.

PR #909, opened 2017-08-14
Drop usage of the TitleAreaDialog

### LINKED ISSUE #902: Drop usage of TitleAreaDialog
Simply use regular dialogs - we don't need the title area gradient.

pmuetschard:
Fixes #902


gapic/src/main/com/google/gapid/widgets/Widgets.java
```diff
@@ -24,6 +24,7 @@
 import com.google.common.util.concurrent.SettableFuture;
 import com.google.gapid.models.Models;
 import com.google.gapid.server.Client;
+import com.google.gapid.views.AtomEditor;
```
ben-clayton: Do you need this new import?

pmuetschard: Yes. The class was moved out of this package into the views package, thus it now needs to be imported.

PR #918, opened 2017-08-15
Ensure GLSL variables are initialized

dsrbecky:
The compat layer sometimes creates code which uses uninitialized
variables.  Although the output is technically correct, it makes
my driver complain a lot.

### DIFF
--- a/gapis/shadertools/cc/spv_manager.cpp
+++ b/gapis/shadertools/cc/spv_manager.cpp
@@ -112,6 +112,55 @@ void SpvManager::renameViewIndex() {
   }
 }
 
+// Explicitly initialize locals. This is a work around for SPIR-V cross problem
+// where it may generate code which reads locals before initializing them.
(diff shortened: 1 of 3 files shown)


gapis/shadertools/cc/spv_manager.cpp
```diff
@@ -112,6 +112,55 @@ void SpvManager::renameViewIndex() {
   }
 }
 
+// Explicitly initialize locals. This is a work around for SPIR-V cross problem
+// where it may generate code which reads locals before initializing them.
+// For example, "v.x = 42.0;" becomes "v = vec4(42.0, v.y, v.z, v.w);"
+void SpvManager::initLocals() {
```
ben-clayton: Looks like a good thing to upstream.

dsrbecky: Well, the proper fix would be in SPIRV-Cross.  I have investigated that approach for a bit, but gave up.  I just ended fixing a bug.

PR #920, opened 2017-08-16
Abort all GL commands without current context

dsrbecky:


dsrbecky:
I am not aware of any checks in the API files - they call GetContext because they need the context - the check is implicit, but the call can not be removed (it could be replaced by your global variable, but let's keep that question separate to this PR).

There are some cases in the go code which check for null context.  Maybe some of those could be simplified, but there is no harm in keeping them for now...

ben-clayton:
This is a steaming hack that adds more overhead and crap into our templates, and affects vulkan and gvr.
I know it is tedious, but I'd really prefer this to be done in the API files.

Issue #872, opened 2017-08-04
One case of replay failure for Vulkan


Qining:
The application is: https://github.com/SaschaWillems/Vulkan/tree/master/scenerendering

The mid-exeuction captured trace can be replayed correct. But replay a full-trace capture shows only a grey plane and the text in the left-top corner.

PR #921, opened 2017-08-16
Fix a bug in Memory tracker

Qining:
When removing a range from tracking, if the range shares pages with
other range, the shared page should not be granted with RW permission.

Also add a test for it.

Two other tests are updated too, as the underlying behaviour changed.


Qining:
Fix issue #872

PR #923, opened 2017-08-16
Default the 'disable PCS' option to true.

pmuetschard:


gapic/src/main/com/google/gapid/models/Settings.java
```diff
@@ -246,12 +246,12 @@ private static long getLong(Properties properties, String name, long dflt) {
     }
   }
 
-  private static boolean getBoolean(Properties properties, String name) {
-    return "true".equalsIgnoreCase(properties.getProperty(name, ""));
-  }
-
   private static boolean getBoolean(Properties properties, String name, boolean dflt) {
-    return "true".equalsIgnoreCase(properties.getProperty(name, dflt ? "true" : "false"));
+    if (dflt) {
```
ben-clayton: This implementation is making me go cross eyed trying to figure out the logic. What was wrong with the old implementation?

pmuetschard: It is now consistent (it was slightly inconsistent before):
if `dflt` is `true` only a value of `"false"` will cause the result to be `false`, similar to how if `dflt` is `false`, only `"true"` will cause the result to be `true`.

PR #925, opened 2017-08-16
Do not bind context if it is already bound on different thread.

dsrbecky:


gapis/api/gles/api/egl.api
```diff
@@ -258,32 +258,36 @@ cmd EGLBoolean eglMakeCurrent(EGLDisplay display,
   if (context == null) {
     SetContext(null)
   } else if context in EGLContexts {
-      ctx := EGLContexts[context]
+    ctx := EGLContexts[context]
+    if ctx.Other.IsBound {
```
dsrbecky: (ws changes only, except for this "if")

PR #919, opened 2017-08-15
Fixes for memory various memory requests

pmuetschard:
- Return a DataUnavailable error when an invalid pool is requested
- Allow requesting of memory of a pool at the command that created the pool
- Fix the UI to handle errors when fetching memory

Fixes #914

PR #926, opened 2017-08-16
Remove unused import and fix build issue from #919

pmuetschard:
I don't understand how the presubmit didn't fail on this...


pmuetschard:
Never mind, the presubmit was green, but the build is red, because there was a merge. Sigh. If only the UI would show this fact.

PR #932, opened 2017-08-17
Make a copy for in-place sorting

Qining:
To make the thing safer.

### DIFF
--- a/gapis/resolve/command_tree.go
+++ b/gapis/resolve/command_tree.go
@@ -352,8 +352,9 @@ func (r *CommandTreeResolvable) Resolve(ctx context.Context) (interface{}, error
 		if v, ok := snc.SubcommandGroups[id]; ok {
 			r := out.root.AddRoot([]uint64{uint64(id)})
 			// subcommands are added before nested SubCmdRoots.
-			sort.SliceStable(v, func(i, j int) bool { return len(v[i]) < len(v[j]) })
-			for _, x := range v {
+			cv := append([]api.SubCmdIdx{}, v...)
+			sort.SliceStable(cv, func(i, j int) bool { return len(cv[i]) < len(cv[j]) })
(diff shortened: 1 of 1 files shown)


gapis/resolve/command_tree.go
```diff
@@ -342,8 +342,9 @@ func (r *CommandTreeResolvable) Resolve(ctx context.Context) (interface{}, error
 	for k, v := range snc.SubcommandGroups {
 		r := out.root.AddRoot([]uint64{uint64(k)})
 		// subcommands are added before nested SubCmdRoots.
-		sort.SliceStable(v, func(i, j int) bool { return len(v[i]) < len(v[j]) })
-		for _, x := range v {
+		cv := append([]api.SubCmdIdx{}, v...)
+		sort.SliceStable(cv, func(i, j int) bool { return len(cv[i]) < len(cv[j]) })
```
ben-clayton: Why don't we just insist on having the provider of the `SubcommandGroups` sort things?

Qining: Actually the provider (currently this is only populated in vulkan.go) generates the subcommand index in the execution order.

ben-clayton: So... do we need this change?

Qining: The following sorting changes the order in-place. So we think it might be safer to add a copy.

e.g. original order:
[411, 0, 3, 2]
[411, 0, 4, 3, 0, 5]
[411, 0, 5, 4]

after the following sorting:
[411, 0, 3, 2]
[411, 0, 5, 4]
[411, 0, 4, 3, 0, 5]

So far the sorting is not causing any issue, but might be in future, some uses of the SubComandGroups may assume the execution order (hmm, but the execution order can be recovered by another sorting...)

Generally this is just make it safer.

PR #933, opened 2017-08-17
Robot: Fix GAPIS busy

baldwinn860:
This moves retry logic from os.shell and into robot.job.worker. This way the actual output of the underlying call can be checked to see if we need to perform a retry. This is still a hack to get around the actual problem, but this bug seems to be too subtle to spend any more time on when it only happens in the test infra. There is logging to provide insight into how many retries are attempted before the task eventually succeeds.

### DIFF
--- a/core/os/shell/local.go
+++ b/core/os/shell/local.go
@@ -17,8 +17,6 @@ package shell
 import (
 	"context"
 	"os/exec"
-	"strings"
-	"time"
 
 	"github.com/google/gapid/core/event/task"
 	"github.com/google/gapid/core/log"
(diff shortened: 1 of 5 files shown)


core/os/shell/local.go
```diff
@@ -58,17 +56,17 @@ func (localTarget) Start(cmd Cmd) (Process, error) {
 	// first child time to close-on-exec all of it's fds.
 	// fix source: https://groups.google.com/forum/#!topic/golang-dev/4efaTJ5uA8Y
 	// context: https://golang.org/src/syscall/exec_unix.go?h=StartProcess#L17
-	for {
-		err := p.exec.Start()
+	// for {
```
ben-clayton: Please don't commit commented-out code. Remove it and rely on version control.

test/robot/replay/client.go
```diff
@@ -44,17 +45,22 @@ func (c *client) replay(ctx context.Context, t *Task) error {
...
-	output, err := doReplay(ctx, t.Action, t.Input, c.store, c.tempDir)
-	status := job.Succeeded
-	if err != nil {
-		status = job.Failed
-		log.E(ctx, "Error running replay: %v", err)
-	} else if output.CallError != "" {
-		status = job.Failed
-		log.E(ctx, "Error during replay: %v", output.CallError)
-	}
+	for {
```
ben-clayton: Bail after N failed attempts, and add a sleep between retries (constant, linear, exponential - don't mind).
Could make a nice `task` package utility function:

```go
// Retry repeatedly calls task until task returns a nil error or the number
// attempts reaches maxAttempts. Retry will sleep for retryDelay between retry
// attempts.
func Retry(ctx context.Context, maxAttempts int, retryDelay time.Duration, task Task) error {
	var err error
	for i := 0; i < maxAttempts; i++ {
		if err = task(ctx); err == nil {
			return nil
		}
		time.Sleep(retryDelay)
	}
	return err
}
```

test/robot/replay/client.go
```diff
@@ -44,17 +45,22 @@ func (c *client) replay(ctx context.Context, t *Task) error {
...
+		output, err := doReplay(ctx, t.Action, t.Input, c.store, c.tempDir)
+		status := job.Succeeded
+		if err != nil {
+			status = job.Failed
+			log.E(ctx, "Error running replay: %v", err)
+		} else if output.CallError != "" {
+			if strings.Contains(output.CallError, "text file busy") {
```
ben-clayton: Put this grossness in a utility function please - or declare a new `ErrTextFileBusy` error type that's returned from `shell.Target.Start()` which we can check for.

Having three magic hardcoded strings makes me three times as sad as having one.

test/robot/report/client.go
```diff
@@ -44,17 +45,22 @@ func (c *client) report(ctx context.Context, t *Task) error {
...
-	output, err := doReport(ctx, t.Action, t.Input, c.store, c.tempDir)
-	status := job.Succeeded
-	if err != nil {
-		status = job.Failed
-		log.E(ctx, "Error running report: %v", err)
-	} else if output.CallError != "" {
-		status = job.Failed
-		log.E(ctx, "Error during report: %v", output.CallError)
-	}
+	for {
```
ben-clayton: Ditto.

core/os/shell/local.go
ben-clayton: I see you've removed it in a later CL. Good. It probably makes sense to squash temporary commits like this before reviewing.

test/robot/replay/client.go
```diff
@@ -44,7 +46,11 @@ func (c *client) replay(ctx context.Context, t *Task) error {
 	if err := c.manager.Update(ctx, t.Action, job.Running, nil); err != nil {
 		return err
 	}
-	output, err := doReplay(ctx, t.Action, t.Input, c.store, c.tempDir)
+	var output *Output
+	err := worker.RetryFunction(ctx, 4, time.Duration(time.Millisecond*100), func() (err error) {
```
ben-clayton: Nit: No need to cast to duration. It already is one.

ben-clayton:
FYI: https://github.com/google/gapid/pull/950/commits/e472288d284353813a171fa1857e801ce01bf623

PR #942, opened 2017-08-21
Add codegen rules for the GVR api.

ben-clayton:
Doesn't actually do anything yet as nothing calls into it.

TODO: gapii codegen / hooks.


ben-clayton:
Now builds. PTAL.

PR #945, opened 2017-08-21
Escape "\" in gapir log path.

AWoloszyn:
This was causing the windows log to get written to C:UsersAndrew....

### DIFF
--- a/gapic/src/main/com/google/gapid/server/GapisProcess.java
+++ b/gapic/src/main/com/google/gapid/server/GapisProcess.java
@@ -96,7 +96,7 @@ public class GapisProcess extends ChildProcess<Integer> {
       args.add("-log-level");
       args.add(Logging.getGapisLogLevel());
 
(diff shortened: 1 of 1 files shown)


gapic/src/main/com/google/gapid/server/GapisProcess.java
```diff
@@ -96,7 +96,7 @@ protected Exception prepare(ProcessBuilder pb) {
       args.add("-log-level");
       args.add(Logging.getGapisLogLevel());
 
-      gapirFlags = "--log " + new File(logDir, "gapir.log").getAbsolutePath() +
+      gapirFlags = "--log " + new File(logDir, "gapir.log").getAbsolutePath().replace("\\", "\\\\") +
```
ben-clayton: Ah, this is getting escaped by `text.SplitArgs`. It took me a while to figure out why this was needed.

PR #955, opened 2017-08-21
gapis/api/sync: Don't explode if GetTerminator returns a nil.

ben-clayton:


gapis/api/sync/sync.go
```diff
@@ -81,7 +81,9 @@ func MutationCmdsFor(ctx context.Context, c *path.Capture, cmds []api.Cmd, id ap
 			if err != nil {
 				return nil, err
 			}
-			terminators = append(terminators, term)
+			if term != nil {
```
AWoloszyn: You probably want to insert the EarlyTerminator in this case (or return it from GetTerminator), otherwise all commands for this API will be kept even if you only want a subset of the commands.

ben-clayton: Done.

PR #956, opened 2017-08-21
api: Remove Span.itemIndex method.

ben-clayton:
It was poorly named and was confusing. Instead just special-case the one place it was used.


ben-clayton:
Windows build was a random build machine failure.

Issue #960, opened 2017-08-22
Meshes don't typically show for Unreal 4 titles


ben-clayton:
GAPIS currently uses name pattern matching to find the 'position' stream.
It would seem that Unreal obfuscates their shader code, including the vertex attribute names:

```
0: in_ATTRIBUTE2, XYZW_U8_NORM
1: in_ATTRIBUTE15, XY_F16
2: in_ATTRIBUTE4, XYZW_F16
3: in_ATTRIBUTE0, XYZ_F32
4: in_ATTRIBUTE1, XYZW_U8_NORM
```

We could get GAPIS to fallback to looking for `XYZ_F32` streams for position, but that's quite a risky solution.

I still like the idea of analyzing the shader to infer the attributes that contributed to `gl_Position`, but that's a fair amount of work.

We should just let the user pick the streams they want to visualize, with some sort of sensible defaults.

dsrbecky:
B is not entirely insane if we use SPIR-V

ben-clayton:
I've tried a few Unreal titles now which seem to mostly work. Removing from milestone.

PR #961, opened 2017-08-22
Guess position & normal streams when name matching fails.

ben-clayton:
Issue: #960


ben-clayton:
> We do something similar for Vulkan that may be more robust.

Good to know. I think if we did shader analysis in SPIRV we could both benefit from the proposed approach in #960.

PR #974, opened 2017-08-22
Migrate build dimension

baldwinn860:
This change removes the gapit and gapid_apk dimensions from the grid and adds a single package dimension. There is a bit of fallout from this change, workers now must match the package that created the input (i.e. replay/report must have the same package that created the tracefile) and packages are treated as complete sets of tools rather than piecemeal subsets.

### DIFF
--- a/test/robot/web/client/data.go
+++ b/test/robot/web/client/data.go
@@ -170,7 +170,7 @@ func clearDimensionData() {
 func newTask(entry map[string]interface{}, kind Item) *task {
 	t := &task{
 		underlying: entry,
-		trace:      traceInfo{target: nilItem, pkg: nilItem, subject: nilItem},
+		trace:      traceInfo{target: nilItem, subject: nilItem},
 		kind:       kind,
 		host:       nilItem,
 		pkg:        nilItem,
--- a/test/robot/web/client/data.go
+++ b/test/robot/web/client/data.go
@@ -21,6 +21,7 @@ import (
 	"html/template"
 	"io/ioutil"
 	"net/http"
+	"reflect"
 	"time"
 
 	"os"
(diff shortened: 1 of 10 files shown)


test/robot/web/client/data.go
```diff
@@ -48,50 +67,39 @@ var (
...
 	}
 
-	dimensions = []*dimension{subjectDimension, targetDimension, hostDimension, kindDimension, gapidApkDimension, gapitDimension}
+	dimensions = []*dimension{kindDimension, subjectDimension, targetDimension, hostDimension, packageDimension}
 )
 
-func itemGetter(idPattern string, displayPattern string) func([]interface{}) enum {
+func isUserType(t reflect.Value) bool {
+	// cannot currently use build.Type_UserType to check the type need to fix that.
+	return t.Kind() == reflect.Float64 && t.Float() == float64(2)
```
baldwinn860: I need to get the actual value for build.Type_UserType, but cannot import the build package as gopherjs doesn't allow importing packages that use cgo, What should we do in this instance?

ben-clayton: I assume the cgo code is in the device package?
You could try using the [build constraint](https://golang.org/pkg/go/build/#hdr-Build_Constraints) `// -build js` in the go file that has the `import "C"` line.

baldwinn860: After looking into this I couldn't get any build constraints to keep GopherJS from trying to build the file that imports the cgo code. One idea I had was to build the proto files into javascript as well, which would not only allow this type of value checking, but would let us use the actual type layouts rather than using text templating and json string to interface mappings to read data. This would probably be another PR though.

Issue #973, opened 2017-08-22
Update glslang to latest


ben-clayton:
Attempting to compile a shader using GLSL 320 ES:

```
ERROR: #version: only version 300 and 310 support the es profile
ERROR: 0:3: 'location' : not supported for this version or the enabled extensions 
ERROR: 0:3: '' : compilation terminated 
ERROR: 3 compilation errors.  No code generated.
```

It looks like support for 320 ES shaders was [added by this CL](https://github.com/KhronosGroup/glslang/commit/53863a3a90be679db4f013e19b54bc21dbf0e113).
Let's update our third_party dependency to get this goodness.

PR #975, opened 2017-08-22
Flesh out support for GLSL 3.2 (#version 320)

ben-clayton:
Fixes: #973

### DIFF
--- a/gapis/api/gles/compat.go
+++ b/gapis/api/gles/compat.go
@@ -440,12 +440,18 @@ func compat(ctx context.Context, device *device.Instance) (transform.Transformer
 			src := ""
 
 			if config.UseGlslang {
-				opts := shadertools.Option{
-					IsFragmentShader: shader.Type == GLenum_GL_FRAGMENT_SHADER,
-					IsVertexShader:   shader.Type == GLenum_GL_VERTEX_SHADER,
+				st, err := shader.Type.ShaderType()
+				if err != nil {
+					log.W(ctx, "%v compat: %v", cmd, err)
 				}
+				opts := shadertools.Option{ShaderType: st}
 
-				res, err := shadertools.ConvertGlsl(shader.Source, &opts)
+				// Trim any prefix whitespace / newlines.
+				// This isn't legal if it comes before the #version, but this
+				// will be picked up by find_issues.go anyway.
+				src = strings.TrimLeft(shader.Source, "\n\r\t ")
+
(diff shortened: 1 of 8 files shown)


gapis/api/gles/compat.go
```diff
@@ -440,12 +440,18 @@ func compat(ctx context.Context, device *device.Instance) (transform.Transformer
 			src := ""
 
 			if config.UseGlslang {
-				opts := shadertools.Option{
-					IsFragmentShader: shader.Type == GLenum_GL_FRAGMENT_SHADER,
-					IsVertexShader:   shader.Type == GLenum_GL_VERTEX_SHADER,
+				st, err := shader.Type.ShaderType()
+				if err != nil {
+					log.W(ctx, "%v compat: %v", cmd, err)
```
pmuetschard: "%v compat: Could not determine shader type: %v" ? The warning seems a bit terse to me otherwise.

gapis/api/gles/glsl.go
```diff
@@ -0,0 +1,40 @@
...
+	case GLenum_GL_TESS_EVALUATION_SHADER:
+		return shadertools.TypeTessEvaluation, nil
+	case GLenum_GL_GEOMETRY_SHADER:
+		return shadertools.TypeGeometry, nil
+	case GLenum_GL_FRAGMENT_SHADER:
+		return shadertools.TypeFragment, nil
+	case GLenum_GL_COMPUTE_SHADER:
+		return shadertools.TypeCompute, nil
+	}
+	return 0, fmt.Errorf("%v is not a shader", e)
```
pmuetschard: 0 here will mean vertex shader, correct? Is that on purpose?

gapis/api/gles/compat.go
ben-clayton: That's pretty much what you'll see as the `err` returned from `ShaderType()` will contain the unknown error.

gapis/api/gles/glsl.go
ben-clayton: What else do you put there? The function returns a value and an error. If the error is non `nil` then you shouldn't be using the value. 0 is the 'default' value for an integer and is as good as any other value that shouldn't be used.

pmuetschard: Ben, I'm sorry if I'm missing something. To me it looks like the result is being used in the error case, which is why I was asking about whether the default is on purpose.

PR #981, opened 2017-08-23
Show the actual generated trace file name in the trace dialog.

### LINKED ISSUE #905: Clear up time stamp added to trace files
It is not obvious from the UI when the time stamps is added to the filename.

pmuetschard:
Fixes #905

### DIFF
--- a/gapic/src/main/com/google/gapid/models/Settings.java
+++ b/gapic/src/main/com/google/gapid/models/Settings.java
@@ -63,7 +63,6 @@ public class Settings {
   public String traceDevice = "";
   public String tracePackage = "";
   public String traceOutDir = "";
-  public String traceOutFile = "";
   public boolean traceClearCache = false;
   public boolean traceDisablePcs = true;
   public String traceExecutable = "";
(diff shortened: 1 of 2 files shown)


ben-clayton:
Ideally I think we should let the user format the file how they like.
Personally I _really_ dislike having the timestamp added to the file name. If I need to know when the file was created, I look at the timestamp on the file.

How easy would it be to let the user put the data specifier in the filename?

Examples:

`my_trace_no_timestamp.gfxtrace`
`my_trace_{date}_{time}.gfxtrace`
`my_trace_{year}{month}{day}_{hour}{minute}.gfxtrace`

pmuetschard:
> Ideally I think we should let the user format the file how they like.

You can still modify the filename, of course. This just changes how the automatically generated name (from the activity or exe) is shown. Before this change, it would not show the timestamp, but add it, if (and only if) the user never changed the filename from the automatically suggested one.

> If I need to know when the file was created, I look at the timestamp on the file.

The timestamp is more there ensure the filename is unique and do so in a more meaningful way then 1, 2, 3..

> How easy would it be to let the user put the data specifier in the filename?

Are you asking for a setting for a pattern of the automatically generated name? We can certainly do that. We should track it as a separate feature request, though, all I'm trying to do here is get rid of the hidden behavior that leads to confusion.

ben-clayton:
> We should track it as a separate feature request

Sure, let me play with this a bit first. I'll see if this fixes my problem with collecting 1000's of `.gfxtrace` files in a directory and then having to manually brain-parse each file name for the most recent one.

PR #982, opened 2017-08-23
Add GVR support

ben-clayton:


gapii/cc/spy.cpp
```diff
@@ -185,8 +188,12 @@ Spy::Spy()
         mCaptureFrames = header.mNumFrames;
         mSuspendCaptureFrames.store((header.mFlags & ConnectionHeader::FLAG_DEFER_START)?
             kSuspendIndefinitely: mSuspendCaptureFrames.load());
+        if (header.mGvrHandle != 0) {
```
ben-clayton: yup. All because it's near impossible to obtain the [`android.context.Context`](https://developer.android.com/reference/android/content/Context.html) from JNI, which is needed to call the GVR loader to get the GVR handle. 😞

PR #993, opened 2017-08-24
Remove protobuf output file.

baldwinn860:
Shouldn't be checked in.


baldwinn860:
done.

.gitignore
```diff
@@ -13,6 +13,7 @@
 *_embed.go
 gapir.log
 gapis.log
+*.pb.go
```
pmuetschard: Thus far, the pattern has been to add the file by name to a .gitignore file in the same dir.

baldwinn860: oh that makes sense, fixed

gapis/api/gvr/.gitignore
```diff
@@ -1 +1,2 @@
 /api.pb.go
+resolvables.pb.go
```
pmuetschard: Add the slash in the front, it means to only exclude it in the current dir rather than in it and it's children.

baldwinn860: done

PR #994, opened 2017-08-24
Add VK_EXT_debug_marker

Qining:
Enables extension when tracing and show the commands in the trace.


Qining:
Tested based on previous commits, as the current master branch (7082544b7a20951dffcc7efefc2c993653b027d2) is not working for Vulkan (some samples replay crash, some samples server crash)

PR #995, opened 2017-08-25
gles: Fix panic in read_framebuffer if there's no bound fb

ben-clayton:


dsrbecky:
I think this should never happen though....

Issue #996, opened 2017-08-25
Display message when attempting to open a too old / new file


ben-clayton:
Currently we just log a bunch of `LoadData` errors, and the UI doesn't update.

PR #998, opened 2017-08-25
Display a sensible error if the file version is too old / new to be read

ben-clayton:
Fixes: #996


ben-clayton:
Would be nice to show a dialog in this case as the error message is displayed in the filmstrip view, which may be hidden.

PR #1006, opened 2017-08-27
gapit: added earlier encoder warning

sjfricke:
It is very disheartening to wait 15 minutes for your trace to finish running `gapit video` and then find out you didn't have a valid encoder in your path, simply checking prior to doing all the actual video logic will save someone a lot of time in the future

### DIFF
--- a/core/video/encoder.go
+++ b/core/video/encoder.go
@@ -40,11 +40,20 @@ func init() {
 	}
 }
 
+// Checks to make sure an encoder has been set
+func GetEncoder() (string, error) {
+	if encoder == "" {
+		return encoder, fmt.Errorf("neither avconv nor ffmpeg was found")
+	} else {
+		return encoder, nil
+	}
+}
+
 // Encode will encode the frames written to the returned chan to a video that
 // can be read from the Reader.
 func Encode(ctx context.Context, settings Settings) (chan<- image.Image, io.Reader, error) {
 	if encoder == "" {
-		return nil, nil, fmt.Errorf("neither avconv or ffmpeg was found")
+		return nil, nil, fmt.Errorf("neither avconv nor ffmpeg was found")
 	}
 
 	in := make(chan image.Image, 64)
--- a/cmd/gapit/video.go
+++ b/cmd/gapit/video.go
@@ -211,6 +211,11 @@ func (verb *videoVerb) Run(ctx context.Context, flags flag.FlagSet) error {
 		return log.Errf(ctx, err, "Finding file: %v", flags.Arg(0))
 	}
 
+	_, err = video.GetEncoder()
+	if err != nil {
+		return log.Errf(ctx, err, "No valid encoder found in path")
+	}
+
 	client, err := getGapis(ctx, verb.Gapis, verb.Gapir)
 	if err != nil {
 		return log.Err(ctx, err, "Failed to connect to the GAPIS server")


core/video/encoder.go
```diff
@@ -40,11 +40,20 @@ func init() {
 	}
 }
 
+// Checks to make sure an encoder has been set
+func GetEncoder() (string, error) {
```
ben-clayton: It's more common in Go to omit the `Get` prefix.

core/video/encoder.go
```diff
@@ -40,11 +40,20 @@ func init() {
 	}
 }
 
+// Checks to make sure an encoder has been set
```
ben-clayton: Godoc comments [start with name of the symbol being documented](https://blog.golang.org/godoc-documenting-go-code).
Also given that the function returns more than just an error, then we should describe what the function does as opposed to its current usage. Example:

```go
// Encoder returns the path to either the 'avconv' or 'ffmpeg' executable used
// for video encoding. If neither can be found on PATH then an error is
// returned.
```

core/video/encoder.go
```diff
@@ -40,11 +40,20 @@ func init() {
...
+		return encoder, fmt.Errorf("neither avconv nor ffmpeg was found")
+	} else {
+		return encoder, nil
+	}
+}
+
 // Encode will encode the frames written to the returned chan to a video that
 // can be read from the Reader.
 func Encode(ctx context.Context, settings Settings) (chan<- image.Image, io.Reader, error) {
 	if encoder == "" {
```
ben-clayton: Use your new function here?

```go
encoder, err := Encoder()
if err != nil {
	return nil, nil, err
}
```

sjfricke: See the whole point is if I don't have an encoder and it spends 15 minutes getting the images ready and then it runs `Encode` it will be a waste of 15 minutes to find out you don't have it. The idea is to check prior. I am not sure if it would make sense to keep the check here as well

ben-clayton: > See the whole point is if I don't have an encoder and it spends 15 minutes getting the images ready and then it runs Encode it will be a waste of 15 minutes to find out you don't have it. 

Understood.

> The idea is to check prior. I am not sure if it would make sense to keep the check here as well

There's nothing stopping a user of the `video` package not calling your new function and just calling `Encode` as we did before. The existing code already checks for `encoder` being empty and returning an identical error message - I'm simply proposing we avoid a duplicated bit of logic and duplicated hard-coded string.

ben-clayton: > I am not sure if it would make sense to keep the check here as well

Fair enough - if that's the case then maybe we should have a single function that does the check, and returns an `Encoder` instance which holds the `Encode` method (or returns the error message). That way there's no duplication of error logic, and you can't forget to call the function that checks there's an encoder.

core/video/encoder.go
```diff
@@ -40,11 +40,20 @@ func init() {
 	}
 }
 
+// Checks to make sure an encoder has been set
+func GetEncoder() (string, error) {
+	if encoder == "" {
+		return encoder, fmt.Errorf("neither avconv nor ffmpeg was found")
+	} else {
```
ben-clayton: Drop the `else` - govet moans if the true block will always return and you have an `else`:

```go
func Encoder() (string, error) {
	if encoder == "" {
		return encoder, fmt.Errorf("neither avconv nor ffmpeg was found")
	}
	return encoder, nil
}
```

PR #1012, opened 2017-08-28
Vulkan: Set object name and tag.

Qining:


Qining:
In vulkan_mid_execution.cpp, the real changes are:
1) add function : `recreateDebugInfo`
2) add function calls to `recreateDebugInfo` after each recreated Vulkan handle.
All other changes in that file are done by clang-format

In vulkan.api, the reason to have two `switch..case` list is that, I cannot get the reference of the DebugInfo pointer in objects in API file.

gapii/cc/vulkan_mid_execution.cpp
```diff
@@ -153,1690 +156,1780 @@ class TemporaryRenderPass {
...
+                        }) != temporary_render_passes_.end();
   }
 
  private:
   CallObserver* observer_;
   VulkanSpy* spy_;
   std::vector<std::shared_ptr<RenderPassObject>> temporary_render_passes_;
 };
+
+template <typename ObjectClass>
```
Qining: Real change starts from here.

gapii/cc/vulkan_mid_execution.cpp
```diff
@@ -153,1690 +156,1780 @@ class TemporaryRenderPass {
...
+      v.mpQueuePriorities = queue_priorities[i].data();
+    }
+    create_info.mqueueCreateInfoCount = queue_create_infos.size();
+    create_info.mpQueueCreateInfos = queue_create_infos.data();
+    create_info.mpEnabledFeatures = &device.second->mEnabledFeatures;
+
+    VkDevice d = device.second->mVulkanHandle;
+    VkPhysicalDevice pd = device.second->mPhysicalDevice;
+    RecreateDevice(observer, pd, &create_info, &d);
+    recreateDebugInfo(
```
Qining: For each recreate vulkan handle, createDebugInfo() is called, although I don't think debug info can be attached to VkInstance, VkPhysicalDevice, VkSurface.

gapis/api/vulkan/vulkan.api
```diff
@@ -8700,6 +8700,229 @@ class VkDebugMarkerMarkerInfoEXT {
...
+	  	Surfaces[as!VkSurfaceKHR(obj)].DebugInfo = updateDebugMarkerObjectName(
+          Surfaces[as!VkSurfaceKHR(obj)].DebugInfo, nameInfo)
+	  case VK_DEBUG_REPORT_OBJECT_TYPE_SWAPCHAIN_KHR_EXT:
+	  	Swapchains[as!VkSwapchainKHR(obj)].DebugInfo = updateDebugMarkerObjectName(
+          Swapchains[as!VkSwapchainKHR(obj)].DebugInfo, nameInfo)
+  }
+}
+
+@override
+@no_replay
```
Qining: Will be changed later to 'custom' once we start to use the debug markers for grouping

gapis/api/vulkan/vulkan.api
```diff
@@ -8709,10 +8932,20 @@ class VkDebugMarkerMarkerInfoEXT {
 cmd VkResult vkDebugMarkerSetObjectTagEXT(
```
Qining: Also will be changed to 'custom' in later CLs

gapii/cc/vulkan_mid_execution.cpp
```diff
@@ -153,1690 +156,1780 @@ class TemporaryRenderPass {
...
+
+template <>
+VkDevice getObjectCreatingDevice<SurfaceObject>(
+    std::shared_ptr<SurfaceObject> obj) {
+  return VkDevice(0);
+}
+
+template <typename ObjectClass>
+void recreateDebugInfo(VulkanSpy* spy, CallObserver* observer,
+                       uint32_t objectType, std::shared_ptr<ObjectClass> obj) {
```
AWoloszyn: You should probably make this a const std::shared_ptr<>& obj, unless you need to modify the object.
shared_ptr copies are more expensive than we want to pay if we don't have to.

Qining: Changed to const ref in a83cc08

PR #1014, opened 2017-08-29
Native bridge

ben-clayton:


core/os/android/adb/device.go
```diff
@@ -254,3 +254,53 @@ func parseDevices(ctx context.Context, out string) (map[string]bind.Status, erro
 	}
 	return devices, nil
 }
+
+// NativeBridgeABI returns the native ABI for the given emulated ABI for the
+// device by consulting the ro.dalvik.vm.isa.<emulated_isa>=<native_isa>
+// system properties.
```
pmuetschard: I think this should also say that it returns the passed in ABI if there is no emulation.

pmuetschard: never mind... for some reason GH was showing me out of date information and this PR was still open.

ben-clayton: Still a good point. Will fix shortly.

ben-clayton: https://github.com/google/gapid/pull/1016/files

PR #1019, opened 2017-08-29
Mesh Statistics

### LINKED ISSUE #904: Geometry view should show some basic geo stats

pmuetschard:
Fixes #904


AWoloszyn:
I checked this on a Vulkan sample, and it seems to be doing the right thing.

Issue #1026, opened 2017-08-30
[installer.cpp:140] Couldn't resolve the interceptor library.


stcdalyc:
![qq 20170830210819](https://user-images.githubusercontent.com/6738727/29873839-63c6b92e-8dc7-11e7-9197-89d3c40bb423.png)

the main error is：
I/GAPID   (32404): [installer.cpp:132] Installing GAPII hooks...
F/GAPID   (32404): [installer.cpp:140] Couldn't resolve the interceptor library.

my device is Redmi 3
related logcat:
[logcat.txt](https://github.com/google/gapid/files/1263660/logcat.txt)

ben-clayton:
Hi @stcdalyc,

Are you able to build from source? If so, are you able to test whether e33a6a1 fixes your issue?

Many thanks for the bug report.
Ben

stcdalyc:
sorry, i can't build successfully.
see #1040

stcdalyc:
@ben-clayton
I'm sure that it has been fixed

ben-clayton:
Okay @stcdalyc. Please reopen if you encounter the issue again.

stcdalyc:
@ben-clayton 
Actually, i can't reopen it, because in github, 

> you cannot re-open your own issues if a repo collaborator closed them

ben-clayton:
Oh, that's a shame. Would you like me to reopen it?

stcdalyc:
No need, because the problem has been fixed

PR #1028, opened 2017-08-30
gapii: Pass the path of libinterceptor down in the header.

ben-clayton:
This moves the loading of `libinteceptor.so` from `jdwp_loader.go` (gapit) to gapii.

It appears that on some devices dlopen() will not find libraries loaded by a prior call to `System.loadLibrary()`. Instead, pass the absolute path of `libinteceptor.so` down to `libgapii.so` and load it directly there.

Fixes: #1026

### DIFF
--- a/core/event/task/task.go
+++ b/core/event/task/task.go
@@ -35,17 +35,17 @@ func Once(task Task) Task {
 	}
 }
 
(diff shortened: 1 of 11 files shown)


AWoloszyn:
Tested on 2 devices with Vulkan. Works for me

PR #1030, opened 2017-08-30
core/image: Flip ASTC decoded images.

ben-clayton:
The tests were wrong because the encoder I used to build the unit tests silently flips the image...

Fixes: #1020


ben-clayton:
Thanks!

PR #1044, opened 2017-08-31
Change the translated 'string' type from 'const char*' to 'std::string' for extern function args

Qining:


gapis/api/templates/cpp_common.tmpl
```diff
@@ -178,6 +178,7 @@
 {{define "C++.CallParameters"}}
   {{AssertType $ "Function"}}
 
+  {{Global "CallFunction" $}}
```
ben-clayton: This will break if you have a call in a call. Use hacky stack to fix this:

```
  {{$prevCallFunction := Global "CallFunction"}}
  {{Global "CallFunction" $}}
  {{ForEach $.CallParameters "C++.Parameter" | JoinWith ", "}}
  {{Global "CallFunction" $prevCallFunction}}
```

gapis/api/templates/specific_gfx_api.cpp.tmpl
```diff
@@ -97,6 +97,7 @@
 
 {{define "CommandHandler"}}
   {{AssertType $ "Function"}}
+  {{Global "CallFunction" $}}
```
ben-clayton: And here

gapis/api/templates/cpp_common.tmpl
Qining: added in 0535813

gapis/api/templates/specific_gfx_api.cpp.tmpl
Qining: added in 0535813

PR #1047, opened 2017-09-01
Add Package Ordering to robot website

baldwinn860:
We want to be able to see at a glance the state of our packages in relation to their predecessors. These changes add several features to the webserver to get us that support. 
 - Tasks know which tasks were spawned from their parent packages.
 - Tasks update their status to "Stale" when they know the status and result of their parents, but not themselves. (as desaturated colors)
 - Tasks update their status to "Changed" when they have a different result than their parents. (as bright colors)
 - Packages are ordered by their ancestry with the newest package listed last.

### DIFF
--- a/test/robot/web/client/widgets/grid/data.go
+++ b/test/robot/web/client/widgets/grid/data.go
@@ -93,6 +93,9 @@ const (
 	// InProgress represents a task that is currently being run.
 	// The task's result will be for data that is not current.
 	InProgress
+
+	// Changed represents a task with a result for the latest data that is different from its stale data
+	Changed
 )
 
 // TaskList is a list of tasks.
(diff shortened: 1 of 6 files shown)


test/robot/web/client/widgets/grid/data.go
```diff
@@ -114,20 +117,24 @@ func (l TaskList) stats() taskStats {
 		numCurrentSucceeded:       l.Count(taskCurrentSucceeded),
 		numStaleSucceeded:         l.Count(taskStaleSucceeded),
 		numInProgressWasSucceeded: l.Count(taskInProgressWasSucceeded),
+		numFixed:                  l.Count(taskFixed),
```
ben-clayton: nit: I'd stick with the existing naming for consistency's sake:
`numSucceededWasFailed`, etc.

Issue #1049, opened 2017-09-01
Failed single frame app trace cause system wide Vulkan crashes


sjfricke:
So I have found this to be for both my old Glaxay S6 and new Glaxay S8 when I try running the [Android Vulkan Tutorials part 5](https://github.com/googlesamples/android-vulkan-tutorials/tree/master/tutorial05_triangle) with:
`do run gapit trace -api vulkan -out <gfxtrace_path> -capture-frames 1 -gapii-device <device_id> -android-package com.google.vulkan.tutorials.five -android-activity android.app.NativeActivity -android-action android.intent.action.MAIN]`

I get 
```
I: [start] <gapit> Connecting to JDWP
F: <gapit> Main failed
Error: Connecting to JDWP
   Cause: EOF
Error running /home/gapid/debug/bin/gapit [trace -api vulkan -out <gfxtrace_path> -capture-frames 1 -gapii-device <device_id> -android-package com.google.vulkan.tutorials.five -android-activity android.app.NativeActivity -android-action android.intent.action.MAIN]: Process returned error
   Cause: exit status 1
exit status 1
```

I then realize all my Vulkan apps crash on trying to make a VkInstance and running the Vulkan CapsViewer App I get `Could not create Vulkan instance! Error: ERROR_LAYER_NOT_PRESENT`

So after restarting my phone all goes back to normal.

sjfricke:
Here is the logcat I pull, it's a debug build of gapid from 2174b0e7bf249878b5cd64f9272cef3818a42914 and this is what it gave

```
I ActivityManager: Start proc 12178:com.google.vulkan.tutorials.five/u0a248 for activity com.google.vulkan.tutorials.five/android.app.NativeActivity
V BoostFramework: BoostFramework() : mPerf = com.qualcomm.qti.Performance@4eaedab
E Zygote  : v2
I libpersona: KNOX_SDCARD checking this for 10248
I libpersona: KNOX_SDCARD not a persona
E Zygote  : accessInfo : 0
W SELinux : SELinux selinux_android_compute_policy_index : Policy Index[2],  Con:u:r:zygote:s0 RAM:SEPF_SECMOBILE_7.0_0009, [-1 -1 -1 -1 0 1]
I SELinux : SELinux: seapp_context_lookup: seinfo=untrusted, level=s0:c512,c768, pkgname=com.google.vulkan.tutorials.five 
[...]

PR #1051, opened 2017-09-01
Fixed setprop after using VkGraphicsSpy

sjfricke:
#1049 

I found this actually lets me trace the [Android Vulkan Tutorials part 5](https://github.com/googlesamples/android-vulkan-tutorials/tree/master/tutorial05_triangle) and also doesn't break all other Vulkan apps. 

I feel this should be verified across multiple devices/OS before merging

### DIFF
--- a/gapii/client/adb.go
+++ b/gapii/client/adb.go
@@ -115,7 +115,7 @@ func StartOrAttach(ctx context.Context, p *android.InstalledPackage, a *android.
 		// Clone context to ignore cancellation.
 		ctx := keys.Clone(context.Background(), ctx)
 		d.RemoveForward(ctx, port)
-		d.Command("shell", "setprop", "debug.vulkan.layers", "\"\"").Run(ctx)
+		d.Command("shell", "setprop", "debug.vulkan.layers", "").Run(ctx)
 	})
 
 	if a != nil {


sjfricke:
Ubuntu with Galaxy phones verified

AWoloszyn:
So this actually fails for me.
Ubuntu + Pixel C.

The call to adb shell setprop 
```
adb shell setprop debug.vulkan.layers ""
usage: setprop NAME VALUE

Sets an Android system property.

setprop: Need 2 arguments
```

I am guessing this is a difference between Samsung and Pixel devices.
If we appended the other type of clear, that should handle both cases.

```go
		// setprop has slightly different semantics on different devices.
		// First try to clear with escaped quotes, which will leave the property
		// set as "" on some devices.
		// Then clear with non-escaped quotes, this clear it on the devices that
		// would have set it as "", but simply fail on others.
		d.Command("shell", "setprop", "debug.vulkan.layers", "\"\"").Run(ctx)
		d.Command("shell", "setprop", "debug.vulkan.layers", "").Run(ctx)
```

sjfricke:
I can't tell if it's the Go code or the phone, when it clears and you go `adb shell getprop | grep vulkan` do you get `[ ]` or `[""]`

Also i'd wait til you can confirm your S6 is the same as I feel having it prompt the `setprop: Need 2 arguments` error should be avoided unless this is the case that Pixel and Samsung devices have a different set method because going `adb shell setprop debug.vulkan.layers ""` works on my Galaxies, just seems the `\"\"\` aren't escaping correctly

AWoloszyn:
So with the existing implementation I get 
`[debug.vulkan.layers]: []` on both my PixelC and my Pixel phone.

sjfricke:
So this is not a phone thing, I got gapid up and running from tip or tree on my Windows 10 laptop and it ran the trace and when it was done it set `[debug.vulkan.layers]: []` properly.

AWoloszyn:
I just got an S6 to test with.


`go version go1.8 linux/amd64 ` -> `"\"\""` works for me
                                                        `""` fails with  `FAILED: adb shell setprop debug.vulkan.layers : exit status 1`


```
adb version                                                                                                                                                               
Android Debug Bridge version 1.0.36
...

adb shell getprop | grep ro.build.description                                                                                                                             
[ro.build.description]: [zerofltebmc-user 7.0 NRD90M G920W8VLS5DQH1 release-keys]
```

Issue #1056, opened 2017-09-05
Better End-of-Frame Handling in the Tree View


pmuetschard:
The end-of-frame markers used in the thumbnail scrubber can be out of sync with what is used to create thumbnails in the tree view.

Maybe GAPIS should tell GAPIC at what command to request the thumbnail for a group?

PR #1062, opened 2017-09-05
Synchronize frame group thumbnails with filmstrip.

ben-clayton:
Fixes: #1056


ben-clayton:
> Doesn't this break the frame thumbnails by undoing the previous improvements?

It shouldn't. The 'look for last draw call' is an improvement for groups that do something nasty, like unbinding the framebuffer after the last draw call. This was important before we had the gvr commands being traced. Now that they are, `gvr_frame_submit` is annotated with `@frame_end`, and the correct frame is shown. In cases were we have user marker groups, we still default to the 'look for last draw call' heuristic, and this works as it did before.

ben-clayton:
PTAL

Issue #966, opened 2017-08-22
Unhandled exception (java.lang.NullPointerException) in the UI thread.


ben-clayton:
Clicking on a command with a state tree node selected.
That node might not have existed for the new selection.

```
org.eclipse.swt.SWTException: Failed to execute runnable (java.lang.NullPointerException)
	at org.eclipse.swt.SWT.error(Unknown Source)
	at org.eclipse.swt.SWT.error(Unknown Source)
	at org.eclipse.swt.widgets.Synchronizer.runAsyncMessages(Unknown Source)
	at org.eclipse.swt.widgets.Display.runAsyncMessages(Unknown Source)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Unknown Source)
	at org.eclipse.jface.window.Window.runEventLoop(Window.java:818)
	at org.eclipse.jface.window.Window.open(Window.java:794)
	at com.google.gapid.Main$UI.show(Main.java:109)
	at com.google.gapid.Main.main(Main.java:67)
Caused by: java.lang.NullPointerException
	at org.eclipse.jface.viewers.CustomHashtable.hashCode(CustomHashtable.java:268)
	at org.eclipse.jface.viewers.CustomHashtable.get(CustomHashtable.java:240)
	at org.eclipse.jface.viewers.TreeSelection$InitializeData.<init>(TreeSelection.java:61)
	at org.eclipse.jface.viewers.TreeSelection$InitializeData.<init>(TreeSelection.java:53)
	at org.eclipse.jface.viewers.TreeSelection.<init>(TreeSelection.java:98)
	at org.eclipse.jface.viewers.TreeSelection.<init>(TreeSelection.java:109)
	at com.google.gapid.views.StateView.lambda$updateSelectionState$8(StateView.java:356)
	at com.google.gapid.util.SelectionHandler$1.lambda$onSuccess$0(SelectionHandler.java:75)
	at com.google.gapid.widgets.Widgets.ifNotDisposed(Widgets.java:118)
	at com.google.gapid.widgets.Widgets.lambda$scheduleIfNotDisposed$1(Widgets.java:146)
	at com.google.gapid.widgets.Widgets.lambda$schedule$0(Widgets.java:138)
	at org.eclipse.swt.widgets.RunnableLock.run(Unknown Source)
	... 7 more
```

Issue #938, opened 2017-08-18
When clicking on a framebuffer before the command-tree is populated causes an exception


AWoloszyn:
I noticed this on a particularly large trace on Windows. (Specifically a vulkan trace).

Note, GAPIC does not crash after dismissing the exception. And continues to work fine.

java.lang.NullPointerException
	at com.google.gapid.models.AtomStream.resolve(AtomStream.java:174)
	at com.google.gapid.models.AtomStream.selectAtoms(AtomStream.java:163)
	at com.google.gapid.views.ThumbnailScrubber.lambda$new$1(ThumbnailScrubber.java:76)
	at org.eclipse.swt.widgets.EventTable.sendEvent(Unknown Source)
	at org.eclipse.swt.widgets.Display.sendEvent(Unknown Source)
	at org.eclipse.swt.widgets.Widget.sendEvent(Unknown Source)
	at org.eclipse.swt.widgets.Display.runDeferredEvents(Unknown Source)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Unknown Source)
	at org.eclipse.jface.window.Window.runEventLoop(Window.java:818)
	at org.eclipse.jface.window.Window.open(Window.java:794)
	at com.google.gapid.Main$UI.show(Main.java:109)
	at com.google.gapid.Main.main(Main.java:67)

PR #1063, opened 2017-09-06
A bunch o' fixes regarding command tree selection.

pmuetschard:
Includes a fix for #938 and #966

### DIFF
--- a/gapis/resolve/events.go
+++ b/gapis/resolve/events.go
@@ -43,6 +43,8 @@ func Events(ctx context.Context, p *path.Events) (*service.Events, error) {
 	events := []*service.Event{}
 
 	s := c.NewState()
+	lastCmd := api.CmdID(0)
+	var pending []service.EventKind
 	api.ForeachCmd(ctx, c.Commands, func(ctx context.Context, id api.CmdID, cmd api.Cmd) error {
 		cmd.Mutate(ctx, id, s, nil)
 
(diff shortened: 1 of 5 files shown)


gapis/resolve/command_tree.go
```diff
@@ -140,12 +141,40 @@ func CommandTreeNodeForCommand(ctx context.Context, p *path.CommandTreeNodeForCo
...
+	}
+
 	return &path.CommandTreeNode{
 		Tree:    p.Tree,
-		Indices: cmdTree.indices(api.CmdID(cmdIdx)),
+		Indices: cmdTree.findNodeForCommand(snc, c.Commands, api.CmdID(cmdIdx)),
 	}, nil
 }
 
+func (t *commandTree) findNodeForCommand(snc *sync.Data, cmds []api.Cmd, cmd api.CmdID) []uint64 {
```
ben-clayton: I'm really not sure about this change. Please can you move this CL into a separate PR?

gapis/resolve/events.go
```diff
@@ -43,13 +43,24 @@ func Events(ctx context.Context, p *path.Events) (*service.Events, error) {
 	events := []*service.Event{}
 
 	s := c.NewState()
+	lastCmd := api.CmdID(0)
+	delayedFirstInFrame := false
```
ben-clayton: This took me a while to understand.

How about:

```go
var pending []*serviceEvent
api.ForeachCmd(ctx, c.Commands, func(ctx context.Context, id api.CmdID, cmd api.Cmd) error {
	cmd.Mutate(ctx, id, s, nil)

	// TODO: Add event generation to the API files.
	if !filter(id, cmd, s) {
		return nil
	}

	events, pending = append(events, pending...), nil
```
...
```go
	if p.FirstInFrame && f.IsEndOfFrame() {
		pending = append(pending, &service.Event{true
			Kind:    service.EventKind_FirstInFrame,		
			Command: p.Capture.Command(uint64(id) + 1),		
		})
	}
```

gapis/resolve/command_tree.go
pmuetschard: Done.

gapis/resolve/events.go
pmuetschard: Done.

Issue #829, opened 2017-07-28
Group render-passes and subpassess within command-buffers


AWoloszyn:
The tree-view would be much easier to navigate if it grouped render-passes, subpasses and draw calls within a submission.

Qining:
Resolved in #1064

PR #1052, opened 2017-09-01
Group subcommands with debug markers

Qining:
Use CmdIDGroup in SubCmdRoot's SubGroup to group the subcommands

PR #1064, opened 2017-09-06
Renderpass grouping

Qining:
Sits on top of #1052

### DIFF
--- a/gapis/api/cmd_id_group.go
+++ b/gapis/api/cmd_id_group.go
@@ -37,7 +37,6 @@ type CmdIDGroup struct {
 	Range    CmdIDRange // The range of commands this group (and items) represents.
 	Spans    Spans      // All sub-groups and sub-ranges of this group.
 	UserData interface{}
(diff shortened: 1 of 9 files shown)


gapis/api/cmd_id_group.go
```diff
@@ -37,6 +37,7 @@ type CmdIDGroup struct {
 	Range    CmdIDRange // The range of commands this group (and items) represents.
 	Spans    Spans      // All sub-groups and sub-ranges of this group.
 	UserData interface{}
+	Parent   SubCmdIdx // SubCmdIdx of the SubCmdRoot if this CmdIDGroup belongs to a SubCmdRoot, otherwise just an empty SubCmdIdx
```
ben-clayton: Please can we discuss this CL before submitting. This code has now passed the threshold of complexity that means I'll never be able to understand it ever again.

Qining: Sure. This is `Parent` is introduced in #1052 . Basically this is used in `Index()` when the CmdIDGroup is a group of subcommands.

Qining: I think another option is to create a new SubCmdIDGroup to group the subcommands, but then we will need to extend the AddGroup() to the new type

Qining:
This resolve the issue #829

PR #1070, opened 2017-09-06
Add a fallback path if the memory tracker is not working.

AWoloszyn:
As we come across new devices/drivers, they may implement coherent
memory differently. When we create the device, check to see if the
coherent memory tracker is working. If not then fallback to the slow
path so at least those devices will work.

### DIFF
--- a/gapis/api/vulkan/templates/vk_spy_helpers.cpp.tmpl
+++ b/gapis/api/vulkan/templates/vk_spy_helpers.cpp.tmpl
@@ -150,7 +150,7 @@ uint32_t VulkanSpy::SpyOverride_vkEnumerateInstanceExtensionProperties(const cha
     return VkResult::VK_SUCCESS;
 }
 
-uint32_t VulkanSpy::SpyOverride_vkEnumerateDeviceExtensionProperties(VkPhysicalDevice physicalDevice, const char *pLayerName, uint32_t *pCount, VkExtensionProperties *pProperties) {    
+uint32_t VulkanSpy::SpyOverride_vkEnumerateDeviceExtensionProperties(VkPhysicalDevice physicalDevice, const char *pLayerName, uint32_t *pCount, VkExtensionProperties *pProperties) {
     //auto next_layer_enumerate_extensions = mImports.mVkInstanceFunctions[PhysicalDevices[physicalDevice]->mInstance].vkEnumerateDeviceExtensionProperties;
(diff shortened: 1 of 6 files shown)


gapis/api/vulkan/templates/vk_spy_helpers.cpp.tmpl
```diff
@@ -540,23 +595,27 @@ void VulkanSpy::readMappedCoherentMemory(CallObserver *observer, VkDeviceMemory
...
+    if (m_coherent_memory_tracking_enabled) {
+        const size_val page_size = mMemoryTracker.page_size();
+        // Get the valid mapped range
+        const auto dirty_pages = mMemoryTracker.GetAndResetDirtyPagesInRange(offset_addr, readSize);
+        for (const void *p : dirty_pages) {
+            uint64_t page_start = (uint64_t)(p);
+            uint64_t page_end = page_start + page_size;
+            observer->read(slice((uint8_t *)page_start, 0ULL, page_size));
+        }
+    return;
```
ben-clayton: nit: bad indent.

gapis/api/vulkan/templates/vk_spy_helpers.cpp.tmpl
```diff
@@ -349,6 +353,56 @@ uint32_t VulkanSpy::SpyOverride_vkCreateDevice(VkPhysicalDevice physicalDevice,
...
+                    VkMemoryPropertyFlagBits::VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT)) ==
+                            (VkMemoryPropertyFlagBits::VK_MEMORY_PROPERTY_HOST_COHERENT_BIT |
+                             VkMemoryPropertyFlagBits::VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT)) {
+                    coherent_bit = i;
+                }
+            }
+            if (coherent_bit == static_cast<uint32_t>(-1)) {
+                return VkResult::VK_ERROR_INITIALIZATION_FAILED;
+            }
+            uint32_t pagesize = 4096;
```
ben-clayton: nit: Magic number alert. There's at lease one other declaration of this in `memory_manager.cc`.
Would be nice if there was one place to hold this.

AWoloszyn: I think I can actually fix this as well by exposing the "PageSize()" method from the memory tracker

AWoloszyn:
Updated based on comments.

sjfricke:
Confirm this fixes the issue on my Galaxy S6 as well!

Issue #1068, opened 2017-09-06
The gapis server has exited with an error code of: 2


ben-clayton:
GAPID Version: 0.6.0:298:3f048c56b628f96f912aac0d27d7e56a3be4f63b
OS: mac os x x86_64

```
panic: runtime error: index out of range [recovered]
	panic: 
--- *resolve.StateTreeResolvable ---
 path:<after:<capture:<id:<data:"-\367y\363\300$F\t\363\246\225\250\313y\350\261\237@L:" > > indices:423452 > > array_group_size:2000 
 Store():
  /tmpfs/src/github/src/github.com/google/gapid/gapis/database/database.go:51  github.com/google/gapid/gapis/database.Store
  /tmpfs/src/github/src/github.com/google/gapid/gapis/resolve/state_tree.go:38  github.com/google/gapid/gapis/resolve.StateTree
  /tmpfs/src/github/src/github.com/google/gapid/gapis/resolve/resolve.go:312  github.com/google/gapid/gapis/resolve.ResolveInternal
  /tmpfs/src/github/src/github.com/google/gapid/gapis/resolve/resolve.go:247  github.com/google/gapid/gapis/resolve.ResolveService
  /tmpfs/src/github/src/github.com/google/gapid/gapis/resolve/get.go:35  github.com/google/gapid/gapis/resolve.(*GetResolvable).Resolve
  /tmpfs/src/github/src/github.com/google/gapid/gapis/database/memory.go:78  github.com/google/gapid/gapis/database.(*record).resolve
  /tmpfs/src/github/src/github.com/google/gapid/gapis/database/memory.go:154  github.com/google/gapid/gapis/database.(*memory).resolveLocked.func1

 Build() #0:
  /tmpfs/src/github/src/github.com/google/gapid/gapis/database/database.go:59  github.com/google/gapid/gapis/database.Resolve
  /tmpfs/src/github/src/github.com/google/gapid/gapis/resolve/state_tree.go:98  github.com/google/gapid/gapis/resolve.StateTreeNode
  /tmpfs/src/github/src/github.com/google/gapid/gapis/resolve/resolve.go:314  github.com/google/gapid/gapis/resolve.ResolveInternal
  /tmpfs/src/github/src/github.com/google/gapid/gapis/resolve/resolve.go:247  github.com/google/gapid/gapis/resolve.ResolveService
[...]

Issue #24, opened 2017-03-06
Tidy the GLES state block


ben-clayton:
Add a global for `CurrentState` so we don't have to always go though the `map!(ThreadID, ref!Context) Contexts` global at the root of the state tree (ensure you update all Link() logic in gles/links.go).

Find a way to hide `GLXContexts`, `WGLContexts`, `CGLContexts`, `EGLContexts`.

Try navigating the state block in GAPIC, make changes to the API file to try and make state navigation not suck.

dsrbecky:
+1 for the global.  Do we want to introduce the concept of thread-local storage, or we just update it before each command?

UI wise:  How about moving all globals to combo box?

ben-clayton:
Given that the state reflects the currently selected command, then the state should probably reflect the current thread of that command.

dsrbecky:
Do we consider this done now?

ben-clayton:
I think we should probably implement `@hidden` first to hide fields we shouldn't be showing.
Then it can probably be removed from the milestone. After that there's still polish I'd like to do.

ben-clayton:
Also - filter to currently selected context.

PR #1086, opened 2017-09-08
Tidy state view

ben-clayton:
Fixes: #1068
Fixes: #24

### DIFF
--- a/gapic/src/main/com/google/gapid/models/ApiState.java
+++ b/gapic/src/main/com/google/gapid/models/ApiState.java
@@ -55,14 +55,14 @@ public class ApiState
   private final ObjectStore<Path.Any> selection = ObjectStore.create();
 
   public ApiState(
-      Shell shell, Client client, Follower follower, AtomStream atoms, ConstantSets constants) {
+      Shell shell, Client client, Follower follower, AtomStream atoms, ApiContext contexts, ConstantSets constants) {
     super(LOG, shell, client, Listener.class);
     this.constants = constants;
 
     atoms.addListener(new AtomStream.Listener() {
       @Override
       public void onAtomsSelected(AtomIndex index) {
-        load(stateTree(index), false);
+        load(stateTree(index, contexts.getSelectedContext()), false);
       }
     });
     follower.addListener(new Follower.Listener() {
--- a/gapic/src/main/com/google/gapid/models/ApiState.java
+++ b/gapic/src/main/com/google/gapid/models/ApiState.java
@@ -76,7 +76,7 @@ public class ApiState
   @Override
   protected ListenableFuture<Node> doLoad(Path.Any path) {
     return Futures.transformAsync(client.get(path),
-        tree -> Futures.transform(client.get(Paths.any(tree.getStateTree().getRoot())),
+        tree -> Futures.transform(client.get(Paths.toAny(tree.getStateTree().getRoot())),
             val -> new RootNode(
                 tree.getStateTree().getRoot().getTree(), val.getStateTreeNode())));
   }
@@ -115,7 +115,7 @@ public class ApiState
 
   public ListenableFuture<Node> load(Node node) {
     return node.load(shell, () -> Futures.transformAsync(
-        client.get(Paths.any(node.getPath(Path.StateTreeNode.newBuilder()))),
+        client.get(Paths.toAny(node.getPath(Path.StateTreeNode.newBuilder()))),
         value -> Futures.transform(constants.loadConstants(value.getStateTreeNode()),
             ignore -> new NodeData(value.getStateTreeNode()))));
   }
--- a/gapic/src/main/com/google/gapid/util/Paths.java
+++ b/gapic/src/main/com/google/gapid/util/Paths.java
@@ -24,6 +24,7 @@ import com.google.gapid.proto.service.Service;
 import com.google.gapid.proto.service.path.Path;
 import com.google.gapid.proto.service.vertex.Vertex;
 import com.google.gapid.views.Formatter;
+import java.util.function.Predicate;
 
 /**
  * Path utilities.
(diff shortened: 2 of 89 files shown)


gapic/src/main/com/google/gapid/util/Paths.java
```diff
@@ -248,82 +272,107 @@ public static int compare(Path.Command a, Path.Command b) {
...
+    public static Path.Any find(Path.Any path, Predicate<Path.Any> predicate) {
+    for (Path.Any p = path; p != null; p = parentOf(p)) {
+      if (predicate.test(p)) {
+        return p;
+      }
+    }
+    return null;
+  }
+
+  public static Path.Any parentOf(Path.Any path) {
```
pmuetschard: This is incomplete and I don't think it's necessary. This is trying to future-proof something without actually making it future-proof. Please undo this and just add the global state case.

gapic/src/main/com/google/gapid/models/ApiState.java
```diff
@@ -53,16 +54,20 @@
...
+      Shell shell, Client client, Follower follower, AtomStream atoms, ApiContext contexts, ConstantSets constants) {
     super(LOG, shell, client, Listener.class);
     this.constants = constants;
+    this.contexts = contexts;
 
     atoms.addListener(new AtomStream.Listener() {
       @Override
       public void onAtomsSelected(AtomIndex index) {
-        load(Paths.any(stateTree(index.getCommand())), false);
+        Path.StateTree path = stateTree(index.getCommand());
```
pmuetschard: Instead of building the proto twice, simply pass the context into the `stateTree` method (see the `commandTree` method)

gapis/service/path/path.proto
```diff
@@ -369,6 +369,8 @@ message Slice {
 // State is a path to a subset of the GlobalState at a point in a capture.
 message State {
     Command after = 1;
+    // If non-nil, then the state is filtered to the specified context.
+    ID context = 2;
```
pmuetschard: Won't this break links to objects in another context which is shared?
```
ctx1 = newContext(nil)
ctx1...[1] = newTexture()
ctx2 = newContext(ctx1)
ctx2.bindTexture(1)
```

If in the UI I am filtering to ctx2, the link will attempt to go to ctx1, which is now no longer part of the tree, right?

gapic/src/main/com/google/gapid/util/Paths.java
ben-clayton: > This is trying to future-proof something without actually making it future-proof

I'd say it is more than just future-proofing. You had an ad-hoc collection of find functions that would need to be duplicated for `findGlobalState`. Yes, what I have changed to is incomplete, but it's as incomplete as your old `findState`. I'd rather implement the full set than roll back to a bunch of random functions. This approach mirrors the go-side, where parent-traversal is used for numerous cases.

gapis/service/path/path.proto
ben-clayton: No it won't - because the server will return absolute GlobalState paths for followed links. The state tree nodes will also use GlobalState paths so that everything is absolute.

ben-clayton: State paths are used as relative paths to the most relevant part of the GlobalState (as considered by GAPIS).

gapic/src/main/com/google/gapid/views/StateView.java
```diff
@@ -531,7 +531,7 @@ public void setHoveredItem(TreeItem hoveredItem, Follower.Prefetcher<Void> follo
     @Override
     protected boolean isFollowable(Object element) {
       ApiState.Node node = (ApiState.Node)element;
-      return node.getData() != null && node.getData().hasValuePath();
+      return node != null && node.getData() != null && node.getData().hasValuePath();
```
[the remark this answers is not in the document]
ben-clayton: Removed.

gapic/src/main/com/google/gapid/models/ApiState.java
```diff
@@ -62,7 +62,7 @@ public ApiState(
     atoms.addListener(new AtomStream.Listener() {
       @Override
       public void onAtomsSelected(AtomIndex index) {
-        load(stateTree(index), false);
```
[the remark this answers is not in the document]
ben-clayton: Done.

gapic/src/main/com/google/gapid/models/ApiState.java
ben-clayton: Done.

gapic/src/main/com/google/gapid/util/Paths.java
```diff
@@ -77,15 +78,10 @@ private Paths() {
     return Path.State.newBuilder().setAfter(command).build();
   }
 
-  public static Path.Any stateTree(AtomIndex atom) {
```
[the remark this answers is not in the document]
ben-clayton: Done.

gapic/src/main/com/google/gapid/util/Paths.java
pmuetschard: >  but it's as incomplete as your old `findState`

I disagree with that assertion. `findState` was incomplete in the sense that the state path was the only path it could find, but it correctly handled every kind of path. `parentOf` pretends to be able to find the parent of any path, but it does not.

I know this is how it's done in go, and it works OK there due to the implicit interfaces. However, in Java the protos are immutable and the constant toBuilder().build() to wrap it in `Path.Any` for it to then just be unwrapped is a lot of overhead for something that is not complete and only used to determine if a path has a state root.

I propose, for now, we simply change `findState` to `isStatePath`, returning a boolean, since that's all that it's actually used for. This makes the change to handle the new global state root very simple. We then look at how we can do the parent relationship better. I would prefer if it was part of the proto.

ben-clayton:
Updated with a final CL that fixes the state tree reparenting. PTAL

PR #1092, opened 2017-09-11
Allow resolving of "hidden" commands to tree nodes.

pmuetschard:
See #1063

### DIFF
--- a/gapis/resolve/command_tree.go
+++ b/gapis/resolve/command_tree.go
@@ -22,6 +22,7 @@ import (
 	"github.com/google/gapid/core/log"
 	"github.com/google/gapid/core/math/interval"
 	"github.com/google/gapid/gapis/api"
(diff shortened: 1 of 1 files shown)


ben-clayton:
I'm sure we must have had a face-to-face discussion about this, but I can't remember the outcome.
Do you still want this change?

pau-baiget:
@pmuetschard @ben-clayton do we still want to merge this PR?

AWoloszyn:
I think this is broken now anyway. I believe we are ALWAYS showing hidden nodes, which we should fix.

pmuetschard:
I think with Q, we will no longer have nested calls (yay) and this is not that important, so closing it.

For reference: this was to fix an issue where the "goto command" UI action would fail, if an ID of a nested command was entered by the user. Since we do show those IDs in the UI, we should allow the user to "goto" them.

PR #1097, opened 2017-09-13
3D and 2D/1D array texture support on the Vulkan side

Qining:


gapis/api/vulkan/resources.go
```diff
@@ -529,17 +540,78 @@ func (t *ImageObject) ResourceData(ctx context.Context, s *api.State) (*api.Reso
...
+		if t.Info.ArrayLayers > uint32(1) {
+			// 2D texture array
+			layers := make([]*api.Texture2D, int(t.Info.ArrayLayers))
+			for layerIndex := range layers {
+				imageLayer := t.Layers.Get(uint32(layerIndex))
+				levels := make([]*image.Info, len(imageLayer.Levels))
+				for levelIndex := range levels {
+					imageLevel := imageLayer.Levels.Get(uint32(levelIndex))
+					img := imageLevel.imageInfo(ctx, s, format)
+					if img == nil {
```
ben-clayton: In GLES we've made an effort to show partially populated textures. It's annoying to only see the image if all levels / layers are populated with data.

Qining: 4334638 changes all the early returns to `continue`.

PR #1100, opened 2017-09-14
Add trackDimension.

baldwinn860:
Now there is a track selector to show just the packages within a single
track. There is currently no ordering between tracks which is most
likely not an issue as there is little information in the relationship
of tracks that would not be better gleaned from git.

### DIFF
--- a/test/robot/web/client/data.go
+++ b/test/robot/web/client/data.go
@@ -29,9 +29,11 @@ import (
 	"github.com/google/gapid/test/robot/web/client/widgets/grid"
 )
 
-type TrackInfo struct {
+type trackInfo struct {
+	track                 Item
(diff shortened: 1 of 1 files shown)


test/robot/web/client/data.go
```diff
@@ -114,35 +158,47 @@ var (
...
+						destTrack.packageList = packageList
+						for _, p := range destTrack.packageList {
+							packageToTrack[p] = destTrack.track
+						}
+						foundTrack = true
+						break
+					}
+				}
+				if !foundTrack {
+					// We just append all tracks to the "auto" track that didn't match an existing track
```
pmuetschard: "...append all packages...", no?

PR #1101, opened 2017-09-14
Fix default group thumbs

ben-clayton:
The `CommandTreeNodeUserData.Thumbnail` is set to `api.CmdNoID` when there's no explicit thumbnail. Don't try to get thumbnails for these no-id IDs.


ben-clayton:
Thanks

PR #1103, opened 2017-09-14
Record the content of external images during trace

dsrbecky:


gapis/api/gles/compat.go
```diff
@@ -1025,6 +1026,18 @@ func compat(ctx context.Context, device *device.Instance) (transform.Transformer
 				// Rebind the currently bound 2D texture.  This might seem like a no-op, however,
 				// the remapping layer will use the ID of the EGL image replacement texture now.
 				out.MutateAndWrite(ctx, dID, cb.GlBindTexture(GLenum_GL_TEXTURE_2D, c.Bound.TextureUnit.Binding2d.ID))
+
+				// Update the content if we made a snapshot.
+				if e := FindEGLImageData(cmd.Extras()); e != nil {
+					_, err := database.Resolve(ctx, e.ID)
```
dsrbecky: This check currently seems to fail, any idea why the resource can not be found?

I failed to debug it so far as the IDs generally print as random ascii art.  Can we just have incrementing integers for the resource IDs please?

ben-clayton: > This check currently seems to fail, any idea why the resource can not be found?

Yes.

The resource IDs are [generated in GAPII using cityhash](https://github.com/google/gapid/blob/536c0cf1f52d79224f48b693d4137954194b6860/gapii/cc/call_observer.cpp#L95). As a memory range is observed, the hash is calculated, and used to determine if the resource [has already been encoded](https://github.com/google/gapid/blob/536c0cf1f52d79224f48b693d4137954194b6860/gapii/cc/call_observer.cpp#L96). If not, the resource data and hash is encoded in the `Capture::Resource` message. An observation always refers to the resource data by this hash.

GAPIS uses SHA1 hashes for all database entries. When you store some data, the message is encoded to binary, SHA1 hashed, and checked to see if it is already in the database. We use this to ensure we only have one copy of any data and also for ensuring that [`Resolvable`](https://github.com/google/gapid/blob/536c0cf1f52d79224f48b693d4137954194b6860/gapis/database/resolvable.go#L32-L35) types always do the work once.

SHA1 is moderately expensive to calculate, hence we use the much faster cityhash in the interceptor. SHA1 is a better hashing function, so we have used it for the server.

So, to avoid having to deal with a cityhash->SHA1 map after load, we [replace the cityhash resource hashes with SHA hashes](https://github.com/google/gapid/blob/536c0cf1f52d79224f48b693d4137954194b6860/gapis/capture/capture.go#L265-L282) while we're deserializing.

Memory observations are explicitly handled, but we now have your new type that is trying to load the resources from the database using the cityhash IDs.

My recommendation for now would be to generalize the hash remapping logic so you can remap the IDs in the `EGLImageData`.

Consider creating this interface in the `api` package:

```go
// ObservationReference is the interface implemented by types that hold references
// to Observation IDs. As these IDs are remapped on load, references must update
// any IDs using the map passed to RemapObservationIDs.
type ObservationReference interface {
    RemapObservationIDs(map[id.ID]id.ID)
}
```

Then you can check for this interface [here and call it](https://github.com/google/gapid/blob/536c0cf1f52d79224f48b693d4137954194b6860/gapis/capture/decoder.go#L134).

And move [this functionality](https://github.com/google/gapid/blob/536c0cf1f52d79224f48b693d4137954194b6860/gapis/capture/capture.go#L267-L269
) into `api.CmdObservation.RemapObservationIDs()`. You'll then be able to do the same for your own new type.

> I failed to debug it so far as the IDs generally print as random ascii art.

`id.ID` should print just fine. Are you referring to printing from the GAPII side?

> Can we just have incrementing integers for the resource IDs please?

Which IDs are you proposing to change? The IDs for Observations in the `.gfxtrace` file? Yes, that could work, but would break file formats again.

If you're referring to all IDs, that breaks a whole bunch of fundamental assumptions, and would be a massive change.

gapis/api/gles/extras.go
```diff
@@ -30,6 +31,16 @@ type ErrorState struct {
 	InterceptorsGlError GLenum
 }
 
+// EGLImageData  is an extra used to store snapshot of external image source.
+type EGLImageData struct {
```
ben-clayton: It's fine to have this type here, but you could just use the proto directly.

gapis/api/gles/compat.go
ben-clayton: Created #1105 to give you this interface.

gapis/api/gles/gles_pb/extras.proto
```diff
@@ -22,3 +22,13 @@ message ErrorState {
 	uint32 TraceDriversGlError = 1;
 	uint32 InterceptorsGlError = 2;
 }
+
+message EGLImageData {
+  string ID     = 1; // Resource ID of the image data
```
ben-clayton: Nit: could just use `bytes` to avoid a ID parse

PR #1111, opened 2017-09-15
Replace builder info, allow BuildBot when git present.

baldwinn860:
We do not need the entire instance information for the builder that came
with every package and in the case of buildbot packages, the builder
info would be incorrect because the uploader is not the builder.
Reducing this information down to just the ABI captures all we need from
the builder info.

Submitting a BuildBot build when git is present would still change the
type to User. Now the test is if a CL is specified in the command line
it keeps the BuildBot type, if prepare needs to ask for the head CL,
then it is a User build.

### DIFF
--- a/cmd/robot/build.go
+++ b/cmd/robot/build.go
@@ -84,6 +84,7 @@ type buildUploadVerb struct {
 	Tag           string `help:"The optional build tag"`
 	Branch        string `help:"The build branch, will be guessed if not set"`
 	Uploader      string `help:"The uploading entity, will be guessed if not set"`
+	BuilderAbi    string `help:"The abi of the builder device, will assume this device if not set"`
 	ServerAddress string `help:"The master server address"`
 
 	store build.Store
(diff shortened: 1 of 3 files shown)


ianthehat:
The original intention is that build bot uploads would not use any of the guessing logic, they should be explicit about the values on the command line as they already know them. It should not even have been trying to run git.
The intent behind the builder information was to be able to uniquely identify the machine that did the build, in case you have a build bot farm with a consistently misbehaving bot so you can identify it.

baldwinn860:
I agree, but the implementation assumed that the build bot is the same machine that does the upload, which is not the case since the build bot cannot access the robot server. So instead I have a local script which watches the build bot artifact location and uploads from my machine. Since I have limited information on the builder instance and I don't want to give false information, I could only submit an ABI.

test/robot/build/build.proto
```diff
@@ -53,8 +53,8 @@ message Information {
   string tag = 4;
   // Description is a user assigned description of the build.
   string description = 5;
-  // Builder is the device information for the machine that made the build.
-  device.Instance builder = 6;
+  // BuilderAbi is the device ABI for the machine that made the build.
+  device.ABI builder_abi = 6;
```
pmuetschard: This change is not really necessary - you can just leave the other fields blank. Also, I just noticed that the ToolSet itself has the ABI, so it is now duplicated.

Instead, for Kokoro builds, we should just leave the builder empty (except maybe put "Kokoro" into the name field) and then use the ABI from the ToolSet to match at test run time.

baldwinn860: The ToolSet populates its ABI based on this ABI when it is created on the server, so we still need that at the very least. I will leave the Instance and put Kokoro in the name anyways.

baldwinn860: Updated.

PR #1115, opened 2017-09-18
Removed most of the hidden state on the Vulkan state block.

AWoloszyn:
This moves us towards being able to transfer the state block
as-is, and greatly improve mid-execution capture speed.

A lot of the hand-crafted templates on the C++ side can/will be removed
in the future once we serialize the entire state block instead of just the
commands needed to recreate it. This gets us on the path there.


gapii/cc/vulkan_inlines.inl
```diff
@@ -29,30 +31,30 @@ private:
 
 ////////// Command Buffer Commands
 template<>
-bool inline CommandListRecreator<std::shared_ptr<RecreateCmdUpdateBufferData>>::operator()(
+bool inline CommandListRecreator<std::shared_ptr<vkCmdUpdateBufferArgs>>::operator()(
```
ben-clayton: I'm sure its just my painkiller fueled state, but I'm confused. This code makes me think we're recreating the original vk commands from the `Args` data (presumably to encode them), but then we have something similar in gapis. Please can you describe in the CL description what this CL is actually doing?

gapis/api/vulkan/externs.go
```diff
@@ -77,68 +77,791 @@ func (e externs) callSub(ctx context.Context, cmd api.Cmd, id api.CmdID, s *api.
 	})
 }
 
-func (e externs) addCmd(commandBuffer VkCommandBuffer, recreate_data interface{}, data interface{}, functionToCall interface{}) {
+func (e externs) addCmd(commandBuffer VkCommandBuffer, data interface{}, functionToCall interface{}) {
 	o := GetState(e.s).CommandBuffers.Get(commandBuffer)
 
+	switch d := data.(type) {
+	case (*VkCmdBindPipelineArgs):
```
ben-clayton: nit: no need for parentheses here.

gapii/cc/vulkan_inlines.inl
AWoloszyn: So the only thing that has actually CHANGED in vulkan_inlines is the name of the structures that we are encoding.
vkCmd*args instead of RecreateCmd...*.

MEC remains the same is always, except we now execute VkCmd* instead of RecreateCMD*, so we can remove all of the RecreateVkCmd* synthetic commands from the API file, and we no longer have to re-write all of the RecreateVkCmd* during state mutation.

And yes, we have identical logic in Go right now (and always) in order for us to cut apart command-buffers, we need to be able to generate a new command-buffer.

Once we only serialize state, all of this logic on the Interceptor side can go away, and will simpify significantly.

I will add some of this into the comment.

gapii/cc/call_observer.h
```diff
@@ -248,7 +248,11 @@ inline Slice<T> CallObserver::copy(const Slice<T>& dst, const Slice<T>& src) {
 template <typename T>
 inline Slice<T> CallObserver::clone(const Slice<T>& src) {
     Slice<T> dst = make<T>(src.count());
-    copy(dst, src);
+    // Make sure that we actually fill the data the first time.
+    // If we use ::copy(), then the copy will only happen if
+    // the observer is active.
```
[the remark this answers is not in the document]
AWoloszyn: So there are a couple of places where this is needed. I will provide an example with VkShaderModule/ShaderModuleObject.

How we used to handle VkCreateShaderModule.

```
//*.api_types.tmpl
{{define "DeclareClass"}}
{{$is_shader_module := (eq $.Name "ShaderModuleObject")}}
{{if $is_shader_module}}
   std::vector<uint32_t> shaderWords;
{{end}}
// C++ Extra
void VulkanSpy::addWords(CallObserver* observer, VkShaderModule module, size_val count, uint32_t* words) {
   auto& m = ShaderModules[module];
   m->shaderWords = std::vector<uint32_t>(words, words+count/4);
}
``` 
So basically, we would create a hidden std::vector on the side of the state object. However this is not "officially" in the state block. This was done because clone() does not actually do the copy when we are not tracking memory operations, so we would end up dropping the words on the floor for MEC. We ended up stuffing this in a side data-structure so that we could guarantee the write.

If we want to remove it, we need that clone to actually copy the bytes, even when observations are turned off. However turning on ALL copies drastically slows down the performance during MEC. (To the point that in some scenes the program is almost unusable), so the compromise was the assumption that a clone is important, and keeps data even when tracing is off, but a copy can be elided.

Another option might be a way to annotate a slice with "Really actually do the copy".

gapii/cc/vulkan_mid_execution.cpp
```diff
@@ -41,8 +43,8 @@ class TemporaryShaderModule {
         VkStructureType::VK_STRUCTURE_TYPE_SHADER_MODULE_CREATE_INFO,  // sType
         nullptr,                                                       // pNext
         0,                                                             // flags
-        module_obj->shaderWords.size() * 4,  // codeSize
-        module_obj->shaderWords.data(),      // pCode
+        static_cast<size_t>(module_obj->mWords.size()),               // codeSize
+        &module_obj->mWords[0],               // pCode
```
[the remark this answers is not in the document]
AWoloszyn: I think I have probably brought this up in the past as well in a code-review. Yes this is bad. I have replaced it with begin(). (This is a slice, so we don't have data(), but begin() is equivalent).

PR #1121, opened 2017-09-19
Fix quirks with the movable tabs.

pmuetschard:
Especially around minimize/maximize behaviour. Now using our own layout logic for finer control.


pmuetschard:
Yes, it is not generic, primarily, because I didn't want to spend that effort now, as it looks like UX is likely wanting to change the whole thing anyways. I just wanted to fix the obviously broken behaviour.

PR #1128, opened 2017-09-21
Store EGLImage resources when saving capture

### LINKED ISSUE #1127: Same EGLImage data when re-saving capture
If the capture is saved the EGLImage resources are omitted.
The fix seems non-trivial due to go package dependencies.

dsrbecky:
Fixes #1127


gapis/capture/decoder.go
```diff
@@ -133,7 +134,17 @@ func (d *decoder) decode(ctx context.Context, in proto.Message) (interface{}, er
 	}
 
 	if r, ok := obj.(api.ResourceReference); ok {
-		obj = r.RemapResourceIDs(d.builder.idmap)
+		err := r.VisitResourceIDs(func(id *id.ID) error {
+			if newID, found := d.builder.idmap[*id]; found {
+				*id = newID
+				return nil
+			} else {
```
ben-clayton: Govet: Drop the else if the true block always returns.

ben-clayton: In this case it's more idiomatic to do the error in conditional block.

ben-clayton:
```go

newID, found := d.builder.idmap[*id]
if !found {
	return fmt.Errorf("Can not remap resource. %v not found.", *id)
}
*id = newID
return nil
```

dsrbecky:
Done

PR #1129, opened 2017-09-21
Added the spy_disabled annotation to slices.

AWoloszyn:
This allows us to avoid creation/copying of any slices that
are not needed during the trace. This should reduce memory pressure
as we no longer need to keep shadow memory around during the trace
as long as we know how to re-create it for mid-execution capture.

### DIFF
--- a/gapii/cc/call_observer.h
+++ b/gapii/cc/call_observer.h
@@ -157,10 +157,8 @@ private:
     // pool and we are supposed to observe application pool.
     template <class T>
     bool shouldObserve(const Slice<T>& slice) const {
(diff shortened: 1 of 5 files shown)


gapii/cc/call_observer.h
```diff
@@ -237,7 +235,7 @@ inline void CallObserver::write(const Slice<T>& dst, uint64_t index,
 template <typename T>
 inline Slice<T> CallObserver::copy(const Slice<T>& dst, const Slice<T>& src) {
     read(src);
-    if (isActive() && !shouldObserve(
+    if (!shouldObserve(
```
ben-clayton: nit: tidy line?

PR #1130, opened 2017-09-21
Robot: Add API option to subject upload.

baldwinn860:
Allows user to specify the API of the uploaded subject as either gles
(the default) or vulkan. Then stores it in the subject's hints to be
used when tracing with gapit.

### DIFF
--- a/cmd/robot/subject.go
+++ b/cmd/robot/subject.go
@@ -36,7 +36,7 @@ func init() {
 		Name:       "subject",
 		ShortHelp:  "Upload a traceable application to the server",
 		ShortUsage: "<filenames>",
-		Action:     &subjectUploadVerb{RobotOptions: defaultRobotOptions},
+		Action:     &subjectUploadVerb{RobotOptions: defaultRobotOptions, API: GLESAPI},
 	})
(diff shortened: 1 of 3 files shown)


cmd/robot/subject.go
```diff
@@ -48,6 +48,7 @@ func init() {
 
 type subjectUploadVerb struct {
 	RobotOptions
+	API       string        `help:"the api to capture, can be either gles or vulkan (default:gles)"`
```
ben-clayton: The flag system [does support enums](https://github.com/google/gapid/blob/e873228976e27d96964158506308962bed7e7db8/core/app/flags/choices.go#L34-L41).

baldwinn860: I was following the pattern that gapit uses for accepting the API seen in cmd/gapit/flags.go. Should I change that instance as well to be an enum?

ben-clayton: You can do, but you don't have to. That flag might become a list, in which case keeping-as-is is fine.

baldwinn860: ok fixed.

baldwinn860:
Commenting to get cla/google to re-run

PR #1131, opened 2017-09-22
Add update verbs

baldwinn860:
Adds ability to update some entities within robot. Specifically Packages and Subjects. This makes it possible to fix mistakes during initial uploads, like the subject's API or the package's description. Also lets us improve the data of these entities, such as better parentage of packages, and less than useful trace times for subjects.

### DIFF
--- a/cmd/robot/build.go
+++ b/cmd/robot/build.go
@@ -87,6 +87,12 @@ func init() {
 		ShortUsage: "<id or name>",
 		Action:     &trackUpdateVerb{RobotOptions: defaultRobotOptions},
 	})
+	setVerb.Add(&app.Verb{
+		Name:       "package",
+		ShortHelp:  "Sets values on a package",
+		ShortUsage: "<id>",
+		Action:     &packageUpdateVerb{RobotOptions: defaultRobotOptions},
+	})
 }
 
 type buildUploadVerb struct {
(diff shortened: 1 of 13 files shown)


cmd/robot/build.go
```diff
@@ -412,3 +418,42 @@ func (v *trackUpdateVerb) Run(ctx context.Context, flags flag.FlagSet) error {
...
+		b := build.NewRemote(ctx, conn)
+		args := flags.Args()
+		pkg := &build.Package{
+			Information: &build.Information{Description: v.Description},
+			Parent:      v.Parent,
+		}
+		if len(args) == 0 {
+			return log.Err(ctx, nil, "Missing argument, must specify a package to update")
+		}
+		err := b.SearchPackages(ctx, script.MustParse("Id == $").Using("$")(args[0]).Query(), func(ctx context.Context, entry *build.Package) error {
```
pmuetschard: you can combine this into:
```
if err := ...; err != nil {
  ...
}
```

cmd/robot/subject.go
```diff
@@ -103,3 +109,46 @@ func (v *subjectSearchVerb) Run(ctx context.Context, flags flag.FlagSet) error {
...
+		subj := &subject.Subject{
+			Hints: &subject.Hints{
+				TraceTime: ptypes.DurationProto(v.TraceTime),
+				API:       v.API,
+			},
+		}
+		if len(args) == 0 {
+			return log.Err(ctx, nil, "Missing argument, must specify a subject to update")
+		}
+		err := s.Search(ctx, idOrApkName(args[0]).Query(), func(ctx context.Context, entry *subject.Subject) error {
```
pmuetschard: you can combine this into:
```
if err := ...; err != nil {
  ...
}
```

cmd/robot/build.go
baldwinn860: ok done.

pmuetschard: uhm... not done.

ben-clayton: When the logic spans multiple lines, like this does - I actually encourage splitting the assignment and condition to separate statements, just for parsabilities sake.

pmuetschard: OK, good to know, thanks.

Nick, it's already approved, so you can just submit as is 😄

baldwinn860: ok I changed it back just to make it all match.

PR #1133, opened 2017-09-23
Optimizations

redmercury:
Optimizing message sending by:

- Batching buffers before calling send.

- Reducing the scope of thread interactions, through local caching of type IDs and reducing mutex scope to the bare minimum.

### DIFF
--- a/core/cc/string_writer.h
+++ b/core/cc/string_writer.h
@@ -1,0 +1,44 @@
+/*
+ * Copyright (C) 2017 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef CORE_STRING_WRITER_H
+#define CORE_STRING_WRITER_H
+
+#include <memory>
+#include <string>
+
+namespace core {
+    class StreamWriter;
+
+// StringWriter is a pure virtual class used to write strings to a StreamWriter.
+class StringWriter {
+public:
+    // write attempts to write the string 'data' to the underlying stream,
+    // returning false upon failure.  'data' may be in an unknown state past
+    // this call, as implementations of this interface may use move semantics
+    // as a memory optimization.
+    virtual bool write(std::string& data) = 0;
+
+    // getStream returns the underlying StreamWriter for the StringWriter.
+    virtual std::shared_ptr<StreamWriter> getStream() = 0;
+
+protected:
+    virtual ~StringWriter() {}
+};
+
+}  // namespace core
+
+#endif  // CORE_STRING_WRITER_H
--- a/gapii/cc/pack_encoder.cpp
+++ b/gapii/cc/pack_encoder.cpp
@@ -19,6 +19,7 @@
 #include "core/data/pack/pack.pb.h"
 
 #include <core/cc/stream_writer.h>
+#include "pack_string_writer.h"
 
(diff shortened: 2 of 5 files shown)


gapii/cc/pack_string_writer.h
```diff
@@ -0,0 +1,37 @@
...
+
+#ifndef GAPII_PACK_STRING_WRITER_H
+#define GAPII_PACK_STRING_WRITER_H
+
+#include "core/cc/string_writer.h"
+
+namespace gapii {
+
+// PackStringWriter is used to write 'Pack' strings to a core::StreamWriter.
+class PackStringWriter : public core::StringWriter {
```
ben-clayton: Might be clearer to call this `ChunkWriter` given that it writes Chunks [as described here](https://github.com/google/gapid/tree/master/core/data/pack#chunkt).

gapii/cc/pack_string_writer.h
```diff
@@ -0,0 +1,37 @@
...
+public:
+    typedef std::shared_ptr<PackStringWriter> SPtr;
```
ben-clayton: Nit: Maybe move this `SPtr` typedef down to the `StringWriter` interface given that this class has no additional methods?

gapii/cc/pack_encoder.cpp
```diff
@@ -30,70 +31,77 @@ using ::google::protobuf::Message;
...
+PackEncoderImpl::Shared::Shared(const std::shared_ptr<core::StringWriter>& writer)
+        : writer(writer)
+        , group_id(0) {}
 
-PackEncoderImpl::PackEncoderImpl(const std::shared_ptr<core::StreamWriter>& writer)
+PackEncoderImpl::PackEncoderImpl(const std::shared_ptr<core::StringWriter>& writer)
         : mShared(new Shared(writer))
         , mGroupId(NO_ID) {
-    writer->write(magic, sizeof(magic) - 1);
+    auto stream = writer->getStream();
```
ben-clayton: You could kill the rather smelly `StringWriter::getStream()` method if you changed this constructor to take the `StreamWriter`, and built the `PackStringWriter::create(stream)` in here.

gapii/cc/pack_encoder.cpp
```diff
@@ -30,70 +31,77 @@ using ::google::protobuf::Message;
...
-const int64_t  TAG_DECLARE_TYPE    = 1;
-const int64_t  TAG_FIRST_OBJECT    = 2;
-const uint64_t NO_ID               = 0xffffffffffffffff;
+const int      VERSION_MAJOR          = 1;
+const int      VERSION_MINOR          = 1;
+const int64_t  TAG_FIRST_GROUP        = -1;
+const int64_t  TAG_GROUP_FINALIZER    = 0;
+const int64_t  TAG_DECLARE_TYPE       = 1;
+const int64_t  TAG_FIRST_OBJECT       = 2;
+const uint64_t NO_ID                  = 0xffffffffffffffff;
```
ben-clayton: Übernit: Undo pointless whitespace changes for git blame's sake.

gapii/cc/pack_encoder.cpp
```diff
@@ -19,6 +19,7 @@
 #include "core/data/pack/pack.pb.h"
 
 #include <core/cc/stream_writer.h>
+#include "pack_string_writer.h"
```
ben-clayton: Nit: Please move next to `#include "pack_encoder.h"`

ben-clayton:
Tested, seems to work well. Thank you!

gapii/cc/pack_encoder.cpp
redmercury: I was re-indenting for some constants that I since removed, left the indent in by mistake.
Moving forward, when adding a new constant that goes past the upper bound of indentation for previous constants, should all constants be re-indented?

core/cc/string_writer.h
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+#ifndef CORE_STRING_WRITER_H
+#define CORE_STRING_WRITER_H
+
+#include <memory>
+#include <string>
+
+namespace core {
+    class StreamWriter;
```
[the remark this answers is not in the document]
redmercury: Changed, but this seems inconsistent with pack_encoder.h which forward-declares using indentation.

PR #1135, opened 2017-09-25
WIP: File format changes

dsrbecky:


gapii/cc/to_proto.h
```diff
@@ -94,53 +72,17 @@ struct ProtoConverter<::google::protobuf::RepeatedField<Out>*, core::StaticArray
     }
 };
 
-template <typename EntryOut, typename KeyOut, typename KeyIn>
```
AWoloszyn: Same as above. The only reason these are unused is that we have not up until now had a map in something that we serialize but with the push to serialize the state-block, many of these types are needed.

gapii/cc/to_proto.h
```diff
@@ -33,34 +34,11 @@ struct ProtoConverter {
     }
 };
 
-// gapii::Slice<T> -> memory_pb::Slice*
```
AWoloszyn: I am making some changes right now to serialize the state-block. 

This means serializing Slices. I would ask that you not remove these, since I have some changes here that will be terribly annoying to try and merge.

dsrbecky: Sure... I dropped the CL that removes this method... so it is back.
Note that part of the reason I removed it is that it does not even compile,
so you need to fix it if you want to use it.

gapii/cc/to_proto.h
dsrbecky: I actually did not remove the map.  I just simplified the code.  Maps are currently used in GLES.

core/data/pack/README.MD
```diff
@@ -1,63 +1,42 @@
-# Proto-Pack format Version 1.1
+# Proto-Pack format Version 2.0
```
ben-clayton: After months of attempting to find a compromise between designs, I believed we had settled on a format we were all happy with. Why are you proposing changing this now?

**There needs to be a really compelling argument to change this format again.**

gapii/cc/to_proto.h
```diff
@@ -45,25 +45,6 @@ struct ProtoConverter<memory_pb::Slice*, gapii::Slice<T>> {
     }
 };
```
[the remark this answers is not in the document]
dsrbecky: Already did.

gapii/cc/to_proto.h
```diff
@@ -23,6 +23,7 @@
 
 #include <type_traits>
 #include <unordered_map>
+#include <google/protobuf/map.h>
```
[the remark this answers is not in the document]
dsrbecky: Oh, fair enough.  I though Ian wanted the stable order just for testing during the switch.

gapii/cc/call_observer.cpp
```diff
@@ -90,20 +90,13 @@ void CallObserver::observePending() {
...
-        }
+        uint8_t* data = reinterpret_cast<uint8_t*>(p.start());
+        uint64_t size = p.end() - p.start();
+        std::string id = mSpy->sendResource(mApi, data, size);
         auto observation = new memory_pb::Observation();
         observation->set_base(p.start());
-        observation->set_size(data.count());
-        observation->set_id(reinterpret_cast<const char*>(id.data), sizeof(id.data));
+        observation->set_size(size);
+        observation->set_id(id.data(), id.size());
```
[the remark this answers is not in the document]
dsrbecky: I tried, but I got stuck in encoder and decoder in go.
We convert the data from proto to go and then remap.  One expects the type to be index and one hash.  So it is tricky when the ID field has just one type.  We could have two fields DatabaseID and FileID maybe.  Or change the decoder logic so the remap can be done as part of the proto/go conversions.

Similarly, the encoder would need to have extra logic to remap from hash to index.

Overall, a lot of extra stuff, so i wanted to check you are on board with the idea before doing anything more.  What I have so far should be hopefully non-controversial and it is good first step either way....

core/data/pack/README.MD
dsrbecky: Well.  I am mostly just trying to simplify it.
It takes me a while to parse the 1.1 format due to all the various sub blocks and +/-1 offsets.
It does not really affect the spirit of the chunks, and it would be minimal changes to the encoder/decoder.
The other things I propose are file format breaking anyway, so might as well try to tweak it so I am happy with it as well.

gapii/cc/to_proto.h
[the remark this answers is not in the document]
dsrbecky: PS: Do we actually have a use case for preserving the order of map data?

gapii/cc/call_observer.cpp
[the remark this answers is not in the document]
ben-clayton: The only places where we deal with resource IDs are with `capture.Resource`, `api.CmdObservation` and your new extra. Each of these have a hand crafted proto to go converter that can deal with the ID translation (most sensibly with the `api.ResourceReference` interface). `capture.Resource` can even have the ID field dropped as it becomes the sequential order in the stream.

Using an index for resource referencing is a sensible optimization, and I'm completely on board with it.

gapii/cc/to_proto.h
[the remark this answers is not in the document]
ben-clayton: Yes. A capture ID is a hash of the capture. Changing a parameter will reserialize the capture. Changing the parameter back again should result in the original capture ID. With this change it won't.

core/data/pack/README.MD
ben-clayton: It'll take you less time to think about the existing 1.1 format than it would to restart this discussion again, to justify why you're trying to push thread-concepts into a file-protocol, reimplement the encoders and decoders, [redesign file level versioning](https://github.com/google/gapid/blob/master/core/data/pack/pack.go), rework [Dan's optimizations](https://github.com/google/gapid/pull/1133), rewrite [the tests](https://github.com/google/gapid/blob/master/core/data/pack/pack_test.go), and reimplement the `gapit unpack` command.

> The other things I propose are file format breaking anyway, so might as well try to tweak it so I am happy with it as well.

Not a very compelling argument so far, I'm afraid.

dsrbecky: Well, I am not really trying to push thread-concepts in.  I happy to drop that part and disallow/reserve positive parent values.  I added that though in maybe cause I realized it fits in nicely and saves space.

ben-clayton: We're weeks away from trying to get a V1 release out. There are plenty of more important things for us to be looking at right now than nitting over something that is working fine for us.

dsrbecky: I do not think the rest of the code would change as significantly.  It is mostly just reordering fields removing the +1/-1 special cases.

Yeah, but after the V1 release, there is going to be argument for preserving backwards compatibility,  so this is the only to make tweaks.

PR #1138, opened 2017-09-26
Various performance related fixes.

pmuetschard:
- Move the RPC initiation off of the UI thread.
- Don't use virtual trees
- Don't create a new texture on every mouse event in the image panels (oops)

### DIFF
--- a/gapic/src/main/com/google/gapid/MainWindow.java
+++ b/gapic/src/main/com/google/gapid/MainWindow.java
@@ -116,13 +116,17 @@ public class MainWindow extends ApplicationWindow {
     Shell shell = getShell();
     Display display = shell.getDisplay();
     while (!shell.isDisposed()) {
-      if (!display.readAndDispatch()) {
-        System.err.println("false");
+      long start = System.nanoTime();
+      boolean sleep = !display.readAndDispatch();
+      long end = System.nanoTime();
+      System.err.println(TimeUnit.NANOSECONDS.toMillis(end - start) + " " + sleep);
+      if (sleep) {
         display.sleep();
       }
-      System.err.println("true");
     }
-    if (!display.isDisposed()) display.update();
+    if (!display.isDisposed()) {
+      display.update();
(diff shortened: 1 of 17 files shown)


gapic/src/main/com/google/gapid/MainWindow.java
```diff
@@ -116,13 +116,17 @@ public int open() {
     Shell shell = getShell();
     Display display = shell.getDisplay();
     while (!shell.isDisposed()) {
-      if (!display.readAndDispatch()) {
-        System.err.println("false");
+      long start = System.nanoTime();
```
ben-clayton: Maybe you want to use Ben's [Awesome Profiling Counters™](https://github.com/google/gapid/blob/master/gapidapk/android/app/src/main/java/com/google/android/gapid/Counter.java).

That or a test that does profiling?

gapic/src/main/com/google/gapid/widgets/ImagePanel.java
```diff
@@ -753,6 +752,7 @@ public void update(Renderer renderer, SceneData newData) {
               .setMinMagFilter(GL11.GL_LINEAR, GL11.GL_NEAREST)
               .setBorderColor(newData.borderColor);
           image.uploadToTexture(texture);
+          imageToTexture.put(image, texture);
```
ben-clayton: Opps. Sorry.

gapic/src/main/com/google/gapid/MainWindow.java
pmuetschard: 😄 Haha, right. I have used a profiler while working on these changes. This is a little commented out code that when un-commented will take over the UI loop and show how much time is spent on the UI thread in each loop iteration. If it gets large the UI is laggy and if it keeps going, even if nothing is changing in the UI, there's some runaway work being scheduled on the UI thread. Useful for a quick check without the profiler setup/overhead...

PR #1142, opened 2017-09-27
NOT FOR MERGING YET.

AWoloszyn:
Mostly putting this up in case I am doing something horrifyingly wrong. Better to get eyes on it earlierm before I pile a ton more work on top.


gapii/cc/to_proto.h
```diff
@@ -25,30 +25,53 @@
...
+                memory_pb::PoolData* pool_data = new memory_pb::PoolData();
+                pool_data->set_identifier(reinterpret_cast<uint64_t>(pool->base()));
+                pool_data->set_size(reinterpret_cast<uint64_t>(pool->size()));
+                std::string* dat = pool_data->mutable_data();
+                const char* base = static_cast<const char*>(pool->base());
+                dat->assign(base, base+pool->size());
+                return pool_data;
+            });
+            out->set_root(address);
+            out->set_base(identifier);
```
ben-clayton: This doesn't look right.

gapis/api/vulkan/vulkan.api
```diff
@@ -2929,9 +2929,9 @@ cmd VkResult vkAllocateMemory(
     MappedOffset:     0,
     MappedSize:       0,
     MappedLocation:   null,
-    MemoryTypeIndex:  allocateInfo.memoryTypeIndex,
-    Data:             make!u8(allocateInfo.allocationSize),
+    MemoryTypeIndex:  allocateInfo.memoryTypeIndex
   )
+  memoryObject.Data = make!u8(allocateInfo.allocationSize)
```
ben-clayton: I assume you've changed this to avoid another template change?

gapii/cc/spy_base.h
```diff
@@ -237,6 +238,40 @@ inline Slice<T> SpyBase::slice(const Slice<T>& src, uint64_t s, uint64_t e) cons
     return src(s, e);
 }
 
+struct StateSerializer {
```
ben-clayton: This class confuses me on several levels.
What's `addr`? The address of a single state object, field?

gapis/api/vulkan/vulkan.api
AWoloszyn: Yeah, I was having a hard time trying to figure out where exactly was the right place to make this change in the templates. It seemed easier to just avoid it, since avoiding the assignment is nice and easy.

core/data/protoconv/protoconv.go
```diff
@@ -49,9 +51,9 @@ func (e ErrNoConverterRegistered) Error() string {
 
 // Register registers the converters toProto and toObject.
 // toProto must be a function with the signature:
-//   func(context.Context, O) (P, error)
+//   func(context.Context, func(interface{}) uin64, O) (P, error)
```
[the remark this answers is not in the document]
AWoloszyn: Will do, I may also switch this to be more like the "updated" C++ side, in 0420bb7e1fc5bdfc03626b662d9b13ddef0fbe96. Which is to say an object that implements an interface that can have actual methods.

AWoloszyn:
Closing this, it has been replaced by another one

PR #1144, opened 2017-09-27
Fix panic during robot trace.

baldwinn860:
Using adb now expects bind to have stored a registry, and if it has not
it will panic. Adding PutRegistry before the adb.Monitor in the trace
client's Run proc fixes this panic.


baldwinn860:
I signed it. A long time ago...

PR #1145, opened 2017-09-28
Fix the disk badge image on macOS kokoro builds.

pmuetschard:
A missing optional dependency PIP package seems to have vanished.


pmuetschard:
I know, I know, minor thing.... When "mounting" the dmg, the drive icon that appears on the desktop is setup to show the GAPID icon. It stopped working and because I've been messing with the DMG stuff, I noticed. At first I thought that it was my signing scripts that were somehow dropping it. It turns out that when we had those "go is on the path twice" issues, there were other strange things going on and this python package disappeared. dmgbuild needs it to do the icon, but considers it optional and just ignores the icon if it cannot "python import" it.

PR #1147, opened 2017-10-02
Fixes to make macOS signing work.

pmuetschard:


kokoro/linux/build.sh
```diff
@@ -23,8 +23,8 @@ SRC=$PWD/github/src/github.com/google/gapid/
 wget -q https://github.com/ninja-build/ninja/releases/download/v1.7.2/ninja-linux.zip
 unzip -q ninja-linux.zip
 
-# Get GO 1.8 - Works around the cgo race condition issues.
-wget -q https://storage.googleapis.com/golang/go1.8.linux-amd64.tar.gz
+# Get GO 1.8.3
```
ben-clayton: Heh - you updated the BUILDING.md file to this a while ago.

kokoro/linux/build.sh
```diff
@@ -23,8 +23,8 @@ SRC=$PWD/github/src/github.com/google/gapid/
...
+wget -q https://storage.googleapis.com/golang/go1.8.3.linux-amd64.tar.gz
 tar -xzf go1.8.linux-amd64.tar.gz
```
ben-clayton: You need to change this line too

pmuetschard: Done, and fixed in a way so it doesn't happen again...

Issue #355, opened 2017-05-15
Improve the interceptor for short functions


ek9852:
If the function is using PLT, then interceptor will fail.
First problem is that the function is just an unconditional branch function, only 4 bytes available.
Error: Interceptor error: Intercepting function at 0x7f488308cc failed: End of function reached after 4 byte when rewriting 12 bytes

After modify some code to using the actually function in plt (glPointSizePointerOES@plt in this example), we got a different error:
Interceptor error: Intercepting function at 0x7f488308cc failed: Instruction not handled yet when one of the operand is IP1 (    ldr     x17, [x16, #4080] (OpcodeId: 1384))

0000000000004308 <glPointSizePointerOES@plt>:
    4308:       b0000030        adrp    x16, 9000 <glGetString+0x3030>
    430c:       f947fa11        ldr     x17, [x16,#4080]
    4310:       913fc210        add     x16, x16, #0xff0
    4314:       d61f0220        br      x17

00000000000058cc <glPointSizePointerOESBounds>:
    58cc:       17fffa8f        b       4308 <glPointSizePointerOES@plt>

dsrbecky:
Removing from the milestone as most functions are traced now.
However, this still needs more as mentioned in #1148

dsrbecky:
Further improved with #1157

ben-clayton:
Would you say that this is now done?

dsrbecky:
Not really

ben-clayton:
We intend to add proper tracing support into Android OS. I do not think investing any more time into `libinterceptor` is worthwhile.

PR #1148, opened 2017-10-03
Load GL commands from libhwgl.so

dsrbecky:
Related to #355

This significantly improves the situation (we can trace and replay).

However many functions still can not be patched (seems likely all just because they are "not-implemented" stubs, so not really essential).

A few valid functions (e.g. glScissor) fail to trace due to use of PC-relative instructions, which is not implemented yet.

### DIFF
--- a/gapii/cc/android/installer.cpp
+++ b/gapii/cc/android/installer.cpp
@@ -81,6 +81,7 @@ void*                                  gInterceptor = nullptr;
 InterceptFunctionFunc*                 gInterceptFunction = nullptr;
 std::unordered_map<std::string, void*> gCallbacks;
 const char*                            gDriverPaths[] = {
+  SYSTEM_LIB_PATH "libhwgl.so", // Huawei specific, must be first.
   SYSTEM_LIB_PATH "libGLES.so",
   SYSTEM_LIB_PATH "libEGL.so",
   SYSTEM_LIB_PATH "libGLESv1_CM.so",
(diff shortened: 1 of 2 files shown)


gapii/cc/android/installer.cpp
```diff
@@ -197,7 +198,19 @@ void Installer::install_gles() {
         const char* name = gapii::kGLESExports[i].mName;
         void* func_export = gapii::kGLESExports[i].mFunc;
         bool import_found = false;
-        for (int i = 0; i < NELEM(gDriverPaths); ++i) {
+        if (drivers[0] != nullptr) { // libhwgl.so
+            // Huawei implements all functions in this library with prefix,
+            // all GL functions in libGLES*.so are just trampolines to his.
```
ben-clayton: Does this also include EGL functions?

dsrbecky: I am not sure.  Should I find out?

ben-clayton: Not really important, I just thought you said it did.

ben-clayton:
> A few valid functions (e.g. glScissor) fail to trace due to use of PC-relative instructions, which is not implemented yet.

So we still can't fully trace on this device?

dsrbecky:
I would say that for practical purposes it should mostly work, but still want to see if I can do better...

ben-clayton:
`glScissor` seems like a fairly important function to trace. I'm not sure we can remove it from the milestone just yet.

PR #1161, opened 2017-10-04
Force us to keep renderpass commands in the dead-code elimination.

AWoloszyn:
This was causing some issues with a trace, where we were dropping
some drawing commands (and subsequently not getting any output).

This needs to be investigated further, but this at least allows these
traces to replay correctly.

### DIFF
--- a/gapis/api/vulkan/footprint_builder.go
+++ b/gapis/api/vulkan/footprint_builder.go
@@ -1992,6 +1992,9 @@ func (vb *FootprintBuilder) BuildFootprint(ctx context.Context,
 			ft.AddBehavior(ctx, cbh)
 		}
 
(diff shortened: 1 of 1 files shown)


gapis/api/vulkan/footprint_builder.go
```diff
@@ -1992,6 +1992,9 @@ func (vb *FootprintBuilder) BuildFootprint(ctx context.Context,
 			ft.AddBehavior(ctx, cbh)
 		}
 
+		bh.Alive = true // TODO(awoloszyn): Investigate why this is needed.
```
ben-clayton: Probably a good idea to put the issue number here.
I could write a tool quite easily that scans our codebase for issues that have since been fixed.

AWoloszyn: Done.

PR #1168, opened 2017-10-06
Add entity previews to object view.

baldwinn860:
In order to preview entities without needing to leave the grid this
change adds a video dom element and modifies the object view to create
previews of logs, reports, and videos in the grid view. Because this
makes the object view larger in size it also made sense to move it to
the right side of the screen and position it absolutely (so you don't
need to scroll to the top of the screen either).

### DIFF
--- a/test/robot/web/client/client.go
+++ b/test/robot/web/client/client.go
@@ -319,6 +319,31 @@ func robotEntityLink(path string, s interface{}) interface{} {
 	return a
 }
 
+func robotTextPreview(path string, s interface{}) interface{} {
+	id := s.(string)
+
+	div := dom.NewDiv()
+	div.Element.Style.MaxWidth = 600
+	div.Element.Style.MaxHeight = 420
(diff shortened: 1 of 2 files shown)


test/robot/web/client/client.go
```diff
@@ -319,6 +319,31 @@ func robotEntityLink(path string, s interface{}) interface{} {
 	return a
 }
 
+func robotTextPreview(path string, s interface{}) interface{} {
+	id := s.(string)
+
+	div := dom.NewDiv()
+	div.Element.Style.MaxWidth = 600
+	div.Element.Style.MaxHeight = 420
```
pmuetschard: why not the full height of the window?

baldwinn860: mostly because the maxheight property takes an int so I couldn't put in "100%". Secondly because my experience with css has been very utilitarian and I just wanted to make sure it didn't scroll off the end of the screen.

PR #1172, opened 2017-10-09
core/fault/stacktrace: Implement compression.

ben-clayton:
The goal here is to get a stacktrace down to something under a 150 bytes so it could be reported in a Google Analytics exception message.

I've implemented a bunch of different compression methods, compared these to flate, gzip, lzw and zlib. All of the custom compressors out-compress the standard compressors.

As compression payloads are small, and time isn't that important, the algorithm tries to compress with 8 different compressors - and encodes the compression type as the first 3 bits of the data.

Currently unused, but hopefully will be soon.

### DIFF
--- a/core/fault/stacktrace/pack.go
+++ b/core/fault/stacktrace/pack.go
@@ -1,0 +1,366 @@
+// Copyright (C) 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package stacktrace
+
+import (
+	"fmt"
+	"sort"
+
+	"github.com/google/gapid/core/data/binary"
+)
+
+// Tweakables
+const (
+	bitChunkMinSize   = 2
+	bitChunkIncrement = 1
+
+	debug = false
+)
+
+func printVLE(v uint64) {
+	fmt.Printf("[")
+	writeVLE(v, bitPrinter{})
+	fmt.Printf("]")
+}
+
+type bitPrinter struct{}
+
+func (bitPrinter) Write(v uint64, c uint32) {
+	for ; c > 0; c-- {
+		fmt.Printf("%v", v&1)
+		v >>= 1
+	}
+}
+func (bitPrinter) WriteBit(v uint64) {
+	if v == 0 {
+		fmt.Print("⁰")
+	} else {
+		fmt.Print("¹")
+	}
+}
+
+type bitwriter interface {
+	Write(bits uint64, count uint32)
+	WriteBit(bit uint64)
+}
+
+func writeVLE(i uint64, w bitwriter) {
+	chunkSize := uint32(bitChunkMinSize)
+	for i != 0 {
+		w.WriteBit(0)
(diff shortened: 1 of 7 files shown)


core/fault/stacktrace/pack.go
```diff
@@ -0,0 +1,374 @@
...
+		bs := binary.BitStream{}
+		writeVLE(val, &bs)
+		return bs.WritePos
+	}
+
+	for i, v := range c {
+		bestValue := v
+		bestIndex := 0
+		bestScore := size(uint64(bestIndex)) + size(bestValue)
+		for idx := 1; idx < p.history; idx++ {
```
dsrbecky: I do not get this.  Why not just: for idx := 0; idx < i; idx++ { ...

core/fault/stacktrace/pack.go
```diff
@@ -0,0 +1,374 @@
...
+	}
+
+	out := make([]uint64, len(c))
+	for i, v := range c {
+		out[i] = table[v]
+	}
+	return out
+}
+
+type compressorList []compressor
```
dsrbecky: That seems like unnecessary complexity.  Does that gain anything at all?

core/fault/stacktrace/pack.go
```diff
@@ -0,0 +1,374 @@
...
+	/* 1 */ xorCompressor{},
+	/* 2 */ compressorList{baseCompressor{}, backRefCompressor{packXor, unpackXor, 3}},
+	/* 3 */ compressorList{baseCompressor{}, backRefCompressor{packXor, unpackXor, 7}},
+	/* 4 */ compressorList{baseCompressor{}, backRefCompressor{packXor, unpackXor, 15}},
+	/* 5 */ compressorList{baseCompressor{}, backRefCompressor{packDiff, unpackDiff, 3}},
+	/* 6 */ compressorList{baseCompressor{}, backRefCompressor{packDiff, unpackDiff, 7}},
+	/* 7 */ compressorList{baseCompressor{}, backRefCompressor{packDiff, unpackDiff, 15}},
+}
+
+const compressorIdxBits = 4 // Must be large enough to hold the index of packer.
```
dsrbecky: Just VLE encode it?

core/fault/stacktrace/pack.go
```diff
@@ -0,0 +1,374 @@
...
+import (
+	"fmt"
+	"sort"
+
+	"github.com/google/gapid/core/data/binary"
+)
+
+// Tweakables
+const (
+	bitChunkMinSize   = 2
```
dsrbecky: Notes:
 - Just increments might not be optimal.  If you build a histogram of the values, you can probably calculate the optimal chunk sizes.
 - Different optimal values might apply to the back-refences and to the deltas.
 - Are all of the addresses aligned by any chance?  (i.e. can we divide all by common factor?)
 - Finally, I do not think it is really worth the time to investigate anything I just said :-D

core/fault/stacktrace/pack.go
ben-clayton: The index used to be fixed size - you right in that the history length is now dynamic, so yes, this is a good suggestion.

core/fault/stacktrace/pack.go
ben-clayton: Yeah - it lets you combine compressors together. Experimenting with different compressor combinations has been really useful.

core/fault/stacktrace/pack.go
ben-clayton: > Just increments might not be optimal. If you build a histogram of the values, you can probably calculate the optimal chunk sizes.

I actually tried this - storing the table outweighed the costs for this case. I was surprised too.

> Are all of the addresses aligned by any chance? (i.e. can we divide all by common factor?)

Nope. Again, something that surprised me. My guess is that these aren't real PC addresses, but contain other metadata somehow.

core/fault/stacktrace/pack.go
ben-clayton: Done

PR #1183, opened 2017-10-10
Add support for multisample FBOs and depth textures

ben-clayton:
Also handle `glDrawRangeElements` for the mesh view.

### DIFF
--- a/gapic/src/main/com/google/gapid/image/ArrayImage.java
+++ b/gapic/src/main/com/google/gapid/image/ArrayImage.java
@@ -435,21 +435,29 @@ public abstract class ArrayImage implements Image {
       if (isRGBA) {
         for (int i = 0, end = buffer.remaining() - 3; i <= end; ) {
           float value = buffer.get(i++);
-          min = Math.min(min, value);
-          max = Math.max(max, value);
+          if (!Float.isNaN(value) && !Float.isInfinite(value)) {
+            min = Math.min(min, value);
+            max = Math.max(max, value);
+          }
           value = buffer.get(i++);
-          min = Math.min(min, value);
-          max = Math.max(max, value);
+          if (!Float.isNaN(value) && !Float.isInfinite(value)) {
+            min = Math.min(min, value);
+            max = Math.max(max, value);
+          }
           value = buffer.get(i++);
-          min = Math.min(min, value);
-          max = Math.max(max, value);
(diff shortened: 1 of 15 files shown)


gapis/api/gles/state.go
```diff
@@ -20,55 +20,90 @@ import (
...
+type fbai struct {
+	width        uint32
+	height       uint32
+	format       GLenum // sized
+	multisampled bool
+}
+
+func attachmentToEnum(a api.FramebufferAttachment) (GLenum, error) {
+	switch a {
+	case api.FramebufferAttachment_Color0:
```
dsrbecky: So you know what although the enums are finite, the number of attachments is, in theory, unbounded. (implementation must support at least 4)

So ideally this should handle Depth and Stencil and do cast and arithmetic for the rest.

Alternatively what do you think of making FramebufferAttachment an integer instead?  (where >= 0 is the respective colour, and negative values are used for depth/stencil)

ben-clayton: Understood, but these are declared in the protobuf. [Negative enums in protos are discouraged](https://developers.google.com/protocol-buffers/docs/proto#enum).
I'll have a think about this, but this issue has not been made better or worse by this PR - opting to defer.

dsrbecky: How about low two bits indicate type?

I am kind ok fine with the current approach, as well.  But let's maybe add a few more?

gapic/src/main/com/google/gapid/image/ArrayImage.java
```diff
@@ -435,21 +435,29 @@ public static PixelInfo compute(FloatBuffer buffer, boolean isRGBA) {
       if (isRGBA) {
         for (int i = 0, end = buffer.remaining() - 3; i <= end; ) {
           float value = buffer.get(i++);
-          min = Math.min(min, value);
-          max = Math.max(max, value);
+          if (!Float.isNaN(value) && !Float.isInfinite(value)) {
```
AWoloszyn: Thank you!

gapis/api/gles/state.go
ben-clayton: If you want more changes, let's do it in a follow up CL. This PR is urgent.

AWoloszyn: Since extending the Enum would require changes to the UI as well, that is probably best done later?

The other changes look fine to me.

PR #1185, opened 2017-10-11
Fix crashes at vkCreateImageView

Qining:


gapis/api/vulkan/footprint_builder.go
```diff
@@ -1990,21 +1994,20 @@ func (vb *FootprintBuilder) BuildFootprint(ctx context.Context,
 			execInfo.beginRenderPass(ctx, vb, cbh, rp, fb)
 			execInfo.renderPassBegin = newForwardPairedLabel(ctx, cbh)
 			ft.AddBehavior(ctx, cbh)
+			cbh.Alive = true // TODO(awoloszyn)(BUG:1158): Investigate why this is needed.
```
AWoloszyn: Do we still need the comment here? or is this now correct.

Qining: Not sure, I think we need to test with DOOM to tell if the dep between image and image view fixes the problem. But for now, this is still a workaround.

Qining: I mean, test with DOOM on Windows to tell if these "keepAlive" can be dropped.

PR #1186, opened 2017-10-11
Fix monitor/scheduler race condition.

baldwinn860:
There was a race condition that caused the scheduler to update robot
tasks with incomplete information. Managers would monitor for changes
inside a goroutine that would compete with the update logic of the data
owner. All initial data would be read in at once, but still in another
thread, which caused the scheduler to update without full knowledge of
entity state, this would cause extraneous tasks to get spawned every
time robot was restarted.

The fix is to initialize the owner's data with a non monitoring search
in the main thread before spawning goroutines to continue monitoring
updates. This ensures we have the complete initial information prior to
the first update. The other changes ensure that the monitoring updates
do not append extraneous entries.

### DIFF
--- a/test/robot/monitor/monitor.go
+++ b/test/robot/monitor/monitor.go
@@ -121,27 +121,52 @@ func Run(ctx context.Context, managers Managers, owner DataOwner, update func(ct
 }
 
 func monitor(ctx context.Context, managers *Managers, owner DataOwner) error {
+	initial := &search.Query{}
 	// TODO: care about monitors erroring
-	all := &search.Query{Monitor: true}
+	monitor := &search.Query{Monitor: true}
 	if managers.Job != nil {
-		crash.Go(func() { managers.Job.SearchDevices(ctx, all, owner.updateDevice) })
-		crash.Go(func() { managers.Job.SearchWorkers(ctx, all, owner.updateWorker) })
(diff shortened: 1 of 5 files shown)


test/robot/monitor/monitor.go
```diff
@@ -121,27 +121,52 @@ func Run(ctx context.Context, managers Managers, owner DataOwner, update func(ct
...
 
 func monitor(ctx context.Context, managers *Managers, owner DataOwner) error {
+	initial := &search.Query{}
 	// TODO: care about monitors erroring
-	all := &search.Query{Monitor: true}
+	monitor := &search.Query{Monitor: true}
 	if managers.Job != nil {
-		crash.Go(func() { managers.Job.SearchDevices(ctx, all, owner.updateDevice) })
-		crash.Go(func() { managers.Job.SearchWorkers(ctx, all, owner.updateWorker) })
+		if err := managers.Job.SearchDevices(ctx, initial, owner.updateDevice); err != nil {
```
ben-clayton: I'm a little upset that I can't find a way to refactor this into something more compact without resorting to reflection. :(

baldwinn860: Yeah, me too, I had written like three different solutions but this was by far the most succinct, all the rest kind of exploded into lots of code changes.

PR #1196, opened 2017-10-13
Guarantee the framebuffer content of the last bound queue is not discarded by DCE

Qining:
Add a "Request()" function in the back propagation machine to differentiate normal alive commands, and request commands where the last bound framebuffer content will be requested.

Keep the last bound framebuffer content for each "behavior" of the normal commands (not subcommands) in the machine. Later when the user requests the command, the "Request()" function is called with the corresponding behavior index, we can mark a use of the last bound framebuffer content, so that all the commands that contributes to the content of the last bound framebuffer will be kept alive.

### DIFF
--- a/gapis/api/transform/dce.go
+++ b/gapis/api/transform/dce.go
@@ -182,9 +182,13 @@ func (t *DCE) backPropagate(ctx context.Context) (
 		if bh.Aborted {
 			continue
 		}
-		if t.requests.contains(fci) ||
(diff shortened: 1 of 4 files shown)


Qining:
Need to check if this CL hurts the performance benefit of DCE badly.

AWoloszyn:
I would be less concerned about replay performance if this improves correctness. If we notice replay performance being a large problem, we can fix it, but I would rather have this correct, than fast. It also looks that it will only affect replay performance when clicking on one of the affected Atoms.

PR #1198, opened 2017-10-16
Fix packed U32 Formats in Vulkan.

AWoloszyn:
Based on how packed formats are represented in Vulkan, we were
interpreting them flipped.


core/stream/datatype.go
```diff
@@ -108,7 +108,10 @@ func (t DataType) Format(f fmt.State, r rune) {
 }
 
 // Bits returns the size of the data type in bits.
-func (t *DataType) Bits() uint32 {
+func (t *DataType) Bits() uint32 {// F10 represents a 10-bit unsigned floating-point number.
+	F10 = DataType{Signed: false, Kind: &DataType_Float{&Float{5, 5}}}
```
ben-clayton: What's this doing here?

AWoloszyn: That is a terrific question, I assume a copy-paste somewhere, will remove.

gapis/api/vulkan/resources.go
```diff
@@ -253,9 +253,9 @@ func getImageFormatFromVulkanFormat(vkfmt VkFormat) (*image.Format, error) {
 	case VkFormat_VK_FORMAT_R32_SFLOAT:
 		return image.NewUncompressed("VK_FORMAT_R32_SINT", fmts.R_F32), nil
 	case VkFormat_VK_FORMAT_B10G11R11_UFLOAT_PACK32:
-		return image.NewUncompressed("VK_FORMAT_B10G11R11_UFLOAT_PACK32", fmts.BGR_F10F11F11), nil
+		return image.NewUncompressed("VK_FORMAT_B10G11R11_UFLOAT_PACK32", fmts.RGB_F11F11F10), nil
 	case VkFormat_VK_FORMAT_E5B9G9R9_UFLOAT_PACK32:
-		return image.NewUncompressed("VK_FORMAT_E5B9G9R9_UFLOAT_PACK32", fmts.RGBE_U9U9U9U5), nil
+		return image.NewUncompressed("VK_FORMAT_E5B9G9R9_UFLOAT_PACK32", fmts.EBGR_U5U9U9U9), nil
```
pmuetschard: Actually, this one was correct, so this change should be undone.

I think it is worth adding a comment to all the "VK_*BGR*" -> "fmts.*RGB*" mappings to communicate that this re-ordering is due to the little endianess or our stream package, to make sure readers don't get confused and think it's a bug.

AWoloszyn: Thanks for catching this the fix is in #1203

PR #1203, opened 2017-10-16
Fixed mistake in previous MR to handle PACK32 formats.

AWoloszyn:
I had flipped EBGR, with the rest of the pack32 formats, but
we had already fixed this specific case previously.


pmuetschard:
Can you also get rid of the now unused ebgr.go, please?

AWoloszyn:
Will do.

PR #1204, opened 2017-10-16
Show memory observations of the initial call for Vulkan Subcommands

Qining:


Qining:
@ben-clayton @AWoloszyn Doubled checked with the trace Ben gave me, all the memory observations are unchanged. This CL show no changes on the GL side.

Issue #1184, opened 2017-10-11
Allow the user to configure tone mapping output.


AWoloszyn:
Either custom curves, custom shaders, or some other method.
Currently any exponential data is hard to interpret in the texture output.

ben-clayton:
Working on it.

Issue #1141, opened 2017-09-27
Cannot see entire framebuffer when zooming in


tomgoogle:
The scroll bars extend over the framebuffer itself when zooming in, partially obscuring it. Most easily visible when trying to look at the "Development Build" text in a Unity dev build - when zooming into the bottom right of the image you cannot see this text.

PR #1206, opened 2017-10-17
gapic: Add histogram and use it for better tone-mapping.

ben-clayton:
Use a percentile to pick upper and lower limits for display tone-mapping when the format is HDR.

Fixes: #1184


ben-clayton:
Rather heavily dependent on #1141 being fixed, as the RHS of the histogram is completely hidden on macOS.

PR #1207, opened 2017-10-17
Switched to using Application.<init> instead of onCreate.

AWoloszyn:
Some applications do not call super().onCreate. The documentation
states that they must, but in reality some do not.

Since the constructor MUST get hit at some point, we can break their
instead. The ClassLoader is already set up at that point anyway, so GL
still works.

### DIFF
--- a/gapii/client/jdwp_loader.go
+++ b/gapii/client/jdwp_loader.go
@@ -52,12 +52,13 @@ func waitForOnCreate(ctx context.Context, conn *jdwp.Connection, wakeup jdwp.Thr
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


cmd/do/config.go
```diff
@@ -223,7 +223,7 @@ func (validator) ValidateAndroidNDKRoot(path file.Path) error {
 			}
 			major, minor := groups[1], groups[2]
 			if major != "15" && major != "16" {
-				return fmt.Errorf("Found NDK %v.%v. Must be r15 or r16", major, minor)
```
ben-clayton: Accidental revert?

AWoloszyn:
Thanks, if someone else could make sure this works on their devices as well, I have tested Pixel C and PIxelXL.

PR #1217, opened 2017-10-18
Fix macOS and Linux launch scripts to work with paths containing spaces.

pmuetschard:


pmuetschard:
> Is this a real problem?

Probably not. I noticed it happened when I had two DMGs mounted on osx and tried running it from within the dmg. The second one was mounted at "/Volumes/GAPID 1/" (or 2, but notice the space) and failed to start.

Issue #1194, opened 2017-10-13
Cannot get the package listing on Pixel C.


AWoloszyn:
Gapic does not give me a package list when asked on the Pixel C.
Tracing/replaying does work if I type in the package name manually.

```
Exception in thread "com.google.gapid.server.GapitPkgInfoProcess$1" java.lang.RuntimeException: The gapit command didn't produce the data marker.
        at com.google.gapid.server.GapitPkgInfoProcess.lambda$createStdoutHandler$0(GapitPkgInfoProcess.java:115)
        at com.google.gapid.server.ChildProcess$BinaryHandler.run(ChildProcess.java:245)
        at com.google.gapid.server.ChildProcess$OutputHandler.lambda$start$0(ChildProcess.java:156)
        at java.lang.Thread.run(Thread.java:745)
W20171013-144211399[server.ChildProcess-gapit][views.ActivityPickerDialog$2.onFailure] Failed to read package info
W20171013-144211399[server.ChildProcess-gapit][views.ActivityPickerDialog$2.onFailure] java.lang.Exception: The gapit command failed:
18:42:11.393 F: <gapit> Main failed
Error: getting package list
   Cause: Connecting to service port
   Cause:
   Cause: Timeout connecting to service
```

ben-clayton:
Can you please look in logcat for the exception in the gapid APK?

AWoloszyn:
```
10-16 15:21:08.060 15875 16410 E gapid-pkginfo: Error occurred
10-16 15:21:08.060 15875 16410 E gapid-pkginfo: java.util.concurrent.ExecutionException: android.content.res.Resources$NotFoundException: Resource ID #0x7f030000
10-16 15:21:08.060 15875 16410 E gapid-pkginfo:         at java.util.concurrent.FutureTask.report(FutureTask.java:94)
10-16 15:21:08.060 15875 16410 E gapid-pkginfo:         at java.util.concurrent.FutureTask.get(FutureTask.java:164)
10-16 15:21:08.060 15875 16410 E gapid-pkginfo:         at com.google.android.gapid.SocketWriter.connectAndWrite(SocketWriter.java:46)
10-16 15:21:08.060 15875 16410 E gapid-pkginfo:         at com.google.android.gapid.PackageInfoService.handleSendPackageInfo(PackageInfoService.java:240)
[...]

PR #1219, opened 2017-10-18
PackageInfoService: Catch Resources.NotFoundException

ben-clayton:
Fixes: #1194


ben-clayton:
@AWoloszyn - do you mind giving this a quick test to see if it fixes the Pixel C?

Issue #1205, opened 2017-10-17
Framebuffer can't be shown after trace UE4 project


tiantian20007:
All Draw call, OpenGL command,  textures, Vertex data, Shaders and has been captured and shown correctly, except the framebuffer. All framebuffer view just showed: 
read tcp 127.0.0.1:64243->127.0.0.1:64241: wsarecv: An existing connection was forcibly closed by the remote host.

I'm using gapid 0.6.2 on win10 for Android project built by UnrealEngine 4.15 and run on Android 7.0 device.

.gfxtrace file:
[DemoProject_20171017_1813.zip](https://github.com/google/gapid/files/1393218/DemoProject_20171017_1813.zip)

![_20171017182304](https://user-images.githubusercontent.com/406098/31659909-406e112c-b368-11e7-9bff-9db61c5e5836.png)
![_20171017182311](https://user-images.githubusercontent.com/406098/31659912-44cbb21a-b368-11e7-8b2c-5a32c9df666b.png)

ben-clayton:
Hi @tiantian20007,

Thank you for the bug report. It looks like the replay has crashed.
Are you happy to share the `.gfxtrace` file with me to investigate further?

Many thanks,
Ben

tiantian20007:
.gfxtrace file attached

ben-clayton:
Hi @tiantian20007 - thank you for the trace. I found and fixed an issue that may have caused the issue you were seeing. Please can you backup and replace the gapis.exe with this version:

[gapis.exe.zip](https://github.com/google/gapid/files/1395291/gapis.exe.zip)

And see if this fixes your issue?

Alternatively, you can wait until we push our next version (likely 0.9.0).

Many thanks,
Ben

tiantian20007:
After replacing gapis.exe, the application cannot be launched by gapid.bat

ben-clayton:
This might be due to Windows security. If you just try running gapis.exe do you get a notification about running the program downloaded from the internet? It so, confined the dialog, close the app and try the batch file again. 
If that doesn't work, we'll be pushing a new release shortly anyway.

ben-clayton:
I've landed a whole bunch of other fixes specific for UE4. I'm moderately confident this (and other issues) are fixed. I'll try to get another release out this week.

ben-clayton:
[v0.9.0 has been released](https://github.com/google/gapid/releases/tag/v0.9.0).
Please let me know if you're still having issues @tiantian20007!

[...]

PR #1220, opened 2017-10-18
gles.moveClientVBsToVAs: Write VAP cmds on draw thread.

ben-clayton:
The calls to `glVertexAttribPointer` may have come from a thread that has since unbound the context. Just use the thread that is making the draw call.

Issue: #1205


ben-clayton:
> Maybe add a comment to the code?

Done.

PR #1223, opened 2017-10-18
Fix indices for subcommand in Vulkan when resolving command groups

### LINKED ISSUE #1214: Vulkan Command Tree Grouping Returns Wrong Command paths.
Given a trace with the following tree:
```
- Frame 1 (0-363)
- Frame 2 (364-372)
  - Draw 1 (364-371)
    - 364: vkWaitForFences
    - ...
    - 371: vkQueueSubmit
      - [371 0]
        - [371 0 0]
          - RenderPass
            - 371.0.0.0: vkCmdBeginRenderPass
            - ...
            - 371.0.0.6: vkCmdEndRenderPass
  - vkQueuePresnetKHR
```

When fetching the nodes from the server the following values are returned:

The node at `[1, 0, 7]` (vkQueueSubmit) is for command: `[371]` - Correct.
The node at `[1, 0, 7, 0]` is for command: `[371, 0, 0]` - Should be `[371, 0]`.
The node at `[1, 0, 7, 0, 0]` is for command `[371, 0, 0, 0]` - Should be `[371, 0, 0]`.
The node at `[1, 0, 7, 0, 0, 0]` (RenderPass) is for commands `[0]-[6]`- Should be `[371, 0, 0]`.
The nodes at `[1, 0, 7, 0, 0, 0, x]` are for commands `[371, 0, 0, x]`, which is correct again.

This results in strange behavior, especially when selecting the RenderPass node.

Qining:
Should fix #1214

Turns out we just need to remove two historical lines of code

### DIFF
--- a/gapis/resolve/command_tree.go
+++ b/gapis/resolve/command_tree.go
@@ -118,8 +118,6 @@ func CommandTreeNode(ctx context.Context, c *path.CommandTreeNode) (*service.Com
 		commandEnd := append([]uint64{}, item.Id...)
 		if len(item.Id) > 1 {
 			g = fmt.Sprintf("%v", item.Id)
-			commandStart = append(commandStart, uint64(0))
-			commandEnd = append(commandEnd, uint64(item.SubGroup.Count()-1))
 			count = uint64(item.SubGroup.Count())
 		}
 		return &service.CommandTreeNode{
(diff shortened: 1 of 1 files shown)


gapis/resolve/command_tree.go
```diff
@@ -118,8 +118,6 @@ func CommandTreeNode(ctx context.Context, c *path.CommandTreeNode) (*service.Com
 		commandEnd := append([]uint64{}, item.Id...)
```
ben-clayton: Do we need these now? Can't we just use item.Id where ever these were used?

PR #1224, opened 2017-10-18
Replace X and Y in geometry view with matrix.

### LINKED ISSUE #1215: Gimbal lock in Geometry view
In Zooshi, try finding the draw call for the Hippo and try to get him standing upright and facing you. It's infuriatingly impossible as you lose an axis of rotation. 

screen shot 2017-10-17 at 5 25 05 pm

AWoloszyn:
This allows unlimited rotation in both axes.
Fixes #1215


AWoloszyn:
As a note, this method doesn't invert the horizontal drag after rotating over the top/bottom.

ben-clayton:
After consideration it's clear that different people prefer different camera modes. Can I suggest we add an option to the settings dialog for camera style? If we do that, can I also add my preferred camera style - WASD?

PR #1225, opened 2017-10-18
If we cannot find the signature for getClassLoader, try an alternate.

AWoloszyn:


ben-clayton:
Shall we land this change? @tomgoogle do you know if this fixes your issue?

AWoloszyn:
I can certainly merge this, it is unlikely to break anything that was not previously broken.

PR #1226, opened 2017-10-19
Add framework for anonymous user statistics (entirely disabled)

ben-clayton:
@tomgoogle


ben-clayton:
I need to look into why this is failing - I think I need to change some environment flags.

gapic/src/main/com/google/gapid/views/SettingsDialog.java
```diff
@@ -72,20 +84,32 @@ protected Control createDialogArea(Composite parent) {
...
+    // sendAnalytics = Widgets.createCheckbox(container, "", settings.analyticsEnabled());
+
+    createLabel(container, "Path to adb:");
     adbPath = withLayoutData(new FileTextbox.File(container, settings.adb) {
       @Override
       protected void configureDialog(FileDialog dialog) {
         dialog.setText("Path to adb:");
       }
     }, new GridData(SWT.FILL, SWT.FILL, true, false));
+    adbPath.addBoxListener(SWT.CHANGED, this::onSettingChanged);
```
pmuetschard: You want SWT.Modify.

gapic/src/main/com/google/gapid/models/Settings.java
```diff
@@ -45,6 +47,11 @@
   private static final Logger LOG = Logger.getLogger(Settings.class.getName());
   private static final String SETTINGS_FILE = ".gapic";
   private static final int MAX_RECENT_FILES = 16;
+  private final List<Listener> listeners = Lists.newArrayList();
```
pmuetschard: Please use Events.ListenerCollection

gapic/src/main/com/google/gapid/models/Settings.java
```diff
@@ -76,6 +83,27 @@
   public boolean autoCheckForUpdates = true;
   public boolean updateAvailable = false;
   public long lastCheckForUpdates = 0; // milliseconds since midnight, January 1, 1970 UTC.
+  public String analyticsClientId = ""; // Empty means do not track.
+
+  /**
+   * listen registers the listener for changes.
+   */
+  public void listen(Listener listener) {
```
pmuetschard: Please call this addListener and add a removeListener method.

gapis/service/service.proto
```diff
@@ -313,6 +313,48 @@ message FindResponse {
...
+  MouseEnter = 9;
+  MouseExit = 10;
+  KeyPressUpOrDown = 11;
+  KeyPressLeftOrRight = 12;
+}
+
+message ClientInteraction {
+  string view = 1;
+  string control = 2;
+  ClientAction action = 3;
```
pmuetschard: The implementation just converts this to a string... so why not just make it a string in the API?

gapic/src/main/com/google/gapid/views/SettingsDialog.java
ben-clayton: Done.

gapic/src/main/com/google/gapid/models/Settings.java
ben-clayton: Done.

gapic/src/main/com/google/gapid/models/Settings.java
ben-clayton: Done.

gapis/service/service.proto
ben-clayton: It's discouraged to expose string fields where unnecessary. This greatly increases the possibility of accidentally leaking PII. The couple of strings I have already are questionable, and probably should be turned into some sort of enum if we can agree on them.

ben-clayton:
PTAL

gapic/src/main/com/google/gapid/models/Settings.java
```diff
@@ -76,6 +82,28 @@
   public boolean autoCheckForUpdates = true;
   public boolean updateAvailable = false;
   public long lastCheckForUpdates = 0; // milliseconds since midnight, January 1, 1970 UTC.
+  public String analyticsClientId = ""; // Empty means do not track.
+
+  /**
+   * addListener registers the listener for changes.
```
pmuetschard: "Registers the listener for changes." and "Removes the listener for chagnes." below. Javadoc automatically adds in the function names and expects the first sentence to be a "verb phrase" for methods and a "noun phrase" for classes. See section 7.2 https://google.github.io/styleguide/javaguide.html#s7.2-summary-fragment

(and TIL, the tip on 7.2, I've been doing that lots...)

ben-clayton: Done

PR #1232, opened 2017-10-19
Fix trace failures not uploading logs.

baldwinn860:
Despite this being against golangs ideomatic style to returning errors
only with nil objects, the clients return "valid" output objects during
an error, this is to communicate logs and error strings that otherwise
could not get uploaded. A fix changed this behavior in the trace client,
and it has not uploaded any logs since.

This fix reverts that change since the function is unexported, and the
alternative would be to populate an output parameter, which also goes
against golang style.

### DIFF
--- a/test/robot/trace/client.go
+++ b/test/robot/trace/client.go
@@ -144,6 +144,8 @@ func (r *runner) trace(ctx context.Context, t *Task) error {
 	return r.manager.Update(ctx, t.Action, status, output)
 }
 
+// doTrace extracts input files and runs `gapit trace` on them, capturing the output. The output object will
+// be partially filled in the event of an upload error from store in order to allow examination of the logs.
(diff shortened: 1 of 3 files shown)


test/robot/trace/client.go
```diff
@@ -207,12 +207,12 @@ func doTrace(ctx context.Context, action string, in *Input, store *stash.Client,
 	log.I(ctx, output)
 	logID, err := store.UploadString(ctx, stash.Upload{Name: []string{"trace.log"}, Type: []string{"text/plain"}}, output)
 	if err != nil {
-		return nil, err
+		return outputObj, err
```
ben-clayton: Please document the function to state that it can return a partially filled object and error.

baldwinn860: ok done.

PR #1233, opened 2017-10-19
Vulkan: cache the memory requirement info for images at creation time

Qining:
Necessary for  #828

### DIFF
--- a/gapis/api/vulkan/templates/vk_spy_helpers.cpp.tmpl
+++ b/gapis/api/vulkan/templates/vk_spy_helpers.cpp.tmpl
@@ -420,18 +420,60 @@ uint32_t VulkanSpy::SpyOverride_vkCreateBuffer(
     }
 }
 
-uint32_t VulkanSpy::SpyOverride_vkCreateImage(
-        VkDevice                     device,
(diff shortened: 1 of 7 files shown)


gapis/api/vulkan/templates/vk_spy_helpers.cpp.tmpl
```diff
@@ -420,18 +420,60 @@ uint32_t VulkanSpy::SpyOverride_vkCreateBuffer(
...
+  override_create_info.musage |=
+      VkImageUsageFlagBits::VK_IMAGE_USAGE_TRANSFER_SRC_BIT;
+  result = mImports.mVkDeviceFunctions[device].vkCreateImage(
+      device, &override_create_info, pAllocator, pImage);
+  if (result != VkResult::VK_SUCCESS) {
+    return result;
+  }
+
+  if (pMemoryRequirements != nullptr) {
+    mImports.mVkDeviceFunctions[device].vkGetImageMemoryRequirements(
```
Qining: This is only called in this override function, so swapchain images won't have valid memory requirement info populated. Their info will be all 0.

PR #1237, opened 2017-10-21
Support sparse binding in Vulkan for buffers and opaque images

Qining:
Resolve partially #828

### DIFF
--- a/core/cc/interval_list.h
+++ b/core/cc/interval_list.h
@@ -118,6 +118,10 @@ public:
     // end() returns the pointer to one-past the last interval in the list.
     inline const T* end() const;
 
+    // operator[] returns the const reference to the element at the specified
+    // location pos.
+    inline const T& operator[](size_t pos) const;
+
 protected:
     // rangeFirst returns the index of the first interval + bias that touches or
     // exceeds start.
@@ -238,6 +242,11 @@ inline const T* CustomIntervalList<T>::end() const {
     } else {
         return nullptr;
     }
+}
+
+template<typename T>
+inline const T& CustomIntervalList<T>::operator[](size_t pos) const {
+  return mIntervals[pos];
 }
 
 template<typename T>
--- a/gapii/cc/vulkan_extras.inl
+++ b/gapii/cc/vulkan_extras.inl
@@ -208,3 +208,45 @@ uint32_t SpyOverride_createImageAndCacheMemoryRequirements(
 void SpyOverride_cacheImageSparseMemoryRequirements(
     VkDevice device, VkImage image, uint32_t count,
     VkSparseImageMemoryRequirements* pSparseMemoryRequirements);
+
+
+class SparseBindingInterval {
+ public:
+  SparseBindingInterval(const VkSparseMemoryBind& bind)
+      : resourceOffset_(bind.mresourceOffset),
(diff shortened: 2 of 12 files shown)


AWoloszyn:
So far I don''t see any problems with this. I will likely look into it a bit more deeply, but it doesn't seem to support mid-execution capture correctly. Handling Sparse Binding with MEC  will be a bit trickier,

Qining:
Right, this so far only support **full trace** and **SPARSE_BINDING_BIT**, **MEC** and **SPARSE_RESIDENCY_BIT** are not there yet.

Actually for **SPARSE_RESIDENCY_BIT**, things can be a little bit complicated, as apps are allowed to bind 'part of a mip image', specified with 'extent3D' and 'offset3D'. So we will have 3D overlapping problem. For buffers and opaque images, we only have 1D overlapping.

Qining:
@AWoloszyn I think we need that in the trace side for MEC. My plan is that, in `EnumerateVulkanResources` we need to use the records to find the memories to bind. And I did this in cpp directly because in API file we don't have `list or vector` and we don't have the `merge, intersect, replace` interfaces.

AWoloszyn:
So for the upcoming change to MEC, we have to have everything in the state block. We will be doing away with the EnumerateVulkanResources in favour of just serializing the state block itself. This means that all data that is tacked on to the side of the Stat block should be removed, unless it can be re-created from the contents of the state block itself. 

In the interest of not having to switch this a second time, I would much rather that we keep the data in the state block itself. We can add externs to manipulate it if we want (merge/intersect/replace) but the data itself should be inside the state block.

Qining:
@AWoloszyn OK. Then I think I just need to keep the sparse binding info maps (the maps in the buffer objects and image objects) updated (like what I've done in externs.go), so that when we serialize the state block we have the latest binding info? Probably for now, I can still utilize the RecreateBindImage/BufferMemory to pass the sparse binding info to the replay side, we can always drop those recreate commands later when we switch to state block serialization?

Qining:
If the sparse bound memories are freed, we need to update the lists/maps to reflect the changes.

Qining:
According to the note under spec 28.7.6

> It is important to note that freeing a VkDeviceMemory object with vkFreeMemory will not cause resources (or resource regions) bound to the memory object to become unbound. Access to resources that are bound to memory objects that have been freed will result in undefined behavior, potentially including application termination.

It seems more reasonable to 1) do not update sparse bindings after vkFreeMemory() calls, 2) when recreate opaque image/buffer sparse bindings, check the liveness of the memories first, report warning/error if those memories do not exist any more.

We probably should not proceed once we found an opaque image or buffer is not entirely bound, because either we stop mutation at RecreateBindingXXX commands, or we will crash when we copy the data to the image or buffer.

AWoloszyn:
As for your second comment, I do not think that is necessarily true. If we only have a partially bound object, we should fill in the parts of it that have been bound, if sparseresidency is there.

Qining:
Right, if SPARSE_RESIDENCY is there. But opaque image sparse memory binding is the case considering the whole image as a black box. So this is only true for opaque image bindings, not true for sparse bindings  with SPARSE_RESIDENCY (not opaque).

PR #1238, opened 2017-10-23
Make map!(K,V) a reference type in C++ to match Go. 

### LINKED ISSUE #569: Support ref!map!T type
It would be occasionally useful in the API file to have by-reference map

dsrbecky:
Fixes #569


gapis/api/gles/gles.api
```diff
@@ -572,14 +572,9 @@ ContextID              NextContextID
...
-  context := Contexts[$Thread]
-  if context == null {
-    _ = newMsg(SEVERITY_ERROR, new!ERR_NO_CONTEXT_BOUND(thread: $Thread))
-    abort
-  }
-  return context
-}
+// Internal "thread-local" global for quick access the current context.
+// It is set in preMutate function (gles.go) before command is mutated.
+@hidden ref!Context ctx
```
ben-clayton: Isn't this exactly what `CurrentContext` is?

dsrbecky: Yes, but I didn't want to go and rename all of references to 'ctx' with 'CurrentContext'.
It is more verbose, and more importantly, it would make one massive diff.
I am happy to do that as a followup.

ben-clayton: Please add that change to this PR. I'm not happy with two copies of the same thing.
Also - aren't you missing logic to set this in GAPII?

dsrbecky:
I have dropped all the commits except for the first two

ben-clayton:
Thank you.
I believed I was happy with the 3rd CL too, but my memory is too crap to remember what it was. :)

dsrbecky:
My memory is bad as well.  I was not 100% sure either.

dsrbecky:
The removed commits live here: https://github.com/dsrbecky/gapid/tree/remove-GetContext

Issue #1134, opened 2017-09-25
Framebuffer view does not always match thumbnail for groups


ben-clayton:
We now do 'smart' analysis to show a sensible thumbnail for a group in the command tree.
Clicking on that group will always show the framebuffer after the last command in the group.

For development Unity builds of GVR applications, the frame group ends with a non-scene framebuffer attached leading to the undefined-framebuffer pattern being shown.

We should probably just change the rules so that the command selection for a group matches the thumbnail.

pmuetschard:
I propose we change the `getFramebufferAttachment` API to use the following request:

```protobuf
message GetFramebufferAttachmentRequest {
  path.Device device = 1;
  api.FramebufferAttachment attachment = 2;
  RenderSettings settings = 3;
  UsageHints hints = 4;

  oneof after {
    path.Command command = 5;
    path.CommandTreeNode command_tree_node = 6;
  }
}
```

What do you think?

ben-clayton:
Not a bad idea. Let me investigate the amount of wiring needed to implement this.

ben-clayton:
I got most the way through implementing this, and then suddenly wondered:

Does it make sense to have the framebuffer view show something mid-way through the group, but have everything else (state, memory, textures, etc) show the state at the end of the group?

What are your thoughts?

dsrbecky:
Fair point, it would make sense to sync everything else to match the point at framebuffer.

In my mind, a group node should have one field which is "the representative command", and we should get the thumbnail using that.  Same goes for framebuffer.  And now that you mention it, it would make sense to apply it to all other state as well.

ben-clayton:
Thanks @dsrbecky. I kinda agree, but I'm also troubled with the idea of the state not being at the end when clicking on a group or a command with children.

@AWoloszyn, @pmuetschard - what are your thoughts?

AWoloszyn:
Sorry, I thought I had already responded.

My #1 concern would be that everything matches. It would be terrible if the State != What created the framebuffer.

That being said, I also like the idea of that being the last command in the group, but I totally understand that in some cases, the "last command" is not actually a useful piece of data to be visualizing.

Would doing something "simple" like a small area that says WHAT command this is the data for be sufficient. (Somewhere global-ish, that is always visible), then we can at least know WHICH command the thumbnail/framebuffer is for.

pmuetschard:
[...]

PR #1240, opened 2017-10-23
service.proto: Add representation field to CommandTreeNode.

ben-clayton:
Use this for picking a command to show data for when selecting a group.

Fixes: #1134


ben-clayton:
@Qining - please can you do a quick sanity check for Vulkan?

Qining:
Tested with Vulkan traces, works now.

PR #1242, opened 2017-10-23
Remove resource IDs on-disk. Reference them by index instead.

dsrbecky:
The first saved resource has index 1, the next has index 2, etc.
Memory observations and others can use the index to reference it.

This saves over 10% of the trace data (excluding raw resource bytes).


ben-clayton:
Please can you add a version code to `capture.proto`'s `Header` message type, and check for a non-0 value? This is a file breaking change, and we should report it with a `service.ErrUnsupportedVersion`.

AWoloszyn:
Can you double check this for vulkan mid-execution capture,  it should be fine, but there is a chance some strangeness comes into play because of how we drop observations.

dsrbecky:
Rebased on top of #1248.
Also passes basic sanity checks.

PR #1245, opened 2017-10-23
Handle copy/paste/cut events on spinners for macOS.

### LINKED ISSUE #1067: Pasting doesn't work in the goto command view
(but middle click works on linux?)

pmuetschard:
Fixes #1067

### DIFF
--- a/gapic/src/main/com/google/gapid/widgets/Widgets.java
+++ b/gapic/src/main/com/google/gapid/widgets/Widgets.java
@@ -24,6 +24,7 @@ import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.SettableFuture;
 import com.google.gapid.models.Models;
 import com.google.gapid.server.Client;
+import com.google.gapid.util.OS;
 import com.google.gapid.views.AtomEditor;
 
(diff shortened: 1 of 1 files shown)


gapic/src/main/com/google/gapid/widgets/Widgets.java
```diff
@@ -342,6 +343,18 @@ public static Spinner createSpinner(Composite parent, int value, int min, int ma
     result.setMinimum(min);
     result.setMaximum(max);
     result.setSelection(value);
+
+    if (OS.isMac) {
+      result.addListener(SWT.KeyUp, e -> {
+        if ((e.stateMask & (SWT.CONTROL | SWT.COMMAND)) != 0) {
+          switch (e.keyCode) {
+            case 'c': result.copy(); break;
```
ben-clayton: We have code like this somewhere else that I wrote for the shader editor - maybe we should unify this so we don't have several places to maintain?

pmuetschard: I think you are referring to the undo/redo keys you added. Also, unfortunately, the copy/paste/cut methods are declared on the Spinner class, so this code can't easily be reused :disappointed:

PR #1248, opened 2017-10-24
Protobuf: Use 'sint' for most fields.

dsrbecky:
This will avoid breaking compatibility in the future
if we change, say GLint to GLuint, somewhere.

'sint' can represent all values including unsigned.
For example, 0xFFFFFFFF can be compactly stored by
by cast to int32 (-1) which is stored in one byte.

All command arguments are in the app pool by definition.
Remove the pool ID as it is redundant and just takes space.

Renumber keys in commands.

### DIFF
--- a/gapii/cc/spy.cpp
+++ b/gapii/cc/spy.cpp
@@ -40,7 +40,7 @@
 
 // CurrentCaptureVersion is incremented on breaking changes to the capture format.
 // NB: Also update equally named field in capture.go
-static const int CurrentCaptureVersion = 0;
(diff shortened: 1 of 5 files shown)


AWoloszyn:
This looks fine to me, as long as we don't remove the pool ID in slices, since we will need them to serialize the state block.

dsrbecky:
Yeap, I am waiting for a point where I can test this.

dsrbecky:
PTAL.
Tested with the unreal app.

PR #1254, opened 2017-10-24
Use original shader if fail at assembling new shader module

### LINKED ISSUE #1247: Modifying a SPIR-V shader to something invalid panics the server
I mistakenly added OpNoop instead of OpNop and booom.

Ben says the offending line is:
https://github.com/google/gapid/blob/cce0f7e1fae67af341726ad6bb8f78f006e89fab/gapis/api/vulkan/resources.go#L724

- You cannot return nil there

Qining:
Also drop stale comments.

Fix #1247

### DIFF
--- a/gapis/api/vulkan/resources.go
+++ b/gapis/api/vulkan/resources.go
@@ -733,22 +733,22 @@ func (cmd *VkCreateShaderModule) Replace(ctx context.Context, c *capture.Capture
 	createInfo.PCode = NewU32ᶜᵖ(code.Ptr())
 	createInfo.CodeSize = memory.Size(len(codeSlice) * 4)
 	newCreateInfo := state.AllocDataOrPanic(ctx, createInfo)
-	newAtom := cb.VkCreateShaderModule(device, newCreateInfo.Ptr(), pAlloc, pShaderModule, result)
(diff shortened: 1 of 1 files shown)


gapis/api/vulkan/resources.go
```diff
@@ -733,17 +732,7 @@ func (cmd *VkCreateShaderModule) Replace(ctx context.Context, c *capture.Capture
...
-	// a 8-byte type and causing padding issues so that won't encode the struct
-	// correctly.
-	// Possible solution: define another type 'size' and handle it correctly in
-	// memory.Write().
-	buf := &bytes.Buffer{}
-	writer := endian.Writer(buf, state.MemoryLayout.GetEndian())
-	memory.Write(memory.NewEncoder(writer, state.MemoryLayout), createInfo)
-	newCreateInfo := state.AllocDataOrPanic(ctx, buf.Bytes())
+	newCreateInfo := state.AllocDataOrPanic(ctx, createInfo)
 	newAtom := cb.VkCreateShaderModule(device, newCreateInfo.Ptr(), pAlloc, pShaderModule, result)
```
ben-clayton: While you're here it might be nice to rename `newAtom` to `newCmd`.

Qining: DONE

PR #1256, opened 2017-10-24
Continue footprint building even if an error occur in another API command

### LINKED ISSUE #1255: DCE crashes for desktop vulkan app

Qining:
Fix #1255 

Turns out that the first command in desktop vulkanBloom trace is `glXMakeContextCurrent`, whose `mutate()` returns with error: `aborted(No context bound)`. Because of this error, the whole footprint was not built then gapis crashes when accessing the footprint.

@ben-clayton I'm wondering what is the most correct way to handle such cases. We definitely don't want to crash the server if we have a bad trace file. So is it right to turn off the whole DCE if an error occurs during footprint building or DCE backpropagation?

### DIFF
--- a/gapis/api/transform/dce.go
+++ b/gapis/api/transform/dce.go
@@ -68,6 +68,7 @@ func (s *commandIndicesSet) contains(fci api.SubCmdIdx) bool {
 type DCE struct {
 	footprint        *dependencygraph.Footprint
 	endBehaviorIndex uint64
+	endCmdIndex      api.CmdID
 	requests         *commandIndicesSet
 	requestCount     uint64
 }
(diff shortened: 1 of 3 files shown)


ben-clayton:
> I'm wondering what is the most correct way to handle such cases.

In GLES we just drop these erroring commands as the spec says that GL errors (with a few exceptions) should result in a no-op.

A single erroring command should not break dependency tracking - in almost all cases of bad input our systems should try to recover and continue. 

In my opinion, I'd drop the warning here - you're likely to spam the logs with a really broken trace. Bad commands should be shown in the report view. Issues with GAPID systems belong in the logs.

Qining:
So I think the right things to do here are
1) Remove the `log.E()` here and continue dependency tracking.
2) Refine the footprint to make sure even if the footprint is empty or some commands are dropped, the server does not crash.

I will work on this. Also will try to see how to move the log to report view.

ben-clayton:
SGTM

PR #1259, opened 2017-10-24
 GAPIC (native) launcher script

pmuetschard:
Adds a new binary `gapid` that will be part of the distributed package. The binary looks for the JVM (bundled or from the system), the GAPIC JAR (bundled or from the build output) and launches GAPIC with the correct JVM flags and environment variables.

The launcher adds the following flags:
```
--vm <path to JVM>: Overrides the JVM search and runs GAPIC with the given JVM.
--vmarg <arg>: Adds the given arg as an argument to the JVM. This flag can be repeated.
--console: (Windows) Causes the launcher to allocate a terminal console.
```

The launcher script also intercepts the `--help` flag and appends its help information after GAPIC's.

The main reason for this launcher is so that we can launch GAPIC on Windows without creating a terminal (the binary is marked as a GUI app and it launches javaw, unless `--console` is given).


cmd/gapid/main.go
```diff
@@ -0,0 +1,225 @@
...
+		case args[i] == "--console":
+			console = true
+			createConsole()
+		default:
+			help = help || args[i] == "--help"
+			gapicArgs = append(gapicArgs, args[i])
+		}
+	}
+
+	if help && !console {
```
ben-clayton: To be slightly safer in avoiding future changes that could cause `createConsole` to be called more than once, I'd recommend removing the call to `createConsole` above, setting `console = 
true` for `--help` and change this line to: 

```go
if console {
    createConsole()
}
```

cmd/gapid/main.go
```diff
@@ -0,0 +1,225 @@
...
+		fmt.Println("Press enter to continue")
+		os.Stdin.Read(make([]byte, 1))
+	}
+
+	if failed {
+		os.Exit(1)
+	}
+}
+
+func locateVM(cwd, vm string, console bool) (string, bool) {
```
ben-clayton: It's more idiomatic to return value, error. Return the error message and have the caller print it.
In the few cases where there are bool return values to denote success/failure, true represents success. Using `error` is unambiguous.

cmd/gapid/main.go
```diff
@@ -0,0 +1,225 @@
...
+		createConsole()
+	}
+
+	java, noVM := locateVM(cwd, vm, console)
+	gapic, noGAPIC := locateGapic(cwd)
+
+	failed := noVM || noGAPIC
+	if !failed {
+		fmt.Println("Starting", java, gapic)
+		if runtime.GOOS == "darwin" {
```
ben-clayton: FYI: go purists hate anything that goes near `GOOS`. They'll always argue for putting OS conditional logic into separate `.go` files with build tag rules. I'm less bothered by this. I do wish the OS was an enum though.

cmd/gapid/main.go
```diff
@@ -0,0 +1,225 @@
...
+	// Looks for the pattern: <product> version "<major>.<minor>.<micro><build>"
+	// Not using regular expressions to avoid binary bloat.
+	versionStr := string(version)
+	if p := strings.Index(versionStr, versionPrefix); p >= 0 {
+		p += len(versionPrefix)
+		if q := strings.Index(versionStr[p:], "."); q > 0 {
+			if r := strings.Index(versionStr[p+q+1:], "."); r > 0 {
+				major, _ := strconv.Atoi(versionStr[p : p+q])
+				minor, _ := strconv.Atoi(versionStr[p+q+1 : p+q+r+1])
+				return major > 1 || (major == 1 && minor >= 8)
```
ben-clayton: Recommend making some `const`s for these magic values at the top of the file.

cmd/gapid/main.go
```diff
@@ -0,0 +1,225 @@
...
+		buildGapic = filepath.Join(cwd, "..", "java", "gapic-osx.jar")
+	default:
+		buildGapic = filepath.Join(cwd, "..", "java", "gapic.jar")
+	}
+
+	if _, err := os.Stat(buildGapic); !os.IsNotExist(err) {
+		return buildGapic, false
+	}
+
+	fmt.Printf("GAPIC JAR \"%s\" not found.\n", gapic)
```
ben-clayton: Protip: Use backticks to avoid nasty escaping:

```go
fmt.Printf(`GAPIC JAR "%s" not found.\n`, gapic)
```

Or do what I typically do, which is use single quotes:

```go
fmt.Printf("GAPIC JAR '%s' not found.\n", gapic)
```

cmd/gapid/main.go
```diff
@@ -0,0 +1,225 @@
...
+}
+
+func locateGapic(cwd string) (string, bool) {
+	gapic := filepath.Join(cwd, "lib", "gapic.jar")
+	if _, err := os.Stat(gapic); !os.IsNotExist(err) {
+		return gapic, false
+	}
+
+	var buildGapic string
+	switch runtime.GOOS {
```
ben-clayton: Nit: maybe use the switch to select the bit that is different between all three?

```go
	jar := "gapic.jar"
	switch runtime.GOOS {
	case "linux", "windows":
 		jar = "gapic-"+runtime.GOOS+".jar"
 	case "darwin":
 		jar = "gapic-osx.jar"
 	}
 	buildGapic := filepath.Join(cwd, "..", "java", jar)
```

cmd/gapid/main.go
pmuetschard: Done.

pmuetschard:
> I'm not seeing any new changes here. Did you forget to push?

Ooops. I had pushed to the wrong branch. Fixed now.

cmake/Modules/GoCompile.cmake
```diff
@@ -27,6 +27,9 @@ if(GO_BUILD)
   endif()
 endif()
 list(APPEND args "-tags" "integration")
+if (WIN32 AND GO_WIN_UI)
+  list(APPEND args "-ldflags" "-H windowsgui")
```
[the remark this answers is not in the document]
ben-clayton: > Aren't those flags to the C compiler and linker, not the go linker? 

You're probably right.

Please add a comment so that people don't have to search in vain like I did.

ben-clayton:
Thanks for all changes!

pmuetschard:
Added a comment in the CMake file, squashed and rebased.

Issue #1261, opened 2017-10-24
Certain frames differ between FBO and Replay


baldwinn860:
Robot has found that FBOs do not match replays on either the first frame for many apks, or during a layout switch (going from portrait to landscape mode). This causes SXS Video to error since one frame has gone over the difference threshold.

This may be a minor problem but it does diminish the usefulness of Robot.

 - Some possible fixes (in preferential order):
   - fix replay so the first frame and layout swaps match observed frames properly
   - detect and ignore the first frame or layout swap frame
   - only have SXS Video error if the average frame difference is above a threshold (this means we may start to get some false positives for potentially bad errors)

ben-clayton:
There shouldn't be any difference at all. We should have identical frames, always.

PR #1266, opened 2017-10-25
gles/compat: Handle glDeleteFramebuffers() where the buffer is bound.

ben-clayton:
See the comment for details.

Fixes: #1261

### DIFF
--- a/gapis/api/gles/compat.go
+++ b/gapis/api/gles/compat.go
@@ -959,6 +959,27 @@ func compat(ctx context.Context, device *device.Instance) (transform.Transformer
 				return
 			}
 
+		case *GlDeleteFramebuffers:
+			// If you delete a framebuffer that is currently bound then the
+			// binding automatically reverts back to the default framebuffer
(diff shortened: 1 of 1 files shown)


ben-clayton:
The API correctly handles the state for this, but it doesn't generate new calls to `glBindBuffer`, which compat looks for to do sRGB trickery.

gapis/api/gles/compat.go
```diff
@@ -959,6 +959,27 @@ func compat(ctx context.Context, device *device.Instance) (transform.Transformer
 				return
 			}
 
+		case *GlDeleteFramebuffers:
+			// If you delete a framebuffer that is currently bound then the
+			// binding automatically reverts back to the default framebuffer
+			// (0). As we do compat for glBindBuffer(), scan the list of
```
dsrbecky: s/glBindBuffer/glBindFramebuffer/

ben-clayton: Done

PR #1274, opened 2017-10-26
Unpack formatting

dsrbecky:
Just a couple of things I found useful when trying to find information from traces.

### DIFF
--- a/core/data/pack/dynamic.go
+++ b/core/data/pack/dynamic.go
@@ -116,6 +116,12 @@ func (d *Dynamic) Unmarshal(data []byte) error {
 				arr = append(arr, el)
 			}
 			val = arr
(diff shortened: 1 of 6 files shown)


core/data/pack/dynamic.go
```diff
@@ -116,6 +121,15 @@ func (d *Dynamic) Unmarshal(data []byte) error {
 				arr = append(arr, el)
 			}
 			val = arr
+		} else if f.GetLabel() == descriptor.FieldDescriptorProto_LABEL_REPEATED {
+			arr := []interface{}{}
```
ben-clayton: Can be simplified down to:

```go
arr, _ := d.Fields[f.GetName()].([]interface{})
```

PR #1276, opened 2017-10-26
GAPID gfxtrace file associations

### LINKED ISSUE #565: Associate GAPIC with .gfxtrace shell extension

pmuetschard:
Associate .gfxtrace files with GAPID on all platforms:
- on Windows it is done via the registry, setup by the MSI installer
- on macOS, GAPIC now listens for the OS event to open the document, and the .app bundle registers GAPID as the owner of gfxtrace files
- on linux, .deb file now registers a mime type that is associated with GAPID in the .desktop file

Fixes #565

### DIFF
--- a/kokoro/windows/gapid.wxs
+++ b/kokoro/windows/gapid.wxs
@@ -55,9 +55,33 @@
 
     <Icon Id="gapid.ico" SourceFile="gapid.ico" />
 
+    <DirectoryRef Id="GAPID">
+      <Component Id="FileAssociation" Guid="*">
+        <RegistryValue Root="HKLM" Key="Software\Classes\GAPID.Trace" Value="Graphics API Trace" Type="string" />
+        <RegistryValue Root="HKLM" Key="Software\Classes\GAPID.Trace" Name="AppUserModelId" Value="GAPID" Type="string" />
+        <RegistryValue Root="HKLM" Key="Software\Classes\GAPID.Trace\Application" Name="AppUserModelId" Value="GAPID" Type="string" />
+        <RegistryValue Root="HKLM" Key="Software\Classes\GAPID.Trace\Application" Name="ApplicationIcon" Value="[!gapid.ico]" Type="string" />
+        <RegistryValue Root="HKLM" Key="Software\Classes\GAPID.Trace\Application" Name="ApplicationName" Value="Graphics API Debugger" Type="string" />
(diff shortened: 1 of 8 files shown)


kokoro/windows/gapid.wxs
```diff
@@ -49,15 +49,39 @@
         <Shortcut Id="ApplicationStartMenuShortcut" Name="GAPID" Description="Graphics API Debugger"
             Target="[!gapid.exe]" WorkingDirectory="APPLICATIONROOTDIRECTORY" Icon="gapid.ico" />
         <RemoveFolder Id="CleanUpShortCut" Directory="ApplicationProgramsFolder" On="uninstall" />
-        <RegistryValue Root="HKCU" Key="Software\Google\GAPID" Name="installed" Type="integer" Value="1" KeyPath="yes" />
+        <RegistryValue Root="HKCU" Key="SOFTWARE\Google\GAPID" Name="installed" Type="integer" Value="1" KeyPath="yes" />
```
ben-clayton: This is an odd change. Completely from memory, isn't "Software" more common? Again, IIRC, the registry is case insensitive, so may not matter.

kokoro/windows/gapid.wxs
```diff
@@ -49,15 +49,39 @@
...
             Target="[!gapid.exe]" WorkingDirectory="APPLICATIONROOTDIRECTORY" Icon="gapid.ico" />
         <RemoveFolder Id="CleanUpShortCut" Directory="ApplicationProgramsFolder" On="uninstall" />
-        <RegistryValue Root="HKCU" Key="Software\Google\GAPID" Name="installed" Type="integer" Value="1" KeyPath="yes" />
+        <RegistryValue Root="HKCU" Key="SOFTWARE\Google\GAPID" Name="installed" Type="integer" Value="1" KeyPath="yes" />
       </Component>
     </DirectoryRef>
 
     <Icon Id="gapid.ico" SourceFile="gapid.ico" />
 
+    <DirectoryRef Id="GAPID">
```
ben-clayton: Does this automatically delete these keys on uninstallation? I vaguely remember there needing to be some additional attribute for telling WiX that these keys are safe to delete on uninstall.

pmuetschard: Yes, I've tested it and the registry values all get cleaned up on uninstall.

kokoro/windows/gapid.wxs
pmuetschard: OK, so I changed this to match what I had below and I had it all uppercase below, because the key is actually all upper case in my registry...

However, it appears you are right, convention now is to use "Software", so I've updated it. Yes, the registry keys are case insensitive.

PR #1279, opened 2017-10-26
Revert shared map

dsrbecky:
These two commits cause trouble with tracing


ben-clayton:
Reverting for now.

I'd like to discuss the fix PR tomorrow - I have some serious concerns about using `unsafe.Pointer`.

dsrbecky:
Concerns are warranted, but after some research, I concluded is fine (e.g. GC is aware of them).

ben-clayton:
> I concluded is fine (e.g. GC is aware of them).

Can you please link to this research?

If we're not 100% sure that this won't break now or in the future, I think the only option is to either pass around pointer to maps, or wrap them in a struct.

dsrbecky:
So first of https://golang.org/pkg/unsafe/ 
"The following patterns involving Pointer are valid."
"(5) Conversion of the result of reflect.Value.Pointer or reflect.Value.UnsafeAddr from uintptr to Pointer."

https://groups.google.com/forum/#!topicsearchin/golang-nuts/subject$3Aunsafe.Pointer$20AND$20subject$3Agarbage
"If the unsafe.Pointer points to a Go object, the garbage collector will know about the object and the memory belonging to the object will not be freed."
"If the garbage collector found the unsafe.Pointer, and the object was allocated by Go, the garbage collector will also traverse all pointers stored in the object."

dsrbecky:
There is also https://groups.google.com/forum/#!searchin/golang-nuts/subject$3Agarbage$20and$20subject$3Acompact%7Csort:date/golang-nuts/KJiyv2mV2pU/PJ8q2S-LBwAJ
Which basically boils down to - go does not move objects in memory.

So for the time being I would say I am double sure:
1) GC will not move objects
2) GC is aware of unsafe.Pointer, so it should work even if GC would move objects.

dsrbecky:
I have also written a little test program to check that the returned Pointer is indeed pointer the the runtime's map structure (go has kind of interesting hashmap implementation by the way: https://golang.org/src/runtime/hashmap.go)

dsrbecky:
Having said all that, going through that hop to get the pointer does indeed pain me.
I have not managed to find a reason why go disallows comparing maps (`map1 == map2` is invalid with the only exception of comparing to `nil`).  Maybe to prevent people shooting themselves in the foot if they think the comparison would do deep equals.

ben-clayton:
I'm torn on this. I agree it's frustrating that Go doesn't provide a way to do map reference equality.
We can dodge all the unsafe nastiness by 'just' using a `*map[X]Y` for fields where we currently have `map[X]Y`. It adds another pointer indirection, but I'm cleaning towards sensible code over performance here.

dsrbecky:
I am that not worried about extra pointer indirect for performance either.
I am a bit more worried if it makes some of the hand-written code icky and large the diff would be.  It might be fine though.
I am happy to go either way.... so?

ben-clayton:
[You have a point](https://play.golang.org/p/uvsetsFl4Z). All the accesses of `A.B.C.MapField[key]` have to be wrapped with a `(*A.B.C.MapField)[key]`. Gross. :(

Humm. When @ianthehat comes online, shall we get his opinion?

dsrbecky:
Well I was thinking we would just add helper function on the pointer type so it would be just `A.B.C.MapField.Get(key)`

ben-clayton:
That could be done if we then wrap each map key-value type with a new corresponding struct type.
Could be done, but I thought that was more tedious than you'd be willing to go.

ben-clayton:
Actually - (I forgot about this) - we already have types being declared for all these map types:

```go
type Stringːboolᵐ map[string]bool

func (m Stringːboolᵐ) Get(key string) bool {
	return m[key]
}
func (m Stringːboolᵐ) Contains(key string) bool {
	_, ok := m[key]
	return ok
}
```

We could easily turn these into:

```go
type Stringːboolᵐ struct {
    m *map[string]bool
}
```

dsrbecky:
I was hopping the helpers would still work with `type Stringːboolᵐ *map[string]bool`

ben-clayton:
Sure - I just omitted them for brevity.

ben-clayton:
Sorry - I just got what you said. Not sure. Let me test it. Go is a bit funny when it comes to pointer type aliases.

ben-clayton:
So you can't put methods on a pointer type alias:

```go
type M *map[int]int
func (m M) Get(k int) int { M[k] }
```

```
invalid receiver type M (M is a pointer type)
```

You could drop the helpers, but then you're no better of than using the map pointer directly.

dsrbecky:
Should I give the "wrapped-struct" approach a go?  (or did you try?)

ben-clayton:
I think the struct approach is the safest way to go. If you don't mind trying it, that would be appreciated.

dsrbecky:
I gave it a fair try - #1286

I would say it is icky, way too icky.

PR #1280, opened 2017-10-26
Fix shared maps

dsrbecky:
Fixes problems introduced in #1238, sorry!


dsrbecky:
It offending PR was reverted for now

PR #1281, opened 2017-10-27
cmd/gapit: Give SxS video exceptions to undefined FBs.

ben-clayton:


ben-clayton:
This doesn't work. Rethinking.

PR #1286, opened 2017-10-27
Wrap maps in a struct to make them referenceable

dsrbecky:


dsrbecky:
So.... pretty massive diff.  It compiles but it does not run - there is still some nil pointer leaking through - I am guess due to structs containing maps being default-initialized by go code.

I still favor the single like unsafe.Pointer instead.  Yes, unsafe is icky, but we use it here and there, and I am not really worried about it ever breaking.

dsrbecky:
Ok, after couple of more fixes, it seems to mostly work.
I get `Value of type "U64ːContextʳᵐ" is not a map-indexable` from the client now.
I still do not like this PR much.

ben-clayton:
This isn't actually as bad as I was expecting. 

> Value of type "U64ːContextʳᵐ" is not a map-indexable

This is going to be coming from [`resolve.MapIndex`](https://github.com/google/gapid/blob/master/gapis/resolve/resolve.go#L211) and the corresponding [set code](https://github.com/google/gapid/blob/master/gapis/resolve/set.go#L253). 

Unfortunately because the map type has functions that take and return key and value types, they don't conform to a generic interface type which we can plug in. Instead we'd have to resort to reflection for these.

I'm assuming you've also resorted to `unsafe.Pointer` [in this change](https://github.com/google/gapid/pull/1290/files#diff-78134865ec99204f30e8cc229c9d39e5) because of these maps?

If I can fix the issues with resolve, can we use this PR to kill all new uses of `unsafe`?

ben-clayton:
Chatted with @ianthehat - he's also strongly advising against going near `unsafe`.
He gave compelling arguments which I won't go into here, but in a nutshell:

> persistent storage of unsafe pointers is not valid
> go is allowed to move memory, there are just no gc's that do it at the moment

He seems to agree that this is one of the most sensible approaches to avoid `unsafe`.

dsrbecky:
Persistent storage?  As in storing them to disk?

Even if Go starts to move memory in the future, it is required to handle unsafe.Pointer correctly (unlike uint - it would be bad idea to use that as key, as far as I can tell, the GC handling is the main difference between uint and unsafe.Pointer).

In terms of reliability, I feel more confident in the unsafe.Pointer. In part, just due to the size of this PR.  I hope I have not misplaced a parenthesis somewhere, and I hope I covered all the use cases (e.g. all Map{} calls need to be converted to NewMap() calls)

Having said that, if you like this PR, it is fine by me.  Could we just pretty please put it just after the other PR?

ben-clayton:
I believe everything in this PR now works correctly. Would appreciate you testing it and seeing if there's anything broken.

ben-clayton:
> Could we just pretty please put it just after the other PR?

Looking at the changes, it seems to me that you're going to have a near identical number of merge conflicts whether you put this at the end of your mega PR, or rebase that PR onto this.

I'd prefer it if we could breakup the mega PR, landing the bits that don't break the file format first - in which case, I think it would make more sense to squash and commit this first.

dsrbecky:
I am 75% leaning towards rebasing this PR on the mega #1290, and then pushing everything in one go.
I can do it in two PRs, as you suggest, if you feel strongly about it though.

PR #1287, opened 2017-10-28
WIP: Remove GetContext() method from the GLES api file

dsrbecky:


dsrbecky:
Drop, or keep?  Your call.

PR #1290, opened 2017-10-30
Several file-format breaking changes.

dsrbecky:
As requested, I set up it, so we can push everything in one go.

The first commit just fixes master.
The next two are the shared-map change with the simple unsafe.Pointer fix.
The remaining commits are various file-breaking changes.

Those are pretty much all the changes I had in mind, I don't plan to do any other.


core/data/pack/README.MD
```diff
@@ -1,63 +1,45 @@
...
+--------- | --------- | ------------
+ `size`   | `sint32`  | Total size of the chunk excluding this size field.
+ `parent` | `sint32`  | If == 0: New root object instance. <br /> If < 0: Relative index of chunk which is the parent. <br /> If > 0: Reserved (potentially stream index for muxing).
+ `type`   | `sint32`  | If > 0: Type index. The object has no children. <br /> If < 0: Negated type index if it may have children. <br /> If == 0: Terminates children list of the `parent`.
+ `data`   | `byte[]`  | Proto message.
 
-A section is either a type or a object:
+Objects may form a tree. Parent object is specified as back-reference to previous
+chunk (-1 is the previous chunk). The referenced chunk must also be an object and
+it must allow addition of children. Values of `parent>=0` define a root object.
```
ben-clayton: > Values of `parent>=0` define a root object.

In the table above parent need to be 0 to define a root object.

core/data/pack/README.MD
```diff
@@ -1,63 +1,45 @@
...
- name  | type             | description
------- | ---------------- | ------------
- size  | `varint`         | Size of the chunk data
- data  | `byte[size]`     | `T` data
+## Object instance (size>0)
 
-## Section
+ name     | type      | description
+--------- | --------- | ------------
+ `size`   | `sint32`  | Total size of the chunk excluding this size field.
```
ben-clayton: You've killed the definition of a chunk. Do you mean object?

core/data/pack/reader.go
```diff
@@ -42,13 +43,12 @@ func Read(ctx context.Context, from io.Reader, events Events, forceDynamic bool)
 		events: events,
 	}
 	r.pb = proto.NewBuffer(r.buf)
-	if err := r.readMagic(); err != nil {
+	if major, _, err := r.readHeader(); err != nil {
 		return err
+	} else if !(MinVersion <= major && major <= MaxVersion) {
+		return ErrUnsupportedVersion{Version: major}
```
ben-clayton: Why did you kill `Version` to then just ignore minor versions?

core/data/pack/reader.go
```diff
@@ -101,34 +88,47 @@ func (r *reader) unmarshal(ctx context.Context) error {
...
+	rawParent, err := r.pb.DecodeZigzag64()
+	if err != nil && err != io.ErrUnexpectedEOF {
+		return err
+	}
+	rawType, err := r.pb.DecodeZigzag64()
+	if err != nil && err != io.ErrUnexpectedEOF {
+		return err
+	}
+	// Protobuf library returns the signed zig-zag-encoded integers as uint64. Meh.
+	hasParent, parent := int(rawParent) < 0, uint64(sint.Abs(int(rawParent)))
```
ben-clayton: Why are you calling abs? The doc states that positive numbers are undefined.

core/data/pack/reader.go
```diff
@@ -101,34 +88,47 @@ func (r *reader) unmarshal(ctx context.Context) error {
...
+	hasChildren, tyIdx := int(rawType) < 0, uint64(sint.Abs(int(rawType)))
```
ben-clayton: Cast to `int64`, not `int`.

core/data/pack/reader.go
```diff
@@ -137,52 +137,45 @@ func (r *reader) unmarshal(ctx context.Context) error {
...
+		str[13] >= '0' &&
+		str[14] == '\n' &&
+		str[15] == 0 {
+		return int(str[11] - '0'), int(str[13] - '0'), nil
+	}
+	return 0, 0, ErrIncorrectMagic
 }
 
-func (r *reader) readChunk() error {
+func (r *reader) readChunk() (chunkSize int64, err error) {
```
ben-clayton: Please either describe what a chunk means in the `README.md`, or rename.

core/data/pack/types.go
```diff
@@ -48,65 +50,79 @@ type types struct {
...
 		desc, _ = protoutil.DescriptorOf(d)
 	}
-	return t.add(name, desc, create)
+	ty := t.add(name, desc)
+	if err := cb(ty); err != nil {
+		return nil, err
+	}
+
+	// Recursively add referenced types.
+	numFields := typ.NumField()
```
ben-clayton: Will panic if the type is not a struct. Why did you need to implement this?
Can we have it tested?

gapis/memory/memory_pb/memory.proto
```diff
@@ -19,32 +19,22 @@ package memory_pb;
...
-    uint64 Address = 1;
-    // The memory pool.
-    uint32 Pool = 2;
-}
-
 // Slice is the common data between all slice types.
 message Slice {
-    // Original pointer this slice derives from.
-    uint64 Root = 1;
+    // The pool identifier.
```
ben-clayton: Don't remove `Root`. As it is currently required by GAPIS, it is required to be encoded by GAPII for MEC. Without this information you cannot always reconstruct the replay memory buffers correctly.

gapis/memory/memory_pb/memory.proto
```diff
@@ -19,32 +19,22 @@ package memory_pb;
...
+    uint64 Pool = 1;
```
ben-clayton: Why is pool now a u64?

core/data/pack/README.MD
dsrbecky: Chunk is still defined in the README:
"The header is followed by arbitrary number of variable-sized chunks."
I will make it a bit more clear.

core/data/pack/README.MD
dsrbecky: The intention is that all '>=0' define root object.
Although we only ever use 0 for now, positive values are still treated a root object by decoder.
The meaning of them is unspecified as can be used for extending in the future.

core/data/pack/reader.go
dsrbecky: Major version breaks file format, so it needs to be checked.
By definition, minion version, can not break format so shouldn't be checked.
Increment minor version can still be used in the future to indicate backward compatible changes.

core/data/pack/types.go
dsrbecky: We didn't save types properly in the capture, so saving capture would result in file which can not be `gapit unpack`ed.  It could still load gapic as it would just use the run-time descriptors.

So I fixed it so we actually save descriptors properly in go (C++ already does this correctly).

gapis/capture/encoder.go
```diff
@@ -11,23 +12,31 @@ import (
...
+	resIDs map[id.ID]int64
 }
 
 func newEncoder(c *Capture, w *pack.Writer) *encoder {
 	return &encoder{
 		c:      c,
 		w:      w,
 		cmdIDs: map[api.Cmd]uint64{},
-		seen:   map[id.ID]bool{},
+		resIDs: map[id.ID]int64{id.ID{}: 0},
```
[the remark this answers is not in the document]
dsrbecky: Nothing uses the resource 0, but I just really wanted to reserve 0 to mean null for robustness.

Say, if we ever encoded extra with empty id.ID in it, it would translate to 0.

core/data/pack/README.MD
ben-clayton: Please be consistent in the definitions. If you state that > 0 is undefined, don't partially define it elsewhere.

gapis/capture/encoder.go
```diff
@@ -11,23 +12,31 @@ import (
...
 	}
 }
 
 func (e *encoder) encode(ctx context.Context) error {
+	// The encoding process modifies resource index of objects that
+	// reference resources. This means we can not encode several
+	// captures at the same time.
+	mutex.Lock()
```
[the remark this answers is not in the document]
dsrbecky: Yeah, we chatted about this.
I would say it is just a quirk of implementation.  The C++ code does not suffer from such ickyness.

My preferred solution would be for the go-stucts to use id.ID only, and the pb-struct to use the index only, and the protoconv converting methods should remap those.  The only trouble is that those methods do not really have access to the right context and maps.

I kind of have a feeling that Andrew might want to propagate some context object all the way from capture to the converters, so if that happens, it is an easy fix.  Alternatively, we could use context.Context to pass the mapping through.  Either way, I don't see it as pressing issue to solve right now.

core/data/pack/reader.go
ben-clayton: The error message displayed to the user includes the minor. You're dropping the minor version read from the file, so you're now potentially displaying an incorrect error message.
As you've just removed code that was perfectly fine, with something that could now display incorrect / confusing error messages, please put it back.

core/data/pack/types.go
ben-clayton: Okay. Please check the type is a struct (anything can implement `proto.Message`).
A simple test to cover this would also be good.

gapis/capture/encoder.go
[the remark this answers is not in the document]
ben-clayton: Okay, SGTM.

gapis/capture/encoder.go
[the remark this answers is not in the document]
ben-clayton: > Alternatively, we could use context.Context to pass the mapping through

That seems like the most sensible solution right now.

> Either way, I don't see it as pressing issue to solve right now.

As the person who has to maintain this moving forward, I'd prefer that it was cleaned up before landing.

gapis/memory/memory_pb/memory.proto
dsrbecky: I know what is the use of Root in gapis, but gapii does not use it, and it does not make any sense to me on-disk, or inside the state block.  I don't really see why it would be needed for MEC, and Andrew say he does not see any need for it in MEC either.

My thinking is that I don't want to add extra stuff in the MEC state block, if we don't use it now.  And it is easy to add the field back later if we find use for it.

gapis/memory/memory_pb/memory.proto
dsrbecky: I don't know.  Andrew changed it, I followed.  For symetry with other fields? :-)

gapis/memory/memory_pb/memory.proto
ben-clayton: If you persisted a state block that has slices to the application pool for vertex attribute pointers, you will need it. That's why it exists.

core/data/pack/reader.go
dsrbecky: The error message does not include the minor version.
Furthermore, error message displayed to the user is just limited to "too new" / "too old".

gapis/memory/memory_pb/memory.proto
dsrbecky: I think we need to persist those.

So, if we keep it in, is the value of Root going to be == Base for all slices for now?
Or do we omit the Root (i.e. 0) if it is not used to save space?

core/data/pack/reader.go
ben-clayton: Humm, perhaps I removed it due to the numbers not matching the releases.
I'd like the version mismatches to be printed to the logs at the very least, so please just keep the old version error type.

gapis/memory/memory_pb/memory.proto
ben-clayton: I'd say Root == Base. I don't think there's any point optimizing something like that for the MEC state block.

dsrbecky: Fair point. ok

gapis/capture/encoder.go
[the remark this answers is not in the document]
dsrbecky: I will try.

core/data/pack/reader.go
```diff
@@ -101,34 +88,50 @@ func (r *reader) unmarshal(ctx context.Context) error {
...
-		tyIdx, isGroup := tyIdxAndGroupFromTag(tag)
+	if tyIdx == 0 { // Null-terminator
+		if hasParent {
+			if err := r.events.EndGroup(ctx, r.id+parent); err != nil {
+				return err
+			}
+		}
+	} else { // New object instance
+		if int64(tyIdx) < 0 {
+			tyIdx = -tyIdx // Absolute value.
```
ben-clayton: TIL: I did not know you could negate an unsigned value.

dsrbecky: Another fun fact: Did you know that: -MIN_INT == MIN_INT

ben-clayton:
Looking much better, thank you for the changes so far.
As I commented on the other WIP PR, we should avoid using `unsafe`. I'm happy to work with you on fixing up `objectKey` in `custom_replay.go`.  The proto conversion stuff is also using unsafe -
 I'm guessing for the same reason? If we fix the former, can we change the latter to `interface{}` ?

dsrbecky:
I think I have addressed all the comments in this PR now.

ben-clayton:
The first two CLs in this PR don't break file formats, correct?
Can we land #1286 to master, take the first CLs from this PR and rebase them into a separate PR?

dsrbecky:
rebased and sanity checked

core/data/id/id_remap.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"context"
+)
+
+// Remapper is an interface which allows remapping between ID to int64.
+// One such remapper can be stored in the current Context.
+// This is used to handle resource when converting to/from proto.
+// It needs to live here to break go package dependency cycles.
+type Remapper interface {
+	GetID(ctx context.Context, index int64) (ID, error)
```
ben-clayton: Nit: I'd probably call these `RemapID` and `RemapIndex` given the interface name.

core/data/id/id_remap.go
```diff
@@ -0,0 +1,48 @@
...
+	GetIndex(ctx context.Context, id ID) (int64, error)
+}
+
+type remapperKeyTy string
+
+const remapperKey = remapperKeyTy("remapper")
+
+// Get returns the Remapper attached to the given context.
```
ben-clayton: `// GetRemapper...`

core/data/id/id_remap.go
```diff
@@ -0,0 +1,48 @@
...
+func GetRemapper(ctx context.Context) Remapper {
+	if val := ctx.Value(remapperKey); val != nil {
+		return val.(Remapper)
+	}
+	panic("remapper missing from context")
+}
+
+// Put amends a Context by attaching a Remapper reference to it.
```
ben-clayton: `// PutRemapper....`

core/data/protoconv/protoconv.go
```diff
@@ -37,6 +37,49 @@ var (
...
+	}
+	if id, ok := ctx.refs[obj]; ok {
+		return id, false
+	}
+	id = int64(len(ctx.refs))
+	ctx.refs[obj] = id
+	return id, true
+}
+
+// ToProtoContext stores internal state for conversion from proto
```
ben-clayton: `// FromProtoContext...`

dsrbecky:
Rebased, and added more code state serialization code from #1142 (which is related to this PR, but it does not break the file format, and it is in-face unused for now; added mostly for visibility)

gapis/capture/capture.go
```diff
@@ -247,13 +247,14 @@ func fromProto(ctx context.Context, r *Record) (out *Capture, err error) {
 	if err := pack.Read(ctx, bytes.NewReader(data.([]byte)), d, false); err != nil {
 		switch err := errors.Cause(err).(type) {
 		case pack.ErrUnsupportedVersion:
+			log.E(ctx, "%v", err)
```
ben-clayton: Nope.

ben-clayton: Sorry, wrong line.

gapii/cc/to_proto.h
```diff
@@ -23,141 +23,131 @@
...
-    }
-};
 
-template <typename EntryOut, typename KeyIn>
-struct ProtoMapKeyConverter<EntryOut, typename KeyIn::ProtoType, KeyIn> {
-    static inline void convert(EntryOut* out, const KeyIn& in) {
-        in.toProto(out->mutable_key());
+    // Remove tailing consecutive keys to save space (only applicable for integral keys).
+    template<bool IsInt, typename T>
+    static typename std::enable_if<IsInt>::type trimKeys(T keys) {
```
AWoloszyn: If you are going to go through the fun of using std::enable_if, you can simplify this as

```
template<typename T>
static typename std::enable_if<std::is_integral<T>>::type trimKeys(T keys) {
....
```

Or, if every compiler we target has C++14 support, you can use enable_if_t<>

PR #1300, opened 2017-10-31
Rework the database so we can store captures > 2GB

ben-clayton:


dsrbecky:
Looks good.

PS: Different approach would have been to encode the capture with a special mode 'omit resource data', which is fine in memory, as the hash-reference inside the blob would still be unique, and the capture could still be even loaded back using the current database.  With my resource-by-index change it would be slightly more complicated, as I would have to add index->hash to capture.Record.  The obvious benefit would be that every save of a capture wouldn't add so much memory overhead (extra copy of all resources).

PR #1302, opened 2017-10-31
gapic: Minor fixes to the AboutDialog.

ben-clayton:
Remove the "A GPU debugger for Android" string - we support desktop too.

Update help URL to point to the gh-pages docs.

Use textboxes instead of labels so that people can copy and paste things like build ID.


pmuetschard:
Submitting, as I will need to merge my changes.

PR #1305, opened 2017-10-31
Remove gles/glsl.

ben-clayton:
This has been completely superseded with glslang.

Goodnight sweet prince, you’ve served us well!


dsrbecky:
PS: I was keeping it around for two reasons:
1) Just in case glslang approach does not work as well as expected... I think that is fine now, I am confident now thta glslang is the way to go.
2) I wanted to do a diff in shader tests passed in dEQP between the old and new.  In part to measure how big improvement it was, and in part to check if there are any regressions.

ben-clayton:
2 is not going to happen any time soon, and I want to simplify the codebase.

PR #1310, opened 2017-11-01
Various api improvements

dsrbecky:
Most of those changes are needed to correctly restore
the state during mid-execution-capture.

Few of those changes are just random improvements as
mid-execution-capture will essentially set the state in stone.


gapil/resolver/inference.go
```diff
@@ -72,13 +72,10 @@ func inferNumber(rv *resolver, in *ast.Number, infer semantic.Type) semantic.Exp
...
 
 func inferUnknown(rv *resolver, lhs semantic.Node, rhs semantic.Node) {
-	// We cannot infer unknown if assigning to _.
-	// This is ok if the unknown is still inferred later.
-	if _, isIgnore := lhs.(*semantic.Ignore); isIgnore {
-		return
-	}
 	u := findUnknown(rv, rhs)
 	if u != nil && u.Inferred == nil {
+		// This might fail to infer the expression (i.e. keeps it at nil).
```
ben-clayton: What is the LHS expression type that isn't `semantic.Ignore` and cannot be inferred?
Doesn't this change mean we can finish resolving without actually inferring an unknown? That seems really bad.

gapis/api/gles/gles.api
```diff
@@ -570,7 +570,25 @@ class Context {
...
+// Internal "thread-local" global for quick access the current context.
+// It is set in the PreMutate() function before any command is mutated.
+ref!Context CurrentContext
+
+// Internal sub-routine which is called before any command is mutated.
+// TODO: The IsGLESCommand() is a hack. We should move all EGL commands to own api.
+// TODO: We should also set the Version and Extension global variables here.
+// TODO: Many commands have @if annotation. We evaluate it and check it here.
+sub void PreMutate() {
+  CurrentContext = Contexts[$Thread]
```
ben-clayton: As this method is never referenced directly by the `.api` file the static analysis will consider this function dead code. All static analysis then becomes heavily dependent on the single assignment to `CurrentContext` in `SetContext`. Without static analysis, things like enum inference will likely break.

This seems really fragile. If we are to really go this way, `PreMutate` should become a first-class language concept. My personal feeling is that we should just call the function at the top of all GL functions, and remove all the hacky GL vs EGL stuff from GAPIS and GAPII.

gapis/api/gles/gles.go
```diff
@@ -111,6 +119,46 @@ func (b *Texture) GetID() TextureId {
 	}
 }
 
+func (b *ImageUnit) GetID() ImageUnitId {
+	if b != nil {
+		return b.ID
+	} else {
```
ben-clayton: [Drop else](https://golang.org/doc/effective_go.html#if).

gapis/api/gles/gles.api
```diff
@@ -732,21 +750,27 @@ sub void ApplyDynamicContextState(ref!Context ctx, ref!DynamicContextState dynam
     stencil_id := backbuffer.StencilAttachment.Renderbuffer.ID
 
     color_buffer := ctx.Objects.Renderbuffers[color_id]
-    color_buffer.Width = dynamicState.BackbufferWidth
-    color_buffer.Height = dynamicState.BackbufferHeight
-    color_buffer.InternalFormat = dynamicState.BackbufferColorFmt
+    color_buffer.Image = new!Image(
```
ben-clayton: Do we not get a `DynamicContextState` whenever we rebind a context?

If so, aren't you nuking the image data?

gapis/api/gles/gles.api
```diff
@@ -58,11 +58,9 @@ class Versions {
...
   bool GLES32 = true
 }
-// TODO: This should be thread local variable
-Versions Version
 
-sub void InitVersions(ref!Context ctx) {
-  Version = Versions(
+sub ref!Versions InitVersions(ref!Context ctx) {
+  return new!Versions(
     GLES10:  VersionGreaterOrEqual(ctx, 1, 0),
```
ben-clayton: I'd really like it if these were called `AtLeastGLESXX`. The way I read them right now is an exact version match.

gapis/api/gles/gles.api
```diff
@@ -596,6 +603,10 @@ sub void PreMutate() {
     _ = newMsg(SEVERITY_ERROR, new!ERR_NO_CONTEXT_BOUND(thread: $Thread))
     abort
   }
+  if (CurrentContext != null) {
+    Version = CurrentContext.Other.SupportedVersions
+    Extension = CurrentContext.Other.SupportedExtensions
```
ben-clayton: Looks like you're missing assignment of these in `SetContext`

gapil/resolver/inference.go
dsrbecky: The problematic code was this:
+  handle := ?
+  ctx := CreateContext(EGLContexts[share_context])
+  EGLContexts[handle] = ctx
+  Context2EGLContext[ctx] = handle
+  return handle

We can always finish without inferring - if there is no expression involving the handle, this code path will not be executed at all.

gapis/api/gles/gles.api
dsrbecky: I will just change the name of the struct to supported versions.
NB: Mind that the primary purpose of this is to be able to add the @if(Version.GLES30) annotation.

gapis/api/gles/gles.api
dsrbecky: The primary purpose is to ensure we match behavior between gapii and gapis.
Secondary on is to allow cleaning of the language.  Calling at the start of every command is messy IMFO.
Having said that, this commit is not important, I just picked it from the other WIP PR as it seems like decent improvement.  If you do not like it, I will rather just drop it.

gapil/resolver/inference.go
ben-clayton: What about code like this:

```
sub void foo(EGLContext share_context) {
    handle := ?
    ctx := CreateContext(EGLContexts[share_context])
    EGLContexts[handle] = ctx
}
```

gapis/api/gles/gles.api
dsrbecky: 1) We never actually set the data, so it will be always empty.
2) Nuking seems like a better option then leaving in data of potentially different renderbuffer, or potentially different size.

The better solution would be to do the same thing we do for textures and set the Data as "GPU produced" at this point.

ben-clayton: > 1. We never actually set the data, so it will be always empty

This is on my TODO list, hence why I brought it up.

> Nuking seems like a better option then leaving in data of potentially different renderbuffer, or potentially different size.

Sure - but only if the framebuffer actually changes dimensions or format. Currently I don't think consider that logic.
Does DCE not use the data field as a key?

dsrbecky: Even if the dimensions and format is same, the underlying renderbuffer may be different.

ben-clayton: Fair point. Perhaps we should only be generating new `DynamicContextState`s if the `EGLDisplay` or `EGLSurface`s change?

I have no idea if they can change dimensions / formats however.

dsrbecky: Yeah, good questions.  It is getting a bit of topic with respect to this PR though - as that just wraps the Data,Width,Height,Format fields in ref!Image, and the Data is still always nil for the default framebuffer.

dsrbecky:
PTAL

gapis/api/gles/gles.api
ben-clayton: Please let's remove this CL from this PR.
I only really want to be landing changes that are bug fixes or critical at this point. This is neither.

ben-clayton:
> PTAL

There are still comments that you have not responded to.

gapil/resolver/inference.go
dsrbecky: Behaves same as 
`sub void foo(EGLContext share_context) {
    handle := ?
    foo := handle
}`
did before.

That is, we never checked if unknowns are assigned, and that continues to be the case.
I think the generated code just files to compile, so it is just unfriendly

gapis/api/gles/gles.api
dsrbecky: Already did

ben-clayton:
Thank you for addressing all the comments.
Looks good. I'd like to try and get V0.9 out before this lands.
I'll get the 0.9 package together tonight, hopefully you can submit this tomorrow.

dsrbecky:
Ok

dsrbecky:
I will hold of submitting anything until Monday.

ben-clayton:
Are you aware that this breaks the file format?
You might as well consider landing the other changes.

dsrbecky:
No, I missed that! (in fact, I still don't see it!)  Apologies.
I will retest the other PR, and land that too.

PR #1313, opened 2017-11-01
Adds crash handling and reporting to GAPIC.

baldwinn860:
This change adds an UncaughtExceptionHandler implementation that uploads a Throwable's
StackTrace to Crash2 for processing. Currently this targets the staging report area
which only accepts crashes from internal IP addresses and in effect is disabled for all
external users.

### DIFF
--- a/gapic/src/main/com/google/gapid/Main.java
+++ b/gapic/src/main/com/google/gapid/Main.java
@@ -26,6 +26,7 @@ import com.google.gapid.models.Settings;
 import com.google.gapid.server.Client;
 import com.google.gapid.server.GapiPaths;
 import com.google.gapid.server.GapisProcess;
+import com.google.gapid.util.Crash2ExceptionHandler;
 import com.google.gapid.util.Flags;
 import com.google.gapid.util.Flags.Flag;
 import com.google.gapid.util.Logging;
@@ -54,6 +55,10 @@ public class Main {
     Display.setAppName(Messages.WINDOW_TITLE);
     Display.setAppVersion(GAPID_VERSION.toString());
     Settings settings = Settings.load();
+
+    if (settings.crashReportingEnabled()) {
+      Crash2ExceptionHandler.registerAsDefault();
+    }
 
     Server server = new Server(settings);
     AtomicReference<UI> uiRef = new AtomicReference<UI>(null);
--- a/gapic/src/main/com/google/gapid/util/Crash2ExceptionHandler.java
+++ b/gapic/src/main/com/google/gapid/util/Crash2ExceptionHandler.java
@@ -1,0 +1,120 @@
+/*
+ * Copyright (C) 2017 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.gapid.util;
+
+import static com.google.gapid.util.GapidVersion.GAPID_VERSION;
+
+import com.google.common.base.Throwables;
+
+import com.squareup.okhttp.OkHttpClient;
+import com.squareup.okhttp.MultipartBuilder;
+import com.squareup.okhttp.Request;
+import com.squareup.okhttp.RequestBody;
+import com.squareup.okhttp.Response;
+
+import okio.Buffer;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.PrintStream;
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+import java.util.logging.Logger;
+import java.util.logging.Level;
+
+/**
+ * Handles uncaught exceptions and sends stacktraces to Crash2 server.
+ */
+public class Crash2ExceptionHandler implements Thread.UncaughtExceptionHandler {
+  private static final Logger LOG = Logger.getLogger(Crash2ExceptionHandler.class.getName());
+
+  // TODO(baldwinn860): Send to production url when we get approval.
+  private static final String CRASH_REPORT_URL_BASE = "https://clients2.google.com/cr/staging_report?";
+  private static final String CRASH_REPORT_PRODUCT = "GAPID";
+  private static final String CRASH_REPORT_VERSION = "Client:" + GAPID_VERSION.toString();
+  private static final String CRASH_REPORT_VERSION_ENCODED;
+  private static final String CRASH_REPORT_URL_FIELDS;
+  private static final String CRASH_REPORT_URL;
+  static { 
+    String temp;
+    try {
+      temp = URLEncoder.encode(CRASH_REPORT_VERSION, "UTF-8");
+    } catch(UnsupportedEncodingException e) {
+      temp = "UnknownVersion";
+    }
+    CRASH_REPORT_VERSION_ENCODED = temp;
+    CRASH_REPORT_URL_FIELDS = "product=" + CRASH_REPORT_PRODUCT + "&version=" + CRASH_REPORT_VERSION_ENCODED;
+    CRASH_REPORT_URL = CRASH_REPORT_URL_BASE + CRASH_REPORT_URL_FIELDS;
+  }
+
+  private final Thread.UncaughtExceptionHandler previousHandler;
+
+  private Crash2ExceptionHandler(Thread.UncaughtExceptionHandler oldHandler) {
+    previousHandler = oldHandler;
+  }
+
+  public static void registerAsDefault() {
+    if (!(Thread.getDefaultUncaughtExceptionHandler() instanceof Crash2ExceptionHandler)) {
+      Crash2ExceptionHandler handler = new Crash2ExceptionHandler(Thread.getDefaultUncaughtExceptionHandler());
+      Thread.setDefaultUncaughtExceptionHandler(handler);
+    }
+  }
+
+  @Override
+  public void uncaughtException(Thread thread, Throwable thrown) {
+    reportException(thrown);
+    // Pass the exception back to the os to get logged
+    previousHandler.uncaughtException(thread, thrown);
+  }
+
+  public static void reportException(Throwable thrown) {
+    Thread uploadThread = new Thread(new StackTraceUploader(thrown));
+    uploadThread.start();
+  }
+
+  private static class StackTraceUploader implements Runnable {
+    private final Throwable thrown;
+
+    public StackTraceUploader(Throwable thrownIn) {
+      thrown = thrownIn;
+    }
+
+    @Override
+    public void run() {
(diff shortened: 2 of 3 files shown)


gapic/src/main/com/google/gapid/util/Crash2ExceptionHandler.java
```diff
@@ -0,0 +1,65 @@
+package com.google.gapid.util;
```
ben-clayton: Missing copyright header

ben-clayton:
Crash reports need to be enabled by the user via a setting.

gapic/src/main/com/google/gapid/util/Crash2ExceptionHandler.java
baldwinn860: Ah, that would be important, fixed.

gapic/src/main/com/google/gapid/util/Crash2ExceptionHandler.java
```diff
@@ -0,0 +1,80 @@
...
+import com.squareup.okhttp.RequestBody;
+import com.squareup.okhttp.Response;
+import okio.Buffer;
+
+/**
+ * Handles uncaught exceptions and sends stacktraces to Crash2 server.
+ */
+public class Crash2ExceptionHandler implements Thread.UncaughtExceptionHandler {
+  private static final String CRASH_REPORT_PRODUCT = "GAPID";
+  private static final String CRASH_REPORT_VERSION = "Client:" + Version.GAPID_VERSION.toString();
```
pmuetschard: You need to sync, this won't build once rebased on top of master. (you need to use GapidVersion.GAPID_VERSION, which is best statically import to not have to be so verbose).

gapic/src/main/com/google/gapid/util/Crash2ExceptionHandler.java
```diff
@@ -0,0 +1,80 @@
...
+  // TODO(baldwinn860): Send to production url when we get approval.
+  private static final String CRASH_REPORT_URL_BASE = "https://clients2.google.com/cr/staging_report?";
+  private static final String CRASH_REPORT_URL_FIELDS = "product=" + CRASH_REPORT_PRODUCT + "&version=" + CRASH_REPORT_VERSION;
```
pmuetschard: The field values should be URL encoded, at least the version.

gapic/src/main/com/google/gapid/Main.java
```diff
@@ -47,6 +48,12 @@
  */
 public class Main {
   public static void main(String[] args) throws Exception {
+    // Don't set the crash handler if it is already set.
+    if (!(Thread.getDefaultUncaughtExceptionHandler() instanceof Crash2ExceptionHandler)) {
+      Crash2ExceptionHandler handler = new Crash2ExceptionHandler(Thread.getDefaultUncaughtExceptionHandler());
+      Thread.setDefaultUncaughtExceptionHandler(handler);
+    }
+
```
pmuetschard: Move this into a static function inside the C2EH class.

gapic/src/main/com/google/gapid/util/Crash2ExceptionHandler.java
```diff
@@ -0,0 +1,80 @@
...
+  private static final String CRASH_REPORT_URL = CRASH_REPORT_URL_BASE + CRASH_REPORT_URL_FIELDS;
+  private static final Logger LOG = Logger.getLogger(Crash2ExceptionHandler.class.getName());
+
+  private Thread.UncaughtExceptionHandler oldDefaultHandler_;
+  public Crash2ExceptionHandler(Thread.UncaughtExceptionHandler oldDefaultHandler) {
+    oldDefaultHandler_ = oldDefaultHandler;
+  }
+
+  @Override
+  public void uncaughtException(Thread t, Throwable th) {
```
pmuetschard: This isn't go. Please use descriptive variable names.

gapic/src/main/com/google/gapid/util/Crash2ExceptionHandler.java
```diff
@@ -0,0 +1,80 @@
...
+    ByteArrayOutputStream traceStream = new ByteArrayOutputStream();
+    th.printStackTrace(new PrintStream(traceStream));
+    String traceString = traceStream.toString();
```
pmuetschard: Use Throwables.getStackTraceAsString (you can then inline it, since it's only used in one place)

gapic/src/main/com/google/gapid/util/Crash2ExceptionHandler.java
```diff
@@ -0,0 +1,80 @@
...
+
+    try {
+      // Creates a connection to crash2
+      OkHttpClient client = new OkHttpClient();
+      RequestBody body = new MultipartBuilder()
+        .type(MultipartBuilder.FORM)
```
pmuetschard: nit: continuation indents should be four spaces.

gapic/src/main/com/google/gapid/util/Crash2ExceptionHandler.java
```diff
@@ -0,0 +1,80 @@
...
+public class Crash2ExceptionHandler implements Thread.UncaughtExceptionHandler {
+  private static final String CRASH_REPORT_PRODUCT = "GAPID";
+  private static final String CRASH_REPORT_VERSION = "Client:" + Version.GAPID_VERSION.toString();
+  // TODO(baldwinn860): Send to production url when we get approval.
+  private static final String CRASH_REPORT_URL_BASE = "https://clients2.google.com/cr/staging_report?";
+  private static final String CRASH_REPORT_URL_FIELDS = "product=" + CRASH_REPORT_PRODUCT + "&version=" + CRASH_REPORT_VERSION;
+  private static final String CRASH_REPORT_URL = CRASH_REPORT_URL_BASE + CRASH_REPORT_URL_FIELDS;
+  private static final Logger LOG = Logger.getLogger(Crash2ExceptionHandler.class.getName());
+
+  private Thread.UncaughtExceptionHandler oldDefaultHandler_;
```
pmuetschard: No underscores or "m", "my", "our", whatever, prefixes/postfixes, please.

Also this can be final.

gapic/src/main/com/google/gapid/util/Crash2ExceptionHandler.java
pmuetschard: This should be done on a separate thread. Use Scheduler.EXECUTOR

gapic/src/main/com/google/gapid/util/Crash2ExceptionHandler.java
```diff
@@ -0,0 +1,80 @@
...
+      OkHttpClient client = new OkHttpClient();
+      RequestBody body = new MultipartBuilder()
+        .type(MultipartBuilder.FORM)
+        .addFormDataPart("product", CRASH_REPORT_PRODUCT)
+        .addFormDataPart("version", CRASH_REPORT_VERSION)
+        .addFormDataPart("exception_info", traceString)
+        .build();
+      Request request = new Request.Builder()
+        .url(CRASH_REPORT_URL)
+        .post(body)
```
pmuetschard: Since you are not using the body anywhere else, you can inline the builder - c.f. proto builders.

gapic/src/main/com/google/gapid/util/Crash2ExceptionHandler.java
```diff
@@ -0,0 +1,80 @@
...
+        .build();
+      Response response = client.newCall(request).execute();
+      if (response.isSuccessful()) {
+        LOG.log(Level.INFO, "Crash Report Uploaded Successfully with " + response.code() +";\nCrash Report ID: " + response.body().string());
```
pmuetschard: No newlines in log messages, please. It messes up the formatting.

baldwinn860:
I believe I've addressed all the review comments @pmuetschard and @ben-clayton. I have the settings class disabling crash reporting until I can get the UI for enabling it together in another pull request.

gapic/src/main/com/google/gapid/util/Crash2ExceptionHandler.java
```diff
@@ -0,0 +1,106 @@
...
+      temp = "UnknownVersion";
+    }
+    CRASH_REPORT_VERSION_ENCODED = temp;
+    CRASH_REPORT_URL_FIELDS = "product=" + CRASH_REPORT_PRODUCT + "&version=" + CRASH_REPORT_VERSION_ENCODED;
+    CRASH_REPORT_URL = CRASH_REPORT_URL_BASE + CRASH_REPORT_URL_FIELDS;
+  }
+
+  private final Thread.UncaughtExceptionHandler previousHandler;
+
+  public static void Register() {
```
pmuetschard: lowercase method names

gapic/src/main/com/google/gapid/util/Crash2ExceptionHandler.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+  }
+
+  public Crash2ExceptionHandler(Thread.UncaughtExceptionHandler oldHandler) {
+    previousHandler = oldHandler;
+  }
+
+  @Override
+  public void uncaughtException(Thread thread, Throwable thrown) {
+    Scheduler.EXECUTOR.submit(new Runnable() {
```
pmuetschard: Since you don't return a result and ignore the returned future, simply use `execute`.

gapic/src/main/com/google/gapid/util/Crash2ExceptionHandler.java
```diff
@@ -0,0 +1,106 @@
...
+          if (response.isSuccessful()) {
+            LOG.log(Level.INFO, "Crash Report Uploaded Successfully; Crash Report ID: " + response.body().string());
+          } else {
+            LOG.log(Level.SEVERE, "Crash Report Not Uploaded; Response Code: " + response.code());
+          }
+        } catch (IOException e) {
+          LOG.log(Level.SEVERE, "Unable to upload exception to crash2", e);
+        } finally {
+          // Pass the exception back to the os to get logged
+          previousHandler.uncaughtException(thread, thrown);
```
pmuetschard: Since this is now happening on a different thread, it's probably better to just invoke it after scheduling the crash2 stuff.

gapic/src/main/com/google/gapid/util/Crash2ExceptionHandler.java
pmuetschard: Also, because this can happen at any time, ignore my previous suggestion to use Scheduler.EXECUTOR and just create a thread manually.

Do we need some rate control?

baldwinn860:
One problem I realized while addressing the feedback is that any call to `Scheduler.EXECUTOR`s `submit` or `schedule` will never hit this `UncaughtExceptionHandler` as `ListenableFuture`s actually wrap their `Runnable`s in a catch block to transform it into a failure return value instead. One possible fix is to add a `onFailure` callback to all futures returned from `submit` and `schedule` calls, is this an acceptable fix @pmuetschard, or do you have another idea?

pmuetschard:
No, not every exception should be reported. Only the "unhandled" ones. This was my point when we were chatting. As is, this CL doesn't actually really do anything (even if we change the settings to enable reporting), because the RPC threads, background threads and the UI thread don't leak exceptions like this (or shouldn't). We have to hook the reporting into a few places where we currently just log or display the exception to the user, but have no better way of handling it.

Since I know best where those places are, please address all the review comments and make the reporting a utility function and I will hook it into the places where we currently just "ignore" unhandled exceptions.

PR #1314, opened 2017-11-01
Fix sparse binding for Vulkan DCE

### LINKED ISSUE #1291: Vulkan DCE does not handle sparse binding

Qining:
Fix #1291


Qining:
Still WIP. So far only have the 'list' based resource binding added in Vulkan's DCE.

Qining:
Ready for review. This CL fixes the crash/replay failure when sparse bindings are used. Note that sparse residency is still not supported yet.

ben-clayton:
How critical is this fix? Should we stall 0.9 to get this in? How confident are you that it doesn't cause more issues than it fixes?

Qining:
This is not that critical, since we didn't see people actually using sparse binding yet, so generally gapid will not crash. And I'm testing it with some samples now.

Issue #1311, opened 2017-11-01
Panic in GAPIS when tracing Vulkan


tomgoogle:
GAPID Version: 0.9.0:462:128e0d8aced0ec2e82edde156e38fdbe9273f4ee
OS: mac os x x86_64

Tracing a Vulkan application (Triangle or Texture from the [Vulkan examples](https://github.com/googlesamples/android-vulkan-tutorials)) causes a panic in GAPIS immediately.

```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x52ce773]

goroutine 74 [running]:
github.com/google/gapid/core/app/crash.Crash.func1()
	/tmpfs/src/github/src/github.com/google/gapid/core/app/crash/crash.go:73 +0x115
sync.(*Once).Do(0x7563470, 0xc42079ae68)
	/tmpfs/src/go/src/sync/once.go:44 +0xbe
github.com/google/gapid/core/app/crash.Crash(0x5cc0dc0, 0x74d4a00)
	/tmpfs/src/github/src/github.com/google/gapid/core/app/crash/crash.go:74 +0x9c
github.com/google/gapid/core/app/crash.handler()
	/tmpfs/src/github/src/github.com/google/gapid/core/app/crash/crash.go:48 +0x52
panic(0x5cc0dc0, 0x74d4a00)
	/tmpfs/src/go/src/runtime/panic.go:489 +0x2cf
github.com/google/gapid/gapis/api/vulkan.(*VkGetPhysicalDeviceMemoryProperties).Mutate(0xc4203eb140, 0x7452720, 0xc420558630, 0xaf, 0xc420a025a0, 0xc4202e3c20, 0x0, 0x0)
	/tmpfs/src/github/src/github.com/google/gapid/gapis/api/vulkan/mutate.go:3729 +0x393
github.com/google/gapid/gapis/replay.(*adapter).MutateAndWrite(0xc4209fa650, 0x7452720, 0xc420558630, 0xaf, 0x748aa80, 0xc4203eb140)
	/tmpfs/src/github/src/github.com/google/gapid/gapis/replay/batch.go:191 +0xcf
[...]

PR #1320, opened 2017-11-02
Prefetch phydev properties

Qining:
Fix Issue #1311

### DIFF
--- a/gapii/cc/vulkan_extras.inl
+++ b/gapii/cc/vulkan_extras.inl
@@ -25,6 +25,13 @@ PFN_vkVoidFunction SpyOverride_vkGetInstanceProcAddr(VkInstance instance,
                                                      const char* pName);
 PFN_vkVoidFunction SpyOverride_vkGetDeviceProcAddr(VkDevice device,
                                                    const char* pName);
+void SpyOverride_prefetchPhysicalDeviceQueueFamilyProperties(
+    VkInstance instance, VkPhysicalDevice physicalDevice,
+    uint32_t* pQueueFamilyPropertyCount,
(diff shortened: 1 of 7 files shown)


Qining:
I think this might not work correctly in the case of multiple GPU yet. But it should fix #1311 for single-GPU case now. Working on a commit to make sure the physical device order is correct.

Qining:
Enumeration order is now recorded in the state block when `prefetchPhysicalDeviceProperties()` is called.

AWoloszyn:
So the actual enumeration order is not important iirc. As long as, during replay, we only call ReplayEnumeratePhysicalDevices, since nothing depends on the actual order that things were returned in. the only thing that matters is preserving the mapping between vkenumeratephysicaldevices and the device/Vendor ID

Qining:
@AWoloszyn My understanding is that we need to make sure the order of vender|deviceIDs for the `replayEnumeratePhysicalDevices` must match with the physical device enumeration order in the trace side. Is this right?

AWoloszyn:
IT doesn't have to match the device enumeration ORDER specifically, but the 
VkPhysicalDevice -> vendor/deviceID     must be correct.

Qining:
@AWoloszyn Thanks for the instruction. So currently the code is not working correctly as the order might be different between the Vender|DeviceIDs and the physical devices sent to the `replayEnumeratePhysicalDevices`. I will work on a fix for this.

Qining:
@AWoloszyn finally I understand why we just need to make sure those two arrays have to match. CL at: https://github.com/google/gapid/pull/1335

Issue #1312, opened 2017-11-01
Levels generated from glGenerateMipmap do not show in the texture view


ben-clayton:
Now that we can pull back texture data from replay, we should do this.

PR #1321, opened 2017-11-02
Fixes for glGenerateMipmap

ben-clayton:
Fixes: #1312

### DIFF
--- a/gapis/api/gles/dependency_graph_behaviour_provider.go
+++ b/gapis/api/gles/dependency_graph_behaviour_provider.go
@@ -215,6 +215,23 @@ func (*GlesDependencyGraphBehaviourProvider) GetBehaviourForAtom(
 			case *GlTexSubImage2D:
 				texData, _ := getTextureDataAndSize(ctx, cmd, id, s, c.Bound.TextureUnit, cmd.Target, cmd.Level)
 				b.Modify(g, texData)
+			case *GlGenerateMipmap:
+				tex, err := subGetBoundTextureOrErrorInvalidEnum(ctx, cmd, id, nil, s, GetState(s), cmd.Thread(), nil, cmd.Target)
+				if err != nil {
+					log.E(ctx, "Can not find bound texture %v", cmd.Target)
+				}
+				for levelIndex, level := range tex.Levels.Range() {
+					for layerIndex := range level.Layers.Range() {
+						data, size := tex.dataAndSize(levelIndex, layerIndex)
+						if levelIndex == 0 {
+							b.Read(g, data)
+							b.Read(g, size)
+						} else {
+							b.Read(g, size)
+							b.Write(g, data)
+						}
+					}
+				}
 			case *GlUniform1fv:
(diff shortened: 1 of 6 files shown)


ben-clayton:
@dsrbecky - PTAL. I'd like to get this into V0.9 so I can start chasing zooshi folks for documentation.

gapis/api/gles/dependency_graph_behaviour_provider.go
```diff
@@ -215,6 +215,22 @@ func (*GlesDependencyGraphBehaviourProvider) GetBehaviourForAtom(
...
+			case *GlGenerateMipmap:
+				tex, err := subGetBoundTextureOrErrorInvalidEnum(ctx, cmd, id, nil, s, GetState(s), cmd.Thread(), nil, cmd.Target)
+				if err != nil {
+					log.E(ctx, "Can not find bound texture %v", cmd.Target)
+				}
+				for levelIndex, level := range tex.Levels.Range() {
+					for layerIndex := range level.Layers.Range() {
+						data, size := tex.dataAndSize(levelIndex, layerIndex)
+						if levelIndex == 0 {
+							b.Read(g, data)
```
dsrbecky: Read both data and size

gapis/api/gles/dependency_graph_behaviour_provider.go
```diff
@@ -215,6 +215,22 @@ func (*GlesDependencyGraphBehaviourProvider) GetBehaviourForAtom(
...
+						} else {
+							b.Read(g, size)
+							b.Write(g, data)
```
dsrbecky: Write both data and size

gapis/api/gles/dependency_graph_behaviour_provider.go
ben-clayton: Okay

gapis/api/gles/dependency_graph_behaviour_provider.go
ben-clayton: Does it write to size?

gapis/api/gles/dependency_graph_behaviour_provider.go
ben-clayton: Done

gapis/api/gles/dependency_graph_behaviour_provider.go
ben-clayton: Writing to size breaks tests and actual texture reads.

dsrbecky: I would say it writes to size, yes.  The spec is not 100% clear though.

"Mipmap generation replaces texture image levels levelbase + 1 through q with
images derived from the levelbase image, regardless of their previous contents."

My interpretation is that the previous non-zero levels are discarded (not-read), and completely replace (defined data+size).  In particular, my reading is that it is valid for the previous definitions of the images to be of wrong size, or absent at all.  Consequently, DCE should be killing any definitions of non-zero levels.

Issue #1303, opened 2017-10-31
Unhandled exception in the UI thread. (Widget is disposed)


ben-clayton:
GAPID Version: 0.9.0:0:developer
OS: linux amd64

Clicking on link in command tree:
```
org.eclipse.swt.SWTException: Failed to execute runnable (org.eclipse.swt.SWTException: Widget is disposed)
	at org.eclipse.swt.SWT.error(Unknown Source)
	at org.eclipse.swt.SWT.error(Unknown Source)
	at org.eclipse.swt.widgets.Synchronizer.runAsyncMessages(Unknown Source)
	at org.eclipse.swt.widgets.Display.runAsyncMessages(Unknown Source)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Unknown Source)
	at org.eclipse.jface.window.Window.runEventLoop(Window.java:818)
	at org.eclipse.jface.window.Window.open(Window.java:794)
	at com.google.gapid.Main$UI.show(Main.java:109)
	at com.google.gapid.Main.main(Main.java:67)
Caused by: org.eclipse.swt.SWTException: Widget is disposed
	at org.eclipse.swt.SWT.error(Unknown Source)
	at org.eclipse.swt.SWT.error(Unknown Source)
	at org.eclipse.swt.SWT.error(Unknown Source)
	at org.eclipse.swt.widgets.Widget.error(Unknown Source)
	at org.eclipse.swt.widgets.Widget.checkWidget(Unknown Source)
	at org.eclipse.swt.widgets.Widget.getData(Unknown Source)
	at com.google.gapid.widgets.LinkifiedTree$LabelProvider.getLabelNoUpdate(LinkifiedTree.java:464)
	at com.google.gapid.widgets.LinkifiedTree$LabelProvider.update(LinkifiedTree.java:382)
	at com.google.gapid.widgets.LinkifiedTree$LabelProvider.lambda$onShow$0(LinkifiedTree.java:285)
	at com.google.gapid.models.ApiState$3.onUiThread(ApiState.java:136)
	at com.google.gapid.models.ApiState$3.onUiThread(ApiState.java:127)
	at com.google.gapid.rpc.UiCallback.lambda$onFinish$0(UiCallback.java:46)
	at com.google.gapid.widgets.Widgets.ifNotDisposed(Widgets.java:119)
	at com.google.gapid.widgets.Widgets.lambda$scheduleIfNotDisposed$1(Widgets.java:147)
[...]

PR #1328, opened 2017-11-03
gapic: Band-aids for widget dispose errors and null states.

ben-clayton:
Not sure if this is the right way to fix these issues, but there seem to be plenty of ways to cause these exceptions when going monkeying with GVR apps.


ben-clayton:
Fixes #1303

pmuetschard:
I prefer my #1334 over this to fix the disposed widget issue, because it limits the disposal check to the RPC->UI boundary.

The null state, I believe is fixed by #1331.

PR #1337, opened 2017-11-06
Enable anonymous user statistics (opt-in)

ben-clayton:


ben-clayton:
Will hold off submitting this until we get sign-off.
Created the PR so you can easily make any additional changes as needed.

ben-clayton:
I've added a new CL that hooks up the crash reporting settings.

gapic/src/main/com/google/gapid/server/GapisProcess.java
```diff
@@ -89,6 +89,11 @@ protected Exception prepare(ProcessBuilder pb) {
 
     String gapirFlags = "";
 
+    if (settings.reportCrashes) {
+      args.add("-crashreport");
+      args.add(settings.analyticsClientId);
```
pmuetschard: -crashreport is a boolean flag, is it not?

ben-clayton: Yes, good spot

PR #1341, opened 2017-11-06
Breakpad exception handler integration

baldwinn860:
This change adds the breakpad crash handling library and integrates it's exception handling and minidump generation functionality into core. Currently only GAPIR creates a crash handler object and only if the `-enable-crash-reporting` flag is passed to gapir. A small unit test is added to check that crashes are handled properly.

### DIFF
--- a/core/cc/android/crash_handler.cpp
+++ b/core/cc/android/crash_handler.cpp
@@ -20,8 +20,7 @@
 
 namespace {
 
-static
-bool handleCrash(const google_breakpad::MinidumpDescriptor& descriptor, void* crashHandlerPtr, bool succeeded) {
+static bool handleCrash(const google_breakpad::MinidumpDescriptor& descriptor, void* crashHandlerPtr, bool succeeded) {
     core::CrashHandler* crashHandler = reinterpret_cast<core::CrashHandler*>(crashHandlerPtr);
     std::string minidumpPath(descriptor.path());
     return crashHandler->handleMinidump(minidumpPath, succeeded);
@@ -39,6 +38,7 @@ CrashHandler::CrashHandler(HandlerFunction handlerFunction) :
 }
 
 // this prevents unique_ptr<CrashHandler> from causing an incomplete type error from inlining the destructor.
+// The incomplete type is the previously forward declared google_breakpad::ExceptionHandler.
 CrashHandler::~CrashHandler() = default;
 
 }
--- a/core/cc/android/crash_handler.cpp
+++ b/core/cc/android/crash_handler.cpp
@@ -1,0 +1,44 @@
+/*
+ * Copyright (C) 2017 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "../crash_handler.h"
+
+#include "client/linux/handler/exception_handler.h"
+
+namespace {
+
+static
+bool handleCrash(const google_breakpad::MinidumpDescriptor& descriptor, void* crashHandlerPtr, bool succeeded) {
+    core::CrashHandler* crashHandler = reinterpret_cast<core::CrashHandler*>(crashHandlerPtr);
+    std::string minidumpPath(descriptor.path());
+    return crashHandler->handleMinidump(minidumpPath, succeeded);
+}
+
+}
+
+namespace core {
+
+CrashHandler::CrashHandler(HandlerFunction handlerFunction) :
+    mHandlerFunction(handlerFunction),
+    mHandler(new google_breakpad::ExceptionHandler(
+            google_breakpad::MinidumpDescriptor("./crashes"),
+            NULL, ::handleCrash, reinterpret_cast<void*>(this), true, -1)) {
+}
+
+// this prevents unique_ptr<CrashHandler> from causing an incomplete type error from inlining the destructor.
+CrashHandler::~CrashHandler() = default;
+
+}
--- a/core/cc/crash_handler.h
+++ b/core/cc/crash_handler.h
@@ -1,0 +1,49 @@
+/*
+ * Copyright (C) 2017 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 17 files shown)


baldwinn860:
The separate commits are a bit of a mess, I'm wondering if it is worth it to go back and revise history to clean them up, or squash the entire PR. Thoughts?

baldwinn860:
Fixed the build on OS X and squashed the change.

core/cc/android/crash_handler.cpp
```diff
@@ -0,0 +1,44 @@
...
+ * limitations under the License.
+ */
+
+#include "../crash_handler.h"
+
+#include "client/linux/handler/exception_handler.h"
+
+namespace {
+
+static
```
ben-clayton: No need for newline here.

core/cc/android/crash_handler.cpp
```diff
@@ -0,0 +1,44 @@
...
+namespace core {
+
+CrashHandler::CrashHandler(HandlerFunction handlerFunction) :
+    mHandlerFunction(handlerFunction),
+    mHandler(new google_breakpad::ExceptionHandler(
+            google_breakpad::MinidumpDescriptor("./crashes"),
+            NULL, ::handleCrash, reinterpret_cast<void*>(this), true, -1)) {
+}
+
+// this prevents unique_ptr<CrashHandler> from causing an incomplete type error from inlining the destructor.
```
ben-clayton: eh? This is new to me. Can't you just put `~CrashHandler() = default;` in the header?

core/cc/crash_handler.h
```diff
@@ -0,0 +1,49 @@
...
+
+#ifndef CORE_CRASH_HANDLER_H
+#define CORE_CRASH_HANDLER_H
+
+#include <functional>
+#include <memory>
+#include <string>
+
+namespace google_breakpad {
+    class ExceptionHandler;
```
ben-clayton: No indents for namespace blocks.

core/cc/crash_handler.h
```diff
@@ -0,0 +1,49 @@
...
+
+// Utility class for attaching a crash handler.
+class CrashHandler {
+public:
+    typedef std::function<bool(const std::string& minidumpPath, bool succeeded)> HandlerFunction;
+
+    CrashHandler(HandlerFunction handlerFunction);
+    ~CrashHandler();
+
+    bool handleMinidump(const std::string& minidumpPath, bool succeeded) {
```
ben-clayton: Does this need to be public?

core/cc/android/crash_handler.cpp
```diff
@@ -0,0 +1,44 @@
+/*
```
ben-clayton: If 99% of this file is identical between Android, macOS and Linux, I'd suggest putting the bulk in the posix directory, and use `static const char*` for for the file location (which seems to be the only thing I can see that's different). You can then define that `static const char*` in these OS specific directories.

third_party/breakpad.cmake
```diff
@@ -0,0 +1,195 @@
...
+
+set(breakpad_dir "third_party/breakpad/src")
+
+if(APPLE)
+    set(breakpad_srcs
+        "common/convert_UTF.c"
+        "common/md5.cc"
+        "common/simple_string_dictionary.cc"
+        "common/string_conversion.cc"
+        "client/minidump_file_writer.cc"
```
ben-clayton: Try to keep these alphabetically sorted.

third_party/breakpad.cmake
```diff
@@ -0,0 +1,195 @@
...
+        find_library(FOUNDATION_LIBRARY Foundation)
+        find_library(SYSTEM_CONFIGURATION_LIBRARY SystemConfiguration)
+        find_library(APPKIT_LIBRARY AppKit)
+        find_library(COCOA_LIBRARY Cocoa)
+
+        set(inspector_srcs
+            "common/convert_UTF.c"
+            "common/md5.cc"
+            "common/string_conversion.cc"
+            "client/minidump_file_writer.cc"
```
ben-clayton: Ditto

core/cc/android/crash_handler.cpp
baldwinn860: No you can't, the unique_ptr needs to know the size of the object in it's destructor, which it can't know until the cpp file due to the forward declaration. Problem is if CrashHandler's default destructor is not manually set in a separate compilation unit it will be inlined by the compiler without knowledge of the size of google_breakpad::ExceptionHandler.
[This](https://stackoverflow.com/questions/13414652/forward-declaration-with-unique-ptr) and [this](https://stackoverflow.com/questions/6012157/is-stdunique-ptrt-required-to-know-the-full-definition-of-t) have more info on the issue.

ben-clayton: Fair enough - you didn't mention forward declaration in your comment.

core/cc/crash_handler.h
baldwinn860: The handleCrash callback needs a method of calling the class's mHandlerFunction, if there is another way to let that happen I would much prefer it.

core/cc/android/crash_handler.cpp
baldwinn860: Android and Linux are indeed nearly identical. MacOS though has different function signatures for the ExceptionHandler constructor and the handleCrash static callback, it takes a MinidumpDescriptor instead of just a plain path.

third_party/breakpad.cmake
baldwinn860: done.

third_party/breakpad.cmake
baldwinn860: done.

core/cc/crash_handler.h
baldwinn860: done.

core/cc/android/crash_handler.cpp
baldwinn860: ah, I'll add that then.

core/cc/crash_handler.h
ben-clayton: Sorry, I thought that was referenced by the constructor. Never mind.

third_party/breakpad.cmake
```diff
@@ -0,0 +1,192 @@
...
+        add_executable(inspector ${inspector_srcs})
+        target_include_directories(inspector PUBLIC
+            ${breakpad_dir}
+            "${breakpad_dir}/common/mac"
+            "${breakpad_dir}/client/apple/Framework"
+        )
+        target_compile_options(inspector PRIVATE "-Wno-deprecated")
+        target_link_libraries(inspector ${CARBON_LIBRARY} ${FOUNDATION_LIBRARY})
+
+        # crash_report_sender_resources
```
ben-clayton: What are these? Are they TODO?

core/cc/android/crash_handler.cpp
baldwinn860: done.

core/cc/android/crash_handler.cpp
baldwinn860: fixed.

third_party/breakpad.cmake
baldwinn860: actually they were unnecessary, I've removed the crash_report_sender and inspector targets.

Issue #407, opened 2017-05-19
Crash reporting


ben-clayton:
Progress:
- [x] GAPIC
  - [x] Crash Handling
  - [x] Reporting
- [ ] GAPIR
  - [ ] Symbols
    - [ ] Windows
    - [ ] Mac OS X
    - [ ] Android
    - [ ] Linux
  - [x] Crash Handling
    - [x] Windows
    - [x] Mac OS X
    - [x] Android
    - [x] Linux
  - [x] Reporting
- [x] GAPIS
  - [x] Crash Handling
  - [x] Reporting
- [X] GAPII (on hold)

PR #1342, opened 2017-11-06
Hook the exception handler into the models and UI thread.

pmuetschard:
Once the UI is up and running, the UI thread captures exceptions and doesn't just let them propagate, as that would cause the app to close. This will cause the UI thread to report these caught exceptions and
similarly caught exceptions in the data models.

For #407


gapic/src/main/com/google/gapid/models/Capture.java
```diff
@@ -105,10 +106,13 @@ public void loadCapture(File file) {
         return success(capturePath);
       }
     } catch (UnsupportedVersionException e) {
+      handler.reportException(e);
```
ben-clayton: humm, we might need to decide which side of the RPC fence this should come from.
I think this might already be logged from GAPIS.
This particular example isn't clear either - is this a crash or a analytics event?

pmuetschard: Good point. Removed.

Issue #1149, opened 2017-10-03
Name variables in GL errors


dsrbecky:
Follow up for #883 

We check arguments using helper functions like this: CheckGE!GLuint(end, start)

This will generate error message which includes the two relevant values, so the user should be able to tell which of the function arguments is incorrect based on that.  It would be more user friendly to also include the argument as string.

If this was C++, we could just use macro CHECK_GE(end, start) which stringifies the argument and includes it in the error message.  Ideally the language should support something similar.

dsrbecky:
As easier first step, it would be nice if we could print the type of the value.
At the moment, we explicitly cast down to u64 as the last step before generating the message.
If we just remove the cast and allow 'interface{}' typed value for the messages:
```glErrorInvalidValueMsg(new!ERR_INVALID_OBJECT_NAME(value: value))```
This would yield slightly nicer error message, such as:
"Invalid value TextureId(42). Object with this name does not exist."

dsrbecky:
Ideally something like ```{{value}}``` would mean untyped message arg.
I am not sure how this interacts with the C++ side of things.
So maybe ```{{value:id}}``` where ```id``` translates to ```uint64_t``` in C++ and ```interface{}``` in Go.

ben-clayton:
> glErrorInvalidValueMsg(new!ERR_INVALID_OBJECT_NAME(value: value))
This would yield slightly nicer error message, such as:
"Invalid value TextureId(42). Object with this name does not exist."

This wouldn't actually work. The message will get serialized to proto, where the `TextureId` (which is an `int` aliased type) will get boxed as an `int`.

> Ideally something like {{value}} would mean untyped message arg.

IIRC it does.

dsrbecky:
Why not serialize the type-name in proto as well then?

I think ```{{value}}}``` is string, and there is no way to do untyped.  Having said that, we could do the conversion to string automatically at message creation side i.e. automatically format and store string "TextureId(42)"

PR #1350, opened 2017-11-08
Add some assertions to Vulkan API

Qining:


gapis/api/vulkan/externs.go
```diff
@@ -1115,3 +1115,11 @@ func bindSparse(ctx context.Context, s *api.GlobalState, binds *QueuedSparseBind
 		}
 	}
 }
+
+func (e externs) notNullOrZero(v interface{}) bool {
```
ben-clayton: Please don't add externs unless they are absolutely necessary.
You should be able to just compare against `null`.

gapis/api/vulkan/vulkan.api
```diff
@@ -2623,6 +2634,8 @@ cmd VkResult vkGetPhysicalDeviceImageFormatProperties(
     VkImageUsageFlags        usage,
     VkImageCreateFlags       flags,
     VkImageFormatProperties* pImageFormatProperties) {
+  assert(notNullOrZero(physicalDevice))
```
ben-clayton: These aren't a good cases for `assert`. `assert` should be used for internal sanity checking. You're trying to use `assert` where proper error checking should be used. Check the parameters and raise a human readable error message.

In GLES we have [errors.go](https://github.com/google/gapid/blob/master/gapis/api/gles/api/errors.api) for raising and reporting human readable errors. This still needs a lot of work:

a) The error messages should include the parameter name. Tracked by #1149.
b) We should make the externs being used part of the API language.
c) The way the stringtable entries are linked in is horrible. Again, we should make these 1st class language features.

That said, if you keep the implementation as close to GLES as possible, I can clean up both together.

Qining:
Replaced by #1355

PR #1354, opened 2017-11-08
Add "create a video" tutorial.

baldwinn860:


tut-create-video.md
```diff
@@ -0,0 +1,35 @@
...
+
+GAPID has the ability to export a captured trace as a video.
+
+This feature is only available via the command line tool `gapit`.
+
+### Prerequisites
+* A previously captured trace file ending in *.gfxtrace
+* Either avconv or ffmpeg encoding libraries on your path (for regular video encoding)
+
+<div class="callouts" markdown="block">
```
ben-clayton: No need for callouts here - they're for pointing out things in an image.

tut-create-video.md
```diff
@@ -0,0 +1,35 @@
...
+
+1. Launch the command line interface.
+
+2. Note the path in which gapid was installed.
+<span class="info">e.g. `C:\Program Files (x86)\gapid\` on Windows; `/Applications/GAPID.app/Contents/MacOS/` on Mac OS X; or `/opt/gapid/` on linux.</span>
+
+3. Run the following command
+```
+$ <gapid_install_path>gapit video -type regular -out <output_path>.mp4 <path_to_tracefile>
```
ben-clayton: No need for `-type regular`, this is implicit, right?

baldwinn860: `-type auto` is the default, if the user somehow created their trace with FrameBufferObservations turned on then it would actually output a side by side video instead.

tut-create-video.md
baldwinn860: ah, fixed.

tut-create-video.md
ben-clayton: Sure, but nobody would do that, right?

ben-clayton: (I'm only nitting on this because fewer options are less intimidating)

baldwinn860: Just wanted to make sure there were no surprises. I've removed it though.

PR #1355, opened 2017-11-09
Report Vulkan errors to the report view

Qining:
Use `vkGetImageMemoryRequirements` as an example.

### DIFF
--- a/gapis/api/vulkan/errors.api
+++ b/gapis/api/vulkan/errors.api
@@ -36,6 +36,8 @@
 // TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
 // MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
 
+// TODO: Change to take message type once the error messages below are moved to
+// en-us.stb.md.
 extern void onVkError(any err)
 
 sub void vkErrorInvalidInstance(VkInstance inst) {
@@ -69,6 +71,9 @@ sub void vkErrorUnsupportedExtension(const char* name) {
   // Continue the mutation as this may not cause problem.
 }
 
+
+// TODO: Add the following classes to messages in messages/en-us.stb.md once
+// the compilation conflict of importing messages.api is resolved.
 class ERR_INVALID_HANDLE {
   string handleType
   u64    handle
--- a/gapis/api/vulkan/errors.api
+++ b/gapis/api/vulkan/errors.api
@@ -66,8 +66,8 @@ sub void vkErrorNullPointer(string pointerType) {
   abort
 }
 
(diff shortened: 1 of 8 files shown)


gapis/api/vulkan/externs.go
```diff
@@ -1115,3 +1117,24 @@ func bindSparse(ctx context.Context, s *api.GlobalState, binds *QueuedSparseBind
...
+	var issue replay.Issue
+	issue.Command = e.cmdID
+	issue.Severity = service.Severity_ErrorLevel
+	switch err := err.(type) {
+	case *ERR_INVALID_HANDLE:
+		issue.Error = fmt.Errorf("Invalid %s: %v", err.HandleType, err.Handle)
+	case *ERR_NULL_POINTER:
+		issue.Error = fmt.Errorf("Null pointer of %s", err.PointerType)
+	case *ERR_UNSUPPORTED_EXTENSION:
+		issue.Error = fmt.Errorf("Unsupported extension: %s", err.Name)
```
AWoloszyn: I would probably make this a warning.
To me ERR_UNSUPPORTED_EXTENSION would be if you use an extension that the driver does not provide.

In THIS case we are talking about an extension that GAPID does not understand. Which is not an issue with the application itself.

gapis/api/vulkan/errors.api
```diff
@@ -0,0 +1,83 @@
...
+  onVkError(new!ERR_NULL_POINTER(pointerType: pointerType))
+  abort
+}
+
+sub void vkErrorUnsupportedExtension(const char* name) {
+  onVkError(new!ERR_UNSUPPORTED_EXTENSION(name: as!string(name)))
+  // Continue the mutation as this may not cause problem.
+}
+
+class ERR_INVALID_HANDLE {
```
ben-clayton: These should be generated by the stringtable system. Declare them in `en-us.stb.md` and `import "../../messages/messages.api"`

Qining: Importing the `../../messages/message.api` causes an error on the CPP side:
```
.../gapid/gapii/cc/vulkan_types.h:1983:27: error: redefinition of 'SEVERITY_DEBUG'
```

`Severity` is defined in both `gles_types.h` and `vulkan_types.h`.

Qining: Is there a way to add a namespace for the types defined in `gles_types.h` and `vulkan_types.h`. That will solve the problem.

Qining:
@ben-clayton Reverted back to defining Vulkan's error message in vulkan/errors.api, and using extern function to receive the messages.
@AWoloszyn Changed from `UnsupportedExtention` to `UnrecognizedExtension`

If this gets in, #1371 should be discarded.

PR #1358, opened 2017-11-09
Go crash reporting.

ben-clayton:


ben-clayton:
Note - I haven't seen a test report appear, so its entirely possible I've screwed up the multipart message somehow. Please let me know if the message looks screwy to you.

core/app/crash/reporting/encoder.go
```diff
@@ -0,0 +1,62 @@
...
+
+type encoder struct {
+	appName    string
+	appVersion string
+	osName     string
+	osVersion  string
+	stacktrace string
+}
+
+const multipartBoundary = "::multipart-boundary::"
```
baldwinn860: Is there a reason we set this manually? Typically a UUID is desired I think.

baldwinn860:
I see the test report appear on the staging server now.

core/app/crash/reporting/encoder.go
ben-clayton: I made it fixed for testing. I figured the chances of `"::multipart-boundary::"` being used in the stack trace or other data was slim to none, so altering the boundary string on tests was fairly pointless.

PR #1346, opened 2017-11-07
gapic: Add various caches to reduce loading times of images.

ben-clayton:
Cache the unboxing of byte arrays so that we don’t end up with a different array for each call to Pods.getBytes().

Add comparable logic to ArrayImage.

Fixes: #655

PR #1361, opened 2017-11-09
Performance improvements for image views.

### LINKED ISSUE #655: Framebuffer takes a lot of time even if it is already cached
I tried to find visual difference between two consecutive commands, so I keep switching between them.
The framebuffer takes non-negligible time to load anyway, and the loading icon eventually show up, which breaks my "human eye diffing algorithm" :-/

pmuetschard:
Fixes #655

### DIFF
--- a/gapic/src/main/com/google/gapid/image/ArrayImage.java
+++ b/gapic/src/main/com/google/gapid/image/ArrayImage.java
@@ -20,13 +20,11 @@ import static com.google.gapid.util.Colors.DARK_LUMINANCE_THRESHOLD;
 import static com.google.gapid.util.Colors.clamp;
 
 import com.google.common.primitives.UnsignedBytes;
-import com.google.common.collect.Sets;
 import com.google.gapid.glviewer.gl.Texture;
-import com.google.gapid.proto.stream.Stream.Channel;
+import com.google.gapid.image.Histogram.Binner;
+import com.google.gapid.proto.stream.Stream;
 import com.google.gapid.util.Colors;
 
-import java.nio.DoubleBuffer;
-import java.util.Set;
 import org.eclipse.swt.graphics.ImageData;
 import org.lwjgl.BufferUtils;
 import org.lwjgl.opengl.GL11;
@@ -36,6 +34,7 @@ import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.nio.FloatBuffer;
 import java.util.Arrays;
+import java.util.Set;
 
 /**
  * An {@link Image} backed by a byte array.
@@ -107,10 +106,10 @@ public abstract class ArrayImage implements Image {
     if (x < 0 || y < 0 || z < 0 || x >= width || y >= height || z > depth) {
       return PixelValue.NULL_PIXEL;
     }
-    return getPixel(x, y, data);
+    return getPixel(x, y);
   }
 
-  protected abstract PixelValue getPixel(int x, int y, byte[] src);
+  protected abstract PixelValue getPixel(int x, int y);
 
   protected static ByteBuffer buffer(byte[] data) {
     return ByteBuffer.wrap(data).order(ByteOrder.LITTLE_ENDIAN);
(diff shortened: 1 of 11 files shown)


gapic/src/main/com/google/gapid/image/ArrayImage.java
```diff
@@ -302,47 +293,38 @@ protected void convert2D(byte[] src, byte[] dst, byte[] alpha, int stride) {
...
-      out.rewind();
-      return out;
+    public boolean isHDR() {
+      return true;
     }
 
     @Override
-    public boolean isHDR() {
-      return true;
+    public void bin(Histogram.Binner binner) {
```
ben-clayton: I'm not sold on this - replacing a very useful 'give me the value of this pixel in a standardized format' with something so heavily tied to the histogram seems like an overall loss. Given that you're calling back to the binner for each pixel component, I can't imagine it's all that much faster.

gapic/src/main/com/google/gapid/image/Histogram.java
```diff
@@ -34,255 +37,271 @@
...
-    for (DoubleBuffer values : allValues.values()) {
-      for (double value : values.array()) {
-        if (!Double.isNaN(value) && !Double.isInfinite(value)) {
-          min = Math.min(min, value);
-          max = Math.max(max, value);
-          average += value;
-          numValues++;
-        }
-      }
+    public double map(double value) {
```
ben-clayton: Why do we have this method declared twice?

gapic/src/main/com/google/gapid/image/Histogram.java
```diff
@@ -34,255 +37,271 @@
...
+    return mapper.range(count);
+  }
 
-    /**
-     * @return the value limited to the min and max values of this range.
-     */
-    public double clamp(double value) {
-      return Math.max(Math.min(value, max), min);
-    }
+  public static class Binner {
```
ben-clayton: Please document all the new code.

gapic/src/main/com/google/gapid/image/ArrayImage.java
```diff
@@ -174,8 +179,10 @@ public Builder flip() {
   public static class RGBA8Image extends ArrayImage {
     private final PixelInfo info;
 
-    public RGBA8Image(int width, int height, int depth, byte[] data) {
-      this(width, height, depth, data, IntPixelInfo.compute(data, true));
+    public RGBA8Image(
+        com.google.gapid.proto.image.Image.ID id, int width, int height, int depth, byte[] data) {
```
ben-clayton: I actually attempted adding an ID into all these images and decided against it once I started encountering multiple images (your `getCombinedId`). Given that you're then making up identifiers, things are just gross.
What are your thoughts on the standard `equals` & `hashCode` approach in my PR?
It certainly seems like far less code...

gapic/src/main/com/google/gapid/widgets/ImagePanel.java
```diff
@@ -1056,9 +1057,11 @@ public void init(Renderer renderer) {
     public void update(Renderer renderer, SceneData newData) {
       // Release textures that are no longer in data.
       Set<Image> newSet = Sets.newHashSet(newData.images);
-      for (Map.Entry<Image, Texture> entry : imageToTexture.entrySet()) {
+      for (Iterator<Map.Entry<Image, Texture>> it = imageToTexture.entrySet().iterator(); it.hasNext(); ) {
```
ben-clayton: While this is marginally faster, I found the implementation in #1346 significantly easier to read.

gapic/src/main/com/google/gapid/image/Histogram.java
pmuetschard: Done.

gapic/src/main/com/google/gapid/image/ArrayImage.java
pmuetschard: Done. I think. It's not really less code, but it certainly is cleaner.

gapic/src/main/com/google/gapid/image/ArrayImage.java
ben-clayton: Fair enough. I can't argue with your profiling results. I guess if we need to sample a pixel, we can always add that back in again.

ben-clayton:
Thank you for the changes!

Issue #1363, opened 2017-11-10
If we get VK_ERROR_DEVICE_LOST, we should not continue to spin on our GetFenceStatus.


AWoloszyn:
We spin on VkGetFenceStatus so that we match the trace perfectly. However, we should instead drop out of our loop if we get a VK_ERROR_DEVICE_LOST.

Otherwise we will loop forever, since there is no way to recover from DEVICE_LOST.

AWoloszyn:
Fixed by #1364

PR #1364, opened 2017-11-10
Stop looping on GetFenceStatus if the return is VK_ERROR_DEVICE_LOST

AWoloszyn:
If we get into this case, we will end up looping forever,
which is not actually useful.


AWoloszyn:
Fixes #1363

Qining:
Is there a way to log error to the server? I guess currently we can only dump to the adb logcat.

AWoloszyn:
We can log from GAPIR->GAPIS, so that might be a good choice, but what will end up happening here ANYWAY, is that the replay will fail. since no further commands will succeed,

PR #1367, opened 2017-11-10
Fix a but in the Vulkan DCE

Qining:


gapis/api/vulkan/footprint_builder.go
```diff
@@ -863,7 +863,10 @@ func (l resBindingList) getSubBindingList(ctx context.Context,
 			if bh != nil {
 				read(ctx, bh, bl[i])
 			}
-			newB := bl[i].newSubBinding(ctx, bh, start, end-start)
+			newB, err := bl[i].newSubBinding(ctx, bh, start-bl[i].span().Start, end-start)
```
Qining: The only real change

Issue #1348, opened 2017-11-07
Add better error reporting if an unsupported Vulkan extension is used.


AWoloszyn:
If an unknown extension is used, it is likely that the application will fail by trying to call a nullptr function pointer, it would be better to at least log an error when we detect an unknown extension in the enabled extension list.

AWoloszyn:
Done

PR #1371, opened 2017-11-10
Log error when unrecognized extensions are found in Vulkan app

Qining:
Mitigate the issue #1348


Qining:
Closed as #1355 gets in.

PR #1386, opened 2017-11-14
Handle SparseResidency sparse binding images

Qining:
We actually can NOT interpret an image sparse binding area (specified with subresource, miplevel, arrayLayer, offset3D and extent3D) into a number of continuous linear ranges.
The reason is that:
- To calculate the address of each texel, we need to call `vkGetImageSubresourceLayout()`, which is only valid if the given image is created with `linear tiling`. However, SparseResidency does not work with `linear tiling` images
> A sparse image created using VK_IMAGE_CREATE_SPARSE_RESIDENCY_BIT supports all non-compressed color formats with power-of-two element size that non-sparse usage supports. Additional formats may also be supported and can be queried via vkGetPhysicalDeviceSparseImageFormatProperties. VK_IMAGE_TILING_LINEAR tiling is not supported.

This CL records all the sparse image bindings in the time order. If a later binding is found to fully overwrite a previous binding, then the previous one will be removed.
For MEC, the `memory` field of each `bind` will be checked with the state block to guarantee the memory exist.
However, this may still have some problems. For example, a not-fully-overwritten previous bound may have its overwritten part bound to something else at the time of resource enumeration, then it can be invalid to recover the whole bound in MEC, as the memory may have already been bound to other resources prior to the `RecreateBindImageSparseMemoryBinding` call.

### DIFF
--- a/gapii/cc/vulkan_mid_execution.cpp
+++ b/gapii/cc/vulkan_mid_execution.cpp
@@ -347,7 +347,9 @@ bool IsFullyBound(VkDeviceSize offset, VkDeviceSize size,
   return true;
 }
 
-// Handy toolbox
+// A helper class to copy the data from a given buffer range or subresources of
+// an image. It handles the creation of extra Vulkan handles to do the job and
+// cleaning of those helper handles.
 class CopyDataHelper {
   public:
    CopyDataHelper(VulkanSpy* spy,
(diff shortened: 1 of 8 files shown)


gapii/cc/vulkan_mid_execution.cpp
```diff
@@ -1152,6 +1153,26 @@ void VulkanSpy::EnumerateVulkanResources(CallObserver* observer) {
...
+        imageSparseBinds.reserve(count);
+        order_keys.reserve(count);
+        for (const auto& b : image.second->mSparseImageMemoryBindings) {
+          order_keys.push_back(b.first);
+        }
+        std::sort(order_keys.begin(), order_keys.end());
+        for (auto k : order_keys) {
+          imageSparseBinds.push_back(image.second->mSparseImageMemoryBindings[k]);
+        }
+        RecreateBindImageSparseMemoryBindings(observer, image.second->mDevice,
```
AWoloszyn: I don't think this will do what you want with sparse residency. 
We try to read the entire image memory, then upload the entire image. Since we cannot copy the entire image, this will lead to undefined behavior.

Qining: @AWoloszyn Just updated with commit: 45f9a91. For partially bound buffers/images, we only recover the bound areas.

PR #1392, opened 2017-11-17
GAPIR crash reporting

baldwinn860:
Now crash dumps in GAPIR will get uploaded to GAPIS via the server connection such that GAPIS can do the final upload to the crash server. Fixes the GAPIR crash reporting sub-task of #407.

### DIFF
--- a/core/app/crash/reporting/encoder.go
+++ b/core/app/crash/reporting/encoder.go
@@ -25,7 +25,6 @@ type encoder struct {
 	appVersion string
 	osName     string
 	osVersion  string
-	stacktrace string
 }
 
 const (
@@ -33,7 +32,7 @@ const (
 	multipartBoundary = "::multipart-boundary::"
 )
 
-func (e encoder) encode() (io.Reader, string, error) {
+func (e encoder) encodeStacktrace(stacktrace string) (io.Reader, string, error) {
 	buf := bytes.Buffer{}
 	w := multipart.NewWriter(&buf)
 	if err := w.SetBoundary(multipartBoundary); err != nil {
@@ -56,8 +55,43 @@ func (e encoder) encode() (io.Reader, string, error) {
 			return nil, "", err
 		}
 	}
-	if e.stacktrace != "" {
-		if err := w.WriteField("exception_info", e.stacktrace); err != nil {
+	if stacktrace != "" {
+		if err := w.WriteField("exception_info", stacktrace); err != nil {
+			return nil, "", err
+		}
+	}
+	return &buf, w.FormDataContentType(), nil
+}
+
+func (e encoder) encodeMinidump(minidumpName string, minidumpData []byte) (io.Reader, string, error) {
+	buf := bytes.Buffer{}
+	w := multipart.NewWriter(&buf)
+	if err := w.SetBoundary(multipartBoundary); err != nil {
+		return nil, "", err
+	}
+	defer w.Close()
+	if err := w.WriteField("product", crashProduct); err != nil {
+		return nil, "", err
+	}
+	if err := w.WriteField("version", e.appName+":"+e.appVersion); err != nil {
+		return nil, "", err
+	}
+	if e.osName != "" {
+		if err := w.WriteField("osName", e.osName); err != nil {
+			return nil, "", err
+		}
+	}
+	if e.osVersion != "" {
+		if err := w.WriteField("osVersion", e.osVersion); err != nil {
+			return nil, "", err
+		}
+	}
+	if minidumpName != "" && minidumpData != nil {
+		filefield, err := w.CreateFormFile("uploadFileMinidump", minidumpName)
+		if err != nil {
+			return nil, "", err
+		}
+		if _, err = io.Copy(filefield, bytes.NewReader(minidumpData)); err != nil {
 			return nil, "", err
 		}
 	}
--- a/core/app/crash/reporting/encoder_test.go
+++ b/core/app/crash/reporting/encoder_test.go
@@ -30,8 +30,7 @@ func TestEncoder(t *testing.T) {
 		appVersion: "V1.2.3",
 		osName:     "BobOS",
 		osVersion:  "10",
-		stacktrace: "foo1.bar 10:20\nfoo2.bar 20:30",
-	}.encode()
+	}.encodeStacktrace("foo1.bar 10:20\nfoo2.bar 20:30")
 	if assert.For(ctx, "err").ThatError(err).Succeeded() {
 		assert.For(ctx, "ty").ThatString(ty).Equals("multipart/form-data; boundary=" + multipartBoundary)
 		bytes, err := ioutil.ReadAll(r)
--- a/core/app/crash/reporting/reporting.go
+++ b/core/app/crash/reporting/reporting.go
@@ -21,6 +21,7 @@ package reporting
 import (
 	"context"
 	"fmt"
+	"io"
 	"net/http"
 	"net/url"
 	"sync"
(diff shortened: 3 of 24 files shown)


core/app/crash/crash.go
```diff
@@ -19,20 +19,33 @@
 package crash
```
ben-clayton: Please don't change this package. Just add another method to the `crash/reporting` package to deal with external process crashes.

core/app/analytics/crash_reporter.go
```diff
@@ -33,11 +33,13 @@ func init() {
 	crash.Register(onCrash)
 }
 
-func onCrash(e interface{}, s stacktrace.Callstack) {
+func onCrash(r crash.Report, appName, appVersion string) {
```
ben-clayton: Please revert this change. See my other comment.

core/app/crash/reporting/encoder.go
```diff
@@ -18,48 +18,60 @@ import (
 	"bytes"
 	"io"
 	"mime/multipart"
+
+	"github.com/google/gapid/core/app/crash"
```
ben-clayton: I'm not convinced that using the `crash` package here is better than no intra-dependencies.

core/app/crash/reporting/encoder_test.go
```diff
@@ -44,19 +41,6 @@ func TestEncoder(t *testing.T) {
 				"Content-Disposition: form-data; name=\"version\"\r\n" +
 				"\r\n" +
 				"TestApp:V1.2.3\r\n" +
```
ben-clayton: And now we're no longer testing a bunch of logic. Please revert.

core/app/crash/reporting/reporting.go
```diff
@@ -47,18 +48,23 @@ var (
...
+					}
 				}
 			}
-			err := reporter{appName, appVersion, osName, osVersion}.report(s, crashURL)
+			rep := reporter{app, version}
+			response, err := rep.report(r, crashURL)
 			if err != nil {
 				log.E(ctx, "%v", err)
+			} else {
+				log.I(ctx, "%s", response)
```
ben-clayton: This for debugging?

gapis/replay/executor/executor.go
```diff
@@ -48,12 +51,14 @@ func Execute(
 	payload protocol.Payload,
 	decoder builder.ResponseDecoder,
 	connection io.ReadWriteCloser,
-	memoryLayout *device.MemoryLayout) error {
+	memoryLayout *device.MemoryLayout,
```
ben-clayton: Just use the `device.Configuration`, it contains both the `MemoryLayout` and `OS`.

core/app/crash/crash.go
baldwinn860: What about the shutdown function setter? Without this change the cancelFunc set in app.handleCrashSignals could preempt any other crash handlers registered and shut down the app before they can report anything.

ben-clayton: > Without this change the cancelFunc set in app.handleCrashSignals could preempt any other crash handlers registered and shut down the app before they can report anything.

Nothing in the crash handler should depend on the context cancellation state. Are you seeing this somewhere?

gapis/replay/executor/executor.go
baldwinn860: device.Configuration will give us the memoryLayout for all ABIs this device supports, but we need the specific one that the capture was taken with.

core/app/crash/reporting/reporting.go
```diff
@@ -74,28 +80,25 @@ func Disable() {
 	}
 }
 
-type reporter struct {
-	appName    string
-	appVersion string
-	osName     string
-	osVersion  string
+func ReportMinidump(r Reporter, minidumpName string, minidumpData []byte) error {
```
ben-clayton: Godocs please

core/app/crash/reporting/reporting.go
```diff
@@ -74,28 +80,25 @@ func Disable() {
...
+		if err := r.reportMinidump(minidumpName, minidumpData, crashURL); err != nil {
+			return err
+		}
+	}
+	return nil
 }
 
-func (r reporter) report(s stacktrace.Callstack, endpoint string) error {
-	stacktrace := s.String()
+type Reporter struct {
```
ben-clayton: Godocs please

core/cc/crash_handler.cpp
```diff
@@ -0,0 +1,43 @@
...
+void CrashHandler::unsetHandlerFunction() {
+    mHandlerFunction = defaultHandlerFunction;
+}
+
+bool CrashHandler::handleMinidump(const std::string& minidumpPath, bool succeeded) {
+    return mHandlerFunction(minidumpPath, succeeded);
+}
+
+
+}
```
ben-clayton: `}  // namespace core`

core/cc/file_reader.cpp
```diff
@@ -0,0 +1,59 @@
...
+        GAPID_ERROR("Failed to get size of file")
+        rewind(mFile);
+        return 0u;
+    }
+    uint64_t size = static_cast<uint64_t>(sizeTemp);
+    rewind(mFile);
+    return size;
+}
+
+}
```
ben-clayton: `}  // namespace core`

core/cc/file_reader.h
```diff
@@ -0,0 +1,50 @@
...
+#define CORE_FILE_READER_H
+
+#include "stream_reader.h"
+
+#include <stdio.h>
+#include <stdint.h>
+
+namespace core {
+
+// FileReader is a pure-virtual interface used to read from binary files.
```
ben-clayton: Doesn't look very pure-virtual interface like to me!

gapir/cc/crash_uploader.cpp
```diff
@@ -0,0 +1,67 @@
...
+        return succeeded;
+    });
+}
+
+CrashUploader::~CrashUploader() {
+    GAPID_INFO("destroying uploader");
+    mCrashHandler.unsetHandlerFunction();
+}
+
+}
```
ben-clayton: `}  // namespace gapir`

gapir/cc/crash_uploader.h
```diff
@@ -0,0 +1,42 @@
...
+  CrashUploader(core::CrashHandler& crash_handler, const ServerConnection& conn);
+  ~CrashUploader();
+
+private:
+
+  core::CrashHandler& mCrashHandler;
+  const ServerConnection& mConnection;
+};
+
+}
```
ben-clayton: `}  // namespace gapir`

gapir/cc/server_connection.cpp
```diff
@@ -126,4 +126,31 @@ bool ServerConnection::post(const void* postData, uint32_t postSize) const {
     return true;
 }
 
+bool ServerConnection::crash(const std::string& filename, const void* crashData, uint32_t crashSize) const {
+    GAPID_DEBUG("CRASH: [%s] %p (%d)", filename.c_str(), crashData, crashSize);
```
ben-clayton: GAPID_WARNING?

gapir/cc/server_connection.h
```diff
@@ -56,11 +56,16 @@ class ServerConnection {
     // the posting was successful false otherwise.
     bool post(const void* postData, uint32_t postSize) const;
 
+    // Post a crash minidump from the given address with the given filename to the server. Returns
+    // true if the posting was successful false otherwise.
+    bool crash(const std::string& filename, const void* crashData, uint32_t crashSize) const;
```
ben-clayton: maybe rename `postCrashdump`? The other methods are verbs, and this doesn't sound like a function you'd want to call. 😄

PR #1395, opened 2017-11-21
Gapir Symbols

baldwinn860:
This change adds GAPIR symbol dumping on the Linux and OS X kokoro scripts. Next will be to add symbol uploading to the release scripts.


kokoro/linux/build.sh
```diff
@@ -62,6 +62,12 @@ echo $(date): Starting build...
 ./do build --test build --buildnum $KOKORO_BUILD_NUMBER --buildsha "$BUILD_SHA"
 echo $(date): Build completed.
 
+# Dump symbols from GAPIR then strip them
+pushd "$BUILD_ROOT/out/release/bin/"
+./dump_syms "gapir" > gapir.sym
+strip "gapir"
```
pmuetschard: you are doing this in the bin folder, however, the package script grabs things from the pkg folder, so stripping this binary has no effect.

Is it possible to make the symbol exporting part of the cmake build, so that before cmake copies the binaries to the pkg folder, the symbols have already been stripped?

PR #1396, opened 2017-11-21
Cache buffer memory requirements and track opaque sparse bindings in blocks

Qining:


gapii/cc/vulkan_mid_execution.cpp
```diff
@@ -331,19 +331,7 @@ uint32_t GetMemoryTypeIndexForStagingResources(
...
-    if (o.start() <= pos) {
-      pos = o.end();
-    } else {
-      return false;
-    }
-  }
-  if (pos < offset + size) {
-    return false;
-  }
+                  const U64ToVkSparseMemoryBind& bindings) {
```
AWoloszyn: This doesn't seem right

PR #1401, opened 2017-11-22
Fall back to /tmp/ if TMPDIR is not defined.

AWoloszyn:


core/cc/linux/crash_handler.cpp
```diff
@@ -32,10 +32,18 @@ static bool handleCrash(const google_breakpad::MinidumpDescriptor& descriptor, v
 
 namespace core {
 
+const char* GetTempDir() {
```
ben-clayton: Probably should be moved to an anonymous namespace.

PR #1410, opened 2017-11-27
gapis: Add --enable-local-files flag.

ben-clayton:
Without this flag, the RPC LoadCapture will return an error.


ben-clayton:
I just realized this is missing from GAPIT. Will update.

ben-clayton:
Now fixed.

PR #1421, opened 2017-11-29
Build GAPIR with MSVC on Windows.

baldwinn860:
This change adds the ability to build GAPIR with msvc on Windows in
order to produce pdb symbols. Most changes have to do with adding an
external project that specifies msvc as the compiler, as well as finding
the correct libraries and setting the correct compiler flags when
running under that project.


do.bat
```diff
@@ -25,10 +25,6 @@ cd %DO_DIR%..\..\..\..\
 SET GOPATH=%DO_DIR%third_party;%CD%
 popd
 
-REM -- Do some env cleanup.
```
ben-clayton: This is a bit scary - these were needed to stop random builds failing due to crap on the env.
That said, we're hopefully moving to Bazel Real Soon™ now, so meh.

baldwinn860: By removing LIB and LIBPATH I could not build with MSVC at all (kernel32.lib and others are on that path), at the very least I would need to store this and set it later.

pmuetschard: Did you test that it will still build (and run OK) after removing these lines using MSYS? I see you made a change to how it finds OpenGL, which was broken if LIB/LIBPATH were set. I'm just wondering if there are any other libs that could be picked up from the wrong place...

baldwinn860:
Yes, building with MSYS still works, at least on my shell which is running with vcvarsall.bat.

PR #1431, opened 2017-11-29
gapis/resolve: Make framebufferAttachmentInfo public.

ben-clayton:
We were exposing the `AttachmentFramebufferChanges` but without any way to use it. Move `getFramebufferAttachmentInfo` as a method of `AttachmentFramebufferChanges`, and make `framebufferAttachmentInfo` public.

I turned out not to need this in the end, but I feel the API is cleaner this way.


ben-clayton:
*ping*

PR #1434, opened 2017-11-29
Add MSVCWinGapir to kokoro script, fix build error.

baldwinn860:
There was a strange build error that suddenly cropped up involving a
logging macro getting called from within a lambda under the core
namespace (msvc only). Sometimes the msvc build flags were getting
ignored as well.

Kokoro now builds gapir with msvc, this adds a call to vcvarsall.bat in
the windows script and adds the MSVCWinGapir flag to the build
configuration.


baldwinn860:
ok, done

Issue #1446, opened 2017-12-02
Missing framebuffer images


bbilodeau:
After successfully capturing OpenGL ES on a Pixel device to a Mac, most frames are not displayed and the following error is shown: "Couldn't get framebuffer attachment Cause: Failed to connect to device Cause: dial tcp [::1]:50334: getsockopt: connection refused". All of the draw calls for the frames appear to be captured correctly.

ben-clayton:
I've reached out to @bbilodeau directly for more information.

ttgdz:
I still found the issue ,I wonder that the issue is solved or not?

ben-clayton:
Hi @ttgdz,

The error described in this bug is extremely generic - it pretty much covers all cases where the replay service crashed unexpectedly. If you are encountering this, please can you file a new bug? If it is possible, please can you include a `.gfxtrace` file and / or test APK that can be used to reproduce the problem?

Many thanks,
Ben

ttgdz:
Hi @ben-clayton 
    yes,the attachment is the trace file.


[gfxtrace.zip](https://github.com/google/gapid/files/1818553/gfxtrace.zip)

PR #1451, opened 2017-12-04
gles: Fix for cross-context deletions.

ben-clayton:
It is a no-op to delete objects that do not exist.
GAPID uses a shared context for replay - while its fine to delete IDs that do not exist in the current context, they might exist in another. Strip out any IDs that do not belong to the current context in the compat layer.

Fixes: #1446


ben-clayton:
`glDeleteProgram`, `glDeleteShader` and `glDeleteSync` will all return `GL_INVALID_VALUE` if the ID is neither 0 or a valid object. If the command errors, it will not get this far.

Added compat for `glDeleteProgramPipelines` and `glDeleteTransformFeedbacks`.

PR #1453, opened 2017-12-04
Improve GAPIR debugging

ben-clayton:
Will not be submitted until after v1.

### DIFF
--- a/core/cc/crash_handler.cpp
+++ b/core/cc/crash_handler.cpp
@@ -17,30 +17,28 @@
 #include "core/cc/crash_handler.h"
 #include "core/cc/log.h"
 
-namespace {
-auto handler = [] (const std::string& minidump_path, bool succeeded) {
+namespace core {
+
+CrashHandler::Handler CrashHandler::defaultHandler =
+        [] (const std::string& minidump_path, bool succeeded) {
             if (!succeeded) {
                 GAPID_ERROR("Failed to write minidump out to %s", minidump_path.c_str());
             }
-            return succeeded;
         };
-}
 
-namespace core {
-
-CrashHandler::HandlerFunction CrashHandler::defaultHandlerFunction = handler;
-
-void CrashHandler::setHandlerFunction(HandlerFunction newHandlerFunction) {
-    mHandlerFunction = newHandlerFunction;
-}
-
-void CrashHandler::unsetHandlerFunction() {
-    mHandlerFunction = defaultHandlerFunction;
+CrashHandler::Unregister CrashHandler::registerHandler(Handler handler) {
+    auto id = mNextHandlerID++;
+    mHandlers[id] = handler;
+    return [this, id] () {
+        mHandlers.erase(id);
+    };
 }
 
 bool CrashHandler::handleMinidump(const std::string& minidumpPath, bool succeeded) {
-    return mHandlerFunction(minidumpPath, succeeded);
+    for (const auto& it : mHandlers) {
+        it.second(minidumpPath, succeeded);
+    }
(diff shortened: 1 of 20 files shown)


core/cc/crash_handler.cpp
```diff
@@ -17,29 +17,28 @@
...
+        mHandlers.erase(id);
+    };
 }
 
 bool CrashHandler::handleMinidump(const std::string& minidumpPath, bool succeeded) {
-    return mHandlerFunction(minidumpPath, succeeded);
+    for (const auto& it : mHandlers) {
+        it.second(minidumpPath, succeeded);
+    }
+    return true;
```
baldwinn860: At the very least this should be `return succeeded` since if the minidump was not created we should let any other error handling mechanisms go to work.
```
if (Debugger::isAttached()) {
    return false;
}
return succeeded;
```
Should allow a debugger to attach to the process on OS X without disabling the handler as well.

core/cc/osx/crash_handler.cpp
```diff
@@ -45,9 +50,16 @@ const char* GetTempDir() {
...
-    mHandler(new google_breakpad::ExceptionHandler(
-        GetTempDir(), NULL, ::handleCrash, reinterpret_cast<void*>(this), true, NULL)) {
+    mNextHandlerID(0),
+    mExceptionHandler(nullptr) {
+
+    if (!Debugger::isAttached()) {
+        mExceptionHandler = std::unique_ptr<google_breakpad::ExceptionHandler>(
+            new google_breakpad::ExceptionHandler(
+                GetTempDir(), nullptr, ::handleCrash,
+                reinterpret_cast<void*>(this), true, nullptr));
```
baldwinn860: This could potentially be prevented in CrashHandler::handleMinidump by returning false, which will allow any other exception handlers (i.e. the debugger) to process the crash. I haven't tested this personally though.

gapir/cc/interpreter.cpp
```diff
@@ -83,6 +90,11 @@ bool Interpreter::run(const uint32_t* instructions, uint32_t count) {
     GAPID_ASSERT(mCurrentInstruction == 0);
     mInstructions = instructions;
     mInstructionCount = count;
+    mCrashHandler.registerHandler([this](const std::string& minidumpPath, bool succeeded) {
+        GAPID_ERROR("--- CRASH DURING REPLAY ---");
+        GAPID_ERROR("LAST COMMAND:     %d", mLabel);
+        GAPID_ERROR("LAST INSTRUCTION: %d", mCurrentInstruction);
+    });
```
baldwinn860: Not capturing the Unregister lambda here, so this handler never gets removed. The Interpreter will get destructed but the crash handler doesn't (until the end of the program at least), so `this->mLabel` will cause a crash within a crash.

core/cc/crash_handler.cpp
ben-clayton: > any other error handling mechanisms go to work

What other handling mechanisms?

core/cc/crash_handler.cpp
```diff
@@ -17,29 +17,28 @@
 #include "core/cc/crash_handler.h"
 #include "core/cc/log.h"
 
-namespace {
-auto handler = [] (const std::string& minidump_path, bool succeeded) {
+namespace core {
```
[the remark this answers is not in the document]
ben-clayton: Buildbot seems happy.

core/cc/osx/crash_handler.cpp
ben-clayton: Done

PR #1454, opened 2017-12-04
Update adb devices in the registry when their status changes. 

### LINKED ISSUE #1445: Device selection has issues
We've had a number of reports now where devices are either never showing up in the UI, or are taking longer than users are expecting and are closing the dialog before they appear.

We should investigate whether there are cases where the device never appears (maybe when authorizing ADB debug?), or if there's clearer hints we could give to the user to wait longer.

pmuetschard:
If a device's status changes in adb, this change will cause it to be removed from the registry and re-added, in order to keep the device configuration information up-to-date. When a device is added to the registry, it's configuration is queried, which fails if the device is not online and the device is added with empty/incomplete information. With this change, such a device is updated once it becomes online. This is especially needed for devices needing user authorization, since we're likely to see the un-authorized device long before the user can give permissions in the dialog.

Fixes #1445

### DIFF
--- a/core/os/android/adb/device.go
+++ b/core/os/android/adb/device.go
@@ -194,11 +194,14 @@ func scanDevices(ctx context.Context) error {
 	defer cacheMutex.Unlock()
 
 	for serial, status := range parsed {
-		device, ok := cache[serial]
-		if !ok {
-			device, err = newDevice(ctx, serial, status)
+		cached, ok := cache[serial]
+		if !ok || status != cached.Status() {
+			device, err := newDevice(ctx, serial, status)
 			if err != nil {
 				return err
+			}
+			if ok {
+				registry.RemoveDevice(ctx, cached)
 			}
 			cache[serial] = device
 			registry.AddDevice(ctx, device)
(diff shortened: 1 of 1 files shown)


core/os/android/adb/device.go
```diff
@@ -195,7 +195,7 @@ func scanDevices(ctx context.Context) error {
 
 	for serial, status := range parsed {
 		device, ok := cache[serial]
-		if !ok {
+		if !ok && status == bind.Status_Online {
```
ben-clayton: I'm a little bit on the fence about this. A few thoughts:

1. The package doesn't say anything about filtering to just online devices. It makes the status field particularly pointless. There's a few scattered comments that should be updated for this change in behavior.
1. I'm not sure if remote devices will always be 'Online' and not sometimes 'Unknown'.
1. Would it not be better to show 'Unauthorized' in the client?

For the sake of the smallest delta, I can approve this on the condition we revisit it. Alternatively, how do you feel about a `registry.RemoveDevice()` then `registry.AddDevice()` if the status changes?

core/os/android/adb/device.go
```diff
@@ -46,13 +46,18 @@ var (
...
 	// cache is a map of device serials to fully resolved bindings.
-	cache      = map[string]*binding{}
+	cache      = map[string]*bindingAndStatus{}
 	cacheMutex sync.Mutex // Guards cache.
 
 	// Registry of all the discovered devices.
 	registry = bind.NewRegistry()
 )
 
+type bindingAndStatus struct {
```
ben-clayton: Do you need this? Couldn't you just get the status from the binding?

pmuetschard: Done. Haha, yes, of course, silly me.

PR #1455, opened 2017-12-05
Output GAPIR symbols on Windows.

baldwinn860:
This change fixes a build error with msvc where a log line could not be
used inside a lambda under the core namespace. Several msvc specific
compile flags were being ignored as well.

Configurations that specify MSVCWinGapir now will create a single
linktime generated pdb file that then is processed with dump_syms to
produce a symbol file that can be uploaded to the crash server.


CMakeLists.txt
```diff
@@ -57,6 +57,11 @@ if(MSVC_GAPIR AND WIN32)
 endif()
 
 if(MSVC)
+    # turn off msvc specific warnings, define platform, turn on pdb generation
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /wd4091 /wd4267 /wd4996 /wd4244 /D__x86_64 /Zi")
```
pmuetschard: Does the build fail if the warnings are not turned off? If so, please document for each warning type, why we can't fix the warnings, otherwise, leave the warnings on.

cmd/gapir/CMakeBuild.cmake
```diff
@@ -31,6 +31,13 @@ if(NOT DISABLED_CXX)
     elseif(NOT GAPII_TARGET)
         add_executable(gapir ${sources})
         install(TARGETS gapir DESTINATION ${TARGET_INSTALL_PATH})
+        if(MSVC)
+            add_custom_command(
```
pmuetschard: Would this be better as an actual target? I'm not a big fan of the "/../../../../" path - reading this, I have no idea where it goes.

baldwinn860: I agree, but I wasn't sure how to specify the gapir-msvc target directory from the custom command. Do you think it would work if I made it a ExternalProject_Step that depended on gapir?

pmuetschard:
Looks OK to me, @ben-clayton?

ben-clayton:
AFAICT, this does not affect the build in any way, unless you explicitly ask for an MSVC build.
Is that correct @baldwinn860 ?

baldwinn860:
Yes currently. I have another cl after this that enables msvc builds in kokoro.

ben-clayton:
Thanks. Landed.

PR #1457, opened 2017-12-05
Actually use the -gapir-device to select the device to run gapir

Qining:
By default, the 'gapir-device' arg is empty, and has no effect on the device selection.
If 'gapir-device' is set to 'host', select only the host if 'host' is one of the compatible devices
if 'gapir-device' is set to 'android', uses only the android devices in the compatible device list
if 'gapir-device' is anything else, consider it as the serial number of an Android device, only the exactly matched Android device will be used.

### DIFF
--- a/cmd/gapit/common.go
+++ b/cmd/gapit/common.go
@@ -28,6 +28,8 @@ import (
 	"github.com/google/gapid/core/app/crash"
 	"github.com/google/gapid/core/log"
 	"github.com/google/gapid/core/os/android/adb"
+	"github.com/google/gapid/core/os/device"
+	"github.com/google/gapid/core/os/device/host"
 	"github.com/google/gapid/gapis/api"
 	"github.com/google/gapid/gapis/client"
 	"github.com/google/gapid/gapis/memory"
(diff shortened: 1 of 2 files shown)


cmd/gapit/common.go
```diff
@@ -111,8 +113,39 @@ func getDevice(ctx context.Context, client client.Client, capture *path.Capture,
...
-	if len(paths) > 0 {
-		return paths[0], nil
+	filteredByFlags := make([]*path.Device, 0, len(paths))
+	for i := 0; i < len(paths); i++ {
+		p := paths[i]
+		o, err := client.Get(ctx, p.Path())
+		if err != nil {
+			return nil, log.Err(ctx, err, "Couldn't resolve device")
+		}
+		d := o.(*device.Instance)
```
Qining: Should we log an Info here about the command line argument value?

ben-clayton: Would it not make more sense to do the device matching inside the vulkan's logic for `GetDevicesForReplay` ? That way this would work for GAPIS too.

Qining: That part will be done in #1501. This part is like, if there are more devices to replay, the user may use `-gapir-device` to specify which one to use.

Qining:
close as #1501 has been submitted.

PR #1458, opened 2017-12-05
Remove boxed string and template from log calls.

baldwinn860:
Every call to the GAPID_LOG macro in C++ instantiated a new template
function to remove some path information from the __FILE__ macro. This
pulluted the function symbols by nearly 87% and slowed down the build by
about 5 minutes (Win32 full clean/rebuild goes from 25m -> 20m).

### DIFF
--- a/core/cc/log.h
+++ b/core/cc/log.h
@@ -54,35 +54,6 @@
   #define PRIsize "zu"
 #endif
 
-// Template meta-programming (tmp) used to strip absolute path from __FILE__.
-// The main difficulty is that we need to make copy of the string since
-// any live reference to __FILE__ will keep the whole path in the binary.
-// We also want to avoid storing the string again for each log message.
-namespace tmp {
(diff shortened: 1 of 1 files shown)


baldwinn860:
I can't seem to add a reviewer to this @ben-clayton. Is this due to removing permissions for code freeze?

ben-clayton:
> I can't seem to add a reviewer to this @ben-clayton. Is this due to removing permissions for code freeze?

Correct.

Is this change absolutely necessary for V1.0?

baldwinn860:
No it isn't required for v1, I was just confused about the reviewers bit.

ben-clayton:
Thanks for the clarification.

While this looks completely harmless, I'd only like to land it if all the other parts of the MSVC build are ready and thoroughly tested.
Given that this is right up to the edge, let's consider this post-v1 for now. I'm happy to be convinced otherwise.

PR #1461, opened 2017-12-06
Fixes for bleeding-edge GVR builds

ben-clayton:
This is a highly risky, but critical bunch of fixes for GVR.

Please review thoroughly - I'm tired and likely to have made mistakes.

### DIFF
--- a/gapis/api/gles/api/extensions.api
+++ b/gapis/api/gles/api/extensions.api
@@ -648,8 +648,10 @@ cmd void glEGLImageTargetTexture2DOES(GLenum target, GLeglImageOES image) {
 
   for _, l, img in eglImage.Images {
     SetTextureImage(t, 0, as!GLint(l), img)
-    t.EGLImage = eglImage
-    GetEGLImageData(eglImage.ID, img.Width, img.Height)
+    if l == 0 {
+      t.EGLImage = eglImage
+      GetEGLImageData(eglImage.ID, img.Width, img.Height)
+    }
   }
 }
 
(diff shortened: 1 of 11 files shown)


gapis/api/gles/api/extensions.api
```diff
@@ -642,9 +645,12 @@ cmd void glEGLImageTargetTexture2DOES(GLenum target, GLeglImageOES image) {
...
   }
   t := GetBoundTextureOrErrorInvalidEnum(target)
-  SetTextureImage(t, 0, 0, eglImage.Image)
-  t.EGLImage = eglImage
-  GetEGLImageData(eglImage.ID, eglImage.Image.Width, eglImage.Image.Height)
+
+  for _, l, img in eglImage.Images {
+    SetTextureImage(t, 0, as!GLint(l), img)
+    t.EGLImage = eglImage
+    GetEGLImageData(eglImage.ID, img.Width, img.Height)
```
dsrbecky: You probably don't want to call this n times...

gapis/api/gles/gles.api
```diff
@@ -119,11 +120,11 @@ class SupportedExtensions {
   bool GL_EXT_primitive_bounding_box                   = true
   bool GL_EXT_raster_multisample                       = true
   bool GL_EXT_robustness                               = true
-  bool GL_EXT_sRGB_write_control                       = true
```
dsrbecky: It does not matter now, just be aware that once we start using MEC, this class (and most state) will be append-only.

gapis/api/gles/api/extensions.api
ben-clayton: Added a `if l == 0` block around these two statements (we still need the image width / height from somewhere)

gapis/api/gles/gles.api
ben-clayton: Ack

PR #1464, opened 2017-12-06
GAPII fixes for S7

ben-clayton:


AWoloszyn:
It is a bit gross, but I think reasonable to work around a fairly common device/issue.

Issue #1356, opened 2017-11-09
Give a sensible error message if the host doesn't support OpenGL


ben-clayton:
Forked from #1349.

It is not clear that this is the reason the client errors on startup.

pmuetschard:
This is done on Windows and macOS. Removing from milestone as only Linux is missing.

PR #1469, opened 2017-12-06
 Don't crash the client on OpenGL init failure.

pmuetschard:
On Windows and OSX, where we are doing our own GL initilization, if we fail to create a GL context, don't SWT.error, but just fallback to a regular canvas and show an error message in the Geometry, Framebuffer and Texture views.

For #1356


ben-clayton:
Ready to submit?

PR #1472, opened 2017-12-07
Update the callouts in vulkan render state tutorial

Qining:


Qining:
oops, not able to submit....repo is locked.

PR #1475, opened 2017-12-07
build: Fix missing gapir.exe

ben-clayton:


cmd/gapir/CMakeBuild.cmake
```diff
@@ -25,6 +25,7 @@ if(MSVC_GAPIR AND WIN32)
         DEPENDEES gapir
         BYPRODUCTS "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/../gapir.sym"
     )
+    install(TARGETS gapir DESTINATION ${TARGET_INSTALL_PATH})
```
AWoloszyn: I think this is not quite correct, given the name of the target.

We should be able to just update add_cmake_target to:

```
add_cmake_target("gapir-msvc" gapir "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}" "gapir.exe"
    DEPENDEES gapir_static  DEPENDEES gapir_static
    DEPENDS ${sources}  DEPENDS ${sources}
    DESTINATION ${TARGET_INSTALL_PATH} 
)
```

PR #1485, opened 2017-12-08
gles: Fixes for unused VAAs that have no data.

ben-clayton:
On some drivers having a vertex attribute array enabled with no bound data will cause `GL_INVALID_OPERATION` errors, even though that vertex attribute is not used by the program.

This is a compat workaround for this case where we temporarily disable the vertex attribute array if we believe the program does not use it.

### DIFF
--- a/gapis/api/gles/compat.go
+++ b/gapis/api/gles/compat.go
@@ -266,6 +266,12 @@ func compat(ctx context.Context, device *device.Instance) (transform.Transformer
 			return
 		}
 
+		if cmd.CmdFlags(ctx, id, s).IsDrawCall() {
+			t := newTweaker(out, dID, cb)
+			disableUnusedAttribArrays(ctx, t)
+			defer t.revert(ctx)
+		}
+
 		switch cmd := cmd.(type) {
 		case *GlBindBuffer:
 			if cmd.Buffer == 0 {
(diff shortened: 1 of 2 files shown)


gapis/api/gles/compat.go
```diff
@@ -1289,6 +1296,48 @@ func canUsePrecompiledShader(c *Context, d *device.OpenGLDriver) bool {
...
+
+	p := c.Bound.Program
+	if p == nil || p.ActiveResources == nil {
+		return nil
+	}
+	inputs := p.ActiveResources.ProgramInputs
+	used := make([]bool, c.Constants.MaxVertexAttribBindings)
+	for _, input := range inputs.Range() {
+		for _, l := range input.Locations.Range() {
+			if l < GLint(len(used)) {
```
dsrbecky: GL can sometimes return locations which are -1, ignore those in the condition? (or make the array into a map)

gapis/api/gles/compat.go
```diff
@@ -1289,6 +1296,48 @@ func canUsePrecompiledShader(c *Context, d *device.OpenGLDriver) bool {
...
+	for l, arr := range c.Bound.VertexArray.VertexAttributeArrays.Range() {
+		if arr.Enabled == GLboolean_GL_TRUE && l < AttributeLocation(len(used)) && !used[l] {
+			out.MutateAndWrite(ctx, dID, cb.GlDisableVertexAttribArray(l))
+			reenable = append(reenable, l)
+		}
+	}
+	if len(reenable) == 0 {
+		return nil
+	}
+	return func() {
```
dsrbecky: Nit: Using tweaker would be a bit more idiomatic in compat.

gapis/api/gles/compat.go
ben-clayton: Done

gapis/api/gles/compat.go
ben-clayton: Done

gapis/api/gles/tweaker.go
```diff
@@ -91,6 +91,12 @@ func (t *tweaker) glDisable(ctx context.Context, name GLenum) {
 	}
 }
 
+func (t *tweaker) glDisableVertexAttribArray(ctx context.Context, l AttributeLocation) {
+	t.doAndUndo(ctx,
```
dsrbecky: Strictly speaking, this needs the "if not already in the desired state", otherwise calling this already disabled array, will incorrectly enable it after the tweaker is reverted.
I know that nn this case you do the check in the caller.

ben-clayton: Done

PR #1486, opened 2017-12-08
WIP: Remove the compile-time dependency to libGL.

AWoloszyn:
This is only on Linux for now. TODO: the rest of the platforms.
Instead of linking at compile-time to the .so file we load the
symbols dynamically. This means we can run on machines that only
have Vulkan.


ben-clayton:
Drop the WIP from the title and I'm happy to approve (post v1)

ben-clayton:
This looks like it has already been merged into master - shall we close this?

PR #1500, opened 2017-12-13
Bazel Work

pmuetschard:
```
$ bazel build cmd/gapis
INFO: Analysed target //cmd/gapis:gapis (118 packages loaded).
INFO: Found 1 target...
Target //cmd/gapis:gapis up-to-date:
 bazel-bin/cmd/gapis/linux_amd64_stripped/gapis
INFO: Elapsed time: 14.568s, Critical Path: 12.70s
INFO: Build completed successfully, 7 total actions
```

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -138,6 +138,13 @@ new_git_repository(
     build_file = "//tools/build/third_party:breakpad.BUILD",
 )
 
+new_git_repository(
+    name = "lss",
+    remote = "https://chromium.googlesource.com/linux-syscall-support",
+    commit = "e6527b0cd469e3ff5764785dadcb39bf7d787154",
+    build_file = "//tools/build/third_party:lss.BUILD",
+)
+
 github_go_repository(
     name = "org_golang_x_crypto",
     organization = "golang",
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -10,7 +10,7 @@ github_repository(
     organization = "bazelbuild",
     project = "rules_go",
     branch = "master",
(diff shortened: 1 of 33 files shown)


gapis/api/gles/gles_pb/BUILD.bazel
```diff
@@ -35,4 +35,5 @@ go_proto_library(
     importpath = "github.com/google/gapid/gapis/api/gles/gles_pb",
     proto = ":gles_pb_proto",
     visibility = ["//visibility:public"],
+    deps = ["//gapis/memory/memory_pb:go_default_library"],  # keep
```
ben-clayton: I think the recommended approach here is to add a dummy import to the `docs.go` file instead of adding explicit `# keep` rules to the bazel rules.

ben-clayton: ( but maybe this is different because of the proto aspect *shrug* )

ianthehat: # keep is better, it just did not exist when I started so I did the doc.go trick instead
the advantage of the #keep it is probably easier to find non real dependancies and fix them up better in the future, the _ import might be because you actaully rely on the packages init functions being called (and also it will force them to be called even if not needed, preventing the compiler from removing the uneeded dependancy)

ianthehat: yes, but don't, because there is an even better way coming next week that means you won't need either...
You can now write code generators that return both source and the deps the sources require, so the apic templates can carry the deps for you.

ianthehat: bazelbuild/rules_go#1131 has most of that work, but it's not submitted yet

PR #1501, opened 2017-12-13
Add VulkanDriver to the device::Instance trace header

Qining:
Sits on top of #1457, but actually is totally independent.

An empty VulkanDriver will be created along with GL driver and other
generic information if Vulkan loader can be found. Real content of
VulkanDriver will be populated in the first vkCreateInstance call of the
tracing application.

When VulkanDriver is populated, another header with updated
device::Instance will be sent (ID rehased also). Decoder of the trace
file should always use the latest header.

The device::Instance is cached in `SpyBase`. Also moved `writeHeader()` from `Spy` to `SpyBase`.

States of VulkanDriver in the header:
1) Does not exist: The tracing device does not support Vulkan.
2) Exist but empty: The trace does not contain vkCreateInstance, the
trace should be replayable on all ABI compatible Vulkan device.
3) Exist and not empty: The trace contains vkCreateInstance, replay will
be initiated only on same vendor/device/api version device.

### DIFF
--- a/cmd/gapit/common.go
+++ b/cmd/gapit/common.go
@@ -28,6 +28,8 @@ import (
 	"github.com/google/gapid/core/app/crash"
 	"github.com/google/gapid/core/log"
 	"github.com/google/gapid/core/os/android/adb"
+	"github.com/google/gapid/core/os/device"
+	"github.com/google/gapid/core/os/device/host"
 	"github.com/google/gapid/gapis/api"
 	"github.com/google/gapid/gapis/client"
 	"github.com/google/gapid/gapis/memory"
@@ -111,8 +113,39 @@ func getDevice(ctx context.Context, client client.Client, capture *path.Capture,
 		return nil, log.Err(ctx, err, "Failed query list of devices for replay")
 	}
 
-	if len(paths) > 0 {
-		return paths[0], nil
+	filteredByFlags := make([]*path.Device, 0, len(paths))
+	for i := 0; i < len(paths); i++ {
+		p := paths[i]
+		o, err := client.Get(ctx, p.Path())
+		if err != nil {
+			return nil, log.Err(ctx, err, "Couldn't resolve device")
+		}
+		d := o.(*device.Instance)
+		switch flags.Device {
+		case "":
+			// empty flag
+			filteredByFlags = append(filteredByFlags, p)
+		case "host":
+			if d == host.Instance(ctx) {
+				filteredByFlags = append(filteredByFlags, p)
+				break
+			}
+		case "android":
+			if d.GetConfiguration().GetOS().GetKind() == device.OSKind_Android {
+				filteredByFlags = append(filteredByFlags, p)
+			}
+		default:
+			// serial number
+			// TODO: Regex matching instead of exact match?
+			if d.GetSerial() == flags.Device {
+				filteredByFlags = append(filteredByFlags, p)
+				break
+			}
+		}
+	}
+
+	if len(filteredByFlags) > 0 {
+		return filteredByFlags[0], nil
 	}
 
 	log.W(ctx, "No compatible devices found. Attempting to use the first device anyway...")
--- a/cmd/gapit/common.go
+++ b/cmd/gapit/common.go
@@ -114,6 +114,7 @@ func getDevice(ctx context.Context, client client.Client, capture *path.Capture,
 	}
 
 	filteredByFlags := make([]*path.Device, 0, len(paths))
+	filteredInsts := map[*path.Device]*device.Instance{}
 	for i := 0; i < len(paths); i++ {
 		p := paths[i]
 		o, err := client.Get(ctx, p.Path())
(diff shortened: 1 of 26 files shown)


Qining:
Oops, if `gapit devices` is called during a trace, the result of `gapit devices` will be wrong.

Qining:
Found a dirty way to set/get system property from here: https://cs.chromium.org/chromium/src/base/sys_info_android.cc?q=__system_property&sq=package:chromium&l=53

cmd/gapit/common.go
```diff
@@ -111,8 +113,39 @@ func getDevice(ctx context.Context, client client.Client, capture *path.Capture,
...
+			// TODO: Regex matching instead of exact match?
+			if d.GetSerial() == flags.Device {
+				filteredByFlags = append(filteredByFlags, p)
+				break
+			}
+		}
+	}
+
+	if len(filteredByFlags) > 0 {
+		return filteredByFlags[0], nil
```
ben-clayton: In the case of `> 1`, perhaps print there were multiple usable devices, and which one was picked.

core/cc/dl_loader.cpp
```diff
@@ -29,23 +29,19 @@
 
 namespace {
 
+// load defs
 #if TARGET_OS == GAPID_OS_WINDOWS
-
-void* load(const char* name) {
+void* load(const char* name, bool must = true) {
```
ben-clayton: Maybe keep the signature of `load` as it was, let it return `nullptr` and add a target-generic `must_load` ?

core/os/device/device.proto
```diff
@@ -204,12 +204,32 @@ message OpenGLDriver {
...
-    // Driver vendor name. e.g. "Qualcomm".
-    string Vendor = 3;
-    // Renderer version. e.g. "OpenGL ES 3.0 V@53.0 AU@  (CL@)".
-    string Version = 4;
+    // Enumerated instance layers.
+    repeated VulkanLayer Layers = 1;
+    // Instance extensions provided by Vulkan implementations and implicit
+    // layers.
+    repeated string IcdAndImplicitLayerExtensions = 2;
+    // Physical devices that have Vulkan supports.
```
ben-clayton: s/supports/support

core/os/device/device.proto
```diff
@@ -204,12 +204,32 @@ message OpenGLDriver {
...
+    repeated VulkanPhysicalDevice PhysicalDevices = 3;
+}
+
+message VulkanLayer {
```
ben-clayton: Docs please

core/os/device/deviceinfo/cc/query.h
```diff
@@ -25,6 +27,45 @@ namespace query {
...
+// will be created with the VkCreateInstance resolved from either the given
+// callback or the Vulkan loader.  The modified device::Instance will have its
+// ID re-hashed with the new content. Returns true if Vulkan physical device
+// info is fetched successfully and device::Instance updated, otherwise returns
+// false and keeps the device::Instance unchanged. Caution: When called with
+// VkGraphicsSpy layer loaded i.e. during tracing, the function pointer to a
+// layer under VkGraphicsSpy must be passed in. Resolving the Vulkan function
+// addresses from loader will cause a infinite calling stack and may deadlock
+// in the loader.
+bool updateVulkanDriver(
```
ben-clayton: Maybe change `getDeviceInstance` to take a struct of options and include vulkan info as part of that? GAPIS also calls this for host info, and it would be a nicer API if there was a single call that can return you the full proto message.

gapis/api/vulkan/replay.go
```diff
@@ -44,10 +44,46 @@ var (
...
+		return 0
+	}
+
+	// The device does not support Vulkan
+	if devVkDriver == nil {
+		return 0
+	}
+
+	for _, abi := range devAbis {
+		if abi.GetMemoryLayout().SameAs(h.GetAbi().GetMemoryLayout()) {
```
ben-clayton: Maybe flip this to:

```go
if !abi.GetMemoryLayout().SameAs(h.GetAbi().GetMemoryLayout()) {
    continue
}
```

to keep the nesting down.

gapis/api/vulkan/templates/vk_spy_helpers.cpp.tmpl
```diff
@@ -254,6 +256,37 @@ uint32_t VulkanSpy::SpyOverride_vkCreateInstance(VkInstanceCreateInfo *pCreateIn
     // Actually call vkCreateInstance, and keep track of the result.
     uint32_t result = create_instance(pCreateInfo, pAllocator, pInstance);
 
+    // Send a header with Vulkan info added if we haven't done so.
```
ben-clayton: Not really introduced in this CL, but I'm not keen on this much bespoke code in a template.
Can it not be moved to a regular C++ file?

core/os/device/deviceinfo/cc/query.h
Qining: Added an option arg to getDeviceInstance() in bc15aaa.
But I think I should keep this 'updateVulkanDriver()'. As when it is called in vkCreateInstance(), it probably will not be valid to get GL driver info again? Calling 'updateVulkanDriver' avoid touching other fields in the device::Instance

gapis/api/vulkan/templates/vk_spy_helpers.cpp.tmpl
Qining: I agree, having such a lot of c++ code in template file is quite weird. I can work on that in another CL

core/os/device/device.proto
```diff
@@ -204,12 +204,34 @@ message OpenGLDriver {
...
-    string Renderer = 2;
-    // Driver vendor name. e.g. "Qualcomm".
-    string Vendor = 3;
-    // Renderer version. e.g. "OpenGL ES 3.0 V@53.0 AU@  (CL@)".
-    string Version = 4;
+    // Enumerated instance layers.
+    repeated VulkanLayer Layers = 1;
+    // Instance extensions provided by Vulkan implementations and implicit
+    // layers.
+    repeated string IcdAndImplicitLayerExtensions = 2;
```
ben-clayton: What does Icd mean / stand for?

core/os/device/deviceinfo/cc/query.h
```diff
@@ -17,13 +17,86 @@
...
+// By default, layers/extensions info and physical devices info will not be
+// queried.
+struct VulkanOption {
+  // Default VulkanOption specifies NOT to query layers/extensions and physical
+  // devices info.
+  VulkanOption()
+      : query_layers_and_extensions(false), query_physical_devices(false) {}
+  // QueryLayerAndExtensions set the flag which specifies layers and extensions
+  // info should be queried.
+  VulkanOption& QueryLayersAndExtensions() {
```
ben-clayton: Aren't we using lower underscore case for methods?

core/os/device/deviceinfo/cc/query.h
ben-clayton: This is almost what I was thinking, but not quite. We can work on this later.

core/os/device/device.proto
Qining: It stands for Vulkan Installable Client Driver, I think this is just the implementation of Vulkan (drivers). Icd seems to be used quite widely in Vulkan docs.

ben-clayton:
Please be sure to squash these changes before merging.

PR #1502, opened 2017-12-13
Gapir crash improvements

baldwinn860:
Just a few improvements for crash reporting.

Logging the report ID is helpful for debugging and is something the client already
does, this just adds it to GAPIS.

Deleting the minidump file keeps the user's disk space from getting cluttered with 
our crashes.

### DIFF
--- a/core/app/crash/reporting/reporting.go
+++ b/core/app/crash/reporting/reporting.go
@@ -66,7 +66,7 @@ func Enable(ctx context.Context, appName, appVersion string) {
 			}.reportStacktrace(s, crashURL)
 			if err != nil {
 				log.E(ctx, "%v", err)
-			} else {
+			} else if res != "" {
 				log.I(ctx, "%v", res)
 			}
 		})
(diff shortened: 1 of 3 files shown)


core/app/crash/reporting/reporting.go
```diff
@@ -57,14 +58,16 @@ func Enable(ctx context.Context, appName, appVersion string) {
...
-			err := Reporter{
+			res, err := Reporter{
 				appName,
 				appVersion,
 				osName,
 				osVersion,
 			}.reportStacktrace(s, crashURL)
 			if err != nil {
 				log.E(ctx, "%v", err)
+			} else {
```
ben-clayton: `} else if res != "" {`

core/app/crash/reporting/reporting.go
```diff
@@ -81,58 +84,66 @@ func Disable() {
 }
 
 // ReportMinidump encodes and sends a minidump report to the crashURL endpoint.
-func ReportMinidump(r Reporter, minidumpName string, minidumpData []byte) error {
+func ReportMinidump(r Reporter, minidumpName string, minidumpData []byte) (string, error) {
 	if disable != nil {
-		if err := r.reportMinidump(minidumpName, minidumpData, crashURL); err != nil {
-			return err
+		if res, err := r.reportMinidump(minidumpName, minidumpData, crashURL); err != nil {
```
ben-clayton: assuming `reportMinidump` returns only a value or error for all codepaths, you can just:

```go
return r.reportMinidump(minidumpName, minidumpData, crashURL)
```

pmuetschard:
FYI, when you update the crash reporting API, you need to fix stubs.go. @ben-clayton should we just get rid of the crashreporting tag, now?

PR #1509, opened 2017-12-15
More Bazel work.

pmuetschard:
- Rebase the bazel branch on top of master (rerun gazelle)
- Get gapir building on Linux and OSX (with some prep for Windows)
  - Fix building of the generated code by creating a separate library target
  - Get mm_library targets to not copy .h files to .h.cpp files
  - Crazy breakpad BUILD rules
  - Fix some include paths
- Get gapii building on Linux and OSX
  - Fix building of the generated code by creating a separate library target
  - Create build rules for the GVR templates
  - Add GVR to gapii
  - Add some missing CC proto targets
  - Fix some include paths

Note, the only significant added commits are:
b67c405: gazelle
82cef76: gapir
a398d7e: gapis


tools/build/rules/mm.bzl
```diff
@@ -25,11 +25,18 @@ def mm_library(name, srcs=[], hdrs = [], copy_hdrs=[], **kwargs):
...
-        copy(
-            name = copy_name,
-            src = src,
-            dst = name+"/"+src+".cc",
-        )
+        if src.endswith(".mm"):
+            copy(
+                name = copy_name,
+                src = src,
+                dst = name+"/"+src+".cc",
```
ben-clayton: This is copying `foo.mm` to `foo.mm.cc`?
Do you understand what this rule is used for? If so, can we put some comments in here?

pmuetschard: Yes, that is exactly what it does, because bazel doesn't like compiling .mm files (it has very strict rules on file extensions). Objective-C/C++ stuff is compiled with the default c++ compiler, but passing the "-x objective-c++" option to clang.

I will add some comments to this.

pmuetschard:
> There is no actual need to copy it in this case, ...

Unfortunately no, because the files are #included without paths and so without the copy, won't be found.

PR #1511, opened 2017-12-15
Unblock robot driven Vulkan trace/replay

### LINKED ISSUE #1448: gapid apk for may crash on Android supporting multiple ABI when querying device info
To reproduce, trace and replay an 32bit arm apk on Pixel C on Android 8.0 (whose preferred abi is 64bit arm. Then delete the 64 bit version of gapid apk.

Do not exit the gapid apk on the device.

Then do another trace for a 32 bit apk. This time, gapid will use the 32 bit version of gapid.apk to query the device info. And gapid.apk on device will crash.

Qining:
Fix #1448. To unblock Robot support of Vulkan.

Two things fixed in this CL:
1. Fall back to other compatible ABIs when the preferred abi apk is not found.
In Robot, only the gapid.apk which matches the application's ABI will be used (unzipped to workspace). For a 64 bit Android that can run 32 bit apk, the preferred abi is 64 bit, lacking of a fallback selection results into test failure at getting the device info (not tracing).
Manual usage of gapic does not have this problem as the 64 bit apk always present (64 bit for device info, 32 bit for tracing a 32 bit application)

2. Stop the gapidapk when we need a fresh run, fixes #1448.
When using armeabi (32 bit) build of gapidapk, starting the device info service after a replay fails (without given any messages, just the port is not opened for listening at all). Force stopping the running gapidapk and rerun it to start the service fixes the problem.


core/os/android/adb/commands.go
```diff
@@ -148,6 +148,31 @@ func (b *binding) StartService(ctx context.Context, a android.ServiceAction, ext
...
+	args := append([]string{
+		"force-stop", pkg,
+	})
+	return b.Shell("am", args...).Run(ctx)
+}
+
+// GetSystemProperty returns the system property in string
+func (b *binding) GetSystemProperty(ctx context.Context, name string) (string, error) {
+	var stdout bytes.Buffer
+	if err := b.Shell("getprop", name).Capture(&stdout, nil).Run(ctx); err != nil {
```
ben-clayton: You can use:

```go
	if res, err := d.Shell("getprop", name).Call(ctx); err == nil {
```

There are plenty of place we do this already. Might be worth moving those calls over to this method.

core/os/android/adb/commands.go
```diff
@@ -148,6 +148,31 @@ func (b *binding) StartService(ctx context.Context, a android.ServiceAction, ext
...
+// ForceStop stops everything associated with the given package.
+func (b *binding) ForceStop(ctx context.Context, pkg string) error {
+	args := append([]string{
+		"force-stop", pkg,
+	})
+	return b.Shell("am", args...).Run(ctx)
+}
+
+// GetSystemProperty returns the system property in string
+func (b *binding) GetSystemProperty(ctx context.Context, name string) (string, error) {
```
ben-clayton: Drop the `Get`.

gapidapk/deviceinfo.go
```diff
@@ -46,12 +151,19 @@ func fetchDeviceInfo(ctx context.Context, d adb.Device) error {
 		return nil
 	}
 
-	action := apk.ServiceActions.FindByName(sendDevInfoAction, sendDevInfoService)
-	if action == nil {
-		return log.Err(ctx, nil, "Service intent was not found")
+	// Remove VkGraphicsSpy in the debug.vulkan.layers property to avoid loading
+	// our spy layer.
+	recover, err := stripVkGraphicsSpyLayer(ctx, d)
```
ben-clayton: Avoid `recover`. It means something completely different in golang.

gapidapk/deviceinfo.go
```diff
@@ -25,15 +28,117 @@ import (
...
+	const propName = "debug.vulkan.layers"
+	const layerName = "VkGraphicsSpy"
+	ctx = log.Enter(ctx, "stripVkGraphicsSpyLayer")
+	log.I(ctx, "Check the existence of %s in %s", layerName, propName)
+	old_layer_str, err := d.SystemProperty(ctx, propName)
+	if err != nil {
+		return nil, log.Errf(ctx, err, "Getting %s.", propName)
+	}
+	should_strip := false
+	old_layers := strings.Split(old_layer_str, ":")
```
ben-clayton: Don't use underscores in variable names. Please use golint and govet as part of your development workflow.

cmd/gapir/cc/main.cpp
```diff
@@ -144,7 +145,28 @@ void android_main(struct android_app* app) {
 
     // Note if you want to create a disk cache create it under:
     // app->activity->internalDataPath
-    listenConnections(std::move(conn), nullptr, nullptr, Connection::NO_TIMEOUT, &memoryManager, crashHandler);
+    std::thread listening_thread([&]() {
+      listenConnections(conn.get(), nullptr, nullptr, Connection::NO_TIMEOUT,
+                        &memoryManager, crashHandler);
+    });
+    while (true) {
```
ben-clayton: Shouldn't this loop exit if `listenConnections` returns?

gapidapk/deviceinfo.go
```diff
@@ -25,15 +28,117 @@ import (
...
+		if se = d.StartService(ctx, *action); se != nil {
+			continue
+		}
+
+		stop_checking_port := make(chan bool)
+		timer := time.AfterFunc(portListeningTimeoutSeconds, func() {
+			stop_checking_port <- true
+			close(stop_checking_port)
+		})
+		func() {
```
ben-clayton: What is the point of this func?

gapidapk/deviceinfo.go
```diff
@@ -25,15 +28,117 @@ import (
...
+	var le error
+	var listening bool
+
+	action := apk.ServiceActions.FindByName(sendDevInfoAction, sendDevInfoService)
+	if action == nil {
+		return log.Err(ctx, nil, "Service intent was not found")
+	}
+
+	// Try to start service.
+	for i := 0; i < startServiceAttempts; i++ {
```
ben-clayton: Consider using `task.Retry`

gapidapk/deviceinfo.go
```diff
@@ -25,15 +28,117 @@ import (
...
+	if le != nil {
+		return le
+	}
+	return log.Errf(ctx, nil, "Run out of attempts: %v", startServiceAttempts)
+}
+
+// Checks the existence of VkGraphicsSpyLayer in the debug.vulkan.layers system
+// property. If found, strip the layer, otherwise keep the property unchanged.
+// Returns a property value recover callback and error.
+func stripVkGraphicsSpyLayer(ctx context.Context, d adb.Device) (func(), error) {
```
ben-clayton: So if I understand this all correctly, this is all attempting to avoid a race (or failed cleanup) of the `debug.vulkan.layers` property, set my tracing, that when left set screws up the device info fetching?

It would appear at first glance that we've added a plaster to try and unset the layer here, but we're still going to have race issues if you try taking a trace at the same time as this logic is executed.

It feels to me that we probably need something more global-lock like. Perhaps we could use the device's properties for a lock on `debug.vulkan.layers`?

gapidapk/gapidapk.go
```diff
@@ -59,88 +59,102 @@ func EnsureInstalled(ctx context.Context, d adb.Device, abi *device.ABI) (*APK,
...
-	}
-
-	ctx = log.V{"abi": abi.Name}.Bind(ctx)
-
-	// Was this recently checked?
-	reg, checkKey := bind.GetRegistry(ctx), lastInstallCheckKey{abi}
-	if res, ok := reg.DeviceProperty(ctx, d, checkKey).(lastInstallCheckRes); ok {
-		if time.Since(res.time) < checkFrequency {
-			return res.apk, nil
+	try_abi := func(ctx context.Context, abi *device.ABI) (*APK, error) {
```
ben-clayton: Move to a standard package private function?

cmd/gapir/cc/main.cpp
Qining: I think we should not. The infinite loop should only terminate when requested by the system. Actually an early return here does not destroy or ends the activity.

I've also tested with early return, the result is, if the activity was ever been started, any 'startservice' calls from adb shell will cause ANR.

gapidapk/deviceinfo.go
Qining: Right. I'm just trying to make sure device info is gathered when 'debug.vulkan.layers' does not contains VkGraphicsSpy. And we still have a race condition.

One problem is that, when we test with Robot tester, a 'gapit trace' and 'gapit devices query' are done with different 'gapis' process. Even if we add a lock in the global level, we still have multiple copies of locks.

To totally avoid race for robot tester, it seems we will need a lock at robot level, generally that means, only one Vulkan operation (trace/replay/report) can be processed at a time, which is not good (two devices can not be used at the same time even if their workload are totally independent from each other).

I think probably we can use 'temporary files' with busy loop as 'locks' for this?

ben-clayton: I'm not sure if you saw or understood my suggestion for using a custom `setprop` variable on the device for the lock. This would mean the lock persists beyond gapit, gapis or robot - which may be slightly dangerous if one of these processes crash with the lock held. 
If we can use a property for a lock, perhaps it can have a periodically updated timestamp so that times greater than some threshold are ignored. 
Of course you could also use a temporary file, but properties have the nice property (pun not intended) of being wiped between reboots.

Qining: Just uploaded 3a5af05 as a preparation of using 'property' for 'lock'. (we don't want 'trace' to block 'deviceinfo query', as 'trace' takes a long time, but we can have 'deviceinfo query' to block 'trace').

Trace is responsible to set and unset 'debug.vulkan.layers'
DeviceInfo query is responsible to set and unset 'devinfo-querying'

Trace can not happen if a 'devinfo-querying' property is set.
DeviceInfo query can not happen if 'debug.vulkan.layers' is set.

But I'm thinking using a system property probably won't resolve the whole problem.
For robot tester, there will never be two trace happening at the same time, but multiple deviceinfo query may happen at the same time (this is because `gapit devices` and its underlying service traverse all the connected devices, and robot tester will spawn individual gapis for each test).

So, A pending 'trace' (which is waiting for devinfo query X may still get a race condition with a 'deviceinfo query' Y, which is initiated after X.

Is there a good solution for this problem, I thought I need an 'atomic getAndSetSystemProperty' thing...

Qining:
Updated with a `flock` package, which implements a inter-process lock by creating/removing files. This can resolve the race condition between device info query and Vulkan trace.

I've tested with manually cancel the context when a FLock is locked, and the underlying file is removed gracefully. But I'm not very sure if those flock files will always be removed given a crash happens. I'm thinking erasing all those flock files before any normal uses of gapid, except when GAPID is initiated from Robotester.

Another way is add a maximum valid time threshold, files created before such a length of time from now will be discarded.

@AWoloszyn Any suggestions?

AWoloszyn:
We could choose to only use the file locks from robotester, then at least if its not perfect only robot will be affected.

(Command-line option to gapis to enable/disable file locks).

I would try inserting some panics () in some places where you expect it might fail, and see if the file gets cleared.

Also you can try to kill-9 from the outside.

Qining:
Dropped as #1449 contains this CL.

PR #1512, opened 2017-12-16
Moar bazel stuff - version and gapic

pmuetschard:
- Carry over defines for swapchain.so from cmake
- Implement the versioning in bazel - we'll edit version.bzl instead of version.cmake
- Add a comment to mm.bzl, pointing out its hackyness
- Fix up the copy actions, since they would corrupt binary files
- Get gapic building
  - GRPC java codegen rule
  - Don't use GeneratedMessage, it's now GeneratedMessageV3 and may soon be something even crazier
  - "Workaround" some error prone warnings
  - Fix a NPE in the crash reporting

I actually managed to run a bazel built gapic, point it at a bazel built gapis & gapir and open a trace. Woot


version.bzl
```diff
@@ -0,0 +1,51 @@
...
+            "@GAPID_BUILD_NUMBER@": ctx.var.get("GAPID_BUILD_NUMBER"),
+            "@GAPID_BUILD_SHA@": ctx.var.get("GAPID_BUILD_SHA"),
+        }
+    )
+
+gapid_version = rule(
+    implementation=_gapid_version,
+    attrs = {
+        "template": attr.label(mandatory = True, allow_files = True, single_file = True),
+        "out": attr.output(mandatory = True),
```
ianthehat: Implicit outputs are bad and wrong...
Hidden outputs are the only clean answer.
If you want to submit this and then I will show you what I mean in a separate PR I am happy to do that.

core/app/BUILD.bazel
```diff
@@ -29,3 +31,9 @@ go_library(
         "@com_github_pkg_errors//:go_default_library",
     ],
 )
+
+gapid_version(
+    name = "version",
+    out = "default_version.go",
+    template = "default_version.go.in",
```
ianthehat: It would probably be cleaner to do this with x_def stamping.

version.bzl
pmuetschard: Please do.

core/app/BUILD.bazel
ianthehat: The problem with this approach is that it modifies a source at the bottom level, which causes a cascade of rebuilds.
The stamping approach just causes a re-link at the top level.

pmuetschard: I'm OK with that.

PR #1528, opened 2018-01-03
Reworked memory functions to rely less on reflection.

AWoloszyn:
This re-orders the methods so that we can decode memory much quicker.
It brings decoding time down for Vulkan traces by ~75%.

When we do have to do reflection, optimize it such that the interfaces
that must be matched are smaller.

### DIFF
--- a/gapis/api/templates/api.go.tmpl
+++ b/gapis/api/templates/api.go.tmpl
@@ -232,6 +232,9 @@
   // IsNullptr returns true if the address is 0 and the pool is ϟmem.ApplicationPool.
   func (p {{$ptr_ty}}) IsNullptr() bool { return p.addr == 0 && p.pool == ϟmem.ApplicationPool }
 
+  // APointer implements the ReflectPointer interface
+  func (p {{$ptr_ty}}) APointer() { return }
+
   // Address returns the pointer's memory address.
   func (p {{$ptr_ty}}) Address() uint64 { return p.addr }
 
--- a/gapis/memory/pointer.go
+++ b/gapis/memory/pointer.go
@@ -31,6 +31,8 @@ const bits32 = uint64(1) << 32
 
 // Pointer is the type representing a memory pointer.
 type Pointer interface {
+	ReflectPointer
+
 	// IsNullptr returns true if the address is 0 and the pool is memory.ApplicationPool.
 	IsNullptr() bool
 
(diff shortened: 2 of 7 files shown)


gapis/memory/pointer.go
```diff
@@ -52,6 +52,18 @@ type Pointer interface {
...
+	// All pointers by definition also implement ReflectPointer
+	APointer() bool
+}
+
+// This is a helper interface, if you want your pointer to be reflected
+// then it must ALSO implement this interface. Since reflection is slow
+// having a much smaller interface to check is significantly better
+// We name this APointer since reflect.Implements checks functions
+// in alphabetical order, meaning this should get hit first (or close to).
+type ReflectPointer interface {
```
ben-clayton: So, if I understand this correctly: You've added `APointer` to `Pointer` so that `reflect.Implements` can early out quicker for non-pointer types?

Given that I see no direct use of `ReflectPointer` (aside from the interface check), why do you need it? Implementers of `Pointer` are required to implement `ReflectPointer` anyway.

AWoloszyn: Unfortunately not quite.

https://github.com/golang/go/blob/master/src/reflect/type.go#L1531

Implements: loops over every method in V (the type) type to see if it matches T (the interface). Even though both lists are in alphabetical order, and the code relies on this fact, there is no short-circuit for when V does not contain a method that is alphabetically <= the next method on T, and so all methods on V will get checked even though there is no possibility of a match

The only short-circuit through that code is when V contains ALL of the methods in T, so if we check against ReflectedPointer, if we DO match, ReflectedPointer, (which is likely the first function match on matching types), then we stop checking.

gapis/api/templates/api.go.tmpl
```diff
@@ -197,6 +197,7 @@
   {{$el_is_ptr  := IsPointer ($p.To | Underlying | Unpack) }}
 
   var _ = ϟmem.Pointer({{$ptr_ty}}{})
+  var _ = ϟmem.ReflectPointer({{$ptr_ty}}{})
```
ben-clayton: Not needed, it is subset of the check above.

gapis/memory/pointer.go
```diff
@@ -52,6 +52,18 @@ type Pointer interface {
 	// ISlice returns a new Slice of elements based from this pointer using
 	// start and end indices.
 	ISlice(start, end uint64, m *device.MemoryLayout) Slice
+
+	// All pointers by definition also implement ReflectPointer
+	APointer() bool
```
ben-clayton: You can just nest the interface with:
```
ReflectPointer
```

gapis/memory/pointer.go
```diff
@@ -52,6 +52,18 @@ type Pointer interface {
...
+}
+
+// This is a helper interface, if you want your pointer to be reflected
+// then it must ALSO implement this interface. Since reflection is slow
+// having a much smaller interface to check is significantly better
+// We name this APointer since reflect.Implements checks functions
+// in alphabetical order, meaning this should get hit first (or close to).
+type ReflectPointer interface {
+	APointer() bool
```
ben-clayton: No need for the bool return type.

gapis/memory/pointer.go
ben-clayton: Gotcha.

gapis/memory/pointer.go
AWoloszyn: Done

gapis/memory/pointer.go
AWoloszyn: Done

PR #1530, opened 2018-01-03
Adds OBB file support to Robot

baldwinn860:
Now specifying `-obb="<obb-path>"` to `do robot upload subject` will upload and associate an obb file with a subject for use during trace. In that trace, gapit will push the OBB file after installing the APK and grant the APK read/write privilege to external storage such that it can access the OBB file. When the trace finishes it then removes the OBB file. Also added the ability to pull the OBB file along with the APK in `do pullapk`.

### DIFF
--- a/core/os/android/adb/commands.go
+++ b/core/os/android/adb/commands.go
@@ -99,6 +99,15 @@ func (b *binding) InstallAPK(ctx context.Context, path string, reinstall bool, g
 	return b.Command("install", args...).Run(ctx)
 }
 
+func (b *binding) PushOBB(ctx context.Context, package_name string, version_code int32, obb_path string) error {
+	storage, err := b.Shell("echo", "$EXTERNAL_STORAGE").Call(ctx)
+	if err != nil {
+		return err
+	}
+	obb_storage_path := fmt.Sprintf("%s/obb/%s/main.%d.%[2]s.obb", storage, package_name, version_code)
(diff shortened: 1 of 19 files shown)


core/os/android/adb/commands.go
```diff
@@ -99,6 +99,36 @@ func (b *binding) InstallAPK(ctx context.Context, path string, reinstall bool, g
 	return b.Command("install", args...).Run(ctx)
 }
 
+// PushOBB places a OBB file to the correct location for an APK to access.
+func (b *binding) PushOBB(ctx context.Context, package_name string, version_code int32, obb_path string) error {
```
ben-clayton: Should this be a method of `adb.InstalledPackage` ?

Don't use underscores in go variable names. Please use golint and govet as part of your development environment.

test/robot/subject/local.go
```diff
@@ -91,7 +91,7 @@ func (s *local) Search(ctx context.Context, query *search.Query, handler Handler
 }
 
 // Add implements Subjects.Add
-func (s *local) Add(ctx context.Context, id string, hints *Hints) (*Subject, bool, error) {
+func (s *local) Add(ctx context.Context, id string, obbId string, hints *Hints) (*Subject, bool, error) {
```
ben-clayton: nit: `obbID`

PR #1532, opened 2018-01-05
LLVM Compiler

ben-clayton:


core/codegen/builder.go
```diff
@@ -0,0 +1,246 @@
...
+		defaultBlock = b.m.ctx.AddBasicBlock(b.function.llvm, "switch_case_default")
+		tests = append(tests, defaultBlock)
+	}
+
+	exit := b.m.ctx.AddBasicBlock(b.function.llvm, "end_switch")
+
+	b.llvm.CreateBr(tests[0])
+
+	for i, c := range cases {
+		i, c := i, c
```
AWoloszyn: Looks weird to me, is this just creating a local copy, so that we don't modify the originals?

core/codegen/types.go
```diff
@@ -0,0 +1,445 @@
...
+	return ok
+}
+
+// Field is a single field in a struct.
+type Field struct {
+	Name string
+	Type Type
+}
+
+// Struct creates a new, empty struct type.
```
AWoloszyn: I dont think this is accurate any more

core/codegen/builder.go
baldwinn860: Its the idiomatic way (unfortunately) to capture range iterators for lambdas and goroutines and the like. `i` and `c` here are more like iterator objects in that they get constructed once for the entirety of the loop, so if you use them in a deferred context (like the function below) you would only get the final value stored inside them. So this way instead you shadow the variable names with true local variables that get reference counted so you can actually use the values of the range within the function.

ben-clayton: It's idiomatic Go for loops containing closures that are executed after the for loop, as you'll get the iterator *pointer* captured, not the value. By creating another shadowed copy with the same name, the value is captured for each loop.

All that said, nothing in this block is executed after the for loop, so this line is actually pointless. Good spot.

core/codegen/types.go
ben-clayton: True, will fix.

Issue #1510, opened 2017-12-15
Use either a pNext or an extension to control the virtual swapchain


AWoloszyn:
We are currently reaching through the opaque pointer of virtual swapchain in gapir. 
We need to fix this so that we control this with either an extension (and a new entrypoint) or a pNext pointer on the swapchain.

Having the dependency is bad, since ABI issues can happen across this boundary.

pmuetschard:
The offending code from `vulkan_gfx_api_extras.tmpl` is:
```
  bool Vulkan::toggleVirtualSwapchainReturnAcquiredImage(Stack* stack) {
    auto pSwapchain = stack->pop<VkSwapchainKHR*>();
    if (stack->isValid()) {
      GAPID_DEBUG("toggleVirtualSwapchainReturnAcquiredImage(%p)", pSwapchain);
      auto virtual_swapchain = reinterpret_cast<swapchain::VirtualSwapchain*>(*pSwapchain);
      virtual_swapchain->SetAlwaysGetAcquiredImage(true);
      return true;
    } else {
      GAPID_WARNING("Error during calling function toggleVirtualSwapchainReturnAcquiredImage");
      return false;
   }
  }
```
`SetAlwaysGetAcquiredImage` simply sets a boolean flag in the swap chain struct:
```
  void SetAlwaysGetAcquiredImage(bool always_get_acquired_image) {
    always_get_acquired_image_ = always_get_acquired_image;
  }
```

AWoloszyn:
Fixed in above commit.

PR #1533, opened 2018-01-05
Updated the virtual swapchain to remove the weird dependency.

AWoloszyn:


AWoloszyn:
Fixes https://github.com/google/gapid/issues/1510

PR #1535, opened 2018-01-08
Remove the build dependency onto opengl32.dll from deviceinfo.

pmuetschard:
This helps in building the windows libgapii.dll with bazel.


pmuetschard:
PTAL. I've made the change to use core::GetGlesProcAddress.

AWoloszyn:
Looks good, but can you also remove opengl32 from the list of cmake dependencies as well, this should help us make sure everything still works as expected.

AWoloszyn:
This looks fine, the build error looks unrelated to this change, but you should probably take a look just in case.

pmuetschard:
The build error is unrelated, but worrisome nonetheless 😨 Looks like the Kokoro image is switching to Java 9, and gradle does not seem to be ready for it. Let's see what happens....

PR #1541, opened 2018-01-11
Close Flock files if cannot acquire the lock

Qining:
When acquiring the flock, the mutex should close the opened file if the lock is not acquired.
When removing lock files, records files that cannot be removed. On Windows, we cannot remove a file which is currently opened.

### DIFF
--- a/core/os/flock/flock.go
+++ b/core/os/flock/flock.go
@@ -15,6 +15,7 @@
 package flock
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"sync"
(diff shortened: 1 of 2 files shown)


core/os/flock/flock.go
```diff
@@ -173,15 +176,33 @@ func (m *mutex) unlock() error {
...
 	}
 	if err != nil {
 		return err
 	}
-	if info.IsDir() {
-		return os.RemoveAll(dir)
+	if !dirInfo.IsDir() {
+		return fmt.Errorf("%v is not a directory", dir)
+	}
+	err = os.RemoveAll(dir)
```
ben-clayton: Überdoobernit:
In the case of short lines like this, it's fine to combine the assignment and if:

```go
if err := os.RemoveAll(dir); err == nil {
    return nil
}
```

core/os/flock/flock.go
```diff
@@ -173,15 +176,33 @@ func (m *mutex) unlock() error {
...
+	if err == nil {
+		return nil
+	}
+	cantRemove := []string{}
+	filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
+		if !info.Mode().IsRegular() {
+			return nil
+		}
+		err = os.Remove(path)
```
ben-clayton: And here.

PR #1543, opened 2018-01-12
Fix a bug in Footprint Builder for VkFlushMappedMemoryRanges

Qining:
In case of the flush target is a coherent memory, the truly overwritten
memory is not the range specified in the memory ranges in the arguments,
because the trace side only capture the memory observation for the truly
modified page. Using the memory ranges in the argument to update the
footprint may cause incorrect overwrite in the unchanged data specified
in the memory ranges.

Also added some api commands not covered in the footprint builder.


Qining:
Locally tested, build on Mac OS is actually fine.

PR #1544, opened 2018-01-12
Fix crash in ApplyLinkProgramExtra

y-novikov:
Access to i.ActiveResources.DefaultUniformBlock used to crash
when glLinkProgram failed, because i.ActiveResources was null.

### DIFF
--- a/gapis/api/gles/api/programs_and_shaders.api
+++ b/gapis/api/gles/api/programs_and_shaders.api
@@ -140,25 +140,27 @@ sub void ApplyLinkProgramExtra(ref!Program p, ref!LinkProgramExtra i) {
 
     // Set uniform values to default (zeroed)
     p.UniformLocations = null
-    for _, uniformIndex, uniform in i.ActiveResources.DefaultUniformBlock {
-      elementInfo := GetUniformTypeInfo(uniform.Type)
-      elementSize := elementInfo.primitiveSize * elementInfo.vectorSize * elementInfo.vectorCount
-      values := make!u8(elementSize * as!u32(uniform.ArraySize))
-      // Each element of the array is assigned a location by compiler.
-      // The location acts as a handle. Locations do not have to be consecutive.
-      for _, elementIndex, loc in uniform.Locations {
-        if (loc != -1) {
-          assert(elementIndex < as!u32(uniform.ArraySize))
-          offset := elementIndex * elementSize
-          p.UniformLocations[as!UniformLocation(loc)] = Uniform(
-            Type:          uniform.Type,
-            Value:         values[offset:offset + elementSize],
-            Values:        values[offset:len(values)],
-            UniformIndex:  uniformIndex
-          )
+    if i.LinkStatus == GL_TRUE {
+      for _, uniformIndex, uniform in i.ActiveResources.DefaultUniformBlock {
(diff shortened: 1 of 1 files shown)


y-novikov:
I'm not sure code flow is even supposed to reach ApplyLinkProgramExtra in the case when glLinkProgram fails.
The check itself should be correct anyway.
Unless you want to change the flow in a way that ApplyLinkProgramExtra is not reached, then it will be better to change the if() to assert().

gapis/api/gles/api/programs_and_shaders.api
```diff
@@ -140,25 +140,27 @@ sub void ApplyLinkProgramExtra(ref!Program p, ref!LinkProgramExtra i) {
...
-        if (loc != -1) {
-          assert(elementIndex < as!u32(uniform.ArraySize))
-          offset := elementIndex * elementSize
-          p.UniformLocations[as!UniformLocation(loc)] = Uniform(
-            Type:          uniform.Type,
-            Value:         values[offset:offset + elementSize],
-            Values:        values[offset:len(values)],
-            UniformIndex:  uniformIndex
-          )
+    if i.ActiveResources != null {
```
ben-clayton: It might be slightly more consistent to use `i.LinkStatus == GL_TRUE` as the conditional here as it then mirrors [the logic in the interceptor here](https://github.com/google/gapid/blob/master/gapii/cc/gles_extras.cpp#L382).

Feel free to ignore.

ben-clayton:
Hi @y-novikov,

Thank you very much for your PR.

> I'm not sure code flow is even supposed to reach ApplyLinkProgramExtra in the case when glLinkProgram fails.

The same API logic is executed for both interception time and for replay, so the command:
```
cmd void glLinkProgram(ProgramId program) {
  p := GetProgramOrError(program)

  ctx := GetContext()
  ApplyLinkProgramExtra(p, GetLinkProgramExtra(ctx, p, null))
}
```

Will call `GetLinkProgramExtra` for interception and replay before passing that into `ApplyLinkProgramExtra`. 
`GetLinkProgramExtra` however is implemented differently for interception and replay.
 * For interception it is implemented in [`gles_extras.cpp`](https://github.com/google/gapid/blob/master/gapii/cc/gles_extras.cpp#L354), which will only fill in the bits of the `LinkProgramExtra` that it can get back from the driver. In this situation, the link failed so we can't get uniform information, hence the null pointer.
 * For replay, we simply return [the data encoded by the interceptor here](https://github.com/google/gapid/blob/master/gapis/api/gles/extras.go#L109).

The logic flow has no way to bail in either case. Just because the intercepted command set a GL error state, that doesn't affect the way we attempt to replay the command stream.

I've considered your fix with a few other approaches, and aside from the single review comment, I think this is good. I'm slightly concerned that there is other code knocking around expecting `ActiveResources` to be non-null, but if this is now working for you, it's a major improvement.

Thank you again,
Ben

y-novikov:
I've changed it to "i.LinkStatus == GL_TRUE".
Could you please merge it?
I don't have write access, I think.

ben-clayton:
Thank you for the change. I've flagged the PR for automated building. Assuming everything is green (should be!) I'll merge ASAP.

Thank you again.

PR #1555, opened 2018-01-17
Bazel build now working on Linux and OSX

pmuetschard:
```
$ bazel build -c opt --strip always install 
```


tools/build/rules/android.bzl
```diff
@@ -19,4 +19,50 @@ def android_native(name, deps=[], **kwargs):
         **kwargs
     )
 
+def _android_ndk_file_impl(ctx):
+    outs = depset()
 
+    for f in ctx.attr.files:
+        out = ctx.new_file(f.rpartition("/")[2])
```
ianthehat: Consider using [skylib](/bazelbuild/bazel-skylib) for this kind of stuff, paths.basename is basically this, but slightly more future proof and easier to read

pmuetschard: Thanks for the pointer. Will look to see what we can use from there.

PR #1571, opened 2018-01-25
Switch robot web client to MVC model, add defVals.

baldwinn860:
This change is largely refactoring and juggling code into a framework
that can be ported more simply over to javascript. Following a MVC model
the page struct and some other methods have been moved and split between
a view and controller struct that separate the concerns. The controller
handles setting values in response to input, the view handles displaying
those values in the selecters and the grid.

defVal() is added to the dimension struct in order to have some method
of defaulting values when first entering the page.

### DIFF
--- a/test/robot/web/client/client.go
+++ b/test/robot/web/client/client.go
@@ -64,13 +64,13 @@ func (e enum) indexOf(s Item) int {
 }
 
 type dimension struct {
-	name       string
-	enumData   enum
-	valueOf    func(*task) Item
-	itemMap    map[string]Item
-	enumSrc    func() enum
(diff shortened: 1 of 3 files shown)


ben-clayton:
Might want to fix the build failure though:
```ninja: error: '/tmpfs/src/github/src/github.com/google/gapid/test/robot/web/www/grid/grid.js.map', needed by '/tmpfs/src/github/src/github.com/google/gapid/test/robot/web/web_embed.go', missing and no known rule to make it```

baldwinn860:
Yeah, I realized I let those files slip in to my cl, they were there for debugging, removed.

PR #1579, opened 2018-02-01
Robot: fail on output to stderr from tasks.

baldwinn860:
An error inside gapit no longer results in a non-zero return code to
robot, this means that the only way robot detects failures is if the
calls to run gapit itself fails. This change now separates capturing
stdout from stderr and any non-empty stderr buffers will cause robot to
report a failure. This fixes some regression in robot causing false
positives.

### DIFF
--- a/test/robot/replay/client.go
+++ b/test/robot/replay/client.go
@@ -15,8 +15,10 @@
 package replay
 
 import (
+	"bytes"
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/google/gapid/core/app/crash"
@@ -35,6 +37,9 @@ import (
 
 const (
 	replayTimeout = time.Hour
+	// this string is returned when GAPIT fails to connect to the GAPIS, particularly due to ETXTBSY
+	// look at https://github.com/google/gapid/pull/933 for more information
+	retryString = "Failed to connect to the GAPIS server"
 )
 
 type client struct {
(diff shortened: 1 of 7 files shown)


baldwinn860:
A much simpler alternative would be to change the final error log in `core/app/run.go: func Run(main task.Task)` to `log.F(ctx, true, "Main failed\nError: %v", err)` which would make it so errors in subcommands that reach the top of the stack will cause a non-zero exit code again. I wasn't sure if this would break other behavior so I tried to localize the affect to robot instead.

test/robot/replay/client.go
```diff
@@ -162,19 +164,21 @@ func doReplay(ctx context.Context, action string, in *Input, store *stash.Client
...
-		if err := worker.NeedsRetry(callErr.Error()); err != nil {
+	if err := cmd.Capture(outBuf, errBuf).Run(ctx); err != nil {
+		if err := worker.NeedsRetry(err.Error()); err != nil {
 			return nil, err
 		}
-		outputObj.CallError = callErr.Error()
+		outputObj.Err = err.Error()
 	}
-	output = fmt.Sprintf("%s\n\n%s", cmd, output)
+	outputObj.Err += strings.TrimSpace(errBuf.String())
```
ben-clayton: Seems like this will cause the strings to join back-to-back with no newlines?

Maybe use a `strings.Join(errs, "\n")`, or add some sort of separator here?

test/robot/replay/client.go
```diff
@@ -162,19 +164,21 @@ func doReplay(ctx context.Context, action string, in *Input, store *stash.Client
...
+	if err := worker.NeedsRetry(outputObj.Err, "Failed to connect to the GAPIS server"); err != nil {
```
ben-clayton: I'd move this magic string up to the top of the file, put it in a const and document why you need it. If there's an underlying bug here that need to be fixed, please create one and link to it in the comment.

PR #1594, opened 2018-02-06
Fix templates.

AWoloszyn:
There was a merge issue with the new MEC stuff and Bazel.
Bazel was missing some template generation that was necessary to
properly read a trace that was created with mid-execution capture.


pmuetschard:
Thanks. Sorry, I forgot to properly check all the CMake changes when rebasing.

PR #1597, opened 2018-02-07
Add core/codegen package.

ben-clayton:
Package codegen provides an interface for code generation using LLVM.

codegen provides a higher-level and simpler interface for generating AOT or JIT code than using the LLVM go bindings directly.

Co-authored-by: AWoloszyn <awoloszyn@google.com>

### DIFF
--- a/core/codegen/arithmetic.go
+++ b/core/codegen/arithmetic.go
@@ -1,0 +1,345 @@
+// Copyright (C) 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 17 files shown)


core/codegen/arithmetic.go
```diff
@@ -0,0 +1,345 @@
...
+	return value
+}
+
+// Not returns !x. The type of x must be Bool.
+func (b *Builder) Not(x *Value) *Value {
+	assertTypesEqual(x.ty, b.m.Types.Bool)
+	return b.val(b.m.Types.Bool, b.llvm.CreateNot(x.llvm, ""))
+}
+
+// Not returns ~x.
```
AWoloszyn: I think this was my fault, but this should be "Invert returns ~x"

core/codegen/output.go
```diff
@@ -0,0 +1,156 @@
...
+	"unsafe"
+
+	"llvm/bindings/go/llvm"
+)
+
+func init() {
+	llvm.InitializeAllTargetInfos()
+	llvm.InitializeAllTargets()
+	llvm.InitializeAllTargetMCs()
+	// llvm.InitializeAllAsmParsers()
```
AWoloszyn: Do we still need this here?

core/codegen/arithmetic.go
ben-clayton: Done

core/codegen/output.go
ben-clayton: Done

PR #1599, opened 2018-02-07
Removed -lGL from deviceinfo linker options.

AWoloszyn:
We no longer statically need -lGL to build, we link to it at
runtime.


AWoloszyn:
The windows build failure is unrelated.

PR #1606, opened 2018-02-09
Adds a runfiles core/app/layout

pmuetschard:
Allows developers to now run `bazel run cmd/gapi[st]`

### DIFF
--- a/core/app/layout/layout.go
+++ b/core/app/layout/layout.go
@@ -15,8 +15,12 @@
 package layout
 
 import (
+	"bufio"
 	"context"
+	"os"
+	"regexp"
 	"runtime"
+	"strings"
 
 	"github.com/google/gapid/core/fault"
 	"github.com/google/gapid/core/log"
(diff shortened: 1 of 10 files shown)


core/app/layout/layout.go
```diff
@@ -198,6 +208,108 @@ func (l binLayout) Json(ctx context.Context, lib LibraryType) (file.Path, error)
...
+}
+
+// runfilesLayout is a layout that uses the bazel generated runfiles manifest
+// to find the various dependencies.
+type runfilesLayout struct {
+	manifest string
+	mapping  map[string]string
+}
+
+var gapiXRegex = regexp.MustCompile("gapid/cmd/(gapi[st])/[^/]+/" + withExecutablePlatformSuffix("gapi[st]"))
```
ben-clayton: Checked this works on windows?

core/app/layout/pkgdata.go
```diff
@@ -70,12 +75,28 @@ func layout(ctx context.Context) (out FileLayout) {
...
+
+	if path := file.Abs(file.ExecutablePath().System() + ".runfiles_manifest"); path.Exists() {
+		if layout, err := RunfilesLayout(path); err == nil {
+			return layout
 		}
 	}
+
+	// We're possibly dealing with a sparse-build, as used by robot.
+	// gapis always has to exist.
+	// TODO: this is kind of dumb, since gapis will always find itself.
```
ben-clayton: Assuming this is the `gapis` process... What about `gapit`?

core/app/layout/layout.go
pmuetschard: yup

core/app/layout/pkgdata.go
pmuetschard: yes, for gapit it sort of makes sense, because it might find gapis. Gapis will, of course always find itself in its own directory. That is why I moved it to be last in priority.

PR #1607, opened 2018-02-09
Allow recursive functions in the API file.

AWoloszyn:
This only allows them in the case where there is no fence
in the function.


gapil/resolver/api.go
```diff
@@ -164,7 +164,8 @@ func resolve(rv *resolver) {
 		removeDeadCode(rv, f.Block)
 	}
 	for _, f := range rv.api.Functions {
-		resolveFenceOrder(rv, f, visitedFuncs{})
+		recursive := false
+		resolveFenceOrder(rv, f, visitedFuncs{}, false, &recursive)
```
ben-clayton: Maybe just add `Recursive` field to `semantic.Function` ?

gapil/resolver/flow.go
```diff
@@ -20,13 +20,18 @@ type visitedFuncs map[*semantic.Function]bool
...
 // fence.
-func resolveFenceOrder(rv *resolver, f *semantic.Function, visitedFuncs visitedFuncs) semantic.LogicalOrder {
+func resolveFenceOrder(rv *resolver, f *semantic.Function, visitedFuncs visitedFuncs, preFence bool, recursive *bool) semantic.LogicalOrder {
 	if f.Order.Resolved() {
 		return f.Order // Already resolved
 	}
 	if visitedFuncs[f] {
-		rv.errorf(f, "Cyclic dependeny found")
-		return 0
+		if !preFence {
```
ben-clayton: I'm really struggling to get my head wrapped around this. Why do you need this error here, won't we catch the fence in recursive functions with the `"Fence in recursive function"` error below?

gapil/resolver/flow.go
```diff
@@ -54,12 +59,18 @@ func resolveFenceOrder(rv *resolver, f *semantic.Function, visitedFuncs visitedF
...
+		rv.errorf(f, "Fence in recursive function")
+		return 0
+	}
 	return f.Order
 }
 
 type fenceTracker struct {
 	resolver     *resolver
 	explicit     bool // explcitly declared fence found.
+	implicit     bool // implicitly declared fence found.
```
ben-clayton: Again, this confuses me. The rules on the implicit fence are done in `resolveFenceOrder` after the call to `analyse` - I'm confused how you're trying to determine the implicit fence status before we know all the facts.

gapil/resolver/flow.go
AWoloszyn: That is true, we can remove that clause, and just handle it elsewhere.

gapil/resolver/api.go
AWoloszyn: Done

gapil/resolver/flow.go
AWoloszyn: Refactored, and fixed now. Tests have also been added

gapil/resolver/api.go
```diff
@@ -172,6 +172,17 @@ func resolve(rv *resolver) {
 	sort.Sort(functionsByName(rv.api.Externs))
 	sort.Sort(slicesByName(rv.api.Slices))
 	sort.Sort(mapsByName(rv.api.Maps))
+
+	for _, f := range rv.api.Functions {
+		if f.Recursive && (f.Order&(semantic.Pre|semantic.Post) != 0) {
```
ben-clayton: Can simplify to:

```go
if f.Recursive && (f.Order.Pre() || f.Order.Post()) {
```

PR #1608, opened 2018-02-09
For android, fail the compilation if we have any outstanding symbols.

### LINKED ISSUE #1605: Link will succeed with unresolved symbols
For Android and macOS, the bazel build will succeed when symbols are unresolved.

For Android this is tragic - if you forget to declare a function, the library will just fail to load. No errors, no reasons given.

I think we need to put a `-Wl,--unresolved-symbols`... somewhere.

AWoloszyn:
This causes android to fail silently, making it very hard to diagnose
issues.

Fixes #1605


ben-clayton:
macOS is still an issue so please reopen the bug after automatic closing on merge.

PR #1609, opened 2018-02-09
Cleanup the vulkan api to reduce external code.

AWoloszyn:
This has now been tested. 
It depends on https://github.com/google/gapid/pull/1607, so ignore that commit here, I will commit 1607 by itself, and then just the one commit from this actual change.


gapis/api/gvr/gvr.go
```diff
@@ -39,6 +39,8 @@ func (s *State) Root(ctx context.Context, p *path.State) (path.Node, error) {
 
 func (*State) SetupInitialState(ctx context.Context) {}
 
+func (s *State) InitializeCustomState() {}
```
ben-clayton: Do we need this and `SetupInitialState`?

AWoloszyn: Unfortunately yes. We need SetupInitialState to take an initial state block, and make sure that it is consistent. 

InitializeCustomState takes an empty state block and sets up all of the objects within.

PR #1610, opened 2018-02-09
Fixes for Robot on Bazel.

baldwinn860:
This change enables embed to use a relative root path in order to
correctly configure filepaths when embedding. Also fixes the globbing in
the robot web build file to be recursive. Some fixes to the replay
client and the web client as well to get better logging information.

NOTE: this does not yet transpile grid.html from gopherjs as the plan is 
to remove the dependency. If you have golang 1.9 you can run 
`go get github.com/gopherjs/gopherjs`, if you have golang 1.10 you 
must manually download and build the [`go1.10` branch of gopherjs](https://github.com/gopherjs/gopherjs/tree/go1.10).
Then run the following command at the project root:
`gopherjs build --minify --output ./test/robot/web/www/grid/grid.js ./test/robot/web/client`

### DIFF
--- a/cmd/embed/main.go
+++ b/cmd/embed/main.go
@@ -111,6 +111,7 @@ func run(ctx context.Context) error {
 			rel := ""
 			if root != "" {
 				// if Rel fails, we just fallback to basename in the loop below
+				root, err = filepath.Abs(root)
 				rel, _ = filepath.Rel(root, path)
 			} else {
 				rel = filepath.Base(path)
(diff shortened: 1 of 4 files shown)


Qining:
@baldwinn860 We need to switch to go1.10?

cmd/embed/main.go
```diff
@@ -111,6 +111,11 @@ func run(ctx context.Context) error {
 			rel := ""
 			if root != "" {
 				// if Rel fails, we just fallback to basename in the loop below
+				root, err = filepath.Abs(root)
+				if err != nil {
+					return err
+				}
+				log.I(ctx, "root is %v", root)
```
ben-clayton: Please remove

PR #1622, opened 2018-02-16
gapil: Remove labels from the language.

ben-clayton:
They were only used in one place (GLboolean).

Add a new optional numerical type specifier to the enum declaration to cover this single usage.


ben-clayton:
Build failures noted, will fix.

PR #1625, opened 2018-02-16
Optimize LLVM libs always

ben-clayton:


ben-clayton:
Submitting as I'm actually spending over half of my day waiting for LLVM to build as I switch between branches.

PR #1627, opened 2018-02-19
Add and integrate gapil/runtime

ben-clayton:
This implements the core logic and bindings for the gapil compiler, and switches gapii over to it.

Co-authored-by: AWoloszyn <awoloszyn@google.com>


core/cc/static_array.h
```diff
@@ -70,6 +77,44 @@ inline StaticArray<T, N>::StaticArray(std::initializer_list<T> l) {
...
+    }
+    return out;
+}
+
+template <typename T, int N>
+inline StaticArray<T, N> StaticArray<T, N>::create(std::initializer_list<T> l) {
+    GAPID_ASSERT(l.size() == N);
+    NEW_UNINITIALIZED(out);
+    for (int i = 0; i < N; i++) {
+        new(&out.mData[i]) T(l.begin()[i]);
```
AWoloszyn: TIL: Initializer lists are guaranteed to be contiguous in memory

gapii/cc/call_observer.h
```diff
@@ -41,10 +41,8 @@ class SpyBase;
 // CallObserver collects observation data in API function calls. It is supposed
 // to be created at the beginning of each intercepted API function call and
 // deleted at the end.
-class CallObserver {
+class CallObserver : public context_t {
```
AWoloszyn: Because the we require that this has no vtable (otherwise it would break context_t) can we mark this as 
__declspec(novtable) on windows just so that we can get errors if something goes wrong. Unfortunately something equivalent does not exist for gcc.
Not a huge priority, but might help catch future issues.

ben-clayton: Would like to do this in follow up CLs.

gapil/runtime/cc/map.h
```diff
@@ -0,0 +1,418 @@
...
+    return it->second;
+}
+
+template <typename K, typename V>
+inline const void* Map<K, V>::instance_ptr() const {
+    return ptr;
+}
+
+////////////////////////////////////////////////////////////////////////////////
+// Map<K, V>::Alocation                                                       //
```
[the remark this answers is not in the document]
ben-clayton: Done

gapii/cc/call_observer.h
AWoloszyn: Sounds good to me.

PR #1635, opened 2018-02-20
Allow DlLoader to have multiple fallback names when searching for a library

### LINKED ISSUE #1634: Look for libvulkan.so.1 on Linux as well.
We should look for libvulkan.so.1 as well as libvulkan.so on Linux as both are valid depending on how the distribution is set up.

Qining:
Fix #1634

### DIFF
--- a/core/cc/dl_loader.cpp
+++ b/core/cc/dl_loader.cpp
@@ -144,7 +144,7 @@ DL(CC _1, CC _2);
 DL(CC _1, CC _2, CC _3);
 DL(CC _1, CC _2, CC _3, CC _4);
 DL(CC _1, CC _2, CC _3, CC _4, CC _5);
-#undef DLLOADER
+#undef DL
 #undef CC
 
 }  // namespace core
--- a/core/cc/dl_loader.cpp
+++ b/core/cc/dl_loader.cpp
@@ -30,9 +30,15 @@
 namespace {
 
 // load defs
-void* load(const char* name) {
+void* load() {
+    return nullptr;
+}
+
+template<typename... ConstCharPtrs>
+void* load(const char* name, ConstCharPtrs... fallback_names) {
+    void* ret = nullptr;
 #if TARGET_OS == GAPID_OS_WINDOWS
-    return reinterpret_cast<void*>(LoadLibraryExA(name, NULL, 0));
+    ret = reinterpret_cast<void*>(LoadLibraryExA(name, NULL, 0));
 #elif TARGET_OS == GAPID_OS_OSX
     if (name == nullptr) {
         return nullptr;
@@ -50,14 +56,19 @@ void* load(const char* name) {
             remove(tmp);
         }
     }
-    return res;
+    ret = res;
 #else
-    return dlopen(name, RTLD_NOW | RTLD_LOCAL);
+    ret = dlopen(name, RTLD_NOW | RTLD_LOCAL);
 #endif // TARGET_OS
+    if (ret == nullptr) {
+        return load(fallback_names...);
+    }
+    return ret;
 }
 
-void* must_load(const char* name) {
-  void* res = load(name);
+template<typename... ConstCharPtrs>
+void* must_load(const char* name, ConstCharPtrs... fallback_names) {
+  void* res = load(name, fallback_names...);
   if (res == nullptr) {
 #if TARGET_OS == GAPID_OS_WINDOWS
     GAPID_FATAL("Can't load library %s: %d", name, GetLastError());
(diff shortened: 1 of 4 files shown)


core/cc/dl_loader.cpp
```diff
@@ -51,6 +54,16 @@ void* load(const char* name) {
...
+#elif TARGET_OS == GAPID_OS_LINUX
+    void* res = dlopen(name, RTLD_NOW | RTLD_LOCAL);
+    if (res != nullptr) {
+        return res;
+    }
+    // Try name + ".1"
+    size_t len = strlen(name) + sizeof(".1");
+    char* fallback_name = new char[len];
+    strcat(fallback_name, ".1");
+    return dlopen(name, RTLD_NOW | RTLD_LOCAL);
```
ben-clayton: ... and don't you mean `fallback_name`?

core/cc/dl_loader.cpp
```diff
@@ -51,6 +54,16 @@ void* load(const char* name) {
...
     return res;
+#elif TARGET_OS == GAPID_OS_LINUX
+    void* res = dlopen(name, RTLD_NOW | RTLD_LOCAL);
+    if (res != nullptr) {
+        return res;
+    }
+    // Try name + ".1"
+    size_t len = strlen(name) + sizeof(".1");
+    char* fallback_name = new char[len];
+    strcat(fallback_name, ".1");
```
ben-clayton: 1. You need to clear the memory.
2. You need to actually copy in name.
3. You need to free the memory.
4. You're better of using `std::stringstream`

ben-clayton:
Note: we already do this for `libGL.so`: https://github.com/google/gapid/blob/master/core/cc/linux/get_gles_proc_address.cpp#L30

Qining:
But the situation for Vulkan is different to libGL.so. At least on my machine, `libvulkan.so` does exist and it is the one should be opened I think.

The current `DlLoader` panics if the given library is not found, so for searching both `libvulkan.so` and `libvulkan.so.1` (and potential future versions), it is not capable to handle.

Qining:
Or do you mean we should always only searching for a **specific** version of Vulkan?

Qining:
Changed to just load `libvulkan.so.1`.

y-novikov:
I think it's working on Android because Android uses core/cc/android/get_vulkan_proc_address.cpp which still has libvulkan.so, which is good, since I don't have libvulkan.so.1 on my Android O Nexus 5X.

y-novikov:
I think we can't be 100% sure that all Linux drivers have libvulkan.so.1. I prefer the version that tries libvulkan.so and if it is not found looks for libvulkan.so.1.

y-novikov:
Also, you probably need to change HasVulkanLoader as well.

Qining:
@y-novikov 5ae192f should have the support to searching across multiple names in order and `HasVulkanLoader` updated

y-novikov:
Thank you, @Qining, looks good. You probably should also update the description to reference issue 1634.

core/cc/dl_loader.cpp
```diff
@@ -125,4 +136,15 @@ bool DlLoader::can_load(const char* lib_name) {
...
 
+#define CC const char*
+#define DL(...) \
+    template DlLoader::DlLoader(__VA_ARGS__)
+DL(CC _1);
+DL(CC _1, CC _2);
+DL(CC _1, CC _2, CC _3);
+DL(CC _1, CC _2, CC _3, CC _4);
+DL(CC _1, CC _2, CC _3, CC _4, CC _5);
+#undef DLLOADER
```
ben-clayton: `#undef DL` ?

core/cc/dl_loader.h
```diff
@@ -22,10 +22,12 @@ namespace core {
...
 public:
-    // Loads the specified dynamic library.
-    // If the library cannot be loaded then this is a fatal error.
-    // For *nix systems, a nullptr can be used to search the application's functions.
-    DlLoader(const char* name);
+    // Loads the dynamic library specified by the given name and fallback names
+    // (if any). Names will be used to try to find the library in order. If the
+    // library cannot be loaded then this is a fatal error. For *nix systems,
+    // a nullptr can be used to search the application's functions.
+    template<typename... ConstCharPtrs>
```
AWoloszyn: Is there a particular reason we are doing this with template magic, rather than something like std::initializer_list?

Qining: hmm, I just feel `DlLoader("libvulkan.so")` looks a little bit better than `DlLoader({"libvulkan.so"})`. But yes, initializer list can do the same and does not need explicit instantiation...Don't have other particular reasons.

AWoloszyn: Ok, as much as I personally love templates, reading

```
void* load() {
    return nullptr;
}

template<typename... ConstCharPtrs>
void* load(const char* name, ConstCharPtrs... fallback_names) {
....
...
    return load(fallback_names...);
...
...
Bunch of macros that will cause a link error if we hit 6 arguments.
...
```
is a lot harder to parse than
```
void* load(std::initializer_list<const char* name> names) {
   for (auto& nm: names) {
   }
```

Issue #826, opened 2017-07-28
Allow passing of additional flags to `am start` when tracing.


digish:
Generally to debug dEQP android CTS graphics tests, we give command line options to the activity which select the testcase and other test parameters.

as the default activity launch will test the full package which will be very large.
so if option to give command line arguments while starting activity will help in such use cases.

example:
we start the dEQP application like below.

adb shell am start -n com.drawelements.deqp/android.app.NativeActivity -e cmdLine "'binaryname --deqp-watchdog=enable --deqp-log-filename=/sdcard/dEQP-Log.qpa  --deqp-gl-config-name=rgba8888d24s8ms0 --deqp-case=dEQP-GLES3.functional.fragment_ops.blend.fbo_srgb.equation_src_func_dst_func.add_constant_alpha_constant_alpha'"

ben-clayton:
Great suggestion @digish.

pmuetschard:
So, this appears to be very specific to dEQP. The parameter being passed to `am start` is `-e`, which is just passing along an [extra](https://developer.android.com/reference/android/content/Intent.html#getExtras()) along with the intent. The fact that `-e cmdline <stuff>` passes command line parameters is specific to dEQP.

Therefore, I think what we can do is provide an input field in the trace dialog "am start flags", which we then pass along. For this specific use case, you'd type in `-e cmdLine "..."`.

ben-clayton:
I still think this is a useful feature to have. Re-opening.

pmuetschard:
Yes, sorry for the confusion. I didn't mean to give the impression that we shouldn't do it, just that it will be more general than "command line arguments", since it's really just an activity extra in dEQP that is used to simulate cli args. That's why I changed the title to be more general.

I was and am still planing on doing this, though I could use a suggestion of what to label the text field, please...

y-novikov:
This is also needed for running Chromium tests, to the extent that the tests crash if intent extras are not provided.
Can this please get a higher priority?

Suggestion for text field label: "Intent Arguments".
From https://developer.android.com/studio/command-line/adb.html#IntentSpec

null77:
+1 to increasing priority for this feature for running ANGLE/Chromium tests.

AWoloszyn:
I just put up a MR for part of this, once it passes the bots, I will merge.
Once it has merged you should be able to trace with the command-line tool and add additional args.

`gapit trace -android-additionalargs "additional flags to am start go here" app_to_trace`

[...]

PR #1637, opened 2018-02-22
Added the ability to supply am start flags to gapit.

AWoloszyn:
The gapit part of what is needed to support #826.

### DIFF
--- a/core/text/split_args.go
+++ b/core/text/split_args.go
@@ -49,3 +49,35 @@ func SplitArgs(s string) []string {
 	flush()
 	return out
 }
+
+// Quote takes a slice of strings, and returns a slice where each
+// string has been quoted and escaped to pass down to the command-line.
+func Quote(s []string) []string {
+	out := []string{}
+
+	for _, str := range s {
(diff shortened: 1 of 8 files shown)


gapii/client/adb.go
```diff
@@ -57,6 +58,16 @@ type Process struct {
 	conn net.Conn
 }
 
+type AdditionalFlags struct {
```
ben-clayton: Maybe add a this to `action.go`:

```go
// CustomExtras is a list of custom intent extras
type CustomExtras []string

// Flags returns the formatted flags to pass to the Android am command.
func (e CustomExtras) Flags() []string { return []string(e) }
```

core/text/split_args.go
```diff
@@ -49,3 +49,33 @@ func SplitArgs(s string) []string {
 	flush()
 	return out
 }
+
+func AddQuotes(s []string) []string {
```
ben-clayton: Maybe call `Quote`?
Please add documentation.

PR #1638, opened 2018-02-22
Cleans up some of the logcat output.

baldwinn860:
Enabling logcat capturing during a trace will cause the whole system to
dump informational levels of logs, and ignores debug and verbose
messages from GAPID, so a quick change makes logcat produce warnings and
above from non-GAPID logs, and verbose and above from GAPID. The other
change is to clear the state of logcat before monitoring, and to force
all logcat messages to be printed on stdout such that it does not
produce an error in robot.

### DIFF
--- a/cmd/gapit/trace.go
+++ b/cmd/gapit/trace.go
@@ -196,6 +196,8 @@ func (verb *traceVerb) captureADB(ctx context.Context, flags flag.FlagSet, start
 
 	if options.monitorLogcat {
 		c := make(chan android.LogcatMessage, 32)
+		// this is to prevent logcat messages from triggering failures in robot
+		ctx := log.PutHandler(ctx, log.Channel(app.Flags.Log.Style.Handler(log.Stdout()), 32))
 		go func() {
 			for m := range c {
 				m.Log(ctx)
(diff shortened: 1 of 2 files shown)


cmd/gapit/trace.go
```diff
@@ -195,13 +195,16 @@ func (verb *traceVerb) captureADB(ctx context.Context, flags flag.FlagSet, start
 	)
 
 	if options.monitorLogcat {
+		d.Command("logcat", "--clear").Run(ctx)
```
ben-clayton: This is what the `-T 0` is supposed to do.

cmd/gapit/trace.go
```diff
@@ -195,13 +195,16 @@ func (verb *traceVerb) captureADB(ctx context.Context, flags flag.FlagSet, start
...
 		c := make(chan android.LogcatMessage, 32)
+		// this is to prevent logcat messages from triggering failures in robot
+		logcatx := log.PutHandler(ctx, log.Channel(app.Flags.Log.Style.Handler(log.Stdout()), 32))
```
ben-clayton: s/logcatx/ctx?

cmd/gapit/trace.go
baldwinn860: ah, I must have been confused then, now removed.

PR #1639, opened 2018-02-23
Add gapil/compiler

ben-clayton:
Implements a true compiler for gapil .api files.


gapil/compiler/map.go
```diff
@@ -0,0 +1,443 @@
...
+// Once the map hits > 80% capacity, we should rehash the map larger.
+//  Otherwise collisions will turn this into a linear search.
+
+// Insertion into the map is similar (assuming it doesn't exist)
+//   1) h = Hashed Key % capacity
+//   2) rehash if necessary
+//   3) find first bucket >= h, where used != full (mod capacity)
+//   4) Insert there, mark used = full
+
+// TODO: Investigate alternative rehash rule.
```
AWoloszyn: This is done now

AWoloszyn: Not the next todo though

gapil/compiler/program.go
```diff
@@ -0,0 +1,132 @@
...
+	Line   int
+	Column int
+}
+
+func (l Location) String() string {
+	return fmt.Sprintf("%v:%v:%v", l.File, l.Line, l.Column)
+}
+
+func (p *Program) Dump() string {
+	//	types := make([]string, 0, len(c.types))
```
AWoloszyn: Do we want to keep these comments around?

gapil/compiler/map.go
ben-clayton: Removed.

gapil/compiler/program.go
ben-clayton: Nope, not even sure what they were. Removed.

PR #1642, opened 2018-02-23
Add Dedupe to core/data.

ben-clayton:
Dedupe returns a new byte slice containing the concatenated bytes of slices with duplicates removed, along with indices that map slices into the into the new slice.

### DIFF
--- a/core/data/dedupe.go
+++ b/core/data/dedupe.go
@@ -1,0 +1,58 @@
+// Copyright (C) 2018 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


core/data/dedupe_test.go
```diff
@@ -0,0 +1,44 @@
...
+	ctx := log.Testing(t)
+
+	B := func(s string) []byte { return ([]byte)(s) }
+	slices := [][]byte{
+		B("cat says meow"),
+		B("says"),
+		B("the cat says meow. the dog says woof. "),
+		B("fish says blub"),
+	}
+	expected := B("the cat says meow. the dog says woof. fish says blub")
```
AWoloszyn: But what does the fox say?

core/data/dedupe.go
```diff
@@ -0,0 +1,58 @@
...
+
+package data
+
+import (
+	"bytes"
+	"sort"
+)
+
+// Dedupe returns a new byte slice containing the concatenated bytes of slices
+// with duplicates removed, along with indices that map slices into the into the
```
pmuetschard: "into the into the"

ben-clayton: Fixed.

PR #1644, opened 2018-02-23
Create aligned pool for image and recover/prime layer by layer

Qining:
This is to unify all the cases.

When recovering the image data from trace side and priming the data to
rebuild the state, we calculate the staging buffer offset by the aligned
size. So the pools must be created with the aligned size.

Also when recover/prime the image data, create image buffer copy layer
by layer so that we don't need to assume the boundaries of multiple layers
in one copy are aligned.

### DIFF
--- a/gapii/cc/vulkan_mid_execution.cpp
+++ b/gapii/cc/vulkan_mid_execution.cpp
@@ -128,14 +128,13 @@ bool IsFullyBound(VkDeviceSize offset, VkDeviceSize size,
 // A helper class that contains a temporary buffer that is bound to
 // hold incomming data from other GPU resources.
 class StagingBuffer {
-public:
-  StagingBuffer(core::Arena* arena,
+ public:
+  StagingBuffer(core::Arena *arena,
                 VulkanImports::VkDeviceFunctions &device_functions,
                 VkDevice device,
                 const VkPhysicalDeviceMemoryProperties &memory_properties,
                 uint32_t size)
       : device_functions_(device_functions), device_(device), size_(size) {
-
     VkBufferCreateInfo staging_buffer_create_info{arena};
     staging_buffer_create_info.msType =
         VkStructureType::VK_STRUCTURE_TYPE_BUFFER_CREATE_INFO;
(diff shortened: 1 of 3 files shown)


gapii/cc/vulkan_mid_execution.cpp
```diff
@@ -658,19 +657,22 @@ void VulkanSpy::prepareGPUBuffers(CallObserver* observer,
...
-                                {
-                                    img->mImageAspect, // aspectMask
-                                    mip_level,
-                                    range.mbaseArrayLayer, // baseArrayLayer
-                                    range.mlayerCount      // layerCount
-                                },
-                                {0, 0, 0},
-                                {e.width, e.height, e.depth}});
-          offset += (e.aligned_level_size * range.mlayerCount);
+          for (size_t j = 0; j < range.mlayerCount; j++) {
```
Qining: Real changes here.

gapis/api/vulkan/state_rebuilder.go
```diff
@@ -1503,27 +1503,31 @@ func (sb *stateBuilder) createImage(img *ImageObject) {
...
-					},
-					VkExtent3D{
-						uint32(e.width),
-						uint32(e.height),
-						uint32(e.depth),
-					},
-				})
-				for l := rng.BaseArrayLayer; l < rng.LayerCount; l++ {
-					data := img.Layers.Get(l).Levels.Get(mipLevel).Data.MustRead(sb.ctx, nil, sb.oldState, nil)
+				for j := uint32(0); j < rng.LayerCount; j++ {
```
Qining: Real changes here.

gapis/api/vulkan/vulkan.api
```diff
@@ -4130,7 +4130,8 @@ cmd VkResult vkCreateImage(
       // Roundup the width and height in the number of blocks.
       widthInBlocks := roundUpTo(width, elementAndTexelBlockSize.TexelBlockSize.Width)
       heightInBlocks := roundUpTo(height, elementAndTexelBlockSize.TexelBlockSize.Height)
-      size := widthInBlocks * heightInBlocks * depth * elementAndTexelBlockSize.ElementSize
+      // Align to the next multiple times of 8
+      size := ((widthInBlocks * heightInBlocks * depth * elementAndTexelBlockSize.ElementSize) + 7) & (0xFFFFFFF8)
```
Qining: Real changes here.

Qining:
All other changes are from clang-format.

PR #1647, opened 2018-02-23
Binary Stripping and Symbols Dumping Rules.

pmuetschard:
- Strip binaries (executables and shared libraries) - all shared libraries and `gapi[rst]`
- Symbol dumping rules for each OS, which will only work if using `--config symbols`, when building.
  - Linux is straight forward, as breakpad's symbol dumping utility can simply read the ELF executable.
  - On macOS:
    - Rename the OS specific `crash_handler.cpp` file, so that `dsymutil` can find the `.o` file, since there would otherwise be two `crash_handler.o` files in the `.a` file, which `dsymutil` can't handle.
    - Remove the sandbox prefix from the debug paths and make them relative, so `dsymutil` doesn't try to access things from a no-longer existing sandbox and choke.
    - Update the symbol dumping rule to first invoke `dsymutil` to create the `.dSYM` directory used to dump the symbols.
  - On Windows:
    - Ask the linker to embed a build-id, which creates a tiny debug ID segment in the output, which is read by the OS when creating a crash dump.
    - Create a repository rule for the breakpad library that applies a patch (if on Windows), which removes the ELF dependency from the symbol dumping code, which is only needed for "split DWARF", which we aren't using anyhow.
     - Add a new symbol dumping utility that supports reading Windows PE files. It looks for the above mentioned debug ID segment for the identifiers and uses the Linux DWARF symbol dumping code to read the symbol information from the debug segments.


tools/build/third_party/breakpad/dump_syms_pe.cc
```diff
@@ -0,0 +1,389 @@
...
+  return nullptr;
+}
+
+// buffer needs to be 9 bytes (or larger).
+static const char* getSectionName(const PEFile& file, const PESectionHeader& section, char* buffer) {
+  if (section.name[0] == '/') {
+    char offS[8];
+    std::memcpy(offS, &section.name[1], 7);
+    offS[7] = 0;
+    int off = atoi(offS);
```
ben-clayton: `int off = std::stoi(std::string(section.name, 1, 7));`

Note that an exception will be thrown if the number cannot be parsed.

tools/build/third_party/breakpad/dump_syms_pe.cc
```diff
@@ -0,0 +1,389 @@
...
+    if (off > file.strings.size) {
+      return "";
+    }
+    return file.strings.data + off;
+  }
+  std::memcpy(buffer, section.name, 8);
+  buffer[8] = 0;
+  return buffer;
```
ben-clayton: `return std::string(section.name, 0, 8)`

tools/build/third_party/breakpad/dump_syms_pe.cc
```diff
@@ -0,0 +1,389 @@
...
+    const char* sectionName = getSectionName(file, *section, nameBuf);
+    if (sectionName == name) {
+      return section;
+    }
+  }
+  return nullptr;
+}
+
+static bool initIdentifiers(PEFile* file) {
+  if (!file->optionalHeader->directory[6].address) {
```
ben-clayton: Please create a sensibly named constant for whatever `6` is supposed to mean.

tools/build/third_party/breakpad/dump_syms_pe.cc
```diff
@@ -0,0 +1,389 @@
...
+    fprintf(stderr, "Debug section too small: %d < %d\n", section->virtualSize, offset + sizeof(DebugDirectory));
+    return false;
+  }
+  if (section->offset + offset + sizeof(DebugDirectory) > file->data.size()) {
+    fprintf(stderr, "Debug section past EOF: %x\n", section->offset);
+    return false;
+  }
+  DebugDirectory* directory = (DebugDirectory*)(file->data.data() + section->offset + offset);
+
+  if (directory->type != 2) {
```
ben-clayton: 2?

tools/build/third_party/breakpad/dump_syms_pe.cc
```diff
@@ -0,0 +1,389 @@
...
+    fprintf(stderr, "Unsupported debug data type: %x\n", directory->type);
+    return false;
+  }
+  if (directory->size < sizeof(DebugData) + 1 /* 0 length string is the min */) {
```
ben-clayton: I'm a bit confused by the +1 here. Is this a null terminator of a string that follows the `DebugData` structure?

tools/build/third_party/breakpad/dump_syms_pe.cc
```diff
@@ -0,0 +1,389 @@
...
+    fprintf(stderr, "Debug data too small: %d\n", directory->size);
+    return false;
+  }
+  if (directory->offset + directory->size > file->data.size()) {
+    fprintf(stderr, "Debug data past EOF: %x\n", directory->offset);
+    return false;
+  }
+  DebugData* debugData = (DebugData*)(file->data.data() + directory->offset);
```
ben-clayton: `reinterpret_cast`

tools/build/third_party/breakpad/dump_syms_pe.cc
```diff
@@ -0,0 +1,389 @@
...
+
+  char debugIdentifier[32 + 8 + 1];
+  snprintf(debugIdentifier, sizeof(debugIdentifier), "%08X%04X%04X%02X%02X%02X%02X%02X%02X%02X%X\0",
```
ben-clayton: No need to null terminate string literals - they already end in a 0. Note your last `%X` should be `%02X`

The C++ way of doing this would be:

```c++
    std::stringstream ss;
    ss << std::setfill('0') << std::hex << std::uppercase
        << std::setw(8) << u.a
        << std::setw(4) << u.b 
        << std::setw(4) << u.c
        << std::setw(2) << int(u.d[0])
        << std::setw(2) << int(u.d[1])
        << std::setw(2) << int(u.d[2])
        << std::setw(2) << int(u.d[3])
        << std::setw(2) << int(u.d[4])
        << std::setw(2) << int(u.d[5])
        << std::setw(2) << int(u.d[6])
        << std::setw(2) << int(u.d[7]);
    file->debugId = ss.str();
```

tools/build/third_party/breakpad/dump_syms_pe.cc
```diff
@@ -0,0 +1,389 @@
...
+      debugData->uuid.a, debugData->uuid.b, debugData->uuid.c,
+      debugData->uuid.d[0], debugData->uuid.d[1], debugData->uuid.d[2], debugData->uuid.d[3],
+      debugData->uuid.d[4], debugData->uuid.d[5], debugData->uuid.d[6], debugData->uuid.d[7],
+      debugData->age);
```
ben-clayton: What is `age` doing here (or is the last %X the age, if so you're missing a byte for the uuid)?

tools/build/third_party/breakpad/dump_syms_pe.cc
```diff
@@ -0,0 +1,389 @@
...
+  file->debugId = debugIdentifier;
+
+  char codeIdentifier[8 + 8 + 1];
+  snprintf(codeIdentifier, sizeof(codeIdentifier), "%08X%X\0",
```
ben-clayton: See above

tools/build/third_party/breakpad/dump_syms_pe.cc
```diff
@@ -0,0 +1,389 @@
...
+  }
+  stream.close();
+
+  int base = getBase(file->data);
+  if (base + sizeof(PEFileHeader) > size) {
+    fprintf(stderr, "Not a PE file: file too small\n");
+    return false;
+  }
+  PEFileHeader* header = file->header = (PEFileHeader*)(file->data.data() + base);
+  if (header->machine != 0x8664) {
```
ben-clayton: Put magic in named constant.

tools/build/third_party/breakpad/dump_syms_pe.cc
```diff
@@ -0,0 +1,389 @@
...
+
+  if (header->extraHeaderSize < sizeof(PEOptionalHeader)) {
+    fprintf(stderr, "Not a PE file: extra header too small\n");
+    return false;
+  } else if (base + sizeof(PEFileHeader) + header->extraHeaderSize > size) {
+    fprintf(stderr, "Not a PE file: invalid extra header size\n");
+    return false;
+  }
+  PEOptionalHeader* optHeader = file->optionalHeader = (PEOptionalHeader*)(file->data.data() + base + sizeof(PEFileHeader));
+  if (optHeader->signature != 0x20b) {
```
ben-clayton: magic

tools/build/third_party/breakpad/dump_syms_pe.cc
```diff
@@ -0,0 +1,389 @@
...
+    byte_reader.SetCFIDataBase(debugFrame->virtualAddress, cfi);
+    dwarf2reader::CallFrameInfo::Reporter dwarf_reporter(name, ".debug_frame");
+    dwarf2reader::CallFrameInfo parser(cfi, debugFrame->virtualSize, &byte_reader, &handler, &dwarf_reporter, false);
+    parser.Start();
+  }
+}
+
+int main(int argc, char **argv) {
+  if (argc != 2) {
+    fprintf(stderr, "Usage: %s <file.exe>\n", argv[0]);
```
ben-clayton: Also include a short description of what this exe does?

tools/build/third_party/breakpad/windows.patch
```diff
@@ -0,0 +1,80 @@
+diff --git a/src/common/dwarf/dwarf2reader.cc b/src/common/dwarf/dwarf2reader.cc
+index f2dc4afe..56bbe98a 100644
+--- a/src/common/dwarf/dwarf2reader.cc
++++ b/src/common/dwarf/dwarf2reader.cc
+@@ -64,7 +64,7 @@ CompilationUnit::CompilationUnit(const string& path,
+       is_split_dwarf_(false), dwo_id_(0), dwo_name_(),
+       skeleton_dwo_id_(0), ranges_base_(0), addr_base_(0),
+       have_checked_for_dwp_(false), dwp_path_(),
+-      dwp_byte_reader_(), dwp_reader_() {}
```
ben-clayton: Why?

tools/build/third_party/breakpad/dump_syms_pe.cc
baldwinn860: Could this be changed to a `<=` and still maintain the same semantics?

tools/build/third_party/breakpad/dump_syms_pe.cc
```diff
@@ -0,0 +1,389 @@
...
+    uint16_t c;
+    uint8_t d[8];
+  } uuid;
+  uint32_t age;
+} DebugData;
+
+static int getBase(const std::vector<char>& data) {
+  int base = 0;
+  if (data.size() >= 64 && data[0] == 'M' && data[1] == 'Z') {
+    int offset = ((uint32_t *)data.data())[15];
```
AWoloszyn: FYI: This is not safe on anything but x86. In general when doing bitcasts for aligned types you should always memcpy in. (The compiler can switch back to a non-memcpy if it is safe to do so)

tools/build/third_party/breakpad/dump_syms_pe.cc
```diff
@@ -0,0 +1,389 @@
...
+  struct {
+    uint32_t a;
+    uint16_t b;
+    uint16_t c;
+    uint8_t d[8];
+  } uuid;
+  uint32_t age;
+} DebugData;
+
+static int getBase(const std::vector<char>& data) {
```
AWoloszyn: Modern C++ suggests putting these in an anonymous namespace instead of making them all static. (Then you don't have to mark them all as static)

tools/build/third_party/breakpad/dump_syms_pe.cc
AWoloszyn: I wish the C++ syntax wasn't so terrible.

tools/build/third_party/breakpad/dump_syms_pe.cc
```diff
@@ -0,0 +1,389 @@
...
+
+  char codeIdentifier[8 + 8 + 1];
+  snprintf(codeIdentifier, sizeof(codeIdentifier), "%08X%X\0",
+      file->header->timeStamp, file->optionalHeader->imageSize);
+  file->codeId = codeIdentifier;
+
+  return true;
+}
+
+static bool readFile(const std::string& in, PEFile* file) {
```
AWoloszyn: Google style guide says naming is TitleCase for functions. Even if I dont like it

tools/build/third_party/breakpad/dump_syms_pe.cc
```diff
@@ -0,0 +1,389 @@
...
+    file_context.AddSectionToSectionMap(getSectionName(file, *section, nameBuffer),
+        (uint8_t*)(file.data.data() + section->offset), section->virtualSize);
+  }
+
+  DumperLineToModule line_to_module(&byte_reader);
+  dwarf2reader::SectionMap::const_iterator debug_info_entry = file_context.section_map().find(".debug_info");
+  assert(debug_info_entry != file_context.section_map().end());
+  const std::pair<const uint8_t *, uint64>& debug_info_section = debug_info_entry->second;
+  assert(debug_info_section.first);
+  uint64 debug_info_length = debug_info_section.second;
```
AWoloszyn: Nit: can also be const

tools/build/third_party/breakpad/dump_syms_pe.cc
```diff
@@ -0,0 +1,389 @@
...
+    dwarf2reader::DIEDispatcher die_dispatcher(&root_handler);
+    dwarf2reader::CompilationUnit reader(name, file_context.section_map(), offset, &byte_reader, &die_dispatcher);
+    offset += reader.Start();
+  }
+}
+
+static void loadDwarfCFI(const std::string& name, const PEFile& file, google_breakpad::Module* module) {
+  PESectionHeader* debugFrame = findSectionByName(file, ".debug_frame");
+  if (debugFrame != nullptr) {
+    vector<string> register_names = google_breakpad::DwarfCFIToModule::RegisterNames::X86_64();
```
AWoloszyn: `std::vector<std::string>` ?

tools/build/third_party/breakpad/dump_syms_pe.cc
pmuetschard: Done.

tools/build/third_party/breakpad/windows.patch
pmuetschard: The dwp_reader is only used for "split dwarf" and assumes ELF. We don't need split dwarf.

AWoloszyn:
I have no more concerns. I will let @ben-clayton  do the final review.

PR #1650, opened 2018-02-26
apic compiler + encoder plugin

ben-clayton:


cmd/apic/compile.go
```diff
@@ -0,0 +1,142 @@
...
+	if err != nil {
+		return err
+	}
+
+	var abi *device.ABI
+	switch v.Target { // Must match values in: tools/build/BUILD.bazel
+	case "":
+		abi = nil // host
+	case "k8":
+		abi = device.LinuxX86_64
```
AWoloszyn: Out of curiosity, is there a windows 64 as well? (Do we assume it is the same as Linux64?)

ben-clayton: It's a good question, and something I'm sure I'll answer when we start using `apic compile` as a build rule.

Issue #1646, opened 2018-02-23
Tracing performance regression from 1.0.3 -> master.


AWoloszyn:
I have tested a relatively heavy application whose frame-rate almost halved with something between 1.0.3 and master.

I will do some more binary searching to figure out where it happened

PR #1654, opened 2018-02-27
Performance fixes.

AWoloszyn:
This fixes https://github.com/google/gapid/issues/1646.

The performance overhead of libgapii is now actually lower than it was in 1.03.

### DIFF
--- a/gapii/cc/BUILD.bazel
+++ b/gapii/cc/BUILD.bazel
@@ -142,7 +142,7 @@ cc_library(
         "//core/memory/arena/cc",
         "//core/memory_tracker/cc",
         "//core/os/device/deviceinfo/cc",
-        "//gapil/runtime/cc",
+        "//gapil/runtime/cc:cc",
         "//gapis/api:api_cc_proto",
         "//gapis/api/gles/gles_pb:gles_pb_cc_proto",
         "//gapis/api/gvr/gvr_pb:gvr_pb_cc_proto",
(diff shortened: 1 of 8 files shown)


gapil/runtime/cc/map.inc
```diff
@@ -22,6 +22,9 @@
 
 #include <unordered_map>
 
+static_assert((GAPIL_MIN_MAP_SIZE & (GAPIL_MIN_MAP_SIZE - 1)) == 0, "Map size must be a power of 2");
```
ben-clayton: Might be worth putting a similar assert in the compiler.

gapii/cc/BUILD.bazel
```diff
@@ -142,7 +142,7 @@ cc_library(
         "//core/memory/arena/cc",
         "//core/memory_tracker/cc",
         "//core/os/device/deviceinfo/cc",
-        "//gapil/runtime/cc",
+        "//gapil/runtime/cc:cc",
```
ben-clayton: Should be no need for the `:cc`

gapil/runtime/cc/BUILD.bazel
```diff
@@ -19,13 +19,25 @@ cc_library(
...
+        "runtime.cpp",
+        "string.cpp",
+    ],
+    hdrs = glob(["*.inc"]),
+    copts = cc_copts(),
+    visibility = ["//visibility:public"],
     deps = [
-        "//core/memory/arena/cc",
+        ":headers",
+        "//core/memory/arena/cc:cc",
```
ben-clayton: No need for `:cc`

PR #1655, opened 2018-02-27
Updated our hash implementations to match getween go and C++.

AWoloszyn:
This sits on top of https://github.com/google/gapid/pull/1654.
Only the last change is relevant.

### DIFF
--- a/gapil/runtime/cc/map.inc
+++ b/gapil/runtime/cc/map.inc
@@ -16,7 +16,7 @@
 #define __GAPIL_RUNTIME_MAP_INC__
 
 #include "map.h"
-#include "string.h"
+#include "hash.h"
 
 #include "core/cc/assert.h"
 #include "core/memory/arena/cc/arena.h"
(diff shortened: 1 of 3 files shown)


gapil/runtime/cc/map.inc
```diff
@@ -25,6 +26,99 @@
...
+        v ^= rotate_right(v, 28);
+        v += shift_left(v, 31);
+        return v;
+    }
+};
+
+template<>
+struct gapil_hash<float> {
+    uint64_t operator() (const float& f) {
+        return gapil_hash<fixedValue>()(fixedValue{static_cast<uint64_t>(*reinterpret_cast<const uint32_t*>(&f))});
```
ben-clayton: I believe for this sort of thing you're supposed to use a union.

gapil/runtime/cc/map.inc
```diff
@@ -25,6 +26,99 @@
...
+
+#define HASH_INTEGER(type) \
+template<> \
+struct gapil_hash<type> {\
+    uint64_t operator() (const type&  v) { \
+        return static_cast<uint64_t>(v); \
+    } \
+};
+
+HASH_INTEGER(bool)
```
ben-clayton: http://en.cppreference.com/w/cpp/types/is_integral ?

gapil/runtime/cc/map.inc
```diff
@@ -25,6 +26,99 @@
...
+HASH_INTEGER(int16_t)
+HASH_INTEGER(uint32_t)
+HASH_INTEGER(int32_t)
+HASH_INTEGER(uint64_t)
+HASH_INTEGER(int64_t)
+
+#undef HASH_INTEGER
+
+template<typename T>
+struct gapil_hash<T*> {
```
ben-clayton: `gapil::hash`?
Maybe put into its own header?

gapil/runtime/cc/map.inc
AWoloszyn: Done

gapil/runtime/cc/map.inc
AWoloszyn: I think actually both end up being wrong for the same reason. (memory alignment :\) if alignof(float) != alignof(uint32_t).

I will do the "right thing" and go through a memcpy, which as far as I can tell is the only safe way to do this in general.

gapil/runtime/cc/map.inc
AWoloszyn: Done

Issue #1617, opened 2018-02-13
Prime Depth images in MEC.


AWoloszyn:
To prime depth images, create a shader + write glFragDepth for each fragment.

Qining:
Fixed by #1739 and #1750.

Issue #1616, opened 2018-02-13
Stop adding TRANSFER_DST_BIT to images in MEC.


AWoloszyn:
We should do different things depending on the contents of img.Info.Usage

This is for single sampled non-depth images.

1) If it contains VK_IMAGE_USAGE_TRANSFER_DST_BIT, then prime images how we are
2) elif it contains VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT,
    Generate a texture + quad + shader + graphics pipeline and prime it by rendering into it
3) elif it contains VK_IMAGE_USAGE_STORAGE_BIT, then create a "compute" pipeline & 
    copy into the image
4) else, there are no valid contents that we can add to the image.

Qining:
PR: #1656, #1739, #1750, #1756 take care of the color/depth/stencil attachment images.

Storage bit images are still not handled.

PR #1656, opened 2018-02-27
Priming color image data by rendering to it

Qining:
If the state block image does not have VK_IMAGE_USAGE_TRANSFER_DST_BIT,
but has VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT or
DEPTH_STENCIL_ATTACHMENT_BIT, we can rendering to it to rebuild the
state.

Contribute to #1616 and #1617 

TODO:
 - [x] Float point bitcast to [4]uint8{}
  - resolved by using the `core/image` package
 - [x] Rendering layer by layer, level by level
 - [x] Supports all common uncompressed formats, except `SCALED` ones.
 - [x] Test the data unpacking logic


gapis/api/vulkan/footprint_builder.go
```diff
@@ -1636,6 +1637,8 @@ func (vb *FootprintBuilder) BuildFootprint(ctx context.Context,
 	// Records the current last draw framebuffer image data, so that later when
 	// the user request a command, we can always guarantee that the last draw
 	// framebuffer is alive.
+	log.W(ctx, "id: %v, cmd: %v", id, cmd)
```
Qining: Drop it later.

gapis/api/vulkan/image_rebuild_shaders.go
```diff
@@ -0,0 +1,960 @@
...
+		VkFormat_VK_FORMAT_R8G8_SNORM,
+		VkFormat_VK_FORMAT_R8G8B8_SNORM,
+		VkFormat_VK_FORMAT_R8G8B8A8_SNORM,
+		VkFormat_VK_FORMAT_B8G8R8_SNORM,
+		VkFormat_VK_FORMAT_B8G8R8A8_SNORM:
+		/*
+			#version 450
+			layout(location = 0) out vec4 out_color;
+			layout(input_attachment_index = 0, binding = 0, set = 0) uniform usubpassInput in_color;
+			float snorm(in uint u) {
```
Qining: This function will be replaced

gapis/api/vulkan/image_rebuild_helper.go
```diff
@@ -0,0 +1,1050 @@
...
+					img.Info.ArrayLayers,
+				},
+			})
+	}
+	outputBarrier := VkImageMemoryBarrier{
+		VkStructureType_VK_STRUCTURE_TYPE_IMAGE_MEMORY_BARRIER,
+		NewVoidᶜᵖ(memory.Nullptr),
+		VkAccessFlags(0),
+		VkAccessFlags(VkAccessFlagBits_VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT),
+		VkImageLayout_VK_IMAGE_LAYOUT_UNDEFINED,
```
AWoloszyn: Can you put a comment here to say these are filled in below

gapis/api/vulkan/image_rebuild_shaders.go
```diff
@@ -0,0 +1,1531 @@
...
+			0x00000009, 0x00000006, 0x00000000, 0x00000000,
+			0x00000000, 0x00000002, 0x00000000, 0x00040020,
+			0x0000000b, 0x00000000, 0x0000000a, 0x0004003b,
+			0x0000000b, 0x0000000c, 0x00000000, 0x00040015,
+			0x0000000e, 0x00000020, 0x00000001, 0x0004002b,
+			0x0000000e, 0x0000000f, 0x00000000, 0x00040017,
+			0x00000010, 0x0000000e, 0x00000002, 0x0005002c,
+			0x00000010, 0x00000011, 0x0000000f, 0x0000000f,
+			0x00040017, 0x00000012, 0x00000009, 0x00000004,
+			0x0004002b, 0x00000009
```
AWoloszyn: You definitely don't need the texcoord/normals here

gapis/api/vulkan/state_rebuilder.go
```diff
@@ -1497,80 +1461,157 @@ func (sb *stateBuilder) createImage(img *ImageObject) {
...
+			return
 		}
+		if len(stagingImgs) == 0 {
+			log.E(sb.ctx, "No staging image created")
+			return
+		}
+		copyDstImgs = stagingImgs
+
+	} else {
+		// There is no way to prime the data
```
AWoloszyn: Maybe we want to issue a warning. (Data in image unprimed, since there was no valid way to add data in the first place)

Qining:
Just realized `image.Convert()` does something like: `float(int(0xf))`, but what we want is `intBitsToFloat(int(0xf))`. Conversion != Bitcast.

So I will need to add `bitcast` support for this CL

AWoloszyn:
Looks great, my only request would be, can we cache the graphics pipelines so that we don't have to create a new one for each image. (We obviously will have to create different ones for different formats, but we should be able to get away with a handful).

Qining:
Request tracked by #1714, I will fix this once the depth and stencil part of the logic are fixed.

PR #1657, opened 2018-02-27
Skip tracking the data dependency of destroyed framebuffer images

Qining:


Qining:
Updated the api file in 595063b. Also the image views used in descriptor sets are invalidated.

PR #1659, opened 2018-02-27
build: Fix gapid.apk hash not being applied on windows.

ben-clayton:
Partially fixes issue where APKs are not being pushed to the device (there's another issue we're trying to uncover)


pmuetschard:
Filed a bug, too: https://github.com/bazelbuild/bazel/issues/4734

PR #1663, opened 2018-02-28
Fix several issues running Vulkan replays on robot.

baldwinn860:
GapidAPK would previously freeze if it was relaunched by GAPIS when it
was already running. An unbroken infinite while caused an ANR issue from
the looks of things. Another issue is that it seems that GAPIS mistimes
sending the connection request when relaunching the apk. The current fix
for this is just a bandaid to get vulkan running on robot, by
force-stopping GapidAPK after the robot task finishes.

### DIFF
--- a/cmd/gapir/cc/main.cpp
+++ b/cmd/gapir/cc/main.cpp
@@ -149,7 +149,8 @@ void android_main(struct android_app* app) {
       listenConnections(conn.get(), nullptr, nullptr, Connection::NO_TIMEOUT,
                         &memoryManager, crashHandler);
     });
-    while (true) {
+    bool alive = true;
+    while (alive) {
       int ident;
       int fdesc;
       int events;
@@ -162,6 +163,7 @@ void android_main(struct android_app* app) {
         }
         if (app->destroyRequested) {
           conn->close();
+          alive = false;
           break;
         }
       }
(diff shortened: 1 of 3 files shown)


cmd/gapir/cc/main.cpp
```diff
@@ -165,6 +165,9 @@ void android_main(struct android_app* app) {
           break;
         }
       }
+      if (app->destroyRequested) {
```
Qining: Prefer to have it like:
```
while (alive) {
  while (ALooper...) {
    if (app->destroyRequested) {
      alive = false;
    }
  }
}
```

test/robot/replay/client.go
```diff
@@ -59,6 +59,11 @@ func Run(ctx context.Context, store *stash.Client, manager Manager, tempDir file
 func (c *client) onDeviceAdded(ctx context.Context, host *device.Instance, target bind.Device) {
 	replayOnTarget := func(ctx context.Context, t *Task) error {
 		job.LockDevice(ctx, target)
+		defer func() {
+			// TEMP: kill gapid.apk manually for now as subsequent reports/replays may freeze the app.
```
ben-clayton: Add `HACK:`, please link to a bug that tracks the issue and gives as much information as possible.

test/robot/report/client.go
```diff
@@ -59,6 +59,11 @@ func Run(ctx context.Context, store *stash.Client, manager Manager, tempDir file
 func (c *client) onDeviceAdded(ctx context.Context, host *device.Instance, target bind.Device) {
 	reportOnTarget := func(ctx context.Context, t *Task) error {
 		job.LockDevice(ctx, target)
+		defer func() {
+			// TEMP: kill gapid.apk manually for now as subsequent reports/replays may freeze the app.
```
ben-clayton: Ditto

Qining:
Works out fine on my side. Tracing/report/replay of Vulkan apps can be started and finished successfully.

baldwinn860:
@ben-clayton or @Qining could I get a final review?

PR #1679, opened 2018-03-01
deviceinfo: Do not explode if there's no GL driver available.

ben-clayton:
Taking the changes in #1658 further still.


AWoloszyn:
Other than the build failures looks good to me.

PR #1685, opened 2018-03-02
Add COMDAT bindings to LLVM

ben-clayton:


core/codegen/function.go
```diff
@@ -45,6 +45,9 @@ func (f Function) Inline() Function {
 // assumption their implementation is identical. Unlike "linkonce" this
 // also preserves inlining.
 func (f Function) LinkOnceODR() Function {
+	c := f.m.llvm.Comdat(f.Name)
```
AWoloszyn: Just want to make sure these are ignored on non-pe files.

ben-clayton: The Android `.so`s built successfully, so I presumed so. I'll wait for the build bot to say it's happy too.

PR #1689, opened 2018-03-02
Logging

ben-clayton:


core/cc/log.cpp
```diff
@@ -70,13 +74,28 @@ void Logger::vlogf(unsigned level, const char* src_file, unsigned src_line, cons
     std::tm* loc = std::localtime(&now);
     auto ms = std::chrono::duration_cast<std::chrono::milliseconds>(t.time_since_epoch());
 
+#define LOG_COMMON "%02d:%02d:%02d.%03d %c %s: [%s:%u] ", loc->tm_hour, loc->tm_min, loc->tm_sec, \
+    static_cast<int>(ms.count() % 1000), "FEWIDV"[level], mSystem, src_file, src_line 
+
+#if TARGET_OS == GAPID_OS_WINDOWS
+    {
```
AWoloszyn: This will help significantly!

We could have also optionally allocated a new console and logged to that (But we would only want to do that for development)

ben-clayton: yeah, I actually find `OutputDebugString` more useful when debugging in VS. Happy to add more logging options though!

PR #1690, opened 2018-03-02
On device destroy clean up presentation and submission images.

AWoloszyn:


AWoloszyn:
Looks like the linux failure is a flake, so I am going to merge as-is. Feel free to revert if anyone finds any issues.

PR #1691, opened 2018-03-02
core/codegen: Only add COMDAT for windows targets

ben-clayton:
macOS doesn't like it.


ben-clayton:
Tested on mac and windows - I don't want to wait all day for the build to catch up. Submitting.

PR #1692, opened 2018-03-02
Init the go rules repositories *after* all our go repo rules.

pmuetschard:
This is so that ours will always take precedence.


pmuetschard:
we have go grpc, go_rules has go grpc. For some strange reason, with my symbol dumping change, bazel starts to prefer go_rules' grpc, which is incompatible with our code.

ben-clayton:
I thought it was related to those build failures, just checkin'.

PR #1693, opened 2018-03-02
Switch from protoc encoding to api compile encoding

ben-clayton:
Optimized windows builds now work.


gapii/cc/BUILD.bazel
```diff
@@ -28,8 +28,16 @@ apic_template(
     ],
 )
 
+apic_compile(
+    name = "gles_compiled",
+    api = "//gapis/api/gles:api", # TODO: Let apic handle multiple api files in one compile.
+#    optimize = True,
```
AWoloszyn: I would assume we actually DO want these on when building -opt?

ben-clayton: One step at a time...

AWoloszyn: Sounds good. It would be good to re-enable them in a followup.

ben-clayton: You were right in that disabling the inlining pass makes a semi-opt build perform fast enough.

AWoloszyn:
Just did some quick performance tests on Android for this.
Test case: Vulkan Mid-Execution capture. Before we have started tracing. This gives us the total mutation + encoding, but does not incur any transfer speed issues.
Also hopefully this won't include encoding forever, but that is a future MR.

Master contains all of the changes to gapil::Ref, gapil::Map

```
Without tracing enabled: 29.8FPS == 33.56ms/frame
1.0.3 << Average: 12.1 FPS == 82.64ms/frame (= 49.09ms/frame overhead)
Master:<< Average: 14.3 FPS == 69.93ms/frame (36.37ms/frame overhead)
This MR: Average 14.9 FPS << 67.11ms/frame (33.56ms/frame overhead)

1.03 -> Master: reduced our tracing overhead by 35%
Master -> This MR: reduced our tracing overhead by a further 5%
```
```
1.03 libgapii size: 9.7M
This MR: ligapii size: 6.3M
```

PR #1698, opened 2018-03-03
Windows Optimized Build

pmuetschard:
- Optimized build and symbol dumping on Windows. 
- Disable symbol dumping on OSX, as it's sadly broken.


ben-clayton:
Yay! (and boo!)

PR #1699, opened 2018-03-05
Add cmd/regres

ben-clayton:
Regress is a tool to display build and runtime statistics over a range of changelists.

### DIFF
--- a/cmd/regres/main.go
+++ b/cmd/regres/main.go
@@ -1,0 +1,239 @@
+// Copyright (C) 2018 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Regress is a tool to display build and runtime statistics over a range of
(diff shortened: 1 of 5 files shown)


cmd/regres/main.go
```diff
@@ -0,0 +1,239 @@
...
+}
+
+type stats struct {
+	SHA                  string
+	BuildTime            float64  // in seconds
+	IncrementalBuildTime float64  // in seconds
+	FileSizes            struct { // in bytes
+		LibGAPII                   int64
+		LibVkLayerVirtualSwapchain int64
+		GAPIDAarch64APK            int64
```
pmuetschard: In addition to tracking the total APK size, I think it would be really useful to track the size of the individual libs inside the APKs.

cmd/regres/main.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+
+	fmt.Printf("-----------------------\n")
+
+	w := tabwriter.NewWriter(os.Stdout, 1, 4, 0, ' ', 0)
+	defer w.Flush()
+
+	fmt.Fprint(w, "sha"+
+		"\t | build_time"+
+		"\t | incremental_build_time"+
```
pmuetschard: nit: `build_time` should be called `incremental_build_time`, since that's what it really is, while `incremental_build_time` should be called something like `gles_api_build_time`.

cmd/regres/main.go
ben-clayton: Good idea. Will consider adding in a followup CL.

cmd/regres/main.go
ben-clayton: Well, yes they're both incremental build times. I'm not sure that there's too much value in `build_time` so I might just remove it. Going to be adding more stuff to this tool, so I'd like to follow up with more CLs to fix this.

PR #1704, opened 2018-03-06
Work around a "feature" in mesa.

AWoloszyn:
In mesa drivers (at least those shipped with Ubuntu 17.10) they
dynamically link against libLLVM-*.so. This means that since we a
static version of LLVM, we get name collisions on llvm, and
subsequent failures. Instead of all of this, preload libLLVM*
with RTLD_DEEPBIND before we even try and load libGL.

This means those llvm symbols will not try and resolve
any of our symbols.

Fixes: #1707

### DIFF
--- a/core/cc/linux/get_gles_proc_address.cpp
+++ b/core/cc/linux/get_gles_proc_address.cpp
@@ -18,11 +18,25 @@
 #include "../get_gles_proc_address.h"
 #include "../log.h"
 
+#include <dlfcn.h>
+
 namespace {
 
 void* getGlesProcAddress(const char *name, bool bypassLocal) {
     using namespace core;
     typedef void* (*GPAPROC)(const char *name);
+
+    // The mesa driver does bad things with LLVM. Since we also use llvm, 
+    // we can't have the mesa driver do bad things to our code.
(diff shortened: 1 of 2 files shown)


Qining:
Cool!

core/cc/linux/get_gles_proc_address.cpp
```diff
@@ -18,12 +18,25 @@
...
 namespace {
 
 void* getGlesProcAddress(const char *name, bool bypassLocal) {
     using namespace core;
     typedef void* (*GPAPROC)(const char *name);
 
+    // The mesa driver does bad things with LLVM. Since we also use llvm, 
+    // we can't have the mesa driver do bad things to our code.
+    // Therefore we should preload any versions of llvm that may be required
+    // into the start of our address space.
```
ben-clayton: Maybe create a bug explaining what the Bad Thing™ is and link to it here. That way, if there's a mesa driver that uses an LLVM version not listed here, people can find the issue and know what they need to do.

AWoloszyn: Good plan! I will do that.

PR #1712, opened 2018-03-06
Allow range selections in the command and state trees.

### LINKED ISSUE #1066: Support command range selections
... if only to copy a bunch of commands to the clipboard.

pmuetschard:
Fixes #1066


gapic/src/main/com/google/gapid/views/AtomTree.java
```diff
@@ -332,7 +332,7 @@ private void updateTree(boolean assumeLoading) {
     private final Map<Long, Color> threadBackgroundColors = Maps.newHashMap();
 
     public CommandTree(Composite parent, Models models, Widgets widgets) {
-      super(parent, SWT.H_SCROLL | SWT.V_SCROLL, widgets);
+      super(parent, SWT.H_SCROLL | SWT.V_SCROLL| SWT.MULTI, widgets);
```
ben-clayton: Nit: whitespace

pmuetschard: Done.

PR #1721, opened 2018-03-06
Switch from protoc encoding to api compile encoding

ben-clayton:


ben-clayton:
Eh? The CL description got replaced... let's try this again.

PR #1722, opened 2018-03-06
cmd/regres: Add trace stats and improve results display

ben-clayton:
Does a capture for one minute and then gathers stats on:
* Number of commands
* Number of draws
* Number of frames

Depends on #1720


ben-clayton:
*ping* ?

PR #1724, opened 2018-03-07
Add an 'ImageAspect' layer to the chain ImageObject->ImageLayer->ImageLevel

Qining:
Now the chain is like: ImageObject->ImageAspect->ImageLayer->ImageLevel.

This is to prepare for copying and priming data of depth stencil images.


AWoloszyn:
Can you make sure this works on all of our existing samples, as well as the UI. (Texture view / framebuffer view).

Qining:
Tested with all the samples we have. All the Color or Depth only textures are working well. Depth+Stencil images are not working perfectly. Will be fixed in a separate CL.

PR #1733, opened 2018-03-08
JDWP overhaul

ben-clayton:
Add a debug mode which prints every send and recv.
Add string helpers for the majority of the message types.
Flesh out thread-reference APIs.
Reimplement the event request API to be more strict about thread suspension.


core/java/jdwp/cmdsets.go
```diff
@@ -36,3 +50,249 @@ const (
...
+var cmdNames = map[cmd]string{}
+
+func init() {
+	register := func(c cmd, n string) {
+		if _, e := cmdNames[c]; e {
+			panic("command already registered")
+		}
+		cmdNames[c] = n
+	}
+	register(cmdVirtualMachineVersion, "Version")
```
pmuetschard: why not just make the name a field in the cmd struct?

ben-clayton: Because if we get a cmdSet and cmdID from somewhere else (say the wire), we can use struct-equality to lookup the name.

pmuetschard: Thanks

PR #1735, opened 2018-03-08
Vulkan: Handle D24 format correctly

Qining:
For D24 format with either X8 or S8 channel, the depth element size is
different for copying through buffer and storing in the image. We need
to manually strip the MSB byte when processing the data if the data is
retrieved through buffer image copy.

### DIFF
--- a/gapii/cc/vulkan_mid_execution.cpp
+++ b/gapii/cc/vulkan_mid_execution.cpp
@@ -329,7 +329,8 @@ void VulkanSpy::prepareGPUBuffers(CallObserver *observer, PackEncoder *group,
     mImports.mVkDeviceFunctions[device.second->mVulkanHandle]
         .vkGetBufferMemoryRequirements(
             device.second->mVulkanHandle, buffer,
-            &mState.TransferBufferMemoryRequirements[device.second->mVulkanHandle]);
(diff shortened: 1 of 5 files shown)


gapii/cc/vulkan_mid_execution.cpp
```diff
@@ -813,12 +828,30 @@ void VulkanSpy::prepareGPUBuffers(CallObserver *observer, PackEncoder *group,
...
 
+        if ((image_info.mFormat == VkFormat::VK_FORMAT_X8_D24_UNORM_PACK32 ||
+          image_info.mFormat == VkFormat::VK_FORMAT_D24_UNORM_S8_UINT) &&
+          (aspect_bit == VkImageAspectFlagBits::VK_IMAGE_ASPECT_DEPTH_BIT)) {
+          // The width of the depth channel are different for img buf copy.
+          byte_size_and_extent copy_e =
+            level_size(copy.mimageExtent, image_info.mFormat, 0, aspect_bit, true);
+          size_t element_size_in_img = 3;
+          size_t element_size_in_buf = 4;
+          // It is always the MSB byte to be stripped.
```
AWoloszyn: Is it? I would have expected LSB. Although I have not double-checked the spec

PR #1738, opened 2018-03-09
gapis/api/gles: Fix panic when calling DataTypeSize(GL_INT)

ben-clayton:
Rename the file, and refactor another place we were calculating type size.


ben-clayton:
Now all fixed

PR #1739, opened 2018-03-09
Vulkan: Prime depth image data

Qining:
If the depth image has TRANSFER_DST usage bit, copy into it through
buffer.

If the depth image can be rendered, render to it using an staginge
image.

This does not prime the stencil aspect data unless the image has
TRANSFER_DST bit.

TODO:
 - [x] Staging image allocation should respect the given aspect bit

### DIFF
--- a/core/stream/fmts/BUILD.bazel
+++ b/core/stream/fmts/BUILD.bazel
@@ -33,6 +33,7 @@ go_library(
         "rgb.go",
         "rgba.go",
         "rgbe.go",
+        "s.go",
         "sd.go",
         "x.go",
         "xy.go",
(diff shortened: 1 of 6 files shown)


gapis/api/vulkan/vulkan.api
```diff
@@ -9676,7 +9676,7 @@ enum SemaphoreUpdate {
   u32         Height
   @unused u32 Depth
   @spy_disabled
-  @hidden @nobox @internal u8[] Data
+  @nobox @internal u8[] Data
```
Qining: @hidden will be added back once the CL is done.

PR #1741, opened 2018-03-12
gles: Log compat errors to the report view.

ben-clayton:
Few people will look in the log view to find the reason why shader conversion is going wrong.


ben-clayton:
It already does. For regular replay (non-find-issues) it'll log. Before we were logging these errors multiple times.

AWoloszyn:
SGTM

PR #1742, opened 2018-03-12
Vulkan: Fix miscalucation in BufferImageCopy

Qining:


Qining:
@AWoloszyn Found this when debugging the depth image data priming CL.

PR #1749, opened 2018-03-15
Update to a newer version of SPIRV-Tools.

pmuetschard:
- Also updated SPIRV-Headers
- Updated the SPIRV-Tools generated code
- Added a readme outlining how to upgrade the generated code
- Moved from spirv/1.2 to spirv/unifed1
- Updated our code to work with the new SPIRV-Tools
  - debug sections were split
  - ir::Module is now wrapped with a ir::Context, which is required to construct instructions
  - it++ -> ++it ... the pain is real
  - set_options was renamed to set_common_options
  - other random small API changes


pmuetschard:
noideadog here, please take a close look at the changes to `gapis/shadertools/cc`.

PR #1750, opened 2018-03-16
Vulkan: Cache the pipelines used for priming image data

Qining:
Also refactored the priming logic

Also for the pool size of image level data, switch back to use the 'real' image size instead of the '8 aligned' size. This fixes the issues that Texture view failed to show small textures (like 1x1x1 size images whose size is smaller than 8 bytes).


gapis/api/vulkan/image_primer.go
```diff
@@ -0,0 +1,1573 @@
...
+		inputViews = append(inputViews, view)
+		if view != nil {
+			defer h.sb.write(h.sb.cb.VkDestroyImageView(dev, view.VulkanHandle, memory.Nullptr))
+		} else {
+			return fmt.Errorf("failed to create image view for input attachment image: %v", input.VulkanHandle)
+		}
+	}
+	outputView := h.createImageView(dev, job.renderTarget, job.aspect, job.layer, job.level)
+	if outputView != nil {
+		defer h.sb.write(h.sb.cb.VkDestroyImageView(dev, outputView.VulkanHandle, memory.Nullptr))
```
AWoloszyn: TIL: defer doesn't care about scoping

PR #1756, opened 2018-03-19
Vulkan: Prime stencil image data

Qining:
Sits on top of #1750


Qining:
Tested with our stencil sample (modified so that stencil buffer won't be re-rendered each frame). No errors found validation layers (except the existing ones with swapchain and presentation engine).

PR #1757, opened 2018-03-19
Refactor our WORKSPACE into callable macros and deal with long Windows paths.

pmuetschard:
- Refactor our WORKSPACE into callable macros. 
  This makes it possible for other projects to inlcude GAPID in their bazel build.
- Deal with long file names on Windows
  - Adds wrappers for gcc and ar that fixup paths into the \\?\ form, so they are not subject to the ~250 char limit.
  - Uses a slightly newer version of rules_go, which has some long path fixes.


tools/build/mingw_toolchain/file_collector.cpp
```diff
@@ -0,0 +1,218 @@
...
+
+namespace {
+
+std::string GetCWD() {
+  char buffer[1024];
+  _getcwd(buffer, sizeof(buffer));
+  return R"(\\?\)" + std::string(buffer) + '\\';
+}
+
+// Attemps to create a new folder in the temporary folder based off the PID.
```
ben-clayton: `// Attempts`

tools/build/mingw_toolchain/file_collector.cpp
```diff
@@ -0,0 +1,218 @@
...
+#include <string>
+#include <vector>
+#include <windows.h>
+
+#include "file_collector.h"
+
+namespace {
+
+std::string GetCWD() {
+  char buffer[1024];
```
ben-clayton: 1024 is a rather arbitrary number, please define a constant.

tools/build/mingw_toolchain/file_collector.h
```diff
@@ -0,0 +1,115 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <string>
+#include <vector>
+
+// Keeps track of tempoarary files and processes @ params files
```
ben-clayton: `temporary`

tools/build/mingw_toolchain/file_collector.cpp
```diff
@@ -0,0 +1,218 @@
...
+
+FileCollector::FileCollector(const std::string& baseName) :
+  baseName_(baseName),
+  cwd_(GetCWD()),
+  tmpDir_(""),
+  counter_(0) {
+}
+
+FileCollector::~FileCollector() {
+}
```
ben-clayton: Automatically clean up in destructor?

tools/build/mingw_toolchain/file_collector.h
```diff
@@ -0,0 +1,115 @@
...
+ private:
+  // Returns the path to a new temporary @ params file.
+  const std::string& NewParamsFile();
+
+  // Handles a line/path read from a @ params file. p should contain the
+  // index of the last '/' in path. If isOutput is true, the path is
+  // processed as an output of the wrapped binary, otherwhise as an input.
+  std::string HandleParam(const std::string& path, std::string::size_type p, bool isOutput);
+
+  typedef struct {
```
ben-clayton: No need for typedef in C++.

```c++
struct Output {
  std::string final_;
  std::string temp_;
}
```

tools/build/mingw_toolchain/file_collector.h
```diff
@@ -0,0 +1,115 @@
...
+  // List of all outputs expected from the wrapped binary and their
+  // final expected location.
+  std::vector<Output> outputs_;
+};
+
+// Collects arguments to the wrapped binary.
+class ArgumentCollector {
+ public:
+  // cmd is the name of the wrapped binary and will be arg[0].
+  ArgumentCollector(const std::string& cmd) {
```
ben-clayton: Any reason why all these methods are inline?

tools/build/mingw_toolchain/file_collector.cpp
```diff
@@ -0,0 +1,218 @@
...
+void FileCollector::Cleanup() {
+  for (auto it = tmpFiles_.begin(); it != tmpFiles_.end(); it++) {
+    DeleteFile(it->c_str());
+  }
+  if (tmpDir_ != "") {
+    RemoveDirectory(tmpDir_.c_str());
+  }
+}
+
+const std::string& FileCollector::NewParamsFile() {
```
ben-clayton: No good reason to use a `const &` here. Performance is not an issue, and you're returning a reference to an element in a vector. If the vector resizes, then the reference can be garbage. Doesn't look like this would happen given it's only usage, but is weird all the same.

tools/build/mingw_toolchain/file_collector.cpp
```diff
@@ -0,0 +1,218 @@
...
+}
+
+void FileCollector::Cleanup() {
+  for (auto it = tmpFiles_.begin(); it != tmpFiles_.end(); it++) {
+    DeleteFile(it->c_str());
+  }
+  if (tmpDir_ != "") {
+    RemoveDirectory(tmpDir_.c_str());
+  }
+}
```
ben-clayton: nit: clear `tmpFiles_` and `tmpDir_`?

tools/build/mingw_toolchain/file_collector.h
```diff
@@ -0,0 +1,115 @@
...
+#include <vector>
+
+// Keeps track of tempoarary files and processes @ params files
+class FileCollector {
+ public:
+  FileCollector(const std::string& baseName);
+  ~FileCollector();
+
+  enum ParamsType {
+    GCC, AR
```
ben-clayton: Please can we have a little bit of documentation here to describe the differences between the two of these?

tools/build/mingw_toolchain/file_collector.cpp
pmuetschard: I didn't do that, because I want it to leave the temp files around if there is an error, so it can be debugged.

tools/build/mingw_toolchain/file_collector.cpp
pmuetschard: Done.

tools/build/mingw_toolchain/file_collector.cpp
pmuetschard: Done.

tools/build/mingw_toolchain/file_collector.h
pmuetschard: Done.

tools/build/mingw_toolchain/file_collector.h
pmuetschard: Done.

tools/build/mingw_toolchain/file_collector.h
pmuetschard: Done.

tools/build/mingw_toolchain/file_collector.h
pmuetschard: Done.

tools/build/mingw_toolchain/file_collector.cpp
```diff
@@ -0,0 +1,257 @@
...
+  if (result.back() != '\\') {
+    result.push_back('\\');
+  }
+  delete cwd;
+  return result;
+}
+
+// Attempts to create a new folder in the temporary folder based off the PID.
+std::string GetTempDir(const std::string& baseName) {
+  TCHAR buffer[MAX_PATH + 1 + baseName.length() + 5 + 1];
```
ben-clayton: Weird mix of `TCHAR` and `char` functions here. I'd just stick with good old `char`, and use the `GetTempPathA` explicit variant.

`MAX_PATH + 1 + baseName.length() + 5 + 1` is a bit weird too. Technically correct, but not sure it is intentionally correct. I assume the `+ 1` is for a null terminator, not sure why we have two. `+ 5` I'm guessing is for `"%s_%04x\\"`, but I count 6 characters.

pmuetschard: Done. `GetTempPath` returns the length excluding the terminator. The docs say it's max return value is `MAX_PATH + 1`. Yes, it should have been `+ 6`, I missed the trailing `\`...

PR #1763, opened 2018-03-27
Tests, tests, tests....

pmuetschard:
Get all our tests building and... passing.

`bazel test tests` is your new best friend.

The https://github.com/google/gapid/commit/2eb009bca3620923653de12912197a49d5a49161 commit needs special attention (fixes boxing/unboxing of dictionaries). Types implementing the `dictionary.Provider` interface are boxed as maps. The unboxing needs to correctly "undo" this.

As do the template changes in commit  https://github.com/google/gapid/commit/ab33dbf01e03dc9b2d48d0ac49af189534aefe72 (fixes (de)serialization of maps to/from proto). Maps are supposed to be reference types, but the reference matching does not treat them that way. Null maps were also not handled correctly.


core/os/device/abi_test.go
```diff
@@ -35,7 +35,7 @@ var abiTestData = []struct {
 	{device.AndroidARMv7a, "armeabi-v7a", 32, 4, 4, 4, device.LittleEndian},
 	{device.AndroidARM64v8a, "arm64-v8a", 64, 8, 8, 8, device.LittleEndian},
 	{device.AndroidX86, "x86", 32, 4, 4, 4, device.LittleEndian},
-	{device.AndroidX86_64, "x86-64", 64, 8, 8, 8, device.LittleEndian},
+	{device.AndroidX86_64, "x86-64", 64, 8, 8, 4, device.LittleEndian},
```
ben-clayton: This test is exceptionally stupid - it is testing that constants defined in one place match those defined here. Happy to just delete this.

PR #1768, opened 2018-03-28
Enable compiling GLSL to SPIR-V

Qining:
So that we can compile shaders in GAPID. The first use of it is generating the shaders used in priming image data for Vulkan mid-execution capture.


Qining:
CL description added.

PR #1770, opened 2018-03-29
Vulkan: Prime data for storage images

Qining:
Prime the data of images which have STORAGE_BIT for Vulkan MEC

TODO:
- ~~handle depth/stencil storage images~~ (we just returns an error and report to the report view to warn the user for such unusual case, currently depth/stencil image won't be supported in GAPID for data priming through imageStore)


Qining:
One question. Can there be a Depth/Stencil image to be used in compute shader as storage image?

AWoloszyn:
I think for now we can just warn about this case if we ever come upon it. I don't expect to ever see it happen, or find a driver that supports it anyway.

gapis/api/vulkan/image_primer.go
```diff
@@ -210,6 +261,519 @@ func (p *imagePrimer) allocStagingImages(img *ImageObject, aspect VkImageAspectF
...
+	// parent store job
+	job *ipStoreJob
+	// descriptor related
+	descSet  VkDescriptorSet
+	pipeline *ComputePipelineObject
+}
+
+// ipStoreDispatchInfo contains the information to submit a dispatch to store
+// staging data to the target image. It is guaranteed that each pixel of the
+// target image will be processed by an local invocation.
```
AWoloszyn: "by a local invocation"

Qining:
Squashed and merged as all build pass for all real changes.

PR #1779, opened 2018-04-05
More robot fixes to get buildbot uploads working

pmuetschard:
- allow querying on enums by representing their names as strings (e.g. `struct.EnumField == "EnumName"`)
- don't require unique tags for every build bot package
- only merge build bot packages based on CL (and fix the test to check for type of both the merger and the mergee)
- allow a `robot upload build` to specify that it is a build bot build and clean up how things are inferred from the local git.

### DIFF
--- a/test/robot/search/eval/eval.go
+++ b/test/robot/search/eval/eval.go
@@ -16,6 +16,7 @@ package eval
 
 import (
 	"context"
+	"fmt"
 	"reflect"
 	"regexp"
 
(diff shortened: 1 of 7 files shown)


test/robot/search/eval/eval.go
```diff
@@ -141,12 +143,22 @@ func compileEqual(ctx context.Context, expr *search.Binary, t reflect.Type) (eva
...
 		return nil, boolType, err
 	}
-	if lt != rt {
-		return nil, boolType, log.Err(ctx, nil, "Types to equal do not match")
+	switch {
+	case lt == rt:
+		return func(ctx context.Context, value interface{}) interface{} {
+			return reflect.DeepEqual(lhs(ctx, value), rhs(ctx, value))
+		}, boolType, nil
+	case lt.Implements(stringerType) && rt.Kind() == reflect.String:
```
ben-clayton: Instead of checking for `stringerType`, you can just use `fmt.Sprint` with the value. It does magic to print the value if it supports [`Stringer`](https://golang.org/pkg/fmt/#Stringer), [`Formater`](https://golang.org/pkg/fmt/#Formatter), and has special cases for most other types.

pmuetschard: Done.

PR #1783, opened 2018-04-09
Restructure dictionaries.

ben-clayton:
These CLs set us up to being able to implement maps in C.


core/data/dictionary/dictionary.go
```diff
@@ -33,7 +33,7 @@ type I interface {
 	Entries() []Entry
 	KeyTy() reflect.Type
 	ValTy() reflect.Type
-	KeysSorted() []interface{}
+	Keys() []interface{}
```
pmuetschard: maybe add a comment here, that the keys are expected to be sorted?

core/data/generic/generic.go
```diff
@@ -0,0 +1,197 @@
...
+	Errors []error
+
+	// Bindings of generic type to matched type.
+	Bindings map[reflect.Type]reflect.Type
+}
+
+// Ok returns true if the type implemented the generic interface.
+func (m Match) Ok() bool { return len(m.Errors) == 0 }
+
+// Implements returns nil if ty implements the generic interface iface,
```
pmuetschard: this function does not return a pointer and thus will never return nil.

core/data/dictionary/dictionary.go
```diff
@@ -23,44 +23,68 @@ import (
...
+	// Add inserts the key-value pair, replacing any existing entry with the
+	// same key.
+	Add(key, val interface{})
+	// Lookup searches for the value of the entry with the given key.
+	Lookup(key interface{}) (val interface{}, found bool)
+	// Contains returns true if the dictionary contains an entry with the given
+	// key.
 	Contains(key interface{}) bool
-	Delete(key interface{})
-	Clear()
```
pmuetschard: Should we keep `Clear`? I know go doesn't have a way to clear a map, but I don't think we should impose that on other implementations. The overhead of delete-one-at-atime to delete-all-at-once, especially with the keys sorting, seems like it can get quite large.

ben-clayton: I'm trying to keep the interface surface small.
There are actually no used of `Clear` in our codebase (with exception to the test). If we find in the future that it is a bottleneck, I'd probably opt for creating a `Clearer` interface that `Clear` uses as an optional fast path.

core/data/dictionary/dictionary.go
ben-clayton: They are not expected to be sorted. There's only expectations in the serialization, which is not part of this contract. However, you are right that there should be comments (which are coming in a later CL).

core/data/generic/generic.go
ben-clayton: Good spot on the stale comment. Will fix.

PR #1785, opened 2018-04-10
Vulkan: Handle the overflow descriptors at vkUpdateDescriptorSets correctly

### LINKED ISSUE #1784: Overflow descriptors are not handled correctly
https://github.com/google/gapid/blob/master/gapis/api/vulkan/vulkan.api#L5917
The binding numbers can be sparse and not consecutive, using `len()` as the `maximum allowed binding + 1` is not correct.

https://github.com/google/gapid/blob/master/gapis/api/vulkan/vulkan.api#L5922
Bindings may not exist in the descriptor set, this can cause a crash in such cases.

Qining:
This is to fix #1784


Qining:
Will work on removing the extern functions in vulkan/errors.api soon, if the conflict of importing `../../messages/message.api` in both gles.api and vulkan.api has been resolved.

PR #1791, opened 2018-04-12
Vulkan: Split the vulkan.api into multiple files

Qining:
Sits on top of #1786 , waits until #1786 is tested and merged.


Qining:
All three build passed before the final rebase, and the final rebase did nothing actually. Merged

PR #1798, opened 2018-04-13
bazel: Don't use named arguments in bazel rules.

ben-clayton:
Fixes build.

### DIFF
--- a/tools/build/rules/common.bzl
+++ b/tools/build/rules/common.bzl
@@ -59,7 +59,7 @@ def _copy_to_impl(ctx):
                 filtered += [src]
     outs = depset()
     for src in filtered:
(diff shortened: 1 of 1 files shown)


tools/build/rules/common.bzl
```diff
@@ -59,7 +59,7 @@ def _copy_to_impl(ctx):
                 filtered += [src]
     outs = depset()
     for src in filtered:
-        dstname = ctx.attr.rename.get(key = src.basename, default = src.basename)
+        dstname = ctx.attr.rename.get(src.basename, src.basename)
```
pmuetschard: please leave `default = ` in.

ben-clayton: Done

PR #1802, opened 2018-04-15
core/cc: Work around #1788

ben-clayton:
Fixes: #1788


ben-clayton:
> Should we keep the bug open, or is there no hope in getting this to work for tests on OSX?

To be completely honest, I haven't done much research on the problem at all. Assuming the build machines test on Linux, I don't mind having the bug closed.

PR #1803, opened 2018-04-15
gapii/cc: Fix assert when MEC tracing.

### LINKED ISSUE #1801: Can't trace mid application with 1.1.0
Trying to use `Trace From Beginning` off and when I run the application just starts up and closes after a frame with nothing traced and get this output

```
<gapit> Capturing: 16B in 0s
<gapit> EOF: 16B
```

`--log-level DEBUG --gapis-log-level DEBUG --gapir-log-level DEBUG` gave me nothing extra to report

This is with `Version 1.1.0:developer`. This feature works correctly with the `Version 1.0.3:release` though, so think its something in current development thats breaking it

----

Host: `Ubuntu 16.04 - Linux 4.4.0`
Device: `Galaxy S8 [Adreno 540]`

ben-clayton:
There was a mismatch of counts when `is_suspended` returned true.
Also added another assert to check the nested counts don't climb something crazy.

Fixes #1801

### DIFF
--- a/gapii/cc/spy.cpp
+++ b/gapii/cc/spy.cpp
@@ -431,10 +431,11 @@ void Spy::onPostDrawCall(CallObserver* observer, uint8_t api) {
 }
 
 void Spy::onPreStartOfFrame(CallObserver* observer, uint8_t api) {
-    if (is_suspended()) {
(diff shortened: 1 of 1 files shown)


ben-clayton:
> Thanks for cleaning up my mess.

About time I returned the favor. 😉

PR #1833, opened 2018-04-24
Update to read_framebuffer to allow compute presentation.

AWoloszyn:
This does 2 things, first it allows us to copy out an image from a Vulkan compute queue.

Secondly it allows us to choose to not resize the images in the server. This is primarily so that the client can resample how it chooses rather than forcing us to do a software resize in GAPIS.


gapis/api/api.go
```diff
@@ -46,7 +46,7 @@ type API interface {
 		after []uint64,
 		state *GlobalState,
 		thread uint64,
-		attachment FramebufferAttachment) (width, height, index uint32, format *image.Format, err error)
+		attachment FramebufferAttachment) (width, height, index uint32, format *image.Format, canResize bool, err error)
```
ben-clayton: There comes a point where a struct is preferred over N return values 😃. I suggest declaring a `FramebufferAttachmentInfo` struct in this file with `Width`, `Height`, `Index`, `Format` and `CanResize` fields.

PR #1840, opened 2018-04-26
Fix the hidden flags and flag help

### LINKED ISSUE #433: core/app flag behavior is funky
I always forget what the precise position, casing, and values of it are, and gapit --help does not help

pmuetschard:
Flags with a help text that starts with an underscore are hidden by default. Currently this is implemented incorrectly, as the flags (sometimes) are printed simply without their help text. With this change,
 - hidden flags are actually hidden and not shown by default
 - if not all flags are shown in the usage text, a message is shown referring to the `-fullhelp` flag
 - Adding of a `-fullhelp` flag (global and on every verb) that is similar to the built-in `-help` flag, causing the usage to be printed with hidden flags shown. (Note, however, that unlike `-help`, flag parsing doesn't immediately stop on `-fullhelp`. This is a limitation of using golang's builtin flag parsing.)

Fixes #433


pmuetschard:
Bug #433 also references the following quirkiness:

```
gapis help

< runs gapis >
```

Fixing that is a larger change (not all binaries use the "verbs" system), which I can do if desired...

core/app/flags.go
```diff
@@ -24,6 +24,7 @@ type (
 		Analytics   string `help:"_If non-empty enable analytics using the specified user-id"`
 		CrashReport bool   `help:"_Automatically send crash reports to Google"`
 		DecodeStack string `help:"_Decode a stackdump generated by this executable"`
+		HelpFull    bool   `help:"_Display the full help"`
```
ben-clayton: Suggest `FullHelp` or `HelpVerbose` to avoid reading like helpful?
Also: Should this be hidden? Bit of a chicken-and-egg here.

pmuetschard: The `-fullhelp` flag, like the built-in `-help` flag is not shown in the help text. It is referred to, though, by the message if there are omitted flags.

PR #1844, opened 2018-04-30
Vulkan: Fix the hang at tracing due to a bug in memory tracker

Qining:
std::prev(some_container.begin(), 1) is an undefined behavior, should
not rely on it.


Qining:
@pmuetschard PTAL, this should fix the hang on our tests.

PR #1845, opened 2018-05-01
Dependency graph prerequisites

bjoeris:
These are a number of bug fixes and changes which are necessary for, but independent of the dependency graph work.

### DIFF
--- a/gapii/cc/vulkan_inlines.inc
+++ b/gapii/cc/vulkan_inlines.inc
@@ -42,3 +42,7 @@ inline void VulkanSpy::vkErrInvalidDescriptorArrayElement(CallObserver*, uint64_
 inline void VulkanSpy::vkErrCommandBufferIncomplete(CallObserver*, VkCommandBuffer cmdbuf) {
     GAPID_WARNING("Error: Executing command buffer %llu was not in the COMPLETED state", cmdbuf)
 }
+
+inline void VulkanSpy::vkErrImageLayout(CallObserver*, uint32_t layout, uint32_t expectedLayout) {
+  GAPID_WARNING("Error: Image was in layout %lu, but was expected to be in layout %lu", layout, expectedLayout);
+}
--- a/gapii/cc/vulkan_inlines.inc
+++ b/gapii/cc/vulkan_inlines.inc
@@ -38,3 +38,7 @@ inline void VulkanSpy::vkErrExpectNVDedicatedlyAllocatedHandle(CallObserver*, st
 inline void VulkanSpy::vkErrInvalidDescriptorArrayElement(CallObserver*, uint64_t set, uint32_t binding, uint32_t array_index) {
   GAPID_WARNING("Error: Invalid descriptor array element specified by descriptor set: %llu, binding: %lu, array index: %lu", set, binding, array_index);
 }
+
+inline void VulkanSpy::vkErrCommandBufferIncomplete(CallObserver*, VkCommandBuffer cmdbuf) {
+    GAPID_WARNING("Error: Executing command buffer %llu was not in the COMPLETED state", cmdbuf)
+}
--- a/gapis/api/vulkan/api/coherent_memory.api
+++ b/gapis/api/vulkan/api/coherent_memory.api
@@ -56,39 +56,46 @@ sub void readCoherentMemory(ref!DeviceMemoryObject memory, VkDeviceSize readOffs
   }
 }
 
-sub void readCoherentMemoryInBuffer(ref!BufferObject buffer, VkDeviceSize readOffset, VkDeviceSize readSize) {
+sub void readMemoryInBuffer(ref!BufferObject buffer, VkDeviceSize readOffset, VkDeviceSize readSize) {
   mem := buffer.Memory
   if mem != null {
-    if as!u64(readSize) == 0xFFFFFFFFFFFFFFFF {
-      readCoherentMemory(mem, buffer.MemoryOffset + readOffset, buffer.Info.Size - readOffset)
-    } else {
-      readCoherentMemory(mem, buffer.MemoryOffset + readOffset, readSize)
+    readBegin := buffer.MemoryOffset + readOffset
+    actualReadSize := switch as!u64(readSize) {
+      case 0xFFFFFFFFFFFFFFFF:
+        buffer.Info.Size - readOffset
+      default:
+        readSize
+    }
+    readEnd := readBegin + actualReadSize
+    readCoherentMemory(mem, readBegin, actualReadSize)
+    if len(mem.Data) != 0 {
+      read(mem.Data[readBegin : readEnd])
     }
   }
 }
 
-sub void readCoherentMemoryInBufferBindings(map!(u32, ref!VkDescriptorBufferInfo) bufferBindings) {
+sub void readMemoryInBufferBindings(map!(u32, ref!VkDescriptorBufferInfo) bufferBindings) {
   for _, _, v in bufferBindings {
     if v.Buffer != as!VkBuffer(0) {
       if (v.Buffer in Buffers) {
-        readCoherentMemoryInBuffer(Buffers[v.Buffer], v.Offset, v.Range)
+        readMemoryInBuffer(Buffers[v.Buffer], v.Offset, v.Range)
       }
     }
   }
 }
 
-sub void readCoherentMemoryBufferViewBindings(map!(u32, VkBufferView) bufferViews) {
+sub void readMemoryBufferViewBindings(map!(u32, VkBufferView) bufferViews) {
   for _, _, v in bufferViews {
     if v in BufferViews {
       view := BufferViews[v]
       if (view.Buffer.VulkanHandle in Buffers) {
-        readCoherentMemoryInBuffer(view.Buffer, view.Offset, view.Range)
+        readMemoryInBuffer(view.Buffer, view.Offset, view.Range)
       }
     }
   }
 }
 
-sub void readCoherentMemoryInDescriptorSet(ref!DescriptorSetObject descriptor_set) {
+sub void readMemoryInDescriptorSet(ref!DescriptorSetObject descriptor_set) {
   for _, _, v in descriptor_set.Bindings {
     switch v.BindingType {
       case
@@ -96,18 +103,18 @@ sub void readCoherentMemoryInDescriptorSet(ref!DescriptorSetObject descriptor_se
         VK_DESCRIPTOR_TYPE_STORAGE_BUFFER,
         VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER_DYNAMIC,
         VK_DESCRIPTOR_TYPE_STORAGE_BUFFER_DYNAMIC:
-          readCoherentMemoryInBufferBindings(v.BufferBinding)
+          readMemoryInBufferBindings(v.BufferBinding)
       case VK_DESCRIPTOR_TYPE_UNIFORM_TEXEL_BUFFER,
         VK_DESCRIPTOR_TYPE_STORAGE_TEXEL_BUFFER:
-          readCoherentMemoryBufferViewBindings(v.BufferViewBindings)
+          readMemoryBufferViewBindings(v.BufferViewBindings)
       default: {}
     }
   }
 }
 
-sub void readCoherentMemoryInBoundDescriptorSets() {
+sub void readMemoryInBoundDescriptorSets() {
   for _, _, s in lastDrawInfo().DescriptorSets {
-    readCoherentMemoryInDescriptorSet(s)
+    readMemoryInDescriptorSet(s)
   }
 }
 
@@ -126,12 +133,12 @@ sub void readCoherentMemoryInImage(ref!ImageObject image) {
   }
 }
 
-sub void readCoherentMemoryInCurrentPipelineBoundVertexBuffers(u32 vertexCount, u32 instanceCount, u32 firstVertex, u32 firstInstance) {
+sub void readMemoryInCurrentPipelineBoundVertexBuffers(u32 vertexCount, u32 instanceCount, u32 firstVertex, u32 firstInstance) {
   for _ , _ , vertex_binding in lastDrawInfo().GraphicsPipeline.VertexInputState.BindingDescriptions {
     if vertex_binding.binding in lastDrawInfo().BoundVertexBuffers {
       bound_vertex_buffer := lastDrawInfo().BoundVertexBuffers[vertex_binding.binding]
       backing_buf := bound_vertex_buffer.Buffer
-      if ((backing_buf.Memory != null) && IsMemoryCoherent(backing_buf.Memory)) {
+      if backing_buf.Memory != null {
         start_vertex := switch vertex_binding.inputRate {
           case VK_VERTEX_INPUT_RATE_VERTEX:
             firstVertex
@@ -151,7 +158,7 @@ sub void readCoherentMemoryInCurrentPipelineBoundVertexBuffers(u32 vertexCount, 
           case false:
             as!VkDeviceSize(num_vertices * vertex_binding.stride)
         }
-        readCoherentMemoryInBuffer(backing_buf, start_offset, num)
+        readMemoryInBuffer(backing_buf, start_offset, num)
       }
     }
   }
--- a/gapis/api/vulkan/api/coherent_memory.api
+++ b/gapis/api/vulkan/api/coherent_memory.api
@@ -143,7 +143,85 @@ sub void readMemoryInDescriptorSet(ref!DescriptorSetObject descriptor_set, map!(
   }
 }
 
-sub void readMemoryInBoundDescriptorSets(
+sub void writeMemoryInBuffer(ref!BufferObject buffer, VkDeviceSize offset, VkDeviceSize size) {
+  mem := buffer.Memory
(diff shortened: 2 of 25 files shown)


pmuetschard:
Please drop `Bazel: Disable binary stripping when building under "-c dbg"` This is already done at head anyways.

bjoeris:
@pmuetschard Done

gapis/api/vulkan/api/queue.api
```diff
@@ -164,6 +171,10 @@ cmd VkResult vkQueueSubmit(
   leaveSubcontext()
   execPendingCommands(queue)
   fence
+  if (fence != 0) {
```
AWoloszyn: I am quite surprised this doesn't collide with the fence keyword above.

gapis/api/vulkan/api/queue.api
```diff
@@ -173,6 +184,7 @@ cmd VkResult vkQueueSubmit(
 cmd VkResult vkQueueWaitIdle(
     VkQueue queue) {
   if !(queue in Queues) { vkErrorInvalidQueue(queue) }
+  Queues[queue] = Queues[queue]
```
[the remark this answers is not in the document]
bjoeris: This is to make vkQueueWaitIdle read and write the relevant queue. Without this, vkQueueWaitIdle doesn't touch the state, will have no dependencies, and will be dropped by the new DCE.

gapis/api/vulkan/api/renderpass_framebuffer.api
```diff
@@ -344,3 +344,22 @@ cmd void vkCmdEndRenderPass(
...
+
+sub void storeImageAttachment(u32 attachmentID) {
+  VK_ATTACHMENT_UNUSED := as!u32(0xFFFFFFFF)
+  if attachmentID != VK_ATTACHMENT_UNUSED {
+    attachment := lastDrawInfo().Framebuffer.ImageAttachments[attachmentID]
+    desc := lastDrawInfo().RenderPass.AttachmentDescriptions[attachmentID]
+    switch desc.storeOp {
+      case VK_ATTACHMENT_STORE_OP_STORE: {
+        if desc.initialLayout != desc.finalLayout {
+          transitionImageLayout(attachment.Image, attachment.SubresourceRange, VK_IMAGE_LAYOUT_UNDEFINED, desc.finalLayout)
```
Qining: So if the `storeOp` is not `VK_ATTACHMENT_STORE_OP_STORE`, can we skip the layout transition?

gapis/api/vulkan/api/queue.api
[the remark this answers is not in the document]
Qining: Nice, nice if we have a comment here.

gapis/api/vulkan/errors.api
```diff
@@ -43,6 +43,7 @@ extern void vkErrNullPointer(string pointerType)
 extern void vkErrUnrecognizedExtension(string name)
 extern void vkErrExpectNVDedicatedlyAllocatedHandle(string handleType, u64 handle)
 extern void vkErrInvalidDescriptorArrayElement(u64 set, u32 binding, u32 arrayIndex)
+extern void vkErrImageLayout(VkImageLayout layout, VkImageLayout expectedLayout)
```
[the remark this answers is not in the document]
ben-clayton: I'm fine with this for now.

gapis/api/vulkan/api/renderpass_framebuffer.api
AWoloszyn: We should do the layout transition regardless, but only write the image subresource if we are storing the image.

gapis/api/vulkan/extensions/khr_swapchain.api
```diff
@@ -266,7 +266,11 @@ cmd VkResult vkQueuePresentKHR(
   if (info.pWaitSemaphores != null) {
     wait_semaphores := info.pWaitSemaphores[0:info.waitSemaphoreCount]
     for i in (0 .. info.waitSemaphoreCount) {
-      Semaphores[wait_semaphores[i]].Signaled = false
+      if (Semaphores[wait_semaphores[i]].Signaled) {
+        Semaphores[wait_semaphores[i]].Signaled = false
+      } else {
+        // TODO: handle case when semaphore is not yet signaled
```
AWoloszyn: This is going to be a weird, (and quite rare) case. In fact since no application I have ever seen uses events, I would posit that we will never actually hit this edge-case.

However we definitely should still handle it.
Would you mind adding a comment here describing the case in some detail (alternatively, opening a github issue, and linking to it).

The only "legal" time this can happen is if we signal a semaphore after a VkCmdWaitForEvents, and then submit this before the associated VkCmdSetEvents/VkSetEvent is hit. If you can see any other legal cases, please also include them (and let me know)

bjoeris: Added bug (https://github.com/google/gapid/issues/1860) and link in the TODO comment.

gapis/api/vulkan/api/queue.api
```diff
@@ -109,6 +110,9 @@ cmd VkResult vkQueueSubmit(
     for j in (0 .. info.commandBufferCount) {
       enterSubcontext()
       cb := CommandBuffers[command_buffers[j]]
+      if cb.Recording != COMPLETED {
+        vkErrorInvalidCommandBuffer(command_buffers[j])
```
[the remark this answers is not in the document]
bjoeris: Added vkErrorCommandBufferIncomplete for this error.

gapis/api/vulkan/api/queue.api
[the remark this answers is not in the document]
bjoeris: Upon further reflection, this is not always adequate to find dependencies on vkQueueWaitIdle. Reverting for now, will address in the dependency graph CL.

gapis/api/vulkan/api/queue.api
bjoeris: Added some comments for the the next time someone has to read this confusing overloading of 'fence'.

gapis/api/vulkan/api/renderpass_framebuffer.api
bjoeris: storeImageAttachment now transitions the image layout regardless of the storeOp.

bjoeris:
A few of my changes had gotten lost in rebasing before initially submitting the pull request. In addition to addressing the comments, the latest version of this pull request made changes in the following places (compared to the version which was reviewed):
- coherent_memory.api: 
  - readmemoryInBufferBindings
  - writeMemoryInBufferBindings
  - readWriteMemoryInBoundDescriptorSets
- properties_features_requirements.api
  - vkGetBufferMemoryRequirements
- renderpass_framebuffer.api
  - dovkCmdBeginRenderPass
  - dovkCmdNextSubpass
  - loadImageAttachment

gapii/cc/vulkan_inlines.inc
```diff
@@ -38,3 +38,11 @@ inline void VulkanSpy::vkErrExpectNVDedicatedlyAllocatedHandle(CallObserver*, st
 inline void VulkanSpy::vkErrInvalidDescriptorArrayElement(CallObserver*, uint64_t set, uint32_t binding, uint32_t array_index) {
   GAPID_WARNING("Error: Invalid descriptor array element specified by descriptor set: %llu, binding: %lu, array index: %lu", set, binding, array_index);
 }
+
+inline void VulkanSpy::vkErrCommandBufferIncomplete(CallObserver*, VkCommandBuffer cmdbuf) {
+    GAPID_WARNING("Error: Executing command buffer %llu was not in the COMPLETED", cmdbuf)
```
AWoloszyn: not in the COMPLETED state"

gapis/api/vulkan/api/draw_commands.api
```diff
@@ -94,7 +94,7 @@ class CmdBindBuffer {
 
 sub void dovkCmdBindVertexBuffers(ref!vkCmdBindVertexBuffersArgs bind) {
   for i , _ , v in bind.Buffers {
-    lastDrawInfo().BoundVertexBuffers[as!u32(i)] = BoundBuffer(
+    lastDrawInfo().BoundVertexBuffers[as!u32(i) + bind.FirstBinding] = BoundBuffer(
```
AWoloszyn: Good catch

gapis/api/vulkan/api/queue.api
```diff
@@ -163,7 +170,11 @@ cmd VkResult vkQueueSubmit(
   }
   leaveSubcontext()
   execPendingCommands(queue)
-  fence
+  fence // 'fence' keyword, marking the point where observed memory writes become visible
+  if (fence != 0) { // 'fence' parameter, unrelated to 'fence' keyword above
```
AWoloszyn: Thanks for the comments

Qining:
Looks good to me

ben-clayton:
Will take a final pass look after lunch.

gapii/cc/vulkan_inlines.inc
bjoeris: Fixed

gapis/api/vulkan/api/coherent_memory.api
```diff
@@ -56,58 +56,170 @@ sub void readCoherentMemory(ref!DeviceMemoryObject memory, VkDeviceSize readOffs
...
+    if len(mem.Data) != 0 {
+      read(mem.Data[readBegin : readEnd])
     }
   }
 }
 
-sub void readCoherentMemoryInBufferBindings(map!(u32, ref!VkDescriptorBufferInfo) bufferBindings) {
-  for _, _, v in bufferBindings {
+sub void readMemoryInBufferBindings(map!(u32, ref!VkDescriptorBufferInfo) bufferBindings, map!(u32, VkDeviceSize) bufferBindingOffsets) {
+  for k, _, v in bufferBindings {
```
ben-clayton: Not sure if the logic is wrong here, but the triple is index, key, value. `k` is an odd variable choice for index, unless `k` always equals `i`.

gapis/api/vulkan/api/coherent_memory.api
```diff
@@ -56,58 +56,170 @@ sub void readCoherentMemory(ref!DeviceMemoryObject memory, VkDeviceSize readOffs
...
+    }
+    end := begin + actualSize
+    if len(mem.Data) != 0 {
+      write(mem.Data[begin : end])
+    }
+  }
+}
+
+sub void writeMemoryInBufferBindings(map!(u32, ref!VkDescriptorBufferInfo) bufferBindings, map!(u32, VkDeviceSize) bufferBindingOffsets) {
+  for k, _, v in bufferBindings {
```
ben-clayton: ditto

gapis/api/vulkan/api/renderpass_framebuffer.api
```diff
@@ -210,39 +210,16 @@ vkCmdBeginRenderPassArgs {
 }
 
 sub void dovkCmdBeginRenderPass(ref!vkCmdBeginRenderPassArgs args) {
-  VK_ATTACHMENT_UNUSED := as!u32(0xFFFFFFFF)
   lastDrawInfo().Framebuffer = Framebuffers[args.Framebuffer]
   lastDrawInfo().LastSubpass = 0
   lastDrawInfo().RenderPass = RenderPasses[args.RenderPass]
   lastDrawInfo().InRenderPass = true
-  for _ , _ , v in lastDrawInfo().Framebuffer.ImageAttachments {
+  for _ , i , v in lastDrawInfo().Framebuffer.ImageAttachments {
```
ben-clayton: humm...

gapis/api/vulkan/api/renderpass_framebuffer.api
```diff
@@ -321,8 +302,8 @@ vkCmdEndRenderPassArgs {
 }
 
 sub void dovkCmdEndRenderPass(ref!vkCmdEndRenderPassArgs unused) {
-  for _ , i , a in lastDrawInfo().RenderPass.AttachmentDescriptions {
-    lastDrawInfo().Framebuffer.ImageAttachments[i].Image.Info.Layout = a.finalLayout
+  for _ , i , _ in lastDrawInfo().RenderPass.AttachmentDescriptions {
```
ben-clayton: again, `i` vs `k`

gapis/api/vulkan/api/coherent_memory.api
Qining: I think @ben-clayton caught a very important point! It looks like what we want here is `descriptor array index`, so we should use `for _, k, v in bufferBindings {`.

I think there may be similar spots around? And I think we should try it through all the samples and tests we have?

gapis/api/vulkan/api/coherent_memory.api
```diff
@@ -56,58 +56,170 @@ sub void readCoherentMemory(ref!DeviceMemoryObject memory, VkDeviceSize readOffs
...
+sub void readMemoryInDescriptorSet(ref!DescriptorSetObject descriptor_set, map!(u32, map!(u32, VkDeviceSize)) bufferBindingOffsets) {
+  for j, _, v in descriptor_set.Bindings {
     switch v.BindingType {
       case
         VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER,
         VK_DESCRIPTOR_TYPE_STORAGE_BUFFER,
         VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER_DYNAMIC,
         VK_DESCRIPTOR_TYPE_STORAGE_BUFFER_DYNAMIC:
-          readCoherentMemoryInBufferBindings(v.BufferBinding)
+          readMemoryInBufferBindings(v.BufferBinding, bufferBindingOffsets[as!u32(j)])
```
Qining: Looks like we want `j` to be the binding number here, but actually `j` is index. Probably we should use `for _, j, v in descriptor_set.Bindings {`?

gapis/api/vulkan/api/coherent_memory.api
bjoeris: The `k` name is due to an earlier version in which the loops in `readWriteMemoryInBoundDescriptorSets` (`i`), `readMemoryInDescriptorSet` (`j`), and `readMemoryInBufferBindings` (`k`) were nested range loops, similar to `dovkCmdBindDescriptorSets`.

I think you're right that using the map iteration index is wrong here. Since this uses Go maps under the hood, the iteration order is undefined, so I think the the index `k` is uncorrelated with the value `v`, correct?

I will switch these cases to `for i in (0 .. len(bufferBindings))` to match the pattern in `dovkCmdBindDescriptorSets` and emphasize that the keys are contiguous.

bjoeris: I found what appear to be similar issues in `dovkCmdBindDescriptorSets`
```
  for i , _ , set in args.DescriptorSets {
    lastDrawInfo().DescriptorSets[args.FirstSet + as!u32(i)] = DescriptorSets[set]
  }
```
and `dovkCmdBindVertexBuffers`
```
  for i , _ , v in bind.Buffers ...
```
In each of these cases, it looks like the index `i` and the value, `set` or `v`, each range over the correct set, but in uncorrelated order, yes? If so I will include a fix of these together with the instances I introduced.

gapis/api/vulkan/api/renderpass_framebuffer.api
bjoeris: The keys in `ImageAttachments` are contiguous in the range `(0 .. let(ImageAttachments))` (see `vkCreateFramebuffer`). I think I will change these cases to use `for i in (0 .. len(imageAttachments))`, if there are no objections.

AWoloszyn:
I can't actually find this comment in the  MR any more, but I try
responding via email.

> The keys in ImageAttachments are contiguous in the range (0 ..
> let(ImageAttachments)) (see vkCreateFramebuffer). I think I will change
> these cases to use for i in (0 .. len(imageAttachments)), if there are no
> objections.
>
 We make this assumption in a lot of places. The api language has no vector
or similar, so we use this pattern in a number of places.

bjoeris:
I have manually tested capture and replay in the UI with several of the SaschaWillems samples on Linux. I have not done any Android testing. The only issues I found were:
1. Mid-execution capture of shadowmappingcascade contain no commands: both the UI and `gapit commands` show that the capture file is empty. However, this behaviour is also present in master.
2. Mid-execution captures made with other gapid versions cannot be loaded by this version, because the state structure has changed.

gapis/api/vulkan/api/descriptor.api
```diff
@@ -618,25 +618,41 @@ vkCmdBindDescriptorSetsArgs {
...
       // binding index, and then array index, we have to loop over all
       // of the BoundBuffers in order of the binding number.
 
-      // Since `for _, k, v in set.BoundBuffers` loops over in an
+      // Since `for _, k, v in setObj.BoundBuffers` loops over in an
       // arbitrary order, instead we loop over all bindings for the layout
       // and see if it is a buffer.
-      for j in (0 .. len(set.Bindings)) {
-        binding := set.Bindings[as!u32(j)]
+      for j in (0 .. len(setObj.Bindings)) {
```
Qining: `setObj.Bindings` is a map of binding number to binding content, not a list of bindings. So for a descriptor set (actually defined in descriptor set layout), the valid bindings might be number 1, 2, 5, for example, then using the `j` here with list looping index: 0, 1, 2, is not correct.

PR #1847, opened 2018-05-01
Allow the user to override the vertex semantics.

### LINKED ISSUE #960: Meshes don't typically show for Unreal 4 titles

pmuetschard:
Adds a new settings-gear button to the geo view, that when clicked, shows a dialog where the user can assign the sementic type to each vertex stream of the current model.

Fixes #960


gapis/service/path/path.proto
```diff
@@ -330,7 +330,13 @@ message Mesh {
 
 // MeshOptions provides parameters for the mesh returned by a Mesh path resolve.
 message MeshOptions {
-    bool faceted = 1; // If true then normals are calculated from each face.
+    // If true then normals are calculated from each face.
+    bool faceted = 1;
+    // If true, the actual data will be missing from the vertex.Buffers.
```
ben-clayton: Maybe include a/the usage case for this?

gapis/service/path/path.proto
```diff
@@ -330,7 +330,13 @@ message Mesh {
...
+    bool exclude_data = 2;
+    // Hints that override the semantic guessing. Maps vertex attribute names to
+    // their intentended semantic type.
+    map<string, vertex.Semantic.Type> vertex_semantics = 3;
```
ben-clayton: We've intentionally avoided proto's map to date. While I am happy to revisit that decision, I'd prefer consistency for now.

gapis/api/gles/draw_call_mesh.go
```diff
@@ -46,9 +46,16 @@ func drawCallMesh(ctx context.Context, dc drawCall, p *path.Mesh) (*api.Mesh, er
 
 	c := GetContext(s, dc.Thread())
 
-	dci, err := dc.getIndices(ctx, c, s)
-	if err != nil {
-		return nil, err
+	noData := p.Options != nil && p.Options.ExcludeData
```
ben-clayton: Possibly the same as `p.GetOptions().GetExcludeData()` ?

gapis/api/gles/draw_call_mesh.go
```diff
@@ -46,9 +46,16 @@ func drawCallMesh(ctx context.Context, dc drawCall, p *path.Mesh) (*api.Mesh, er
...
+
+	var dci drawCallIndices
+	if noData {
+		dci = drawCallIndices{nil, dc.getDrawMode(), false}
```
ben-clayton: If we're not getting data, is the draw mode important? If not, can we drop the new interface method?

gapis/api/gles/guess_semantics.go
```diff
@@ -69,11 +78,15 @@ func guessSemantics(vb *vertex.Buffer) {
...
+			if needsGuess(s) && s.Format.String() == p.format.String() {
 				s.Semantic.Type = p.semantic
 				taken[p.semantic] = true
 				break
 			}
 		}
 	}
 }
+
+func needsGuess(s *vertex.Stream) bool {
```
ben-clayton: Isn't this already covered with the `taken` map?

gapis/api/vulkan/draw_call_mesh.go
```diff
@@ -81,20 +81,25 @@ func drawCallMesh(ctx context.Context, dc *VkQueueSubmit, p *path.Mesh) (*api.Me
 
 	stats := &api.Mesh_Stats{}
 
+	noData := p.Options != nil && p.Options.ExcludeData
```
ben-clayton: See above

gapis/api/vulkan/draw_call_mesh.go
```diff
@@ -452,7 +470,20 @@ func guessSemantics(vb *vertex.Buffer) {
 	}
 
 	taken := map[vertex.Semantic_Type]bool{}
+	if len(hints) > 0 {
+		for _, s := range vb.Streams {
+			if t, ok := hints[s.Name]; ok && !taken[t] {
```
ben-clayton: @AWoloszyn do streams typically have names in vulkan?

gapis/service/path/path.proto
ben-clayton: Do you think that name to type is the right association here?
I was thinking that stream index to type might be more explicit and less error prone (like when there is no name).

PR #1849, opened 2018-05-02
Actually construct the final command if it failed in the middle.

AWoloszyn:
We actually add partial commands to the command list.
This means we can actually display the command that caused a crash.

### DIFF
--- a/gapis/capture/decoder.go
+++ b/gapis/capture/decoder.go
@@ -72,6 +72,7 @@ func (d *decoder) EndGroup(ctx context.Context, id uint64) error {
 
 	switch obj := obj.(type) {
 	case *cmdGroup:
+		obj.invoked = true
 		id := d.builder.addCmd(ctx, obj.cmd)
 		for _, c := range obj.children {
 			c.SetCaller(id)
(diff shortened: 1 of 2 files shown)


gapis/capture/decoder.go
```diff
@@ -184,3 +184,15 @@ func (d *decoder) decode(ctx context.Context, in proto.Message) (interface{}, er
 
 	return obj, nil
 }
+
+func (d *decoder) flush(ctx context.Context) {
+	for _, g := range d.groups {
+		if obj, ok := g.(*cmdGroup); ok {
```
ben-clayton: Maybe just call `d. EndGroup` so that you have less copy-paste code, and so it will also work with future group types?

AWoloszyn:
Fixed.

gapis/capture/decoder.go
```diff
@@ -72,6 +72,7 @@ func (d *decoder) EndGroup(ctx context.Context, id uint64) error {
 
 	switch obj := obj.(type) {
 	case *cmdGroup:
+		obj.invoked = true
```
ben-clayton: This line is not actually needed, as the object falls out of scope in this function anyway. There's no harm in it though.

PR #1851, opened 2018-05-02
Small fix and warnings from the trace UI changes.

pmuetschard:
Removes the duplicate field from the `AndroidTraceRequest` class that is already present in its base class. Sorry I missed this during the review. The rename in the `TracerDialog` is to silence a name-shadowing warning. I know in this case it is annoying, having to come up with two names for the same thing, however, the warning has saved me a bunch from silly mistakes (like spelling errors) other times, so I put up with it.

### DIFF
--- a/gapic/src/main/com/google/gapid/server/Tracer.java
+++ b/gapic/src/main/com/google/gapid/server/Tracer.java
@@ -122,7 +122,8 @@ public class Tracer {
     public final boolean midExecution;
     public final boolean disableBuffering;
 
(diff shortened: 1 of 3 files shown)


gapic/src/main/com/google/gapid/server/Tracer.java
```diff
@@ -174,10 +175,10 @@ public String toString() {
     public final String intentArgs;
     public final boolean clearCache;
     public final boolean disablePcs;
-    public final boolean disableBuffering;
```
AWoloszyn: Thanks, sorry for missing this myself!

PR #1852, opened 2018-05-02
gles: Fix for GPU texture readbacks

ben-clayton:
The size field of slices was not being set.
This got broken in eca98c386


ben-clayton:
**ping**

PR #1858, opened 2018-05-04
Use the cached initial commands for dependency graph generation.

AWoloszyn:


ben-clayton:
Please can we have a bit more context? Was this broken, an enhancement, etc?

AWoloszyn:
This is purely an enhancement. We use the initial commands for a couple of things. We added them to the database for things like resources, but did not for the Dependency graph. This means that we would be generating the commands at least twice, when it was un-necessary.

gapis/resolve/dependencygraph/dependency_graph.go
```diff
@@ -162,14 +163,14 @@ func (r *DependencyGraphResolvable) Resolve(ctx context.Context) (interface{}, e
 	cmds := c.Commands
 	behaviourProviders := map[api.API]BehaviourProvider{}
 
-	// If the capture contains initial state, prepend the commands to build the state.
-	initialCmds, ranges := c.GetInitialCommands(ctx)
-	if len(initialCmds) > 0 {
-		cmds = append(initialCmds, cmds...)
+	initCmds, ranges, err := initialcmds.InitialCommands(ctx, r.Capture)
```
ben-clayton: I wasn't even aware of the `initialcmds` package. Given the name `GetInitialCommands` it suggests to me that this is fetching some precalculated commands, not doing work to generate them.
Would it make more sense to do the work of `GetInitialCommands` once, store it in the capture, make `GetInitialCommands` return them, and kill `initialcmds`?

AWoloszyn: That would make sense, where do you think the right place for that is. I would like to avoid doing it on load. Specifically because a lot of operations do not need the initial commands.

If you are just running gapit state, we never have to generate them, and they can be incredibly expensive to generate. 

I could have GetInitialCommands be lazy, and only generate them the first time it is called, or we could generate them in gapit during capture time. (Post-process the trace file, take the time then, add them to the capture file, then never have to deal with it again).

Let me know what you think.

Andrew

ben-clayton: Okay, the 'lot of operations do not need the initial commands' argument is a good one.

Keep as-is if you want, we can talk about refactoring later. (very least, I'd like `GetInitialCommands` renamed to `BuildInitialCommands`)

ben-clayton:
> This is purely an enhancement...

Can we have this in the CL description please?

gapis/resolve/dependencygraph/dependency_graph.go
AWoloszyn: Done

PR #1861, opened 2018-05-04
gRPC for GAPIR

Qining:
- **Requires** Bazel to be upgraded to **0.13.0**, this is to utilize
  the new flag: `--experimental_shortened_obj_file_path=true` to
  workaround the 260 MAX_PATH limit on Windows when compiling with Android
  NDK prebuilts. This solution might be a temporary workaround, in future
  we need something like the mingw wrapper for Android NDK prebuilts.
- On Android, use file system socket for communication. GAPIS uses 'ls'
  to check if the socket file exists or not before dialing to connect.
- Do not build C-ARES for Windows as it is not supported (C-ARES is
  dependent by gRPC).
- Overwrite the gRPC's build file with our own, to fix build for Android
  and Windows.
- Force static link for build on Windows by adding `-static` to the
  linker flags for mingw toolchain. Otherwise, the bulit 'protoc` won't
  start due to `not found dll` (because the gRPC Bazel function does not
  pass any ENV to the `protoc` when using it to generate cpp/h files)
 - It seems this also makes gapis.exe fully statically linked, so libstdc++.dll and 
   libwinpthread.dll might not be required for shipping.
- On Windows, run `pacman -Sy mingw-w64-x86_64-crt-git` when fixing up
  the env, this is to fix the '__imp___* undefined reference' error.
  (The real reason of the error might be an incomplete `pacman -Syu` call)


WORKSPACE
```diff
@@ -25,3 +25,14 @@ load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies")
...
+load("@com_github_grpc_grpc//:bazel/grpc_deps.bzl", "grpc_deps")
+grpc_deps()
+bind(
+    name = "grpc_cpp_plugin",
+    actual = "@com_github_grpc_grpc//:grpc_cpp_plugin",
+)
+bind(
+    name = "grpc++_codegen_proto",
+    actual = "@com_github_grpc_grpc//:grpc++_codegen_proto",
+)
```
Qining: @pmuetschard Do you have a better place in mind to place this piece of code?

core/cc/socket_connection.cpp
```diff
@@ -258,20 +258,81 @@ std::unique_ptr<Connection> SocketConnection::createSocket(
...
+  FILE* f = fopen(portFile, "w");
+  if (f == nullptr) {
+    return false;
+  }
+  fprintf(f, "Bound on port '%d'", port);
+  fclose(f);
+  return true;
+}
+
+uint32_t SocketConnection::getFreePort(const char* hostname) {
```
Qining: This is just a copy of the createSocket() function, except it just returns the free port assigned by OS, the socket will be closed.

gapir/cc/memory_manager.cpp
```diff
@@ -75,19 +75,41 @@ MemoryManager::MemoryManager(const std::vector<uint32_t>& sizeList) : mConstantM
...
-        return false;
-    }
-
-    mReplayData = {align(mMemory.get() + mSize - size), size};
-    GAPID_DEBUG("Replay range: [%p,%p]",
-        mReplayData.base, mReplayData.base + mReplayData.size - 1);
-    return true;
+bool MemoryManager::setReplayDataSize(uint32_t constantMemorySize,
+                                      uint32_t opcodeMemorySize) {
+  GAPID_INFO("in MemoryManager::setReplayDataSize(%d, %d)", constantMemorySize, opcodeMemorySize);
```
Qining: Remove all those GAPID_INFO calls.

gapis/replay/builder/builder.go
```diff
@@ -580,17 +585,13 @@ func (b *Builder) Build(ctx context.Context) (protocol.Payload, ResponseDecoder,
...
-	payload := protocol.Payload{
-		StackSize:          uint32(512), // TODO: Calculate stack size
-		VolatileMemorySize: uint32(vml.size),
-		Constants:          b.constantMemory.data,
-		Resources:          b.resources,
-		Opcodes:            opcodes.Bytes(),
-	}
+	// TODO: This does not looks good
+	payload := *gapir.NewPayload(
+		uint32(512), uint32(vml.size), b.constantMemory.data, b.resources, opcodes.Bytes())
```
Qining: Suggestion needed. Is this the correct way to avoid extra copy? Is it fine to make NewPayload return the struct, instead of a pointer to the struct?

gapis/replay/executor/executor.go
```diff
@@ -65,207 +60,91 @@ func Execute(
...
-	}
-	w.Uint32(uint32(len(e.payload.Opcodes)))
-	w.Data(e.payload.Opcodes)
-
-	data := buf.Bytes()
-
-	// Store the payload to the database
-	id, err := database.Store(ctx, data)
+func (e executor) execute(ctx context.Context, connection *gapir.Connection) error {
+	pp := e.payload.ToProto()
```
Qining: Suggestion needed, this is making an extra copy?

cmd/gapir/cc/main.cpp
```diff
@@ -132,25 +129,24 @@ void android_main(struct android_app* app) {
...
-                        &memoryManager, crashHandler);
-    });
-    bool alive = true;
-    while (alive) {
+            uri.c_str(), core::supportedABIs());
+
+    std::unique_ptr<Server> server = Setup(
+        uri.c_str(), nullptr, nullptr, 0 /*no timeout?*/, &crashHandler, &memoryManager);
+    std::thread waiting_thread([&]() { server.get()->wait(); });
+    if (chmod(socket_file_path.c_str(), S_IRUSR|S_IWUSR|S_IXUSR|S_IROTH|S_IWOTH|S_IXOTH)) {
```
AWoloszyn: Do we really need executable flags here? That seems overly-broad.

tools/build/workspace.bzl
```diff
@@ -62,8 +63,10 @@ def gapid_dependencies(android = True, java_client = True, mingw = True, locals
         locals = locals,
         organization = "grpc",
         project = "grpc",
-        commit = "fa301e3674a1cc786eb4dd4253a0e677f2eb68e3",
+        commit = "d45132a2e9246b11ddd0b70c07160076d5cbbb12",
+        build_file = "@gapid//tools/build/third_party:grpc_c++.BUILD",
     )
+    grpc_deps()
```
Qining: Duplicated with the one in Workspace.bzl

gapis/replay/executor/executor.go
AWoloszyn: @ben-clayton  Please correct me if I am wrong, but I think there will be no extra copies. When you take the address of a go object, the compiler does its magic to make sure it will be on the heap, and no copy will be made.

cmd/gapir/cc/main.cpp
Qining: This definitely should be a TODO

gapir/service/service.proto
```diff
@@ -0,0 +1,97 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package service;
```
Qining: @ben-clayton This is causing duplicated registered proto name with `gapis/service/service.proto`, which has  `PingRequest/Response etc`, should I rename it to `gapir_service` or `GapirService` or any other suggestions?

AWoloszyn:
This looks like a real failure in the Window build.

Qining:
@AWoloszyn Looks like the Windows failure is caused by compiling the gRPC library. I think I need to use a newer gRPC commit later to try to fix the build.

gapir/cc/mock_replay_connection.h
```diff
@@ -0,0 +1,52 @@
...
+ public:
+  MockReplayConnection() : ReplayConnection(nullptr) {}
+  MOCK_METHOD0(getPayload, std::unique_ptr<ReplayConnection::Payload>());
+  MOCK_METHOD1(mockedGetResources, std::unique_ptr<ReplayConnection::Resources>(
+                                       ReplayConnection::ResourceRequest*));
+  std::unique_ptr<ReplayConnection::Resources> getResources(
+      std::unique_ptr<ReplayConnection::ResourceRequest> req) override {
+    return mockedGetResources(req.get());
+  }
+  // MOCK_METHOD0(sendReplayFinished, bool());
```
AWoloszyn: We shouldn't leave commented out code in I don't think

gapir/cc/test_utilities.h
```diff
@@ -29,13 +30,13 @@
 #include <utility>
 #include <vector>
 
-namespace core {
-namespace test {
+// namespace core {
```
AWoloszyn: Remove commented out code

gapir/cc/test_utilities.h
```diff
@@ -66,17 +67,26 @@ void pushUint32(std::vector<uint8_t>* buf, uint32_t v);
...
+    const std::vector<Resource>& resources,
+    const std::vector<uint32_t>& instructions);
+
+std::unique_ptr<ReplayConnection::Resources> createResources(
+    const std::vector<uint8_t>& data);
 
-std::unique_ptr<ServerConnection> createServerConnection(core::test::MockConnection* connection,
-                                                         const std::string& replayId,
-                                                         uint32_t replayLength);
+// std::unique_ptr<ServerConnection> createServerConnection(
```
AWoloszyn: Remove

gapir/client/session.go
```diff
@@ -167,31 +189,50 @@ func (s *session) newADB(ctx context.Context, d adb.Device, abi *device.ABI) err
...
+	apkDir, err := apk.FileDir(ctx)
+	if err != nil {
+		return log.Errf(ctx, err, "Getting gapid.apk files directory")
 	}
-	s.onClose(func() { d.RemoveForward(ctx, localPort) })
+	// Wait for the socket to be opened
+	socketPath := strings.Join([]string{apkDir, socket}, "/")
+	err = task.Retry(ctx, maxCheckSocketFileAttempts, checkSocketFileRetryDelay,
+		func(ctx context.Context) (bool, error) {
+			str, err := d.Shell("run-as", apk.Name, "cat", "/proc/net/unix", "|", "grep", socket).Call(ctx)
```
AWoloszyn: Do we need to run this as the application?

tools/bazel.rc
```diff
@@ -7,6 +7,8 @@ run --direct_run
 # Adding in c11 compiling, this really should be done in the android toolchains
 build --cxxopt="-std=c++11"
 
+build --copt -DGRPC_BAZEL_BUILD
```
AWoloszyn: I don't know bazel that well, but does this need to be added to every c/cpp file that ever gets built?

Qining: Right, I think I'd better make it only used for GRPC library

gapir/client/session.go
Qining: Nope, so do you prefer to drop the `run-as`? It looks like harmless.

cmd/gapir/cc/main.cpp
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017 Google Inc.
+ * Copyright (C) 2018 Google Inc.
```
ben-clayton: Please don't bump copyright years (unless you're making a new file)

gapir/cc/test_utilities_test.cpp
```diff
@@ -75,44 +75,31 @@ void pushString(std::vector<uint8_t>* buf, const char* str) {
...
+    const std::vector<uint8_t>& constantMemory,
+    const std::vector<Resource>& resources,
+    const std::vector<uint32_t>& instructions) {
+  auto p = std::unique_ptr<service::Payload>(new service::Payload);
+  p->set_stack_size(stackSize);
+  p->set_volatile_memory_size(volatileMemorySize);
+  p->set_constants(constantMemory.data(), constantMemory.size());
+  p->set_opcodes(instructions.data(), instructions.size() * sizeof(uint32_t));
+  for (size_t i = 0; i < resources.size(); i++) {
+    p->add_resources();
```
ben-clayton: Doesn't the `add_` functions return a pointer to the newly added object? If so, slightly nicer to have:

```c++
auto r = p->add_resources();
r->set_id(resources[i].id);
r->set_size(resources[i].size);
```

gapir/client/connection.go
```diff
@@ -0,0 +1,259 @@
...
+	c.servClient = nil
+	c.stream = nil
+}
+
+func (c *Connection) Ping(ctx context.Context) error {
+	if c.servClient == nil {
+		return log.Err(ctx, nil, "Gapir not connected")
+	}
+	// If there is an authentication token, attach it in the metadata
+	if len(c.authToken) != 0 {
```
ben-clayton: Helper function, perhaps?

gapir/client/connection.go
```diff
@@ -0,0 +1,259 @@
...
+	payload Payload,
+	handleResourceRequest func(context.Context, *ResourceRequest, *Connection) error,
+	handleCrashDump func(context.Context, *CrashDump, *Connection) error,
+	handlePostData func(context.Context, *PostData, *Connection) error,
+	handleNotification func(context.Context, *Notification, *Connection) error) error {
+	ctx = log.Enter(ctx, "HandleReplayCommunication")
+	if c.conn == nil || c.servClient == nil {
+		return log.Errf(ctx, nil, "Gapir not connected")
+	}
+	// Drop the unfinished replay with on connection.
```
ben-clayton: eh?

gapir/service/service.proto
ben-clayton: As we discussed, I suggest `package gapir` and `service Service`

gapis/replay/builder/builder.go
```diff
@@ -603,18 +604,26 @@ func (b *Builder) Build(ctx context.Context) (protocol.Payload, ResponseDecoder,
...
-		var d binary.Reader
-		if r != nil {
-			d = endian.Reader(r, byteOrder)
+	responseDecoder := func(pd *gapir.PostData) {
+		// TODO: should we skip it instead of return error?
+		ctx = log.Enter(ctx, "responseDecoder")
+		if pd == nil {
+			log.E(ctx, "Cannot handle nil PostData")
 		}
 		go func() {
```
ben-clayton: Looks like this should be using `crash.Go`

gapir/cc/test_utilities_test.cpp
Qining: You are right, I just realized it today. I will change it to use the returned pointer.

Qining:
@ben-clayton Unfortunately using `package gapir and service Service` won't compile.
The generated c++ code is like below:
```
class Service final {
...
    class Service : public ::grpc::Service {
    ....
    };
....
};
```
And the error message from compiler:
In file .../genfiles/gapir/service/service.grpc.pb.cc:6:
.../genfiles/gapir/service/service.grpc.pb.h:125:9: error: member 'Service' has the same name as its class
  class Service : public ::grpc::Service {
        ^

gapir/cc/memory_manager.cpp
```diff
@@ -75,19 +75,36 @@ MemoryManager::MemoryManager(const std::vector<uint32_t>& sizeList) : mConstantM
...
-bool MemoryManager::setReplayDataSize(uint32_t size) {
-    if (size > mSize) {
-        return false;
-    }
-
-    mReplayData = {align(mMemory.get() + mSize - size), size};
-    GAPID_DEBUG("Replay range: [%p,%p]",
-        mReplayData.base, mReplayData.base + mReplayData.size - 1);
-    return true;
+bool MemoryManager::setReplayDataSize(uint32_t constantMemorySize,
```
Qining: setReplayDataSize() is changed to accept two parameters. And there won't be a setConstantMemory() anymore

cmd/gapir/cc/main.cpp
```diff
@@ -71,44 +73,42 @@ std::unique_ptr<ResourceInMemoryCache> createResourceProvider(
...
-            GAPID_INFO("Shutting down");
-            break;
-        }
-        std::unique_ptr<CrashUploader> crash_uploader = std::unique_ptr<CrashUploader>(new CrashUploader(crashHandler, *acceptedConn));
+// Setup creates and starts a replay server at the given URI port. Returns the
+// created and started server.
+std::unique_ptr<Server> Setup(const char* uri, const char* authToken,
+                              const char* cachePath, int idleTimeoutSec,
+                              core::CrashHandler* crashHandler,
+                              MemoryManager* memMgr) {
```
Qining: Add TODO
This memMgr is not thread-safe. The original replay server does not handle multiple client connections, so it was fine to have just a single memory manager.

But with gRPC, the server is able to handle multiple connections. This results into multiple independent replay streams using the same memory manager, it can result into crashes/wrong result.

So far for each GAPIR, as we always have only one GAPIS talking to it, it won't cause an issue. But need note for future.

Qining: Added at https://github.com/google/gapid/pull/1861/commits/2ea8a8635f2614c9b57130dd36be447c9a1d876c#diff-4f71d0359f7b5b6da6691abe3a4599e2R78

gapir/cc/context_test.cpp
```diff
@@ -46,79 +45,58 @@ class ContextTest : public ::testing::Test {
...
     EXPECT_THAT(context, NotNull());
 }
 
 TEST_F(ContextTest, CreateErrorReplayRequest) {
     // Failed to load
-    Resource res(REPLAY_ID, 10);
-    EXPECT_CALL(*mResourceProvider, get(_, _, _, _, 10))
-            .With(Args<0, 1>(ElementsAre(res)))
-            .WillOnce(Return(false));
+    EXPECT_CALL(*mConn, getPayload()).WillOnce(Return(ByMove(nullptr)));
```
AWoloszyn: ByMove(nullptr))

Is that right? It doesn't seem right, since you cannot move a nullptr.

gapir/cc/replay_connection.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+replay_service::ResourceRequest*
+ReplayConnection::ResourceRequest::release_to_proto() {
+  auto ptr = mProtoResourceRequest;
+  mProtoResourceRequest = nullptr;
+  return ptr;
+}
+
+ReplayConnection::ResourceRequest::ResourceRequest()
+    : mProtoResourceRequest(new replay_service::ResourceRequest()) {}
```
AWoloszyn: Can we please put this in a unique_ptr. 
new/delete make me unhappy.

gapir/cc/replay_connection.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+// Posts member methods
+
+ReplayConnection::Posts::~Posts() {
+  if (mProtoPostData != nullptr) {
+    delete mProtoPostData;
```
AWoloszyn: Same here

gapir/cc/replay_connection.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+const void* ReplayConnection::Posts::piece_data(int index) const {
+  return mProtoPostData->post_data_pieces(index).data().data();
+}
+
+uint64_t ReplayConnection::Posts::piece_id(int index) const {
+  return mProtoPostData->post_data_pieces(index).id();
+}
+
+ReplayConnection::Posts::Posts() : mProtoPostData(new replay_service::PostData()) {}
```
AWoloszyn: And here

gapir/cc/replay_connection.cpp
```diff
@@ -0,0 +1,251 @@
...
+  replay_service::ReplayResponse res;
+  res.set_allocated_post_data(posts->release_to_proto());
+  return mGrpcStream->Write(res);
+}
+
+bool ReplayConnection::sendNotification(uint64_t id, uint32_t api_index,
+                                        uint64_t label, const std::string& msg,
+                                        const void* data, uint32_t data_size) {
+  replay_service::ReplayResponse res;
+  res.mutable_notification()->set_id(id);
```
AWoloszyn: auto& notification = res.mutable_notification();
notification->set_id(id);
...

gapir/cc/replay_request.h
```diff
@@ -26,20 +26,18 @@
...
 
 // Class for storing the information about a replay request that came from the server
 class ReplayRequest {
 public:
     // Creates a new replay request and loads it content from the given gazer connection. Returns
     // the new replay request if loading it was successful or nullptr otherwise. The memory manager
     // is used to store the content of the replay request
-    static std::unique_ptr<ReplayRequest> create(const ServerConnection& gazer,
-                                                 ResourceProvider* resourceProvider,
+    static std::unique_ptr<ReplayRequest> create(ReplayConnection* conn,
```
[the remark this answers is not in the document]
AWoloszyn: I am not familiar with this part of the code-base. I defer to @ben-clayton

gapir/cc/replay_request.h
```diff
@@ -60,27 +58,16 @@ class ReplayRequest {
...
 
     // The size of the volatile memory required by the replay
     uint32_t mVolatileMemorySize;
 
-    // The base address and the size of the constant memory
+    // The base address and the size in bytes of the constant memory
     std::pair<const void*, uint32_t> mConstantMemory;
 
-    // The base address and the size of the instruction list
+    // The base address and the count of the instructions
```
AWoloszyn: number of instructions

gapir/cc/replay_connection.cpp
Qining: done

gapir/cc/replay_connection.cpp
Qining: done

gapir/cc/replay_request.h
Qining: done

gapir/cc/context_test.cpp
Qining: It is quite interesting here. gmock, without `ByMove()` inside of `Return()`, tries to make a copy when returning the value to caller (not sure if this is due to compiler). Because `getPayload()` returns `unique_ptr`, compilation will fail due to deleted copy-ctor of `unique_ptr`.

I've changed this to `Return(ByMove(std::unique_ptr<...>(nullptr)))`. This should make more senses.

gapir/cc/replay_request.h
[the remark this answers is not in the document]
ben-clayton: Sounds reasonable.

gapir/cc/resource_cache.cpp
```diff
@@ -25,21 +28,21 @@ ResourceCache::ResourceCache(std::unique_ptr<ResourceProvider> fallbackProvider)
...
         if (size < resource.size) {
-            return false; // Not enough space
+            return false;  // Not enough space
         }
         // Try fetching the resource from the cache.
         if (getCache(resource, dst)) {
             // In cache. Flush the pending requests.
-            if (!batch.flush(*this, server)) {
-                return false; // Failed to get resources from fallback provider.
+            if (!batch.flush(*this, conn)) {
```
ben-clayton: Note: This implementation can result in many round trips to the server, and its design was based around the idea that we could load the resource directly into the destination buffer without temporary copies. Given that gRPC forces us to have temporary copies, this implementation should be changed to have a single fetch.
Not suggesting this should be implemented in this PR, but a TODO would be nice.

gapir/cc/resource_cache.h
```diff
@@ -22,6 +22,8 @@
 #include <memory>
 #include <vector>
 
+#include "replay_connection.h"
+
 namespace gapir {
 
 // ResourceCache is an abstract base class for caching resource providers.
```
ben-clayton: ... in fact all of these cache classes could be massively simplified now.

gapir/client/connection.go
```diff
@@ -0,0 +1,262 @@
...
+type ResourceInfoList []*replaysrv.ResourceInfo
+
+func (l *ResourceInfoList) Append(resourceID string, size uint32) {
+	*l = append(*l, &replaysrv.ResourceInfo{
+		Id:   resourceID,
+		Size: size,
+	})
+}
+
+type Payload replaysrv.Payload
```
ben-clayton: Is there any benefit to redeclaring all these `replay_service` types?
If you want just some syntactic sugar here, you can [use type aliases](https://golang.org/doc/go1.9#language) (new in go 1.9) which does not create a new type, and so doesn't need any casting:

```
type (
	Payload   = replaysrv.Payload
	Resources = replaysrv.Resources
	...
)
```

Please check with the vetting tools to see if these should be documented.

gapir/client/connection.go
```diff
@@ -0,0 +1,262 @@
...
+	"github.com/google/gapid/core/log"
+	replaysrv "github.com/google/gapid/gapir/replay_service"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+const (
+	gapirAuthTokenMetaDataName = "gapir-auth-token"
+)
+
```
ben-clayton: Docs please

gapir/client/connection.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+	return nil
+}
+
+func (c *Connection) HandleReplayCommunication(
+	ctx context.Context,
+	replayID string,
+	payload Payload,
+	handleResourceRequest func(context.Context, *ResourceRequest, *Connection) error,
+	handleCrashDump func(context.Context, *CrashDump, *Connection) error,
```
ben-clayton: Consider putting the `Connection` into the context. This'll simplify the callbacks, and let you fetch it much deeper down in a call chain.

gapir/client/session.go
```diff
@@ -167,31 +189,50 @@ func (s *session) newADB(ctx context.Context, d adb.Device, abi *device.ABI) err
...
+	apkDir, err := apk.FileDir(ctx)
+	if err != nil {
+		return log.Errf(ctx, err, "Getting gapid.apk files directory")
 	}
-	s.onClose(func() { d.RemoveForward(ctx, localPort) })
+	// Wait for the socket file to be created
+	socketPath := strings.Join([]string{apkDir, socket}, "/")
+	err = task.Retry(ctx, maxCheckSocketFileAttempts, checkSocketFileRetryDelay,
+		func(ctx context.Context) (bool, error) {
+			str, err := d.Shell("run-as", apk.Name, "ls", socketPath).Call(ctx)
```
ben-clayton: Has this been tested on rooted / non rooted / non-google devices?

gapir/replay_service/service.proto
```diff
@@ -0,0 +1,102 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package replay_service;
+
+option go_package = "github.com/google/gapid/gapir/replay_service";
+
```
ben-clayton: Docs please

gapir/replay_service/service.proto
```diff
@@ -0,0 +1,102 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package replay_service;
```
ben-clayton: Sorry to go back to this again, but what was the original issue with this being called `service`? Did you say there was a collision between this and gapis service?

gapir/replay_service/service.proto
```diff
@@ -0,0 +1,102 @@
...
+    uint64 id = 1;
+    bytes data = 2;
+}
+
+message PostData{
+    repeated PostDataPiece post_data_pieces = 1;
+}
+
+message Notification {
+  uint64 id = 1;
```
ben-clayton: Weird mix of indenting in this file

gapis/replay/builder/builder.go
ben-clayton: Why not make `Build` return a `*Payload`?

ben-clayton: Please move the 512 to a named local variable, and keep the TODO

gapis/replay/builder/builder.go
```diff
@@ -603,20 +605,28 @@ func (b *Builder) Build(ctx context.Context) (protocol.Payload, ResponseDecoder,
...
+		ctx = log.Enter(ctx, "responseDecoder")
+		if pd == nil {
+			log.E(ctx, "Cannot handle nil PostData")
 		}
-		go func() {
-			for _, p := range b.decoders {
-				err = p(d, err)
-				if err != nil {
-					d = nil
+		crash.Go(func() {
```
ben-clayton: Not entirely sure about spawning a new go routine now. 
This function is called whenever a new post gRPC call is made, right? Assuming the rpc is non blocking, we can probably safely drop this.

gapis/replay/executor/executor.go
ben-clayton: I don't know what the compiler will do, but I don't think you have to care.
Why are you worried? If it's regarding the payload data - they're in slices that will not be deep copied.

gapis/replay/executor/executor.go
```diff
@@ -65,207 +60,85 @@ func Execute(
...
-	// Decode and handle postbacks as they are received
-	e.decoder(responseR, nil)
-
-	err = <-comErr
-	if closeErr := responseR.Close(); closeErr != nil {
-		log.W(ctx, "Replay execute pipe reader Close failed: %v", closeErr)
-	}
+	err = connection.HandleReplayCommunication(
+		ctx, id.String(), e.payload,
+		e.handleResourceRequest,
```
ben-clayton: An alternative to N `funcs` here is to use an interface. Not arguing that it is a better design, but at least it would let you document the functions clearer.

gapis/replay/executor/executor.go
```diff
@@ -65,207 +60,85 @@ func Execute(
...
-	if r.Error() != nil {
-		return r.Error()
-	}
-
-	n := r.Uint32()
-	if r.Error() != nil {
-		return r.Error()
-	}
+func (e executor) handleNotification(ctx context.Context, notification *gapir.Notification, conn *gapir.Connection) error {
+	return nil
```
ben-clayton: TODO?

tools/build/rules/grpc_c++.bzl
```diff
@@ -0,0 +1,250 @@
+# Copyright (C) 2018 Google Inc.
```
ben-clayton: @pmuetschard please can you review this (likely post submission) when you're back?

gapis/replay/builder/builder.go
```diff
@@ -49,7 +50,12 @@ type marker struct {
...
 // then the postback data was absent or corrupted and err holds the error.
-type ResponseDecoder func(r io.Reader, err error)
+type ResponseDecoder func(p *gapir.PostData)
+
+type postBackDecoder struct {
+	expectedSize int
+	decode       Postback
+}
 
 // Postback decodes a single commands's postback, returning and carrying over errors.
```
ben-clayton: I think we should remove the `error` return value from this function signature as it is completely unused, and will simplify the implementations.
Please also update docs.

gapir/replay_service/service.proto
Qining: `package service` => name conflict with gapis service
`package gapir` and `service Service` => C++ won't compile, and results into weird folder naming like `gapir/gapir/service.proto`, that `gapir/gapir` hierarchy is terrible.

So to keep the folder's name and the package name the same (so the generated code location will be consistent), renamed the package to replay_service. I think this makes sense, as replay_service is part of the functionalities of GAPIR.

cmd/gapir/cc/main.cpp
```diff
@@ -71,44 +73,49 @@ std::unique_ptr<ResourceInMemoryCache> createResourceProvider(
...
-        }
-        std::unique_ptr<CrashUploader> crash_uploader = std::unique_ptr<CrashUploader>(new CrashUploader(crashHandler, *acceptedConn));
+// Setup creates and starts a replay server at the given URI port. Returns the
+// created and started server.
+// Note the given memory manager, it will be used for multiple connections, and
+// this can result into wrong result or crash when there are more than one
+// client connected and requesting replay. The memory manager is kept here
+// to minimized the required modification to adopt gRPC, and this is good enough
+// when there is only one GAPIS talking to this GAPIR.
+// TODO: Make memory manager thread-safe, or create individual memory manager
```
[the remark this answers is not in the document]
Qining: Not sure if mutex helps. The problem is the 'Constant' and 'Opcodes' segment, with two client requesting different replay request, that segment can be corrupted, that will result in a crash? or fatal error?

I think I'd better keep it a TODO, and come back later with a more mature solution.

gapir/replay_service/service.proto
ben-clayton: When you say it collides with gapis service, when? I assume gapir has no reason to import the gapis protos. So is this a collision in GAPIS at runtime?

gapir/cc/post_buffer.cpp
```diff
@@ -19,11 +19,14 @@
...
+#include "replay_connection.h"
+
 namespace gapir {
 
 PostBuffer::PostBuffer(uint32_t desiredCapacity, PostBufferCallback callback) :
-        mBuffer(new (std::nothrow) uint8_t[desiredCapacity]),
-        mCapacity(mBuffer ? desiredCapacity : 0), mCallback(callback), mOffset(0) {
+        mPosts(ReplayConnection::Posts::create()),
+        mPostPieceCount(0),
+        mCapacity(mPosts ? desiredCapacity : 0), mCallback(callback), mOffset(0) {
```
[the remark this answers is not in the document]
Qining: Right, Posts::create() will never return nullptr. I will just use the desiredCapacity to initialize mCapacity.

gapir/replay_service/service.proto
Qining: And the `RegisterType`:
https://github.com/golang/protobuf/blob/master/proto/properties.go#L482

core/os/android/adb/commands.go
```diff
@@ -120,6 +120,7 @@ func (b *binding) StartActivity(ctx context.Context, a android.ActivityAction, e
 	args := append([]string{
 		"start",
 		"-S", // Force-stop the target app before starting the activity
+		"-W", // Wait until the launch finish
```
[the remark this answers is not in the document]
Qining: Done

core/os/android/adb/forward.go
```diff
@@ -64,6 +64,12 @@ func (p NamedAbstractSocket) adbForwardString() string {
 	return fmt.Sprintf("localabstract:%s", p)
 }
 
+type NamedFileSystemSocket string
```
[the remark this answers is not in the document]
Qining: Done

gapis/replay/executor/executor.go
Qining: Changed to an "handler" interface.

gapis/replay/builder/builder.go
Qining: Done, also changed a lot of Post() caller sites in gapis/api/gles/find_issues.go.

gapir/cc/resource_cache.h
Qining: TODO added

gapir/client/connection.go
Qining: This is wonderful!

gapir/client/connection.go
Qining: Done

gapir/replay_service/service.proto
Qining: Done

gapir/replay_service/service.proto
Qining: Done

Qining:
Rebased and tested with current master, didn't see any performance impact

Qining:
For the performance impact, I should also specify the changes in binary sizes.
On Linux:
sha                         | 550570          | a58ef4
gapid-aarch64               | 8241652         | 9851597      (+19.5%)
gapid-armeabi               | 7772398         | 9235743      (+18.8%)
gapid-x86                   | 9516494         | 11306454     (+18.8%)
gapir                       | 1648712         | 7706128      (+367.4%)
gapis                       | 93492120        | 93710360     (+0.2%)
gapit                       | 15448504        | 15502584     (+0.4%)

`gapir` gets bloated quite a lot, as it now has gRPC++.

gapir/client/session.go
Qining: Tested with Pixel C (rooted and unrooted), Pixel 2 (unrooted)

Qining: Just tested with XiaoMi 5S (Android 7.0), it works! (Just need to enable all USB debugging options)

gapis/api/gles/find_issues.go
```diff
@@ -198,22 +199,24 @@ func (t *findIssues) Transform(ctx context.Context, id api.CmdID, cmd api.Cmd, o
 		out.MutateAndWrite(ctx, dID, cb.GlGetShaderSource(cmd.Shader(), buflen, memory.Nullptr, tmp.Ptr()))
 		out.MutateAndWrite(ctx, dID, cb.Custom(func(ctx context.Context, s *api.GlobalState, b *builder.Builder) error {
 			b.ReserveMemory(tmp.Range())
-			b.Post(value.ObservedPointer(tmp.Address()), buflen, func(r binary.Reader, err error) error {
+			b.Post(value.ObservedPointer(tmp.Address()), buflen, func(r binary.Reader, err error) {
 				if err != nil {
-					return err
+					t.onIssue(cmd, id, service.Severity_FatalLevel, fmt.Errorf("Failed to decode glGetShaderSource postback: %v", err))
```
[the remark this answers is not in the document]
ben-clayton: Thanks!

PR #1866, opened 2018-05-07
Drop debug logs in image primer

Qining:


Qining:
Looks like the windows build is still a little bit flaky. This CL only drops several `log.W()` calls and is surly not causing build failure. Merged.

PR #1869, opened 2018-05-07
Add context destruction blacklist.

AWoloszyn:
This will cause a context and window leak, which is bad, but
only for certain hardware combinations.

We can remove once we figure out exactly waht is causing
this behavior.


core/os/device/deviceinfo/cc/query.cpp
```diff
@@ -160,7 +160,17 @@ void buildDeviceInstance(const query::Option& opt, void* platform_data,
   instance->set_allocated_configuration(configuration);
   deviceInstanceID(instance);
 
-  query::destroyContext();
+  // Blacklist of OS/Hardware version that mean we cannot safely
```
ben-clayton: s/mean/means

Link to bug in comment?

PR #1872, opened 2018-05-09
Replace API direct field access with getter / setters

ben-clayton:
Required for the new compiler generated Mutate logic.


ben-clayton:
As discussed, even with the hit on performance / size this is a stepping stone to what we believe to be far greater performance improvements.

PR #1873, opened 2018-05-09
Add replace_resource verb (shaders only) to gapit

apazylbe:


gapis/api/vulkan/resources.go
```diff
@@ -773,7 +773,17 @@ func (cmd *VkCreateShaderModule) Replace(ctx context.Context, c *capture.Capture
...
 	shader := data.GetShader()
-	codeSlice := shadertools.AssembleSpirvText(shader.Source)
+	var codeSlice interface{}
+	var codeSize int
+	if shader.GetType() == api.ShaderType_Spirv {
+		assembledCode := shadertools.AssembleSpirvText(shader.Source)
+		codeSlice = assembledCode
+		codeSize = len(assembledCode) * 4
+	} else {
+		codeSlice = []byte(shader.Source)
```
AWoloszyn: We should probably make sure this is a multiple of 4 bytes, or else we would be creating an invalid SPIR-V program / Vulkan trace. 

Nicer to tell the user they are doing something wrong earlier.

cmd/gapit/replace_resource.go
```diff
@@ -0,0 +1,115 @@
...
+		}
+		verb.At = int(boxedCapture.(*service.Capture).NumCommands) - 1
+	}
+
+	handleRegex := regexp.MustCompile("<(.*?)>")
+
+	for _, types := range resources.GetTypes() {
+		if types.Type == api.ResourceType_ShaderResource {
+			for _, v := range types.GetResources() {
+				handleHexStr := handleRegex.FindStringSubmatch(v.GetHandle())[1]
```
AWoloszyn: If we can't match the handle in the "<*>" form, we should probably test against the full returned handle.

I think handles MAY come back with different names if the user has debug names attached to the handles, so matching against a "real" name may be helpful.

ben-clayton: I agree, making any sort of assumptions on the format of the handle string here seems fragile.
If the typical usage case is purely to find a shader by handle in hex, perhaps we can just make this a substring match (which would also then work for debug annotated names), and error if you find multiple of the same substring match?

gapis/server/server.go
```diff
@@ -198,6 +198,18 @@ func (s *server) LoadCapture(ctx context.Context, path string) (*path.Capture, e
 	return p, nil
 }
 
+func (s *server) SaveCapture(ctx context.Context, c *path.Capture, path string) error {
+	ctx = log.Enter(ctx, "SaveCapture")
+	if !s.enableLocalFiles {
+		return fmt.Errorf("Server not configured to allow reading of local files")
```
ben-clayton: s/reading/writing.

gapis/api/vulkan/resources.go
apazylbe: Done

cmd/gapit/replace_resource.go
apazylbe: Done

gapis/server/server.go
apazylbe: Done

apazylbe:
Updated for comments

cmd/gapit/replace_resource.go
```diff
@@ -0,0 +1,117 @@
...
+		Action:    verb,
+	})
+}
+
+func (verb *replaceResourceVerb) Run(ctx context.Context, flags flag.FlagSet) error {
+	if flags.NArg() != 3 {
+		app.Usage(ctx, "Exactly 3 arguments: resource handle, resource path, capture, got %d", flags.NArg())
+		return nil
+	}
+	resourceHandle := flags.Arg(0)
```
ben-clayton: Nit: We typically use flags for non-capture gapit parameters (keeping the non-flag args to 1). Fine to keep as-is, but be aware that we might change this later.

gapis/server/server.go
```diff
@@ -198,6 +198,18 @@ func (s *server) LoadCapture(ctx context.Context, path string) (*path.Capture, e
 	return p, nil
 }
 
+func (s *server) SaveCapture(ctx context.Context, c *path.Capture, path string) error {
+	ctx = log.Enter(ctx, "SaveCapture")
+	if !s.enableLocalFiles {
+		return fmt.Errorf("Server not configured to allow writing of local files")
+	}
+	buf := bytes.Buffer{}
```
ben-clayton: You can avoid a temporary copy of the capture here:

```go

	f, err := os.OpenFile(path, os.O_CREATE|os.O_WRONLY, 0666)
	if err != nil {
		return err
	}
	defer f.Close()
	if err := capture.Export(ctx, c, &buf); err != nil {
		return err
	}
```

cmd/gapit/replace_resource.go
apazylbe: That's what I wrote initially, then changed, cause I thought flags are the optional arguments. Changed it back, to follow the convention. The resourcePath and handle are required however, I didn't see a pattern in gapit words for required flags.

gapis/server/server.go
apazylbe: Done

apazylbe:
Updated for review comments

PR #1391, opened 2017-11-17
WIP: Get data from GPU at the start of MEC in gapii.

dsrbecky:
In theory, finished.
In practice, I did not get around to testing it.

PR #1875, opened 2018-05-10
GLES Mid Execution Capture

pmuetschard:
Some refactoring to share code between Vulkan and GLES for state serialization and reworking of #1391 to use the same approach as Vulkan for serializing GPU buffers.

This should also improve GLES capture performance due to the use of `@spy_disable`.


pmuetschard:
Please be as picky about C++ design and syntax as you can, thanks!

pmuetschard:
Testing showed that commit 1aa941b has multiple issues and is not yet ready.

pmuetschard:
Not ready in its current state.

PR #1883, opened 2018-05-16
Elide ALL commands that we are not trying to trace.

AWoloszyn:
We were ignoring some of the data, but not all of it. We would
end up adding the commands to the tree, but no observations,
this caused havok with the new memory work, since replaying
those commands would cause OOB memory accesses.

### DIFF
--- a/gapii/cc/call_observer.cpp
+++ b/gapii/cc/call_observer.cpp
@@ -61,6 +61,7 @@ CallObserver::CallObserver(SpyBase* spy, CallObserver* parent, uint8_t api)
           [](uint8_t* buffer) { return releaseBuffer(buffer); }),
       mError(0 /*GL_NO_ERROR*/),
       mApi(api),
+      mShouldTrace(false),
       mCurrentThread(core::Thread::current().id()) {
 
     // context_t initialization.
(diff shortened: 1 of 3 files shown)


gapii/cc/call_observer.cpp
```diff
@@ -118,18 +118,31 @@ void CallObserver::observePending() {
...
 
 void CallObserver::encode(const ::google::protobuf::Message* cmd) {
+    if (!mSpy->should_trace(mApi)) {
+        return;
+    }
     encoder()->object(cmd);
 }
 
 void CallObserver::exit() {
+    if (!mSpy->should_trace(mApi)) {
```
ben-clayton: This scares me. If the `should_trace` return value changes between `enter` and `exit` then You're Going To Have A Bad Time. 
Are there cases where `should_trace` should change for the lifetime of a `CallObserver`? If not, can we fetch it once and store it as a member?

AWoloszyn: There are no cases (that I know) that would make this change during a call.
The only place we change ShouldTrace is during MEC start

AWoloszyn: I can definitely store it off as a member, that will prevent this from accidentally happening.

gapii/cc/spy_base.h
```diff
@@ -89,6 +89,10 @@ class SpyBase {
         return should_trace(api) ? mEncoder : mNullEncoder;
     }
 
+    std::shared_ptr<gapii::PackEncoder> parentEncoder(uint8_t api, CallObserver* parent) {
```
ben-clayton: This feels like a bit of a clunky method. There's nothing particularly parent-specific about the implementation, and the name has no real indication of what it does. 
I wonder if it would be clearer to just expose `mNullEncoder` and let the caller do this logic. This would also deobfuscate the call to `should_trace`.

PR #1886, opened 2018-05-16
WIP: Replace memory ranges with actual byte slices.

AWoloszyn:
This has one major piece of work left to do, which is to handle
OOB reads, and "virtual" writes, as in writes that require
replays to read.

They should not be a significant amount of extra work, but will
at least be some.

I also intend on fixing all of the DO NOT CHECK THIS IN, and re-enabling all of the tests before even trying to check this in.

Preliminary performance numbers vs master:
```
report-time                 | 35.665628305    | 22.059362156  (-38.1%)
linearize-time              | 51.404897128    | 34.734433885  (-32.4%)
```

### DIFF
--- a/core/memory/arena/arena.go
+++ b/core/memory/arena/arena.go
@@ -41,7 +41,8 @@ type Arena struct{ Pointer unsafe.Pointer }
 // You must call Dispose to free the arena object and any arena-owned
 // allocations.
 func New() Arena {
-	return Arena{Pointer: unsafe.Pointer(C.arena_create())}
+	a := Arena{Pointer: unsafe.Pointer(C.arena_create())}
+	return a
 }
 
 // Dispose destructs and frees the arena and all arena-owned allocations.
@@ -52,14 +53,14 @@ func (a Arena) Dispose() {
 
 // Allocate returns a pointer to a new arena-owned, contiguous block of memory
 // of the specified size and alignment.
-func (a Arena) Allocate(size, alignment int) unsafe.Pointer {
+func (a Arena) Allocate(size, alignment uint32) unsafe.Pointer {
 	a.assertNotNil()
 	return C.arena_alloc((*C.arena)(a.Pointer), C.uint32_t(size), C.uint32_t(alignment))
 }
 
 // Reallocate reallocates the memory at ptr to the new size and alignment.
 // ptr must have been allocated from this arena.
-func (a Arena) Reallocate(ptr unsafe.Pointer, size, alignment int) unsafe.Pointer {
+func (a Arena) Reallocate(ptr unsafe.Pointer, size, alignment uint32) unsafe.Pointer {
 	a.assertNotNil()
 	return C.arena_realloc((*C.arena)(a.Pointer), ptr, C.uint32_t(size), C.uint32_t(alignment))
 }
@@ -89,7 +90,7 @@ func (a Arena) Stats() Stats {
 	return Stats{int(numAllocs), int(numBytes)}
 }
 
-func (a Arena) assertNotNil() {
+func (a *Arena) assertNotNil() {
 	if a.Pointer == nil {
 		panic("nil arena")
 	}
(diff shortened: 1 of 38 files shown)


gapis/api/state.go
```diff
@@ -142,15 +150,24 @@ func (s GlobalState) MemoryDecoder(ctx context.Context, d memory.Data) *memory.D
...
+func (s GlobalState) MemoryEncoder(ctx context.Context, p memory.PoolID, rng memory.Range) *memory.Encoder {
+	return memory.NewEncoder(s.MemoryWriter(ctx, p, rng), s.MemoryLayout)
 }
 
 // Alloc allocates a memory range using the Allocator associated with
 // the given State, and returns a AllocResult that can be used to access the
 // pointer, and range.
 func (s *GlobalState) Alloc(ctx context.Context, size uint64) (AllocResult, error) {
-	at, err := s.Allocator.Alloc(size, 8)
+	// Force at least one byte to be allocated,
```
AWoloszyn: We could also just return nullptr here, and handle frees of nullptr specially. (a-la malloc/free)

gapis/config/config.go
```diff
@@ -18,7 +18,7 @@ package config
 const (
 	DebugReplay                = false
 	DebugReplayBuilder         = false
-	DisableDeadCodeElimination = false
+	DisableDeadCodeElimination = true
```
AWoloszyn: Definitely have to remove this

gapis/memory/blob.go
```diff
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io"
 
+	// REMOVE THIS BEFORE CHECKING IN
```
AWoloszyn: Just the comment... don't remove anything else but the comment that says "Remove"

gapis/memory/pool.go
```diff
@@ -62,65 +69,223 @@ const (
...
+		panic(fmt.Sprintf("Could not find allocation in pool: %+v, %+v, %+v", base, allocation, m.allocations))
+	}
+	return i
+}
+
+func (m *Pool) mustGetOffsetAndAllocation(rng Range) (int, uint64) {
+	i := m.mustGetAllocation(rng.Base)
+	allocation := m.allocations[i]
+	if allocation.offset+allocation.size < rng.Base+rng.Size {
+		panic(fmt.Sprintf("Slicing outside of range %+v %+v %+v \n\n---------------------------\n%+v\n----------------------\n", rng, rng.Size, allocation, m.allocations))
```
AWoloszyn: We probably want to make this a bit of a nicer message, it is useful for debug, but prints out EVERY valid range in the pool.

gapii/cc/vulkan_mid_execution.cpp
```diff
@@ -293,7 +293,7 @@ void VulkanSpy::prepareGPUBuffers(CallObserver *observer, PackEncoder *group,
     pool->buffer = nullptr;
 
     memory::Observation observation;
-    observation.set_base(0);
+    observation.set_base(pool->size - 1);
```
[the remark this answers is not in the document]
AWoloszyn: The point of these observations was to have at least one observations in every pool, so that we can create them correctly on replay side, without having to go through the state tree, find all of the pools, and create them that way.

But since now we have to pre-specify the size of a pool, now these observations occur at the end of the pool, so that the first observations we have encompass the entire pool.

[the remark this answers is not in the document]
ben-clayton: Ah okay, it's this weird 0 sized observation thing. Fair enough.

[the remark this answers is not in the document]
ben-clayton: A comment here explaining what you just said would be lovely.

core/memory/arena/arena.go
```diff
@@ -39,33 +41,41 @@ type Arena struct{ Pointer unsafe.Pointer }
 // New constructs a new arena.
 // You must call Dispose to free the arena object and any arena-owned
 // allocations.
-func New() Arena {
-	return Arena{Pointer: unsafe.Pointer(C.arena_create())}
+func New() *Arena {
+	a := &Arena{Pointer: unsafe.Pointer(C.arena_create())}
+	runtime.SetFinalizer(a, func(a *Arena) {
+		numBytes := a.Stats().NumBytesAllocated
```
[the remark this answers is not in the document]
AWoloszyn: (forcing a double indirection for debug reasons is not good).
This was not intended to be done for debug reasons, but specifically so we could call Dispose at a "reasonable" time. (i.e. when the arena is no longer referenced).

If we passed this around as a non-pointer, then we would run into the issue where we cannot automatically dispose of the object. That being said we could move SetFinalizer() into pool, and then destroy the arena when the pool goes out of scope.

I was not sure the best place to have the automatic cleanup take place.

gapis/memory/blob.go
AWoloszyn: Well Remove this comment of course

AWoloszyn:
This is so out of date now, its not worth keeping around. I will resurrect if needed in the future.

PR #1893, opened 2018-05-18
Fix tracking of image layout

### LINKED ISSUE #1850: Vulkan Image layout not tracked at the correct entity
Layouts can be different for each aspect, each layer and each mip level. Currently the entire image has one single layout, this is not correct.

Qining:
Track and check image layout at each image aspect/layer/level.

Fix #1850 

@bjoeris Sorry for disturbing, but hope you can take a look of this later, as this might affect the data flow analysis.

### DIFF
--- a/gapii/cc/vulkan_extras.cpp
+++ b/gapii/cc/vulkan_extras.cpp
@@ -27,9 +27,12 @@ struct destroyer {
     std::function<void(void)> destroy;
 };
 
+// Declared in api_spy.h.tmpl
 bool VulkanSpy::observeFramebuffer(CallObserver* observer,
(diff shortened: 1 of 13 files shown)


bjoeris:
Changing to per-level layout certainly makes sense, and I don't think it should mess up anything on my end.

I noticed that ImageInfo still has the old Layout field. Should it be removed? It seems that any use of the old per-image layout is probably incorrect.

Qining:
@bjoeris I think that 'Layout' in ImageInfo is just for showing the 'initial Layout' used for creating the image, I will change the name to 'initial layout' to make it clearer.

PR #1894, opened 2018-05-22
Fix tests.

ben-clayton:
Most were broken by the new gRPC replay changes.
`TestNewContextUndefined` gives slightly different results on different GPUs, so I've bumped the threshold a little.


ben-clayton:
You can run all the tests with:
`bazel test tests`

PR #1897, opened 2018-05-22
Allow packed arguments into our applications.

AWoloszyn:


AWoloszyn:
Did something similar, but not exactly what you had suggested. I added -packedargs to the command-list, but handle it before the rest of the command-line parsing. It means we can handle it recursively, without any weirdness going on.

core/app/run.go
```diff
@@ -131,9 +132,31 @@ func Run(main task.Task) {
 	rootCtx := prepareContext(&Flags.Log)
 
 	// parse the command line
+	args := os.Args[1:]
```
ben-clayton: Perhaps place into a private function to keep the length of this function sane?

PR #1901, opened 2018-05-22
Add frame range options to gapit stats

iburinoc:
Options emulate those of gapit video:
  - frame-start is the frame to start at
  - frame-count is the number of frames to consider after the start

### DIFF
--- a/cmd/gapit/stats.go
+++ b/cmd/gapit/stats.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"math"
 	"path/filepath"
+	"sort"
 
 	"github.com/google/gapid/core/app"
 	"github.com/google/gapid/core/log"
@@ -87,8 +88,11 @@ func (verb *infoVerb) getEventsInRange(ctx context.Context, client service.Servi
 		}
 	}
 
-	if verb.Frames.Start > len(fifIndices) {
-		return nil, log.Errf(ctx, nil, "Captured only %v frames, less than start frame %v", len(fifIndices), verb.Frames.Start)
+	if verb.Frames.Start < 0 {
+		return nil, log.Errf(ctx, nil, "Negative start frame %v is invalid", verb.Frames.Start)
+	}
+	if verb.Frames.Start >= len(fifIndices) {
+		return nil, log.Errf(ctx, nil, "Captured only %v frames, not greater than start frame %v", len(fifIndices), verb.Frames.Start)
 	}
 
 	startIndex := fifIndices[verb.Frames.Start]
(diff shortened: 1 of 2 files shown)


cmd/gapit/stats.go
```diff
@@ -62,20 +64,62 @@ func loadCapture(ctx context.Context, flags flag.FlagSet, gapisFlags GapisFlags)
...
+	begin, end := len(events), len(events)
+	for i, e := range events {
+		if i < begin && e.Command.Indices[0] >= startIndex {
+			begin = i
+		}
+		if i < end && e.Command.Indices[0] >= endIndex {
+			end = i
+		}
+	}
+	fmt.Println(startIndex, endIndex, begin, end)
```
ben-clayton: Is this debug?

iburinoc: Yep, oops, will fix.

cmd/gapit/stats.go
```diff
@@ -62,20 +64,61 @@ func loadCapture(ctx context.Context, flags flag.FlagSet, gapisFlags GapisFlags)
...
+
+	startIndex := fifIndices[verb.Frames.Start]
+	endIndex := uint64(math.MaxUint64)
+	if verb.Frames.Count >= 0 &&
+		verb.Frames.Start+verb.Frames.Count < len(fifIndices) {
+
+		endIndex = fifIndices[verb.Frames.Start+verb.Frames.Count]
+	}
+
+	begin, end := len(events), len(events)
```
ben-clayton: I think this is equivalent to:

```go

begin := sort.Search(len(events), func(i int) bool {
	return events[i].Command.Indices[0] >= startIndex
})
end := sort.Search(len(events), func(i int) bool {
	return events[i].Command.Indices[0] >= endIndex
})
```

cmd/gapit/stats.go
```diff
@@ -62,20 +64,61 @@ func loadCapture(ctx context.Context, flags flag.FlagSet, gapisFlags GapisFlags)
...
+}
+
+func (verb *infoVerb) Run(ctx context.Context, flags flag.FlagSet) error {
+	client, capture, err := loadCapture(ctx, flags, verb.Gapis)
+	if err != nil {
+		return err
+	}
+	defer client.Close()
+
+	events, err := verb.GetEventsInRange(ctx, client, capture)
```
ben-clayton: Shouldn't this be `getEventsInRange` ?

cmd/gapit/stats.go
iburinoc: Changed, thanks

cmd/gapit/stats.go
iburinoc: Oops, fixed, thanks

PR #1902, opened 2018-05-22
 Vulkan: Move most SpyOverride functions out of template

Qining:
Sits on top of #1893 

Most of the override override API functions and extern functions of
Vulkan are not generated from the template system. So move them to
vulkan_extras.cpp from vk_spy_helpers.cpp.tmpl.


ben-clayton:
Please hold off submitting this if approved, I'd like to properly review it tomorrow.

PR #1908, opened 2018-05-23
Encode physical device properties in extra

### LINKED ISSUE #1766: Instead of inserting synthetic "GetPhysicalDeviceProperties" use an Extra
We currently add synthetic GetPhysicalDeviceProperties after calls to VkEnumeratePhysicalDevices. Instead of doing this, Embed the important data into an extra attached to VkEnumeratePhysicalDevices.

Qining:
Fixes #1766 

So that we won't insert extra `vkGetPhysicalDeviceProperties` after `vkEnumeratePhysicalDevices`.

### DIFF
--- a/gapis/api/vulkan/api/physical_device.api
+++ b/gapis/api/vulkan/api/physical_device.api
@@ -72,8 +72,7 @@ cmd VkResult vkEnumeratePhysicalDevices(
       devices[i] = device
     }
     pPhysicalDeviceCount[0] = count
-    props := new!PhyDevProperties()
-    fetchPhysicalDeviceProperties(instance, devices, props)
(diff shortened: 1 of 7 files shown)


gapis/api/vulkan/api/physical_device.api
```diff
@@ -47,6 +47,10 @@
   @unused VkPhysicalDeviceProperties         PhysicalDeviceProperties
 }
 
+@internal class PhyDevProperties{
```
AWoloszyn: Can we use the full name here. Since this will show up in the state block, it would be good to keep it sane.

AWoloszyn:
I tested this out on a machine with multiple physical devices, and the remapping still works.

PR #1909, opened 2018-05-23
 Remove inserted commands for fetching image/buffer memory requirements

Qining:
Sits on top of #1908 

Remove inserted command for fetching image/buffer memory requirements  …
So that we won't see an extra vkGetImage/BufferMemoryRequirements
command inserted by GAPID after each vkCreateImage/Buffer command in the
command tree.

### DIFF
--- a/gapis/api/vulkan/api/image.api
+++ b/gapis/api/vulkan/api/image.api
@@ -115,6 +115,11 @@
   VkSparseMemoryBindFlags Flags
 }
 
+@internal class ImageMemoryRequirements {
+  VkMemoryRequirements MemoryRequirements
+  map!(u32, VkSparseImageMemoryRequirements) AspectBitsToSparseMemoryRequirements
+}
+
 @threadSafety("system")
 @indirect("VkDevice")
 @override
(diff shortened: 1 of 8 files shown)


gapis/api/vulkan/api/image.api
```diff
@@ -210,6 +217,17 @@ cmd VkResult vkCreateImage(
...
     }
   }
+
+  memRequirements := fetchImageMemoryRequirements(device, handle, hasSparseBit)
+  // If the vkCreateImage is inserted by GAPID (e.g. the staging image for
+  // reading framebuffer), NO memory requirements will be returned.
+  if memRequirements != null {
+    object.MemoryRequirements = memRequirements.MemoryRequirements
+    if (hasSparseBit) {
+      object.SparseMemoryRequirements = memRequirements.AspectBitsToSparseMemoryRequirements
```
Qining: I'm not very sure about this 'map to map' assignment, is it valid/allowed in API files?

ben-clayton: It is allowed, but be aware maps are ref types. If this map is coming from an extra, you'll need to clone it so that modifications don't alter the original deserialized instance. See gles/extras.go for examples.

Qining:
All three builds passed before the rebase.

PR #1915, opened 2018-05-24
gapis/api: Add arena parameter to all API type constructors.

ben-clayton:
Currently unused.

### DIFF
--- a/gapis/api/vulkan/command_buffer_rebuilder.go
+++ b/gapis/api/vulkan/command_buffer_rebuilder.go
@@ -55,6 +55,8 @@ func allocateNewCmdBufFromExistingOneAndBegin(
 	modelCmdBuf VkCommandBuffer,
 	s *api.GlobalState) (VkCommandBuffer, []api.Cmd, []func()) {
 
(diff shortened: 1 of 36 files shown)


ben-clayton:
Another thrilling one for you @AWoloszyn

gapis/api/vulkan/command_buffer_rebuilder.go
```diff
@@ -55,6 +55,8 @@ func allocateNewCmdBufFromExistingOneAndBegin(
 	modelCmdBuf VkCommandBuffer,
 	s *api.GlobalState) (VkCommandBuffer, []api.Cmd, []func()) {
 
+	a := s.Arena // TODO: Should this be a seperate temporary arena?
```
AWoloszyn: I think it could be, but definitely not for this MR. It's probably best to avoid that kind of optimization for now. I worry that there might be some tricky business with allocations.

ben-clayton: Completely agree. This is going to require a lot of iteration.

PR #1919, opened 2018-05-25
Fix the crash of Vulkan MEC due to fetching physical device properties

Qining:


Qining:
Build passed for all three platforms, tested manually.

PR #1924, opened 2018-05-28
Add enum annotation @analyze_usage to opt-in to usage analysis

iburinoc:
Currently only enums that get used as parameters or return values from
API functions get added to constant_sets, which doesn't work well for
Vulkan since so many of the enums are used indirectly through parameter
structs.
This allows any enum to be added to constant_sets by annotating it with
@unique.

Tested by adding @unique to VkMemoryPropertyFlagBits, needed for a WIP
memory breakdown feature.

### DIFF
--- a/gapil/semantic/type.go
+++ b/gapil/semantic/type.go
@@ -512,6 +512,30 @@ func UnsignedInteger(size int32) Type {
 	}
 }
 
+// AsUint64 converts a sized integer value to a uint64.
+func AsUint64(val Expression) uint64 {
+	switch v := val.(type) {
+	case Int8Value:
+		return uint64(v)
+	case Uint8Value:
+		return uint64(v)
+	case Int16Value:
+		return uint64(v)
+	case Uint16Value:
+		return uint64(v)
+	case Int32Value:
+		return uint64(v)
+	case Uint32Value:
+		return uint64(v)
+	case Int64Value:
+		return uint64(v)
+	case Uint64Value:
+		return uint64(v)
+	default:
+		panic(fmt.Sprintf("Attempted to convert %v to uint64", val.ExpressionType()))
+	}
+}
+
 // Underlying returns the underlying type for ty by recursively traversing the
 // pseudonym chain until reaching and returning the first non-pseudoym type.
 // If ty is not a pseudonym then it is simply returned.
--- a/gapil/semantic/type.go
+++ b/gapil/semantic/type.go
@@ -513,26 +513,26 @@ func UnsignedInteger(size int32) Type {
 }
 
 // AsUint64 converts a sized integer value to a uint64.
-func AsUint64(val Expression) uint64 {
+func AsUint64(val Expression) (uint64, bool) {
 	switch v := val.(type) {
 	case Int8Value:
-		return uint64(v)
+		return uint64(v), true
 	case Uint8Value:
-		return uint64(v)
+		return uint64(v), true
 	case Int16Value:
-		return uint64(v)
+		return uint64(v), true
 	case Uint16Value:
-		return uint64(v)
+		return uint64(v), true
 	case Int32Value:
-		return uint64(v)
+		return uint64(v), true
 	case Uint32Value:
-		return uint64(v)
+		return uint64(v), true
 	case Int64Value:
-		return uint64(v)
+		return uint64(v), true
 	case Uint64Value:
-		return uint64(v)
+		return uint64(v), true
 	default:
-		panic(fmt.Sprintf("Attempted to convert %v to uint64", val.ExpressionType()))
+		return uint64(0), false
 	}
 }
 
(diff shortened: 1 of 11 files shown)


gapil/template/constant_sets.go
```diff
@@ -188,6 +188,24 @@ func buildConstantSets(api *semantic.API, mappings *resolver.Mappings) *Constant
 			constsets[p] = b.addLabels(ev.Labels, ev.Ty.IsBitfield)
 		}
 	}
+	for _, e := range api.Enums {
+		unique := false
+		for _, annotation := range e.Annotations {
+			if annotation.Named == semantic.Named("unique") {
+				unique = true
```
apazylbe: break?

gapil/template/constant_sets.go
```diff
@@ -188,6 +188,25 @@ func buildConstantSets(api *semantic.API, mappings *resolver.Mappings) *Constant
...
+				break
+			}
+		}
+		if unique {
```
AWoloszyn: You can simplify this with
if unique := e.Annotations.GetAnnotation("unique"); unique != nil {

gapil/template/constant_sets.go
```diff
@@ -188,6 +188,25 @@ func buildConstantSets(api *semantic.API, mappings *resolver.Mappings) *Constant
...
+			labels := analysis.Labels{}
+			for _, entry := range e.Entries {
+				// Assume the enums are uint32
```
AWoloszyn: Enums only default to U32, they can be specified to be anything.

`Enum.NumberType` specifies the type of the enum. If we want to only support 32-bit for now, that's fine but we should error out of we can't handle it, rather than silently continue.

gapil/template/constant_sets.go
```diff
@@ -188,6 +188,25 @@ func buildConstantSets(api *semantic.API, mappings *resolver.Mappings) *Constant
...
+				if v, ok := entry.Value.(semantic.Uint32Value); ok {
+					labels[uint64(uint32(v))] = string(entry.Named)
```
AWoloszyn: Do we really need the double-cast here?

iburinoc:
Squashed all commits.

AWoloszyn:
Awesome thanks!
It looks great to me, but I would like @ben-clayton  to take a look at this to make sure. He understands the compiler side better than I.

iburinoc:
Done, this change also somewhat depends on #1930, I think the increase in number of enums processed and added to constant_sets made #1929 show up more.

gapis/api/vulkan/api/enums.api
```diff
@@ -91,6 +91,7 @@ enum VkResult {
 }
 
 /// Structure type enumerant
+@analyze_usage
```
AWoloszyn: Is there any reason you are specifically keeping analyze_usage on this enum (and the next)?

iburinoc: They have duplicate values (both have some labels that have <label> and <label>_KHR versions), and so as is constant_sets.go will panic when it encounters them.  I'm open to better solutions to enums with duplicates though.

AWoloszyn: That's fair. This should be ok for now. Those 2 enums are cases where we probably SHOULD be analyzing usage anyway.

AWoloszyn:
Looks good to me. Any comments @ben-clayton

gapil/template/constant_sets.go
```diff
@@ -175,6 +175,41 @@ func buildConstantSets(api *semantic.API, mappings *resolver.Mappings) *Constant
 
 	constsets := map[semantic.Node]*int{}
 
+	for _, e := range api.Enums {
+		if e.Annotations.GetAnnotation("analyze_usage") == nil {
+			labels := analysis.Labels{}
+			for _, entry := range e.Entries {
+				value := uint64(0)
+				switch v := entry.Value.(type) {
```
ben-clayton: This switch also [exists here](https://github.com/google/gapid/blob/master/gapil/analysis/value.go#L413). Please add a `func AsUint64(Expression) (uint64, bool)` function to `gapil/semantic/type.go` next to the other helpers (`IsInteger()`, `IsUnsigned()`, etc)

gapil/template/constant_sets.go
```diff
@@ -175,6 +175,41 @@ func buildConstantSets(api *semantic.API, mappings *resolver.Mappings) *Constant
 
 	constsets := map[semantic.Node]*int{}
 
+	for _, e := range api.Enums {
```
ben-clayton: A comment here explaining what this is doing would be great.

gapil/semantic/type.go
```diff
@@ -512,6 +512,30 @@ func UnsignedInteger(size int32) Type {
 	}
 }
 
+// AsUint64 converts a sized integer value to a uint64.
+func AsUint64(val Expression) uint64 {
```
ben-clayton: Please migrate the other switch statement over to call this function, and return a `(uint64, bool)` pair, where the second is a success or failure.

iburinoc: I did, I think the PR didn't update to show the other file.  Changed to `(uint64, bool)` now 👍

PR #1930, opened 2018-05-29
Don't build enum for nil node (fix #1929)

### LINKED ISSUE #1929: buildConstantSets sometimes creates incorrect enums
Sometimes (not sure what determines whether it will happen or not yet), `gapil/templates/constant_sets.go::buildConstantSets` creates an extra enum for Vulkan containing
`{ QUERY_STATUS_INACTIVE: 0, VK_IMAGE_ASPECT_COLOR_BIT: 1, QUERY_STATUS_COMPLETE: 2 }`.
It does not occur every time, and since buildConstantSets gets run twice (once to generate the `<api>/constant_sets.go` and once for the `<api>/api.go`, it can cause a discrepancy between a constant set ID and then strings associated with it.

It seems to be caused by `nil` being used as a key in `func (nl *nodeLabeler) addLabels(n semantic.Node, v analysis.Value)`, although I'm not sure what's causing the nondeterminism.

iburinoc:
It seems to be created non-deterministically, which can cause issues
with getting the right constant set ID value.

This change causes a constant set that was previously added to vulkan/constant_sets.go to be removed (`{QUERY_STATUS_INACTIVE, QUERY_STATUS_ACTIVE, QUERY_STATUS_COMPLETE}`), but I checked and this constant set was never actually referenced in vulkan/api.go.

This bug can be found (sometimes) with:
```
diff --git a/gapil/template/constant_sets.go b/gapil/template/constant_sets.go
index 58e821ff..83d33542 100644
--- a/gapil/template/constant_sets.go
+++ b/gapil/template/constant_sets.go
@@ -16,6 +16,7 @@ package template
 
 import (
 	"sort"
+	"strings"
 
 	"fmt"
 
@@ -66,6 +67,9 @@ func (b *constsetPackBuilder) addLabels(labels analysis.Labels, isBitfield bool)
 	}
 	set.idx = new(int)
 	b.sets[key] = set
+	if strings.Contains(key, "VK_IMAGE_ASPECT") && strings.Contains(key, "QUERY_STATUS") {
+		panic(":(")
+	}
 	return set.idx
 }
 ```
which checks for a specific bad enum that was being created.

### DIFF
--- a/gapil/template/constant_sets.go
+++ b/gapil/template/constant_sets.go
@@ -158,8 +158,8 @@ func (nl *nodeLabeler) traverse(n semantic.Node, v analysis.Value) {
 		}
 	case *analysis.MapValue:
 		for k, v := range v.KeyToValue {
-			nl.traverse(nil, k)
-			nl.traverse(nil, v)
+			nl.traverse(n, k)
+			nl.traverse(n, v)
 		}
 	}
 }

### CODE REVIEW

File: gapil/template/constant_sets.go [hunk 1]
```diff
@@ -137,6 +137,9 @@ func (nl *nodeLabeler) addLabels(n semantic.Node, v analysis.Value) {
 	if !ok || len(ev.Labels) == 0 {
 		return // Parameter wasn't an enum, or had no labels.
 	}
+	if n == nil {
```
AWoloszyn: I think the proper solution here is in 
```
 func (nl *nodeLabeler) traverse(n semantic.Node, v analysis.Value) {
```
traverse with `(n, k)  and (n, v)` instead of `(nil, k), (nil, v)`.

PR #1931, opened 2018-05-30
Pipe validation output to report view

Qining:
This CL does not package Vulkan validation layers in GAPID. It only pipes the output validation layers to the report view, if the validation layers are present and can be used to create instance.

For initial commands, their IDs are assigned to 0, and linearized ID
(the label) will be specified in the message.

GAPIS and GAPIR use same Severity level, which is extracted from
gapis/service/service.proto.

Some validation errors relevant to the virtual swapchain are deferred to
next CL to fix
 - vkGetPhysicalDeviceSurfaceFormatsKHR, 'count' arg not match
 - vkGetPhysicalDeviceSurfacePresentModesKHR, 'count' arg not match
 - vkCreateSwapChainKHR, 'imageFormat' does not match with 'colorSpace' member


gapis/api/vulkan/find_issues.go
```diff
@@ -55,12 +61,78 @@ func (t *findIssues) reportTo(r replay.Result) { t.res = append(t.res, r) }
...
+			b.RegisterNotificationReader(func(n gapir.Notification) {
+				vkApi := API{}
+				if uint8(n.GetApiIndex()) != vkApi.Index() {
+					return
+				}
+				var issue replay.Issue
+				// TODO: For MEC, we need to minus the number of initial commands
+				// TODO: Figure out what to do for commands inserted by GAPID, whose label is ~0.
+				issue.Command = api.CmdID(n.GetLabel())
+				issue.Severity = service.Severity(uint32(n.GetSeverity()))
```
Qining: Do we have a better way to enforce this? I found in gapis/service/service.proto, and core/log, we also just note in the comments to say their value should be the same.

gapis/api/templates/vulkan_gfx_api_extras.tmpl
```diff
@@ -76,66 +77,144 @@
...
     // null it out.
-    pCreateInfo->pNext = nullptr;
+    pInfo->pNext = nullptr;
 
+    if (!has_validation && addValidationLayer) {
+      layers[layer_count++] = validationLayerName;
+    }
+    pInfo->enabledLayerCount = layer_count;
+    if (!has_debug_report && addValidationLayer) {
+      exts[ext_count++] = debugReportName;
```
Qining: VK_EXT_debug_report is an instance extension so we should not enable it for device. This code is just added here to test the fallback route. Will be removed later in this PR.

gapir/client/connection.go
```diff
@@ -21,6 +21,7 @@ import (
 	"github.com/google/gapid/core/app/auth"
 	"github.com/google/gapid/core/log"
 	replaysrv "github.com/google/gapid/gapir/replay_service"
+	"github.com/google/gapid/gapis/service/severity"
```
Qining: Using the original 'Severity' enum defined in 'gapis/service/service.proto' results into cyclic dependency. So I have to extract out the severity to its own package.

gapir/replay_service/service.proto
```diff
@@ -86,12 +88,16 @@ message PostData {
 // build time of the replay instruction.
 message Notification {
   uint64 id = 1;
+  // Serverity defines the severity of a notification message.
+  // The values are identical to values in core/log/serverity.go and
+  // gapis/service/service.proto
```
Qining: Update the blurbs

gapis/api/vulkan/find_issues.go
```diff
@@ -16,31 +16,62 @@ package vulkan
...
+	"VK_LAYER_LUNARG_core_validation",
+	"VK_LAYER_GOOGLE_unique_objects",
+}
+
+const (
+	validationMetaLayer  = "VK_LAYER_LUNARG_standard_validation"
+	debugReportExtension = "VK_EXT_debug_report"
+)
+
+func isValidationLayer(n string) bool {
```
Qining: blurbs

gapis/replay/batch.go
```diff
@@ -164,8 +164,9 @@ func (m *Manager) execute(
 	}
 
 	var payload gapir.Payload
-	var decoder builder.ResponseDecoder
-	builderBuildTimer.Time(func() { payload, decoder, err = b.Build(ctx) })
+	var postResp builder.PostDataResponsor
```
Qining: Responsor? Responser? Receiver? Consumer?

gapis/replay/builder/builder.go
```diff
@@ -47,10 +47,16 @@ type marker struct {
...
 
-// ResponseDecoder decodes all postback responses from the replay virtual machine.
-// If err is nil, then r is the Reader to the sequential postback data. If r is nil,
-// then the postback data was absent or corrupted and err holds the error.
-type ResponseDecoder func(p *gapir.PostData)
+// NotificationResponsor takes a Notification message from the replay virtual
+// machine.
+type NotificationResponsor func(p *gapir.Notification)
+
+type ReadNotification func(p gapir.Notification)
```
Qining: Blurb

gapis/replay/builder/builder.go
```diff
@@ -566,10 +573,14 @@ func (b *Builder) Write(rng memory.Range, resourceID id.ID) {
 	b.ReserveMemory(rng)
 }
 
+func (b *Builder) RegisterNotificationReader(reader ReadNotification) {
```
Qining: Blurb

tools/build/rules/grpc_c++.bzl
```diff
@@ -110,6 +110,156 @@ def grpc_cc_library(name, srcs = [], public_hdrs = [], hdrs = [],
...
+    implementation = generate_cc_impl,
+)
+
+def generate_cc(well_known_protos, **kwargs):
+  if well_known_protos:
+    _generate_cc(well_known_protos="@com_google_protobuf//:well_known_protos", **kwargs)
+  else:
+    _generate_cc(**kwargs)
+
+def cc_grpc_library(name, proto, dep_cc_proto_libs, **kwargs):
```
Qining: `cc_grpc_library` is overwritten here (together with generate_cc to avoid dependency issue) because the gRPC original one depends on their in-house defined 'intermediate' targets when generating gRPC code for multiple proto files, makes the rule not good for external use, e.g. inferencing the original proto_library target names from proto-generated cpp file group target, but that rule of inference is only followed in gRPC repository.

Qining:
@AWoloszyn This CL is ready for review

core/vulkan/vk_virtual_swapchain/cc/swapchain.cpp
```diff
@@ -115,7 +115,7 @@ VKAPI_ATTR VkResult VKAPI_CALL vkGetPhysicalDeviceSurfaceCapabilitiesKHR(
   // TODO(awoloszyn): Handle all of the composite types.
 
   pSurfaceCapabilities->supportedUsageFlags =
-      VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT;
+      VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT|VK_IMAGE_USAGE_TRANSFER_SRC_BIT;
```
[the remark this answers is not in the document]
AWoloszyn: Yep, that is fair, although PROBABLY should be another CL.

gapir/replay_service/service.proto
```diff
@@ -86,12 +88,14 @@ message PostData {
 // build time of the replay instruction.
 message Notification {
```
AWoloszyn: The general rule for protobuf is that you never re-number fields. You only add to the end.
However since we ship the client&server in one package, this won't actually be a problem.

gapis/api/vulkan/find_issues.go
```diff
@@ -16,31 +16,64 @@ package vulkan
...
 	"github.com/google/gapid/gapis/api"
 	"github.com/google/gapid/gapis/api/transform"
 	"github.com/google/gapid/gapis/capture"
 	"github.com/google/gapid/gapis/replay"
 	"github.com/google/gapid/gapis/replay/builder"
 	"github.com/google/gapid/gapis/replay/value"
+	"github.com/google/gapid/gapis/service"
 )
 
+var validationLayers = [...]string{
```
AWoloszyn: That's new to me

gapis/api/vulkan/find_issues.go
```diff
@@ -55,17 +88,193 @@ func (t *findIssues) reportTo(r replay.Result) { t.res = append(t.res, r) }
...
+		info.SetPpEnabledLayerNames(NewCharᶜᵖᶜᵖ(layersData.Ptr()))
+		info.SetEnabledExtensionCount(uint32(len(exts)))
+		info.SetPpEnabledExtensionNames(NewCharᶜᵖᶜᵖ(extsData.Ptr()))
+		infoData := s.AllocDataOrPanic(ctx, info)
+
+		newCmd := cb.VkCreateInstance(infoData.Ptr(), cmd.PAllocator(), cmd.PInstance(), cmd.Result())
+		for _, d := range validationLayersData {
+			newCmd.AddRead(d.Data())
+		}
+		newCmd.AddRead(debugReportExtNameData.Data())
```
AWoloszyn: I think AddRead() returns the cmd.

newCmd.AddRead().AddRead().AddRead()....

gapis/api/vulkan/find_issues.go
```diff
@@ -55,17 +88,193 @@ func (t *findIssues) reportTo(r replay.Result) { t.res = append(t.res, r) }
...
+		validationLayersData := []api.AllocResult{}
+		for _, v := range validationLayers {
+			d := s.AllocDataOrPanic(ctx, v)
+			validationLayersData = append(validationLayersData, d)
+			layers = append(layers, NewCharᶜᵖ(d.Ptr()))
+		}
+		layersData := s.AllocDataOrPanic(ctx, layers)
+		info.SetEnabledLayerCount(uint32(len(layers)))
+		info.SetPpEnabledLayerNames(NewCharᶜᵖᶜᵖ(layersData.Ptr()))
+		infoData := s.AllocDataOrPanic(ctx, info)
```
AWoloszyn: We need to clean this up somewhere.

gapis/api/vulkan/find_issues.go
```diff
@@ -55,17 +88,193 @@ func (t *findIssues) reportTo(r replay.Result) { t.res = append(t.res, r) }
...
+			debugReportExtNameData = s.AllocDataOrPanic(ctx, debugReportExtension)
+			exts = append(exts, NewCharᶜᵖ(debugReportExtNameData.Ptr()))
+		}
+		extsData := s.AllocDataOrPanic(ctx, exts)
+
+		info.SetEnabledLayerCount(uint32(len(layers)))
+		info.SetPpEnabledLayerNames(NewCharᶜᵖᶜᵖ(layersData.Ptr()))
+		info.SetEnabledExtensionCount(uint32(len(exts)))
+		info.SetPpEnabledExtensionNames(NewCharᶜᵖᶜᵖ(extsData.Ptr()))
+		infoData := s.AllocDataOrPanic(ctx, info)
```
AWoloszyn: We have to clean this up somewhere

gapis/api/vulkan/find_issues.go
```diff
@@ -55,17 +88,193 @@ func (t *findIssues) reportTo(r replay.Result) { t.res = append(t.res, r) }
...
+		inst := ci.PInstance().MustRead(ctx, cmd, s, nil)
+		callbackHandle := VkDebugReportCallbackEXT(newUnusedID(true, func(x uint64) bool {
+			for _, cb := range t.reportCallbacks {
+				if uint64(cb) == x {
+					return true
+				}
+			}
+			return false
+		}))
+		callbackHandleData := s.AllocDataOrPanic(ctx, callbackHandle)
```
AWoloszyn: Need to clean these up.

gapis/api/vulkan/find_issues.go
Qining: Fixed in 2b9a068

gapis/api/vulkan/find_issues.go
Qining: Fixed in 2b9a068.

Similar to read_framebuffer.go, the memory is cleaned after at the return of find_issue's Transform call. It is one-command later then freeing just after the call to mutate() for `vkCreateInstance`, because I inserted a `vkCreateDebugReportCallbackEXT` after it. But I think it should be fine, using defer looks nicer than calling free() manually after each mutate().

gapis/api/vulkan/find_issues.go
Qining: Fixed in 2b9a068

gapis/api/vulkan/find_issues.go
Qining: Fixed in 2b9a068

gapir/cc/replay_connection.cpp
```diff
@@ -228,12 +230,40 @@ bool ReplayConnection::sendPostData(std::unique_ptr<Posts> posts) {
   return mGrpcStream->Write(res);
 }
 
-bool ReplayConnection::sendNotification(uint64_t id, uint32_t api_index,
-                                        uint64_t label, const std::string& msg,
+bool ReplayConnection::sendNotification(uint64_t id, int severity,
+                                        uint32_t api_index, uint64_t label,
+                                        const std::string& msg,
                                         const void* data, uint32_t data_size) {
+  using severity::Severity;
```
AWoloszyn:
```
const Severity logLevels[] = {
   Severity::FatalLevel,
   Severity::ErrorLevel,
   ...
}
if (severity <= LOG_LEVEL_DEBUG) {
   sev = logLevels[severity];
}
```

gapir/cc/replay_connection.h
```diff
@@ -225,9 +226,9 @@ class ReplayConnection {
   // Sends post data. Returns true if succeeded, otherwise returns false.
   virtual bool sendPostData(std::unique_ptr<Posts> posts);
   // Sends notification. Returns true if succeeded, otherwise returns false.
-  virtual bool sendNotification(uint64_t id, uint32_t api_index, uint64_t label,
-                                const std::string& msg, const void* data,
-                                uint32_t data_size);
+  virtual bool sendNotification(uint64_t id, int severity, uint32_t api_index,
```
AWoloszyn: Can we make severity unsigned? Is there any reason it can ever be negative?

tools/build/rules/grpc_c++.bzl
```diff
@@ -110,6 +110,156 @@ def grpc_cc_library(name, srcs = [], public_hdrs = [], hdrs = [],
     alwayslink = alwayslink,
   )
 
+
```
AWoloszyn: @pmuetschard Can you take a look at this part?

gapir/cc/replay_connection.h
Qining: Changed to uint32_t in c0e75da

tools/build/rules/grpc_c++.bzl
```diff
@@ -110,6 +110,156 @@ def grpc_cc_library(name, srcs = [], public_hdrs = [], hdrs = [],
...
+"""Generates C++ grpc stubs from proto_library rules.
+
+This is an internal rule used by cc_grpc_library, and shouldn't be used
+directly.
+"""
+
+def generate_cc_impl(ctx):
```
pmuetschard: Names that start with an underscore are considered private. So if you don't want functions to be used outside of this file, name them `_foo`. This is typically always done for rule implementations.

tools/build/rules/grpc_c++.bzl
pmuetschard: Done. Only one minor comment, otherwise looks fine.

tools/build/rules/grpc_c++.bzl
Qining: Done in 4caa404

PR #1948, opened 2018-06-01
Cleaned up a bunch of error message in Vulkan.

AWoloszyn:
Instead of returning a normal fmt.Errorf(), we now return
a proper service.ErrDataUnavailable. This allow the
UI to show a proper error message unstead of just a
panic in the log view.


AWoloszyn:
They are all certainly POSSIBLE with some user data. We don't expect most of them in a correctly written program, but I think we can hit them all normally.

With Qining getting the validation layers together, I am hoping that we can pipe any of the other errors that are "real" errors out more cleanly, and in one spot.

PR #1949, opened 2018-06-01
Dependency graph

bjoeris:
This is an initial commit of the Vulkan dependency graph code, and dead code elimination using the dependency graph.

The new DCE can work as a drop-in replacement for the existing Vulkan DCE, but this is disabled by default, using a config flag (`NewDeadCodeElimination`). This new DCE has been tested successfully against each vulkan_test_applications and SaschaWillems sample.

I'm working on a doc describing the design.

### DIFF
--- a/cmd/gapit/flags.go
+++ b/cmd/gapit/flags.go
@@ -265,4 +265,16 @@ type (
 		Gapis GapisFlags
 		At    flags.U64Slice `help:"command/subcommand index to get the memory after. Empty for last"`
 	}
+	TrimFlags struct {
+		Gapis         GapisFlags
+		Gapir         GapirFlags
+		Commands      bool           `help:"Treat every command as its own frame"`
+		ExtraCommands flags.U64Slice `help:"Additional commands to include (along with their dependencies)"`
+		Frames        struct {
+			Start int `help:"first frame to include (default 0)"`
+			Count int `help:"number of frames to include: -1 for all frames (default -1)"`
+		}
+		Out string `help:"gfxtrace file to save the trimmed capture"`
+		CommandFilterFlags
+	}
 )
--- a/cmd/gapit/trim.go
+++ b/cmd/gapit/trim.go
@@ -1,0 +1,149 @@
+// Copyright (C) 2018 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"context"
+	"flag"
+	"io/ioutil"
+	"path/filepath"
+
+	"github.com/google/gapid/core/app"
+	"github.com/google/gapid/core/log"
+	"github.com/google/gapid/gapis/service"
+	"github.com/google/gapid/gapis/service/path"
+)
+
+type trimVerb struct{ TrimFlags }
+
+func init() {
+	verb := &trimVerb{}
+	verb.Frames.Count = allTheWay
+
+	app.AddVerb(&app.Verb{
+		Name:      "trim",
+		ShortHelp: "(WIP) Trims a gfx trace to the dependencies of the requested frames",
+		Action:    verb,
+	})
+}
+
+func (verb *trimVerb) Run(ctx context.Context, flags flag.FlagSet) error {
+	if flags.NArg() != 1 {
+		app.Usage(ctx, "Exactly one gfx trace file expected, got %d", flags.NArg())
+		return nil
+	}
+
+	filepath, err := filepath.Abs(flags.Arg(0))
+	if err != nil {
+		return log.Errf(ctx, err, "Finding file: %v", flags.Arg(0))
+	}
+
+	client, err := getGapis(ctx, verb.Gapis, verb.Gapir)
+	if err != nil {
+		return log.Err(ctx, err, "Failed to connect to the GAPIS server")
+	}
+	defer client.Close()
+
+	capture, err := client.LoadCapture(ctx, filepath)
+	if err != nil {
+		return log.Errf(ctx, err, "LoadCapture(%v)", filepath)
+	}
+
+	eofEvents, err := verb.eofEvents(ctx, capture, client)
+	if err != nil {
+		return err
+	}
+
+	dceRequest := verb.getDCERequest(eofEvents, capture)
+	if len(dceRequest) > 0 {
+		capture, err = client.DCECapture(ctx, capture, dceRequest)
+		if err != nil {
+			return log.Errf(ctx, err, "DCECapture(%v, %v)", capture, dceRequest)
+		}
+	}
+
+	data, err := client.ExportCapture(ctx, capture)
+	if err != nil {
+		return log.Errf(ctx, err, "ExportCapture(%v)", capture)
+	}
+
+	output := verb.Out
+	if output == "" {
+		output = "trimmed.gfxtrace"
+	}
+	if err := ioutil.WriteFile(output, data, 0666); err != nil {
+		return log.Errf(ctx, err, "Writing file: %v", output)
+	}
+	return nil
+}
+
+func (verb *trimVerb) eofEvents(ctx context.Context, capture *path.Capture, client service.Service) ([]*service.Event, error) {
+	filter, err := verb.CommandFilterFlags.commandFilter(ctx, client, capture)
+	if err != nil {
+		return nil, log.Err(ctx, err, "Couldn't get filter")
+	}
+	requestEvents := path.Events{
+		Capture:     capture,
+		LastInFrame: true,
+		Filter:      filter,
+	}
+
+	if verb.Commands {
+		requestEvents.LastInFrame = false
+		requestEvents.AllCommands = true
+	}
+
+	// Get the end-of-frame events.
+	eofEvents, err := getEvents(ctx, client, &requestEvents)
+	if err != nil {
+		return nil, log.Err(ctx, err, "Couldn't get frame events")
+	}
+
+	lastFrame := verb.Frames.Start
+	if verb.Frames.Count > 0 {
+		lastFrame += verb.Frames.Count - 1
+	}
+	if lastFrame >= len(eofEvents) {
+		return nil, log.Errf(ctx, nil, "Requested frame %d, but capture only contains %d frames", lastFrame, len(eofEvents))
+	}
+
+	return eofEvents, nil
+}
+
+func (verb *trimVerb) getDCERequest(eofEvents []*service.Event, p *path.Capture) []*path.Command {
+	frameCount := verb.Frames.Count
+	if frameCount < 0 {
+		frameCount = len(eofEvents) - verb.Frames.Start + 1
+	}
+	dceRequest := make([]*path.Command, 0, frameCount+len(verb.ExtraCommands))
+	for i := 0; i < frameCount; i++ {
+		indices := eofEvents[verb.Frames.Start+i].Command.Indices
+		newIndices := make([]uint64, len(indices))
+		copy(newIndices, indices)
+		cmd := &path.Command{
+			Capture: p,
+			Indices: newIndices,
+		}
+		dceRequest = append(dceRequest, cmd)
+	}
+	for _, id := range verb.ExtraCommands {
+		cmd := &path.Command{
+			Capture: p,
+			Indices: []uint64{id},
+		}
+		dceRequest = append(dceRequest, cmd)
+	}
+	return dceRequest
+}
--- a/gapis/api/cmd.go
+++ b/gapis/api/cmd.go
@@ -18,6 +18,7 @@ import (
 	"context"
 
 	"github.com/google/gapid/core/fault"
+	"github.com/google/gapid/core/memory/arena"
 	"github.com/google/gapid/gapis/replay/builder"
 )
 
@@ -61,6 +62,9 @@ type Cmd interface {
 	// If the StateWatcher argument is not nil, then the StateWatcher
 	// methods will be called at the appropriate points during execution.
 	Mutate(context.Context, CmdID, *GlobalState, *builder.Builder, StateWatcher) error
+
+	// Clone makes a shallow copy of this command.
+	Clone(arena.Arena) Cmd
 }
 
 const (
--- a/gapis/api/cmd.go
+++ b/gapis/api/cmd.go
@@ -55,9 +55,12 @@ type Cmd interface {
 	// Extras returns all the Extras associated with the command.
 	Extras() *CmdExtras
 
-	// Mutate mutates the State using the command. If the builder argument is
-	// not nil then it will call the replay function on the builder.
-	Mutate(context.Context, CmdID, *GlobalState, *builder.Builder) error
+	// Mutate mutates the State using the command.
+	// If the builder argument is not nil then it will call the replay
+	// function on the builder.
+	// If the StateWatcher argument is not nil, then the StateWatcher
+	// methods will be called at the appropriate points during execution.
+	Mutate(context.Context, CmdID, *GlobalState, *builder.Builder, StateWatcher) error
 }
 
 const (
--- a/gapis/api/cmd.go
+++ b/gapis/api/cmd.go
@@ -65,6 +65,9 @@ type Cmd interface {
 
 	// Clone makes a shallow copy of this command.
 	Clone(arena.Arena) Cmd
+
+	// Alive returns true if this command should be marked alive for DCE
+	Alive() bool
 }
 
 const (
--- a/gapis/api/cmd_extras.go
+++ b/gapis/api/cmd_extras.go
@@ -37,6 +37,19 @@ func (e *CmdExtras) Add(es ...CmdExtra) {
 	}
 }
 
+// Replace replaces the `old` CmdExtra with the `new` CmdExtra.
+// If `old` is not found, this does nothing.
+func (e *CmdExtras) Replace(old, new CmdExtra) {
+	if e != nil {
+		for i, x := range *e {
+			if x == old {
+				(*e)[i] = new
+				return
+			}
+		}
+	}
+}
+
 // MustClone clones all of es, adding them to e.
 // if there was an error, a panic is raised
 func (e *CmdExtras) MustClone(es ...CmdExtra) {
--- a/gapis/api/gles/compat.go
+++ b/gapis/api/gles/compat.go
@@ -272,7 +272,7 @@ func compat(ctx context.Context, device *device.Instance, onError onCompatError)
 		switch cmd := cmd.(type) {
 		case *GlBindBuffer:
 			if cmd.Buffer() == 0 {
-				buf, err := subGetBoundBuffer(ctx, nil, api.CmdNoID, nil, s, GetState(s), cmd.Thread(), nil, cmd.Target())
+				buf, err := subGetBoundBuffer(ctx, nil, api.CmdNoID, nil, s, GetState(s), cmd.Thread(), nil, nil, cmd.Target())
 				if err != nil {
 					log.E(ctx, "Can not get bound buffer: %v ", err)
 				}
@@ -337,11 +337,11 @@ func compat(ctx context.Context, device *device.Instance, onError onCompatError)
 
 		case *GlBindBufferBase:
 			if cmd.Buffer() == 0 {
-				genBuf, err := subGetBoundBuffer(ctx, nil, api.CmdNoID, nil, s, GetState(s), cmd.Thread(), nil, cmd.Target())
+				genBuf, err := subGetBoundBuffer(ctx, nil, api.CmdNoID, nil, s, GetState(s), cmd.Thread(), nil, nil, cmd.Target())
 				if err != nil {
 					onError(ctx, id, cmd, fmt.Errorf("Can not get bound buffer: %v", err))
 				}
-				idxBuf, err := subGetBoundBufferAtIndex(ctx, nil, api.CmdNoID, nil, s, GetState(s), cmd.Thread(), nil, cmd.Target(), cmd.Index())
+				idxBuf, err := subGetBoundBufferAtIndex(ctx, nil, api.CmdNoID, nil, s, GetState(s), cmd.Thread(), nil, nil, cmd.Target(), cmd.Index())
 				if err != nil {
 					onError(ctx, id, cmd, fmt.Errorf("Can not get bound buffer: %v", err))
 				}
@@ -379,10 +379,10 @@ func compat(ctx context.Context, device *device.Instance, onError onCompatError)
 		case *GlVertexAttrib4fv:
 			if c.Vertex().Attributes().Contains(cmd.Location()) {
 				oldAttrib := c.Vertex().Attributes().Get(cmd.Location())
-				oldValue := oldAttrib.Value().MustRead(ctx, cmd, s, nil /* builder */)
-				cmd.Mutate(ctx, id, s, nil /* no builder, just mutate */)
+				oldValue := oldAttrib.Value().MustRead(ctx, cmd, s, nil /* builder */, nil /* watcher */)
+				cmd.Mutate(ctx, id, s, nil /* no builder, just mutate */, nil /* wacher */)
 				newAttrib := c.Vertex().Attributes().Get(cmd.Location())
-				newValue := newAttrib.Value().MustRead(ctx, cmd, s, nil /* builder */)
+				newValue := newAttrib.Value().MustRead(ctx, cmd, s, nil /* builder */, nil /* watcher */)
 				if reflect.DeepEqual(oldValue, newValue) {
 					// Ignore the call if it is redundant.
 					// Some applications iterate over all arrays and explicitly initialize them.
@@ -412,7 +412,7 @@ func compat(ctx context.Context, device *device.Instance, onError onCompatError)
 			// command.
 			// This is so we can grab the source string from the Shader object.
 			// We will actually provide the source to driver at compile time.
-			cmd.Mutate(ctx, id, s, nil /* no builder, just mutate */)
+			cmd.Mutate(ctx, id, s, nil /* no builder, just mutate */, nil /* watcher */)
 			return
 
 		case *GlCompileShader:
@@ -479,7 +479,7 @@ func compat(ctx context.Context, device *device.Instance, onError onCompatError)
 			// Currently the default framebuffer for replay is single-buffered
 			// and so we need to transform any usage of GL_BACK to GL_FRONT.
 			cmd.Extras().Observations().ApplyReads(s.Memory.ApplicationPool())
-			bufs := cmd.Bufs().Slice(0, uint64(cmd.N()), s.MemoryLayout).MustRead(ctx, cmd, s, nil)
+			bufs := cmd.Bufs().Slice(0, uint64(cmd.N()), s.MemoryLayout).MustRead(ctx, cmd, s, nil, nil)
 			for i, buf := range bufs {
 				if buf == GLenum_GL_BACK {
 					bufs[i] = GLenum_GL_FRONT
@@ -893,7 +893,7 @@ func compat(ctx context.Context, device *device.Instance, onError onCompatError)
 				// framebuffer. The state is mutated so that when a non-default
 				// framebuffer is bound later on, FRAMEBUFFER_SRGB will be enabled.
 				// (see GlBindFramebuffer below)
-				cmd.Mutate(ctx, id, s, nil /* no builder, just mutate */)
+				cmd.Mutate(ctx, id, s, nil /* no builder, just mutate */, nil /* watcher */)
 				return
 			}
 
@@ -956,7 +956,7 @@ func compat(ctx context.Context, device *device.Instance, onError onCompatError)
 
 		case *GlDeleteBuffers:
 			cmd.Extras().Observations().ApplyReads(s.Memory.ApplicationPool())
-			ids, err := cmd.Buffers().Slice(0, uint64(cmd.Count()), s.MemoryLayout).Read(ctx, cmd, s, nil)
+			ids, err := cmd.Buffers().Slice(0, uint64(cmd.Count()), s.MemoryLayout).Read(ctx, cmd, s, nil, nil)
 			if err == nil {
 				deleteCompat(ctx, ids, dID, dictionary.From(c.Objects().Buffers()), s, out,
 					func(cnt GLsizei, buf memory.Pointer) api.Cmd { return cb.GlDeleteBuffers(cnt, buf) })
@@ -970,7 +970,7 @@ func compat(ctx context.Context, device *device.Instance, onError onCompatError)
 			// framebuffers that are being deleted, and forward them to a fake
 			// call to glBindFramebuffer(XXX, 0) if we find any.
 			cmd.Extras().Observations().ApplyReads(s.Memory.ApplicationPool())
-			fbs, err := cmd.Framebuffers().Slice(0, uint64(cmd.Count()), s.MemoryLayout).Read(ctx, cmd, s, nil)
+			fbs, err := cmd.Framebuffers().Slice(0, uint64(cmd.Count()), s.MemoryLayout).Read(ctx, cmd, s, nil, nil)
 			if err == nil {
 				for _, fb := range fbs {
 					if fb == c.Bound().DrawFramebuffer().ID() {
@@ -988,7 +988,7 @@ func compat(ctx context.Context, device *device.Instance, onError onCompatError)
 
 		case *GlDeleteProgramPipelines:
 			cmd.Extras().Observations().ApplyReads(s.Memory.ApplicationPool())
-			ids, err := cmd.Pipelines().Slice(0, uint64(cmd.N()), s.MemoryLayout).Read(ctx, cmd, s, nil)
+			ids, err := cmd.Pipelines().Slice(0, uint64(cmd.N()), s.MemoryLayout).Read(ctx, cmd, s, nil, nil)
 			if err == nil {
 				deleteCompat(ctx, ids, dID, dictionary.From(c.Objects().Pipelines()), s, out,
 					func(cnt GLsizei, buf memory.Pointer) api.Cmd { return cb.GlDeleteProgramPipelines(cnt, buf) })
@@ -997,7 +997,7 @@ func compat(ctx context.Context, device *device.Instance, onError onCompatError)
 
 		case *GlDeleteQueries:
 			cmd.Extras().Observations().ApplyReads(s.Memory.ApplicationPool())
-			ids, err := cmd.Queries().Slice(0, uint64(cmd.Count()), s.MemoryLayout).Read(ctx, cmd, s, nil)
+			ids, err := cmd.Queries().Slice(0, uint64(cmd.Count()), s.MemoryLayout).Read(ctx, cmd, s, nil, nil)
 			if err == nil {
 				deleteCompat(ctx, ids, dID, dictionary.From(c.Objects().Queries()), s, out,
 					func(cnt GLsizei, buf memory.Pointer) api.Cmd { return cb.GlDeleteQueries(cnt, buf) })
@@ -1006,7 +1006,7 @@ func compat(ctx context.Context, device *device.Instance, onError onCompatError)
 
 		case *GlDeleteRenderbuffers:
 			cmd.Extras().Observations().ApplyReads(s.Memory.ApplicationPool())
-			ids, err := cmd.Renderbuffers().Slice(0, uint64(cmd.Count()), s.MemoryLayout).Read(ctx, cmd, s, nil)
+			ids, err := cmd.Renderbuffers().Slice(0, uint64(cmd.Count()), s.MemoryLayout).Read(ctx, cmd, s, nil, nil)
 			if err == nil {
 				deleteCompat(ctx, ids, dID, dictionary.From(c.Objects().Renderbuffers()), s, out,
 					func(cnt GLsizei, buf memory.Pointer) api.Cmd { return cb.GlDeleteRenderbuffers(cnt, buf) })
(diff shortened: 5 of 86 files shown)


gapis/api/cmd_extras.go
```diff
@@ -37,6 +37,18 @@ func (e *CmdExtras) Add(es ...CmdExtra) {
 	}
 }
 
+// Remove deletes a CmdExtra from the list of CmdExtras
```
Qining: Blurb does not match

gapis/api/reference.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"sync/atomic"
+)
+
+type RefID uint64
+
+const NilRefID = RefID(0)
+
+type Reference interface {
```
Qining: Blurb

gapis/api/vulkan/custom_replay.go
```diff
@@ -283,16 +283,16 @@ func (a *VkCreateDevice) Mutate(ctx context.Context, id api.CmdID, s *api.Global
...
 			for _, d := range allocated {
 				d.Free()
 			}
 		}()
 		extensionCount := uint64(createInfo.EnabledExtensionCount())
 		newExtensionNames := []memory.Pointer{}
-		for _, e := range createInfo.PpEnabledExtensionNames().Slice(0, extensionCount, s.MemoryLayout).MustRead(ctx, a, s, nil) {
-			extensionName := string(memory.CharToBytes(e.StringSlice(ctx, s).MustRead(ctx, a, s, nil)))
+		for _, e := range createInfo.PpEnabledExtensionNames().Slice(0, extensionCount, s.MemoryLayout).MustRead(ctx, a, s, nil, nil) {
+			extensionName := string(memory.CharToBytes(e.StringSlice(ctx, s).MustRead(ctx, a, s, nil, nil)))
```
Qining: If I remember it correctly, you can use `e.String()` directly.

gapis/resolve/dependencygraph2/dependency_graph.go
```diff
@@ -0,0 +1,425 @@
...
+	ForeachNode(cb func(NodeID, Node) error) error
+
+	// ForeachDependency iterates over all pairs (src, tgt), where src depends on tgt
+	ForeachDependency(cb func(NodeID, NodeID) error) error
+
+	// ForeachDependencyFrom iterates over all the nodes tgt, where src depends on tgt
+	ForeachDependencyFrom(src NodeID, cb func(NodeID) error) error
+
+	// ForeachDependencyTo iterates over all the nodes src, where src depends on tgt.
+	// If Config().ReverseDependencies is false, this will return an error.
```
iburinoc: `DependencyGraphConfig` doesn't currently have a `ReverseDependencies` field

gapis/capture/capture.go
```diff
@@ -362,12 +368,13 @@ type builder struct {
...
-		cmds:     []api.Cmd{},
-		resIDs:   []id.ID{id.ID{}},
-		arena:    arena.New(),
+		apis:         []api.API{},
+		seenAPIs:     map[api.ID]struct{}{},
+		observed:     interval.U64RangeList{},
+		cmds:         []api.Cmd{},
+		resIDs:       []id.ID{id.ID{}},
+		arena:        arena.New(),
+		initialState: &InitialState{APIs: map[api.API]api.State{}},
```
Qining: Maybe one day we can merge the `initialState` and `apis`. But probably too much for this CL.

cmd/gapit/flags.go
```diff
@@ -256,4 +256,22 @@ type (
...
 		}
 	}
+	TrimFlags struct {
+		Gapis    GapisFlags
+		Gapir    GapirFlags
+		Commands bool `help:"Treat every command as its own frame"`
+		No       struct {
+			DCE bool `help:"Prevent dead code elimination"`
+			New struct {
+				State bool `help:"Prevent the construction of a new initial state"`
```
AWoloszyn: It would be good to document somewhere (even if its just a comment) what this actually means.
What is the end-user impact of creating an initial state.
Is it the case that NOT specifying this flag will result in incorrect traces being generated?
If the last part is true, I would rather the correct trace be generated in the initial case, and a "smaller but possibly broken" trace being generated when specifying a flag.

gapis/resolve/dependencygraph2/builder.go
```diff
@@ -0,0 +1,303 @@
+// Copyright (C) 2018 Google Inc.
```
Qining: Maybe we can rename this file to `dependency_graph_builder.go`, as we have two different builders in this directory.

gapis/resolve/dependencygraph2/dce.go
```diff
@@ -0,0 +1,380 @@
...
+// If the old CmdID refers to a dead command, the returned command will refer to the next live command; if there is no next live command, api.CmdNoID is returned.
+func (b *DCEBuilder) LiveCmdID(oldCmdID api.CmdID) api.CmdID {
+	liveCmdID, ok := b.liveCmdIDs[oldCmdID]
+	if !ok {
+		return api.CmdNoID
+	}
+	return liveCmdID
+}
+
+// OriginalCmdIDs a live CmdID to the CmdID of the corresponding command in the original capture
```
Qining: OriginalCmdIDs maps a ....

gapis/resolve/dependencygraph2/dependency_graph.go
```diff
@@ -0,0 +1,425 @@
...
+	initialCmdRanges  interval.U64RangeList
+	nodes             []Node
+	numDependencies   int
+	dependenciesFrom  [][]NodeID
+	dependenciesTo    [][]NodeID
+
+	config DependencyGraphConfig
+}
+
+const initCmdBit = api.CmdID(1 << 62)
```
Qining: Should we mention that this is the same to api/cmd_id.go: `derivedBit`? Or some comment why choose the same value, or why it does not matter to use the same value?

gapis/api/vulkan/replay.go
```diff
@@ -479,17 +481,32 @@ func (a API) Replay(
...
+					cfg := dependencygraph2.DependencyGraphConfig{
+						MergeSubCmdNodes:       true,
+						IncludeInitialCommands: true,
+					}
+					graph, err := dependencygraph2.GetDependencyGraph(ctx, capture.Get(ctx), cfg)
+					if err != nil {
+						return 0, fmt.Errorf("Could not build dependency graph for DCE: %v", err)
+					}
+					dceInfo.newDce = dependencygraph2.NewDCEBuilder(graph)
+					// numInitialCommands = len(graph.InitialCommands())
```
AWoloszyn: Commented code.

AWoloszyn:
Would it be possible to get some information about the performance / size impact of this code. Since it is such a large change.

Just outputting the results from `cmd/regres` would be sufficient.
The simplest way to do this is to (on a separate branch) squash all of your commits, capture a rather large (Mid-Execution Capture) trace, and then run

`regres -trace <path/to/capture.gfxtrace> -optimize`

Which should output a bunch of information about times for compile  /  report (mutation) / linearize_trace(state rebuiding), as well as binary size comparisons.

gapis/resolve/dependencygraph2/dependency_graph.go
```diff
@@ -0,0 +1,425 @@
...
+type dependencyGraph struct {
+	capture           *capture.Capture
+	cmdNodeIDs        *api.SubCmdIdxTrie
+	initMemoryNodeIDs obsNodeIDs
+	initCmdObsNodeIDs []obsNodeIDs
+	cmdObsNodeIDs     []obsNodeIDs
+	initialCommands   []api.Cmd
+	initialCmdRanges  interval.U64RangeList
+	nodes             []Node
+	numDependencies   int
```
Qining: Any reason to not use unsigned integer?

gapis/resolve/dependencygraph2/builder.go
```diff
@@ -0,0 +1,303 @@
...
+}
+
+// EndCmd is called by the API emulator at the end of each API call
+func (b *dependencyGraphBuilder) OnEndCmd(ctx context.Context, cmdID api.CmdID, cmd api.Cmd) {
+	debug(ctx, "OnEndCmd [%d] %v", cmdID, cmd)
+	if len(b.currentDependencies) > 0 {
+		sortedDeps := make([]NodeID, len(b.currentDependencies), len(b.currentDependencies)+b.currentCreateCount)
+		copy(sortedDeps, b.currentDependencies)
+		b.currentDependencies = b.currentDependencies[:0]
+		sort.Slice(sortedDeps, func(i, j int) bool { return sortedDeps[i] < sortedDeps[j] })
```
Qining: Do we really need to sort here?

gapis/resolve/dependencygraph2/dependency_graph.go
```diff
@@ -0,0 +1,425 @@
...
+	return g.initialCmdRanges
+}
+
+// Config returns the config used to create this graph
+func (g *dependencyGraph) Config() DependencyGraphConfig {
+	return g.config
+}
+
+func (g *dependencyGraph) setDependencies(src NodeID, targets []NodeID) {
+	g.numDependencies += len(targets) - len(g.dependenciesFrom[src])
```
Qining: Is it possible that: `len(targets) < len(g.dependenciesFrom[src])`?

gapis/resolve/dependencygraph2/builder.go
```diff
@@ -0,0 +1,303 @@
...
+	// currentCmdID is the CmdID of the command currently being processed
+	currentCmdID api.CmdID
+
+	// currentCmd is the command currently being processed
+	currentCmd api.Cmd
+
+	// currentNodeID is the NodeID of the dependency graph node corresponding to the currentCmdID
+	currentNodeID NodeID
+
+	// currentDependencyExists[i] is true if currentNodeID depends on node i
```
Qining: Why don't we use `map[NodeID]struct{}` to store current dependencies, it seems such an unordered_set can replace `currentDependencyExists` and `currentDependencies` and also avoid the sorting and panic hassle.

gapis/resolve/dependencygraph2/dependency_graph.go
```diff
@@ -0,0 +1,425 @@
...
+func (g *dependencyGraph) InitialCmdRanges() interval.U64RangeList {
+	return g.initialCmdRanges
+}
+
+// Config returns the config used to create this graph
+func (g *dependencyGraph) Config() DependencyGraphConfig {
+	return g.config
+}
+
+func (g *dependencyGraph) setDependencies(src NodeID, targets []NodeID) {
```
Qining: Maybe we can have `addDependencies(src NodeID, tgts ...NodeID)` instead of two functions `setDependencies()` and `addDependency`

gapis/resolve/dependencygraph2/builder.go
```diff
@@ -0,0 +1,303 @@
...
+		}
+	} else if writeNode, ok := writes[frag]; ok {
+		writeEffect := WriteFragmentEffect{writeNode, frag}
+		b.addDependency(writeEffect, readEffect)
+	}
+}
+
+func (b *dependencyGraphBuilder) OnSet(ctx context.Context, owner api.Reference, frag api.Fragment, oldValueRef api.Reference, newValueRef api.Reference) {
+	debug(ctx, "  OnSet (%T %d)%v : %d → %d", owner, owner.RefID(), frag, oldValueRef.RefID(), newValueRef.RefID())
+	if _, ok := frag.(api.CompleteFragment); ok {
```
Qining: Actually I don't find any case of OnSet(..., api.CompleteFragment{}...). Feel like this is actually not a valid case, do we really want to handle such cases? Or we should panic or emit error?

gapis/resolve/dependencygraph2/dce.go
```diff
@@ -0,0 +1,380 @@
...
+		if !b.isLive[nodeID] {
+			b.isLive[nodeID] = true
+			nodes = append(nodes, nodeID)
+		}
+	}
+	t0 := dCE2Counter.Start()
+	b.markDependencies(nodes)
+	dCE2Counter.Stop(t0)
+
+	b.graph.ForeachNode(func(nodeID NodeID, node Node) error {
```
Qining: I think we need to document the traversing order for `ForeachNode` and how `addCmoNodes` makes sure the order requirement is satisfied, and how the use of `orphanObs` interact with the order. This is an implicit state machine.

gapis/resolve/dependencygraph2/dependency_graph.go
```diff
@@ -0,0 +1,425 @@
...
+func (g *dependencyGraph) GetNodeID(node Node) NodeID {
+	if cmdNode, ok := node.(CmdNode); ok {
+		index := cmdNode.Index
+		if len(index) == 0 {
+			return NodeNoID
+		}
+		if !g.config.IncludeInitialCommands && !api.CmdID(index[0]).IsReal() {
+			return NodeNoID
+		}
+		if g.config.MergeSubCmdNodes {
```
Qining: Do we actually support subcommands? It looks like we never put a command id with len() > 1 to the `cmdNodeIDs` trie.

Qining:
I think there are several things we don't handle in CL.

1) Subcommands, generally, if a queue submission is alive, all the commands submitted in that queue submission will be marked as alive, even though some of the command does not contribute to the final output.

2) VkDestroyXXX commands might all be killed.

For 1) I think it is fine for now. For 2), we need something like, take VkImage as an example:
 * If the VkCreateImage is alive, keep the VkDestroyImage alive also.
 * If the VkCreateImage is not alive, drop the VkDestroyImage also.
Seems like using the `OpenForwardDependencyEffect ` can fix those cases.

gapil/template/arith.go
```diff
@@ -0,0 +1,23 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package template
+
+func (*Functions) Add(x int, y int) int {
```
ben-clayton: Godocs please. Ubernit, but you can also share the types in parameters: `Add(x, y int)`

gapis/capture/capture.go
```diff
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io"
 	"sync"
+	"yeti/third_party/gapid/core/data/deep"
```
ben-clayton: Not used, please remove.

gapis/api/watcher.go
```diff
@@ -0,0 +1,94 @@
...
+	FieldName() string
+	ClassName() string
+}
+
+// StateWatcher provides callbacks to track state effects
+type StateWatcher interface {
+	// OnBeginCmd is called at the beginning of each API call
+	OnBeginCmd(ctx context.Context, cmdID CmdID, cmd Cmd)
+
+	// OnEndCmd is called by the API emulator at the end of each API call
```
ben-clayton: The term used for command 'emulation' I'm sticking with is 'executor'.
Regardless, maybe match the documentation of `OnBeginCmd`:

`// OnEndCmd is called at the end of each API call`

gapis/api/watcher.go
```diff
@@ -0,0 +1,94 @@
...
+	// OnReadSlice is called when reading from a slice
+	OnReadSlice(ctx context.Context, s memory.Slice)
+
+	// OnWriteObs is called when a memory write observation becomes visible
+	OnWriteObs(ctx context.Context, index int, obs CmdObservation)
+
+	// OnReadObs is called when a memory read observation becomes visible
+	OnReadObs(ctx context.Context, index int, obs CmdObservation)
+
+	// OnCreateResource is called when a resource is created
```
ben-clayton: Godoc and function name mismatch.

gapis/api/watcher.go
```diff
@@ -0,0 +1,94 @@
...
+	OnSet(ctx context.Context, owner Reference, f Fragment, old Reference, new Reference)
+
+	// OnWriteSlice is called when writing to a slice
+	OnWriteSlice(ctx context.Context, s memory.Slice)
+
+	// OnReadSlice is called when reading from a slice
+	OnReadSlice(ctx context.Context, s memory.Slice)
+
+	// OnWriteObs is called when a memory write observation becomes visible
+	OnWriteObs(ctx context.Context, index int, obs CmdObservation)
```
ben-clayton: For performance reasons, would it make sense to pass all the observations in this call?

gapis/api/cmd.go
```diff
@@ -57,7 +57,7 @@ type Cmd interface {
 
 	// Mutate mutates the State using the command. If the builder argument is
 	// not nil then it will call the replay function on the builder.
-	Mutate(context.Context, CmdID, *GlobalState, *builder.Builder) error
+	Mutate(context.Context, CmdID, *GlobalState, *builder.Builder, StateWatcher) error
```
ben-clayton: Please document that `StateWatcher` can be nil

gapis/api/templates/api.go.tmpl
```diff
@@ -162,14 +162,18 @@ import (
 
   type {{$name}} struct {
     data *[{{$ty.Size}}]{{$el}}
+    ϟrefID ϟapi.RefID
```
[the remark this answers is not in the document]
ben-clayton: IMO this should go away once we start using arena allocated data - pointers then become stable.

gapis/api/cmd.go
```diff
@@ -58,6 +59,9 @@ type Cmd interface {
 	// Mutate mutates the State using the command. If the builder argument is
 	// not nil then it will call the replay function on the builder.
 	Mutate(context.Context, CmdID, *GlobalState, *builder.Builder, StateWatcher) error
+
+	// CloneCmd makes a copy of this command.
```
ben-clayton: `// Clone`

Please document what is deep copied vs shallow.

gapis/api/gles/compat.go
```diff
@@ -209,7 +209,7 @@ func compat(ctx context.Context, device *device.Instance, onError onCompatError)
 			// TODO(dsrbecky): This might make some commands valid for replay which were invalid on trace.
 			scs := FindStaticContextState(s.Arena, cmd.Extras())
 			if !scs.IsNil() && !version.IsES && scs.Constants().MajorVersion() < 3 {
-				clone := cmd.Clone(s.Arena)
+				clone := cmd.Clone(s.Arena).(*EglMakeCurrent)
```
ben-clayton: Consider adding a `clone(arena.Arena) *T` method (note the lowercase), which `Clone(arena.Arena) api.Cmd` delegates on to.

gapis/api/reference.go
```diff
@@ -0,0 +1,40 @@
...
+	RefID() RefID
+}
+
+var lastRefID uint64
+
+func NewRefID() RefID {
+	id := RefID(atomic.AddUint64(&lastRefID, 1))
```
ben-clayton: Nit: No need for local.

gapis/config/config.go
```diff
@@ -18,8 +18,9 @@ package config
 const (
 	DebugReplay                = false
 	DebugReplayBuilder         = false
-	DisableDeadCodeElimination = false
+	DisableDeadCodeElimination = true
```
ben-clayton: intentional?

gapis/resolve/dependencygraph2/dce.go
```diff
@@ -0,0 +1,380 @@
...
+	dCE2DataLiveCounter = benchmark.Integer("DCE2.data.live")
+)
+
+// DCECapture returns a new capture containing only the requested commands and their dependencies.
+func DCECapture(ctx context.Context, name string, p *path.Capture, requestedCmds []*path.Command) (*path.Capture, error) {
+	c, err := capture.ResolveFromPath(ctx, p)
+	if err != nil {
+		return nil, err
+	}
+	fmt.Printf("DCE\n")
```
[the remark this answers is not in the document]
ben-clayton: Yes, please do not print directly to stdout.

gapil/template/arith.go
```diff
@@ -0,0 +1,23 @@
+// Copyright (C) 2017 Google Inc.
```
ben-clayton: 2018

gapil/template/BUILD.bazel
```diff
@@ -18,6 +18,7 @@ go_library(
     name = "go_default_library",
```
ben-clayton: Candidate for separate PR.

gapis/capture/capture.go
```diff
@@ -20,6 +20,7 @@ import (
 	"fmt"
```
ben-clayton: Candidate for separate PR.

gapis/api/vulkan/linux/vulkan_linux.api
```diff
@@ -74,6 +74,7 @@ cmd VkBool32 vkGetPhysicalDeviceXlibPresentationSupportKHR(
     u32              queueFamilyIndex,
```
ben-clayton: Candidate for separate PR.

gapis/api/cmd.go
```diff
@@ -18,6 +18,7 @@ import (
 	"context"
```
ben-clayton: Candidate for separate PR.

gapis/api/cmd_extras.go
```diff
@@ -37,6 +37,18 @@ func (e *CmdExtras) Add(es ...CmdExtra) {
 	}
```
ben-clayton: Candidate for separate PR.

bjoeris:
Thank you all for these detailed reviews. I'm still in MTV this week, and I've lost connectivity to my Linux workstation, so I think I'll need to wait until I get back next week to address these comments.

cmd/gapit/trim.go
```diff
@@ -0,0 +1,178 @@
...
+	if err != nil {
+		return log.Errf(ctx, err, "LoadCapture(%v)", filepath)
+	}
+
+	eofEvents, err := verb.eofEvents(ctx, capture, client)
+	if err != nil {
+		return err
+	}
+
+	// begin, count := verb.getRange(eofEvents)
```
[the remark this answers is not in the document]
bjoeris: Correct, until I can rebuild the captures InitialState, TrimCapture cannot work correctly on anything but very simple captures. I have decided to leave this functionality out for now, so `gapit trim` just takes the eof commands, and all of their dependencies, which may include earlier, unrequested frames. I will remove the commented out code.

cmd/gapit/flags.go
bjoeris: `no-new-state` is only relevant with with the InitialState rebuilding code, which is currently broken and not included in this PR. I will also remove this flag from this PR.

Once we have InitialState rebuilding, the user is not likely to care about this flag, but it is useful for testing/debugging.

gapis/resolve/dependencygraph2/dependency_graph.go
bjoeris: The only reason it is redefined here is that `derivedBit` is not exported. Upon further reflection, this is a bad coupling between these packages; I will move the bit manipulations into `CmdID` methods.

gapis/api/templates/api.go.tmpl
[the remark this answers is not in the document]
bjoeris: Yes, the `RefID`s are intended primarily to avoid the `DependencyGraphBuilder` holding onto references to every object created. 

@ben-clayton Does the arena get around the problem of pointer re-use that @AWoloszyn mentioned?

gapis/resolve/dependencygraph2/dependency_graph.go
```diff
@@ -0,0 +1,425 @@
...
+
+	// Capture returns the capture whose dependencies are stored in this graph
+	Capture() *capture.Capture
+
+	// GetCommand returns the command identified by the given CmdID
+	GetCommand(api.CmdID) api.Cmd
+
+	// InitialCommands returns the initial commands, which reconstruct the
+	// initial state before the first command in the capture.
+	InitialCommands() []api.Cmd
```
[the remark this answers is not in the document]
bjoeris: I think I was originally holding onto these to avoid recomputing them. Now that they are cached, I will remove them from the dependency graph.

gapis/resolve/dependencygraph2/dependency_graph.go
bjoeris: No, there is no reason for it to be signed. I'll also make it explicitly 64 bit, since this can easily exceed 2^32.

gapis/resolve/dependencygraph2/builder.go
bjoeris: You're correct that the sort is not currently needed. The sort will be needed if we want to query whether one node is a dependency of another node. I will take the sort out for now, and put it back if/when it is actually needed.

gapis/resolve/dependencygraph2/dependency_graph.go
bjoeris: Yes this is possible. I will split this into two operations to be safe with the change of numDependencies to unsigned.

gapis/resolve/dependencygraph2/builder.go
bjoeris: Using the two slices significantly improved performance over a map. The dependency graph can easily have billions of dependencies, and, when there is a dependency between two nodes, there are often 10s or 1000s of different reads/writes that give the same dependency. Each of these redundant dependencies result in an array lookup in `currentDependencyExists`, rather than an attempted map insertion.

I don't think I kept the profiling data around (this was when I was doing a lot of experimentation with this code), but as I recall the overall improvement in the time to construct large dependency graphs improved by 10-20%.

gapis/resolve/dependencygraph2/dependency_graph.go
bjoeris: `setDependencies` exists for performance.

Using `addDependency` instead would result in `dependenciesFrom[src]` to be dynamically grown by `append`. This would be slow, and also result in `dependenciesFrom[src]` being significantly larger than necessary, on average. 

Instead, the `DependencyGraphBuilder` creates a slice with almost* exactly the right capacity, and assigns dependenciesFrom directly, using `setDependencies`. (*forward dependencies make this slightly more complicated, but are much more rare than backward dependencies).

gapis/resolve/dependencygraph2/builder.go
bjoeris: You're right that `CompleteFragment` is not currently used with `OnSet`, but this is a perfectly valid case. I think it is better to keep this case, so that `OnSet` can correctly handle all types of `Fragments`.

This was used previously when non-reference structs contained their data directly, rather than through a `data` pointer. At that point, assigning to a struct was the same as assigning to all of it's fields, which was represented by `OnSet` with a `CompleteFragment`.

I can also imagine other scenarios that might use `CompleteFragment` with `OnSet`: e.g., `Clear` method for maps.

bjoeris: RE: `oldValueRef` and `newValueRef`, these are helpful for debugging. They can also be used to reconstruct a human readable path to the fragments (to get something like `Buffers[0x1234].Info.Size` instead of `0x6789.Size`). However, maintaining enough data to reconstruct these human readable paths is expensive, and unnecessary for DCE, so I have not included that code in this PR.

gapis/resolve/dependencygraph2/dce.go
```diff
@@ -0,0 +1,380 @@
...
+				Reads: newReads,
+			}
+			extras.Add(newObs)
+		} else {
+			newReads := make([]api.CmdObservation, 0, len(b.orphanObs)+len(obs.Reads))
+			for _, o := range b.orphanObs {
+				newReads = append(newReads, o.CmdObservation)
+			}
+			newReads = append(newReads, obs.Reads...)
+			newWrites := make([]api.CmdObservation, len(obs.Writes))
```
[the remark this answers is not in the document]
bjoeris: I want to be careful not to modify any of the existing data, which is still associated with the old capture. That's why there are all these explicit new slices.

gapis/resolve/dependencygraph2/dce.go
bjoeris: I will add documentation to clarify this order, and the algorithm used by DCE more broadly.

gapis/resolve/dependencygraph2/dependency_graph.go
bjoeris: At this point, the dependency graph and DCE do not handle subcommands. I hope to add this capability soon, but I am worried about memory and performance. This is likely to at least double the number of nodes--each `vkCmd...` is will have a node when recorded to the command buffer, and a node each time the command buffer is submitted.

gapis/api/cmd_extras.go
bjoeris: Done

gapis/api/reference.go
bjoeris: Done

gapis/resolve/dependencygraph2/dependency_graph.go
bjoeris: Done

cmd/gapit/trim.go
```diff
@@ -0,0 +1,178 @@
...
+		cmd := &path.Command{
+			Capture: p,
+			Indices: newIndices,
+		}
+		dceRequest = append(dceRequest, cmd)
+	}
+	for _, id := range verb.ExtraCommands {
+		cmd := &path.Command{
+			Capture: p,
+			Indices: []uint64{id},
```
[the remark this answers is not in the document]
bjoeris: Removed code referring to startCmd until I can do proper trimming (including initial state rebuilding)

gapis/api/watcher.go
```diff
@@ -0,0 +1,94 @@
...
+	OnReadObs(ctx context.Context, index int, obs CmdObservation)
+
+	// OnCreateResource is called when a resource is created
+	OpenForwardDependency(ctx context.Context, dependencyID interface{})
+
+	// OnDestroyResource is called when a resource is destroyed
+	CloseForwardDependency(ctx context.Context, dependencyID interface{})
+}
+
+type Fragment interface {
```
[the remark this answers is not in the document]
bjoeris: Done

gapis/resolve/dependencygraph2/dce.go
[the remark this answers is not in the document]
bjoeris: Done

gapis/resolve/dependencygraph2/dce.go
```diff
@@ -0,0 +1,380 @@
...
+				}
+			}
+			liveCmdID := b.LiveCmdID(obsNode.CmdID)
+			fmt.Printf("[ %s ] (%v [%s] / %v) Range: %v  Pool: %v  ID: %v\n", status, obsNode.CmdID, cmdStatus, liveCmdID, obsNode.CmdObservation.Range, obsNode.CmdObservation.Pool, obsNode.CmdObservation.ID)
+		}
+		return nil
+	})
+}
+
+// func (b *DCEBuilder) printLiveCmds() {
```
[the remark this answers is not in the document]
bjoeris: Removed

gapis/resolve/dependencygraph2/dce.go
```diff
@@ -0,0 +1,380 @@
...
+		100*b.numDead/numCmd, b.numDead, b.deadMem/1024/1024,
+		100*b.numLive/numCmd, b.numLive, b.liveMem/1024/1024,
+		dCE2Counter.Get())
+}
+
+func (b *DCEBuilder) markDependencies(nodes []NodeID) {
+	// TODO: See if a more efficient queue is necessary
+	queue := make([]NodeID, len(nodes))
+	copy(queue, nodes)
+	// nodeStr := func(nodeID NodeID) string {
```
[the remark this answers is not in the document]
bjoeris: Removed

gapis/resolve/dependencygraph2/builder.go
bjoeris: Done

gapis/resolve/dependencygraph2/dce.go
bjoeris: Done

gapis/api/vulkan/replay.go
bjoeris: Removed

gapis/resolve/dependencygraph2/dependency_graph.go
```diff
@@ -0,0 +1,425 @@
...
+
+	// InitialCmdRanges returns the memory ranges used by the initial commands.
+	InitialCmdRanges() interval.U64RangeList
```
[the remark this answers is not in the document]
bjoeris: Removed

gapis/resolve/dependencygraph2/memory.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+type memoryWrite struct {
+	effect WriteMemEffect
+	span   interval.U64Span
+}
+
+type memoryWriteList []memoryWrite
+
+func (l *memoryWriteList) Length() int {
```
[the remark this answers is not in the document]
bjoeris: Done

gapis/resolve/dependencygraph2/builder.go
```diff
@@ -0,0 +1,303 @@
...
+}
+
+// OnReadSlice is called when reading from a slice
+func (b *dependencyGraphBuilder) OnReadSlice(ctx context.Context, slice memory.Slice) {
+	debug(ctx, "  OnReadSlice: %v", slice)
+	b.addMemoryRead(ReadMemEffect{b.currentNodeID, slice})
+}
+
+// ObservationSlice constructs a Slice from a CmdObservation
+func ObservationSlice(obs api.CmdObservation) memory.Slice {
```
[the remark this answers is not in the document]
bjoeris: Done

gapis/resolve/dependencygraph2/dce.go
```diff
@@ -0,0 +1,380 @@
...
+		cloneCmd()
+		cmd.SetCaller(b.LiveCmdID(cmd.Caller()))
+	}
+	if len(b.orphanObs) > 0 {
+		cloneCmd()
+		extras := make(api.CmdExtras, len(cmd.Extras().All()))
+		copy(extras, cmd.Extras().All())
+		obs := extras.Observations()
+		if obs == nil {
+			newReads := make([]api.CmdObservation, 0, len(b.orphanObs))
```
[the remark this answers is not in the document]
bjoeris: I don't want to modify Observations() directly, because it is shared with the old command, in the old capture. I also want the orphan observations to come before the command's own read observations because it is possible that the command had a different, later observation of an overlapping memory range, in which case the order of the observations matters.

gapil/template/arith.go
bjoeris: Removed this file, as I realized I am no longer using it.

gapis/api/watcher.go
bjoeris: Done

gapis/api/watcher.go
bjoeris: Done

gapis/api/watcher.go
bjoeris: Done. I don't think it will have a significant performance impact, but it does make the interface cleaner by removing the `index` parameter.

bjoeris:
@AWoloszyn Here is the regres output for a 20 frame mid-execution capture of vkquake. The patch for regres included enabling the new dead code elimination, which is disabled by default in this PR.

I'll also try to get data from a bigger capture. Any suggestions for an application to try? I couldn't get a capture from 3dmark on my Linux box.

Also, it seems like frames, draws and commands stats are missing because I didn't use a pkg.

```
sha                         | 4cc41d          | dc3349        
incremental-build           | 193.790890192   | 199.272569855 (+2.8%)
libgapii                    | 7998136         | 8010616       (+0.2%)
libVkLayer_VirtualSwapchain | 43096           | 43096         
gapid-aarch64               | 7171720         | 7174237       (+0.0%)
gapid-armeabi               | 6198934         | 6201629       (+0.0%)
gapid-x86                   | 8184919         | 8188600       (+0.0%)
gapid                       | 1633534         | 1633534       
gapir                       | 3326520         | 3326520       
gapis                       | 91830504        | 113162824     (+23.2%)
gapit                       | 14028696        | 14053336      (+0.2%)
frames                      | 0               | 0             
draws                       | 0               | 0             
commands                    | 0               | 0             
report-time                 | 1.196185294     | 7.010618572   (+486.1%)
linearize-time              | 54.285633154    | 12.079515047  (-77.7%)
```

gapis/resolve/dependencygraph2/dependency_graph_builder.go
```diff
@@ -0,0 +1,303 @@
...
+		b.addObs(i, o, false)
+	}
+}
+
+// OpenForwardDependency is called to begin a forward dependency.
+// See `StateWatcher.OpenForwardDependency` for an explanation of forward dependencies.
+func (b *dependencyGraphBuilder) OpenForwardDependency(ctx context.Context, dependencyID interface{}) {
+	debug(ctx, "  OpenForwardDependency: %v", dependencyID)
+	if _, ok := b.openForwardDependencies[dependencyID]; ok {
+		log.I(ctx, "OpenForwardDependency: Forward dependency opened multiple times before being closed. DependencyID: %v, close node: %v", dependencyID, b.currentNodeID)
```
iburinoc: `StateWatcher.OpenForwardDependency` says there should be exactly one close per open, however this log line is getting called quite frequently in my testing.  Is this intended behaviour, and if so then perhaps it would be better at debug instead of info?

gapis/api/templates/api.go.tmpl
```diff
@@ -340,18 +485,65 @@ import (
...
+
   // Clone makes a deep copy of this map.
   func (m {{$name}}) Clone(ϟa arena.Arena, ϟseen ϟapi.CloneContext) {{$name}} {
-    if existing, ok := ϟseen[m.Map]; ok {
-      return {{$name}}{Map: existing.(*map[{{$key}}]{{$value}}), a: ϟa}
+    if existing, ok := ϟseen[m.ϟrefID]; ok {
+      return existing.({{$name}})
     }
     out := make(map[{{$key}}]{{$value}}, len(*m.Map))
-    ϟseen[m.Map] = &out
```
pmuetschard: This was here on purpose - before the recursion - to handle cycles.

gapis/api/templates/api.go.tmpl
```diff
@@ -396,21 +589,55 @@ import (
...
   // Clone makes a deep copy of this reference.
   func (c {{$name}}) Clone(ϟa arena.Arena, ϟseen ϟapi.CloneContext) {{$name}} {
     if c.IsNil() { return Nil{{$name}} }
-    if existing, ok := ϟseen[c.data]; ok {
-      return {{$name}}{data: existing.(*{{$data}})}
+    if existing, ok := ϟseen[c.ϟrefID]; ok {
+      return existing.({{$name}})
     }
-    d := &{{$data}}{}
-    ϟseen[c.data] = d
```
pmuetschard: This was here on purpose - before the recursion - to handle cycles.

bjoeris: Understood. I have changed the `Clone` templates so that `ϟseen` is set before the recursive calls.

bjoeris:
@AWoloszyn The results of regres on the capture from @Qining is [here](https://gist.github.com/bjoeris/8c576604980168940459c20f1fb5cb78). I ran it with both the old DCE (still used by default) and new dependency graph DCE (turned on in config.go).

gapis/api/templates/api.go.tmpl
bjoeris: I have changed all the `ϟseen` values like this to be set in `Clone` before any recursion.

bjoeris:
I found a bug apparently introduced by my latest changes. Please do not merge yet.

AWoloszyn:
Can we close this now as I think this was merged in a different MR? (I just wanted to get to 100 comments)

PR #1950, opened 2018-06-01
Fix command tree grouping by frame.

pmuetschard:
Fixes a bug introduced by commit 62fa80ab0, where the "last in frame" events inferred from "start of frame" commands where added after the "start of frame" event instead of before it, essentially ending frames on the same command as starting them.


pmuetschard:
+CC @iburinoc

PR #1954, opened 2018-06-04
Vulkan memory breakdown

iburinoc:
This creates a GAPIS service endpoint to get memory layout information at a given point in a trace, and implement it for Vulkan.  It also adds a gapit verb to print out this information textually, along with information about aliased regions.  No visual representation exists for now, just a text report.


gapis/resolve/metrics.go
```diff
@@ -0,0 +1,68 @@
...
+	switch p.Type {
+	case path.Metrics_MEMORY:
+		return memoryMetrics(ctx, p)
+	default:
+		return nil, fmt.Errorf("Metrics type %v not implemented", p.Type)
+	}
+
+}
+
+// FIXME DO NOT MERGE: Not sure where the best spot for this is, it can't go in gapis/api because api can't depend on service
```
iburinoc: This is likely not the best spot for this interface, maybe it should go in `gapis/service` along with the MemoryLayout proto definition?

cmd/gapit/memory.go
```diff
@@ -0,0 +1,232 @@
...
+				}
+			}
+		}
+	}
+	return nil
+}
+
+type bindingSlice []*service.MemoryBinding
+
+func (bindings bindingSlice) BindingLess(i, j int) bool {
```
apazylbe: Does it need to be public? If so, consider adding documenting.

cmd/gapit/memory.go
```diff
@@ -0,0 +1,232 @@
...
+	if bindings[i].Offset < bindings[j].Offset {
+		return true
```
apazylbe: Could simply, up to you
if bindings[i].Offset != bindings[j].Offset {
       return bindings[i].Offset < bindings[j].Offset
}

gapis/api/vulkan/memory_layout.go
```diff
@@ -0,0 +1,114 @@
...
+)
+
+// MemoryLayout stores an overview of the state's memory layout into a
+// service.MemoryLayout object.  The layout includes data on memory types,
+// allocations, and which resources are bound to which locations in memory.
+func (a API) MemoryLayout(st *api.GlobalState) (*service.MemoryLayout, error) {
+	s := GetState(st)
+	allocations := []*service.MemoryAllocation{}
+	// Serialize data on all allocations into protobufs
+	for handle, info := range *(s.DeviceMemories().Map) {
```
Qining: Prefer to `s.DeviceMemories().All()`

gapis/api/vulkan/memory_layout.go
```diff
@@ -0,0 +1,114 @@
...
+		var size VkDeviceSize
+		var typ service.MemoryBinding_BindingType
+		if buffer, ok := s.Buffers().Lookup(VkBuffer(handle)); ok {
+			typ = service.MemoryBinding_BUFFER
+			size = buffer.Info().Size()
+		} else if image, ok := s.Images().Lookup(VkImage(handle)); ok {
+			typ = service.MemoryBinding_IMAGE
+			size = image.MemoryRequirements().Size()
+		} else {
+			return nil, fmt.Errorf("Bound object %v is not a buffer or an image", handle)
```
Qining: I remember for sparse image and buffers, we didn't register the binding info on the memory object, only the buffer/image object. So we probably want to have a TODO or Note here to say some cases are not handled yet.

gapis/service/service.proto
```diff
@@ -671,6 +672,76 @@ message MemoryRange {
...
 }
 
+message Metrics {
+  oneof metrics {
+    MemoryLayout memory_layout = 1;
+  }
+}
+
+// The description of the memory layout of the API state
+message MemoryLayout {
```
Qining: Feel `MemoryLayout` name is a little bit confusing in GAPID, as we use this name for 'alignment, data width' things also. May be something like: `MemoryUsage`?

@AWoloszyn @ben-clayton @pmuetschard Any better idea or just using `MemoryLayout` is fine?

gapis/service/service.proto
```diff
@@ -671,6 +672,76 @@ message MemoryRange {
...
+  // The memory allocations that are currently in use.
+  repeated MemoryAllocation allocations = 3;
+}
+
+// A single memory allocation
+message MemoryAllocation {
+  // The device this allocation was made on
+  uint64 device = 1;
+  // The memory type this allocation is from
+  uint32 memory_type = 2;
```
Qining: Looks like this is more like `memory_type_index`? But `index` also sounds too Vulkan specific, I'm not sure if in GL world there is a thing like `type index`.

cmd/gapit/memory.go
```diff
@@ -0,0 +1,232 @@
...
+		e, _ := endsAt[p]
+		for _, handle := range e {
+			delete(active, handle)
+		}
+		s, _ := startsAt[p]
+		for _, handle := range s {
+			active[handle] = struct{}{}
+		}
+
+		if len(active) > 1 && i < len(points)-1 {
```
apazylbe: Could also do
for _, p := range points[:len(points)-1] above

gapis/api/vulkan/memory_layout.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/google/gapid/gapis/service"
+	"github.com/google/gapid/gapis/service/path"
+)
+
+// MemoryLayout stores an overview of the state's memory layout into a
+// service.MemoryLayout object.  The layout includes data on memory types,
+// allocations, and which resources are bound to which locations in memory.
+func (a API) MemoryLayout(st *api.GlobalState) (*service.MemoryLayout, error) {
+	s := GetState(st)
+	allocations := []*service.MemoryAllocation{}
```
apazylbe: Consider pre-allocating with an appropriate length if it is known

gapis/service/service.proto
```diff
@@ -671,6 +672,76 @@ message MemoryRange {
...
+  // Flags describing properties of the allocation, e.g. whether it's device
+  // local, whether it's host visible, etc.
+  uint32 flags = 3;
+  // Constant index for looking up the flag values with the path.ConstantSet endpoint
+  uint32 constant_set = 4;
+  // The API specific ID for this allocation
+  uint64 id = 5;
```
Qining: Is this Id used? And should we name it 'handle'?

gapis/service/service.proto
Qining: What about 'MemoryBreakdown'?

gapis/service/service.proto
iburinoc: Yes, it is memory type index in Vulkan, I didn't want to make it overspecific to Vulkan though.

gapis/service/service.proto
iburinoc: @ben-clayton suggested separated id and name for bindings, wasn't sure if it would also be useful for allocations.  Open to using either id or handle, not sure which is a more descriptive name.

gapis/resolve/metrics.go
AWoloszyn: It most likely belongs in `gapis/api`. This is similar to the Mesh API: `gapis/api/mesh.go`.

gapis/service/service.proto
AWoloszyn: GL works a bit differently, since all memory is implicit. The allocations will be 1:1 with the resources.
I would expect something like "memory_type" to be "GL_IMAGE or GL_BUFFER" most likely.

gapis/service/service.proto
AWoloszyn: I have a small preference for Handle. It more closely maps to how I view things. (That is not to say my view is right).

gapis/service/service.proto
```diff
@@ -671,6 +672,76 @@ message MemoryRange {
...
+}
+
+// A mapping from part of a memory allocation into host memory
+message MemoryMapping {
+  // The size in bytes of the mapping
+  uint64 size = 1;
+  // The offset of the mapping into the source allocation, in bytes
+  uint64 offset = 2;
+  // The address on the host that this memory is bound to
+  uint64 host_address = 3;
```
AWoloszyn: I would prefer `mapped_address` rather than `host_address` since the host address doesn't have to necessarily be fixed for the entire applications lifetime (The driver is free to return different addresses for different mappings).

gapis/service/service.proto
```diff
@@ -671,6 +672,76 @@ message MemoryRange {
...
+}
+
+// A buffer/image that is bound to part of a memory allocation
+message MemoryBinding {
```
AWoloszyn: Before we finalize this API we should figure out how this should look for sparse binding, since a single image can be bound to multiple locations.

AWoloszyn: This might mean another binding type of 'SparseImageBlock', which will define 
`x, y, w, h, mip_level, array_layer`
Then we would also need `SparseImageMetadata` and `SparseImageMipTail`
Lastly we would need `SparseBufferBlock`

gapis/resolve/metrics.go
iburinoc: Ok, and then the Metrics proto definition should be moved to `gapis/api/service.proto`?

gapis/service/service.proto
iburinoc: Would you prefer to also keep `name` or is just handle good enough?

gapis/service/service.proto
AWoloszyn: That is a fair point, overloading terms may not serve us well.
MemoryBreakdown is fine with me. Even a MemoryUsage (although people may expect a single number out of that).

gapis/service/service.proto
AWoloszyn: I think keeping the name is fine. It means we can use debug names if assigned.

gapis/service/service.proto
iburinoc: We also need some way to describe opaque sparse image binding blocks.  Could probably do that by setting `x, y, w, h = (~0, ~0, 0, 0)`.

iburinoc: SparseImageBlock also needs an aspect flags field

iburinoc:
Added proto definition for sparse types, implementation isn't done yet

gapis/service/service.proto
iburinoc: Hmm, I think it's actually better to have a separate type for opaque image block bindings

gapis/api/service.proto
```diff
@@ -238,3 +238,117 @@ message CubemapLevel {
...
+  // The height of the bound block
+  uint32 height = 4;
+
+  // The mip level of the bound block
+  uint32 mip_level = 5;
+  // The array layer the bound block is in
+  uint32 array_layer = 6;
+
+  // The aspect(s) of the image this block is from (colour, depth, etc.)
+  uint32 aspects = 7;
```
AWoloszyn: Can we use something other than a uint32_t here. Preferably a repeated field so we can have multiple aspects that have a proper name.

gapis/api/service.proto
```diff
@@ -238,3 +238,117 @@ message CubemapLevel {
...
+  string name = 2;
+
+  // The size of the binding, in bytes
+  uint64 size = 3;
+  // The offset into the bound allocation, in bytes
+  uint64 offset = 4;
+
+  // Which type of object is using this memory, and any extra associated data
+  // with that.
+  oneof type {
```
AWoloszyn: I am not sure if we can do this in Proto, but for consistency (and forward compatibility) can we make a BufferBinding and ImageBinding message, even if they are empty.

gapis/api/service.proto
iburinoc: Something like `repeated string` or do we still want to keep the values?  `repeated Aspect; message Aspect { uint32 value; string name; }`

AWoloszyn: I think? this will work (we might be using an old version of proto but

enum AspectType {
    COLOR,
    DEPTH,
    STENCIL
}

repeated AspectType

AWoloszyn: COLOR, DEPTH, STENCIL are pretty standard across all graphics APIs so I dont mind having them in the proto.

iburinoc: Alright, metadata is technically also a vulkan aspect but I suppose it shouldn't show up except in `SparseImageMetadataMipTail`, where we've already distinguished it from regular mip tail.

AWoloszyn: Sounds reasonable.

AWoloszyn:
Thanks!

iburinoc:
Example outputs of `gapit memory`:

```
$ gapit memory -at 382 alias.gfxtrace
4 memory allocations
Name: 37942576
        Device:       38013824
        Memory Type:  9
        Size:         256
        Flags:
                VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT
                VK_MEMORY_PROPERTY_HOST_COHERENT_BIT
        1 bindings:
        Buffer: 42453424
                Offset: 0
                Size:   192
        No aliased regions
Name: 42427504
        Device:       38013824
        Memory Type:  9
        Size:         2048
        Flags:
                VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT
                VK_MEMORY_PROPERTY_HOST_COHERENT_BIT
        3 bindings:
        Buffer: 42600560
                Offset: 0
                Size:   1024
        Buffer: 42599920
                Offset: 256
                Size:   512
        Buffer: 42600208
                Offset: 512
                Size:   512
        3 aliased regions:
        0:
                Offset:  256
                Size:    256
                Shared by:
                        42599920
                        42600560
        1:
                Offset:  512
                Size:    256
                Shared by:
                        42599920
                        42600208
                        42600560
        2:
                Offset:  768
                Size:    256
                Shared by:
                        42600208
                        42600560
Name: 42454400
        Device:       38013824
        Memory Type:  7
        Size:         256
        Flags:
                VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT
        1 bindings:
        Buffer: 42416976
                Offset: 0
                Size:   80
        No aliased regions
Name: 42565280
        Device:       38013824
        Memory Type:  7
        Size:         256
        Flags:
                VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT
        1 bindings:
        Buffer: 37943008
                Offset: 0
                Size:   12
        No aliased regions
```

```
$ gapit memory -at 500 sparse.gfxtrace
10 memory allocations
Name: 94322983582016
	Device:       94322983530432
	Memory Type:  9
	Size:         256
	Flags:
		VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT
		VK_MEMORY_PROPERTY_HOST_COHERENT_BIT
	1 bindings:
	Buffer: 94322983621376
		Offset: 0
		Size:   148
	No aliased regions
Name: 94322983608800
	Device:       94322983530432
	Memory Type:  9
	Size:         256
	Flags:
		VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT
		VK_MEMORY_PROPERTY_HOST_COHERENT_BIT
	Mapped into host memory at 0x7f5c27549c00
		Offset: 0
		Size:   256
	1 bindings:
	Buffer: 94322983608384
		Offset: 0
		Size:   66
	No aliased regions
Name: 94322983634352
	Device:       94322983530432
	Memory Type:  9
	Size:         512
	Flags:
		VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT
		VK_MEMORY_PROPERTY_HOST_COHERENT_BIT
	Mapped into host memory at 0x7f5c27549e00
		Offset: 0
		Size:   512
	1 bindings:
	Buffer: 94322988450096
		Offset: 0
		Size:   280
	No aliased regions
Name: 94322988016288
	Device:       94322983530432
	Memory Type:  7
	Size:         8102784
	Flags:
		VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT
	1 bindings:
	Image: 94322988011232
		Offset: 0
		Size:   8102784
	No aliased regions
Name: 94322988084640
	Device:       94322983530432
	Memory Type:  7
	Size:         131072
	Flags:
		VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT
	1 bindings:
	Image: 94322988090272
		Offset: 0
		Size:   131072
	No aliased regions
Name: 94322988408688
	Device:       94322983530432
	Memory Type:  7
	Size:         351744
	Flags:
		VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT
	1 bindings:
	Image: 94322983617152
		Offset: 0
		Size:   351744
	No aliased regions
Name: 94322988424912
	Device:       94322983530432
	Memory Type:  7
	Size:         65536
	Flags:
		VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT
	1 bindings:
	Sparse Image Mip Tail: 94322983582384
		Offset: 0
		Size:   65536
		Array Layer:     0
		Mip Tail Offset: 0
		Aspects:      Color
	No aliased regions
Name: 94322988429376
	Device:       94322983530432
	Memory Type:  7
	Size:         65536
	Flags:
		VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT
	1 bindings:
	Sparse Image Block: 94322983582384
		Offset: 0
		Size:   65536
		Block Offset: (0, 0)
		Block Extent: (128, 128)
		Mip Level:    6
		Array Layer:  0
		Aspects:      Color
	No aliased regions
Name: 94322988432800
	Device:       94322983530432
	Memory Type:  7
	Size:         2097152
	Flags:
		VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT
	1 bindings:
	Buffer: 94322988432448
		Offset: 0
		Size:   2097152
	No aliased regions
Name: 94322988436272
	Device:       94322983530432
	Memory Type:  7
	Size:         387328
	Flags:
		VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT
	1 bindings:
	Buffer: 94322988435920
		Offset: 0
		Size:   387096
	No aliased regions
```

gapis/api/service.proto
```diff
@@ -238,3 +238,127 @@ message CubemapLevel {
 	image.Info negative_z = 5;
 	image.Info positive_z = 6;
 }
+
+message Metrics {
+  oneof metrics {
```
ben-clayton: Instead of having this as a `oneof`, I'd suggest having `memory_breakdown` as a regular field.
Then instead of having an enum for the single metrics you want, the interface can be a list of bool flags for the bits of info the client wants. That way the client can fetch everything it cares about in a single RPC.

gapis/api/service.proto
```diff
@@ -238,3 +238,127 @@ message CubemapLevel {
...
+
+// A single memory allocation
+message MemoryAllocation {
+  // The device this allocation was made on
+  uint64 device = 1;
+  // The memory type this allocation is from
+  uint32 memory_type = 2;
+  // Flags describing properties of the allocation, e.g. whether it's device
+  // local, whether it's host visible, etc.
+  uint32 flags = 3;
```
ben-clayton: Part of the goal with this RPC interface is that the client should never have to have intimate knowledge of the actual graphics API. This breaks the abstraction.
I can see two ways of solving this:
1. Have a repeated string field of API flags as names.
2. Making another message with fields like `IsHostVisible`, etc. However this makes it hard to reproduce the true vulkan names.

cmd/gapit/memory.go
```diff
@@ -0,0 +1,293 @@
...
+
+			fmt.Println("\t\tOffset:", binding.Offset)
+			fmt.Println("\t\tSize:  ", binding.Size)
+
+			stringifyAspects := func(aspects []api.AspectType) string {
+				if len(aspects) == 0 {
+					return ""
+				}
+				strs := make([]string, len(aspects))
+				for i, asp := range aspects {
```
ben-clayton: You can add a `Format()` method on the type so that this automatically handled by the `fmt` package.
There are many examples of this [here](https://github.com/google/gapid/blob/master/gapis/service/path/path.go#L156).

cmd/gapit/memory.go
```diff
@@ -0,0 +1,293 @@
...
+			}
+			switch val := binding.Type.(type) {
+			case *api.MemoryBinding_SparseImageBlock:
+				info := val.SparseImageBlock
+				fmt.Printf("\t\tBlock Offset: (%v, %v)\n",
+					info.XOffset, info.YOffset)
+				fmt.Printf("\t\tBlock Extent: (%v, %v)\n",
+					info.Width, info.Height)
+				fmt.Printf("\t\tMip Level:    %v\n", info.MipLevel)
+				fmt.Printf("\t\tArray Layer:  %v\n", info.ArrayLayer)
```
ben-clayton: Nit: There is the [`text/tabwriter`](https://golang.org/pkg/text/tabwriter/) package that does automatic alignment of text into columns. It also deals with emitting whitespace instead of tab characters (which many terminals will struggle with).

gapis/api/service.proto
iburinoc: It was originally a repeated string field of API flag names, but @AWoloszyn suggested using `uint32` along with the `allocation_flags_index` field to fetch the flag names from `ConstantSets`

ben-clayton: Ah okay, please can you document that the flags relate to bitfields in the constant sets then?
This was not obvious.

iburinoc: Good point, sure.

gapis/api/service.proto
```diff
@@ -238,3 +238,127 @@ message CubemapLevel {
...
+  // Constant index for looking up the flag values with the path.ConstantSet endpoint
```
ben-clayton: Of course, I now see it is mentioned here. I of course didn't look here. :)

iburinoc: Although now that you point this out I think this field is duplicating the work of `allocation_flags_index` in `MemoryBreakdown`.  That should be fixed.

PR #1959, opened 2018-06-05
Allow state cloning to handle references properly.

AWoloszyn:
State cloning was broken with the getter/setter MR.
Track references, and make sure that we are keeping multiple
references to the same object together rather than uniquing them.


AWoloszyn:
Test added

AWoloszyn:
Closing in favor of #1963

PR #1960, opened 2018-06-05
Fix for linear image with preinitialized intial layout

Qining:
1) If an image has linear tiling and has initial layout set to
PREINITIALIZED, all its subresources data pool will be linked back to
the bound device memory, no new pool will be created for those
subresources.

2) If an image has linear tiling and has initial layout set to
PREINITIALIZED, and it can not be used as TRANSFER_DST, render target or
imageStore target, its data will be primed by maping memory and flush
map memory ranges.

### DIFF
--- a/gapii/cc/vulkan_extras.cpp
+++ b/gapii/cc/vulkan_extras.cpp
@@ -504,6 +504,39 @@ VkMemoryRequirements VulkanSpy::fetchBufferMemoryRequirements(
   return reqs;
 }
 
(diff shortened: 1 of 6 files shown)


Qining:
Most of the changes in image_primer.go is just to return 'better' errors. The only real change in that file is `func primeByPreinitialization()`.

Qining:
Windows and Mac builds are tested locally, all passed.

Issue #1934, opened 2018-05-30
GAPIR should be restarted by GAPIS in case of crash


Qining:
Since the gRPC cl, this feature is lost, we should add it back.

Qining:
This is actually a ping interval issue. As go-grpc does not have a 'signal' for connection failure, we may have to rely on ping requests to check if the connection is alive. With a shorter ping interval, such a failure can be found sooner, then the next request can initiate a new session and connection earlier.

Qining:
A newer version of go-grpc may have 'GetState()' interface for client
connection, with which we can set a go routine to pull and set signal
for connection failure. This may not use communication bandwidth
so we don't need to worry about performance. But the general idea
is similar.

PR #1962, opened 2018-06-06
Response for failed connection quicker so we can restart gapir earlier

Qining:
Mitigate #1934 

A replay session will closed if heartbeat PING failed. If the interval
is too long, the user need to wait for a longer time before the next
request can be issued and restart a new replay session and connection.
With a shorter PING interval, we find such connectionf failure sooner so
we can start to process next comming request sooner.

This is probably not a perfect fix. But the go-grpc currently does not
have a 'signal' for connection failure.

A newer version of go-grpc may have 'GetState()' interface for client
connection, with which we can set a go routine to pull and set signal
for connection failure. This may not occupy communication bandwidth. But
the general idea is similar.

### DIFF
--- a/gapir/client/session.go
+++ b/gapir/client/session.go
@@ -45,6 +45,7 @@ const (
 	maxCheckSocketFileAttempts = 10
 	checkSocketFileRetryDelay  = time.Second
 	connectTimeout             = time.Second * 10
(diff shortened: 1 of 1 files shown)


Qining:
Tracked in #1934

PR #1963, opened 2018-06-06
A few fixes mostly related to references in the API language

pmuetschard:
- Don't lose the arena when creating/cloning.
 - Fix the cloning of API references and maps such that the share-tree is maintained in the cloned copy.

### DIFF
--- a/gapis/api/api.go
+++ b/gapis/api/api.go
@@ -106,3 +106,6 @@ func Register(api API) {
 func Find(id ID) API {
 	return apis[id]
 }
+
+// CloneContext is used to keep track of references when cloning API objects.
+type CloneContext map[interface{}]interface{}
--- a/gapis/api/templates/api.go.tmpl
+++ b/gapis/api/templates/api.go.tmpl
@@ -224,7 +224,7 @@ import (
   {{end}}
 
   // Clone makes a deep copy of this static array, using ϟa for allocations.
-  func (m {{$name}}) Clone(ϟa arena.Arena) {{$name}} {
+  func (m {{$name}}) Clone(ϟa arena.Arena, ϟseen ϟapi.CloneContext) {{$name}} {
     out := {{$name}}{ data: &[{{$ty.Size}}]{{$el}}{} }
     for i := 0; i < {{$ty.Size}}; i++ {
       out.Set(i, {{Template "Clone" "Type" $ty.ValueType "Src" "m.Get(i)"}})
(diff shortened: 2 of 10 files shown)


gapis/api/templates/api.go.tmpl
```diff
@@ -341,12 +342,29 @@ import (
   }
 
   // Clone makes a deep copy of this map.
-  func (m {{$name}}) Clone(ϟa arena.Arena) {{$name}} {
+  func (m {{$name}}) Clone(ϟa arena.Arena, ϟseen ϟapi.CloneContext) {{$name}} {
+    if done, ok := ϟseen[m.Map]; ok {
```
AWoloszyn: Done is probably not the best name here.
`existing` maybe?

gapis/api/templates/api.go.tmpl
```diff
@@ -249,10 +249,11 @@ import (
   type {{$name}} struct {
     Map *map[{{$key}}]{{$value}}
     a   arena.Arena
+    uncompareable
```
AWoloszyn: Are we sure we never want to compare map references?

ben-clayton: It's a good question - I believe it to be completely valid to use the equality operator on a map in the API language.
What exactly is this fixing? Do you have a situation where two maps are in two different arenas or something?

gapis/api/gles/custom_replay.go
```diff
@@ -39,39 +39,39 @@ type objectKey struct {
 func (i BufferId) remap(cmd api.Cmd, s *api.GlobalState) (key interface{}, remap bool) {
```
ben-clayton: > The reference tracking on maps should be on the actual map.

Should it? The only other field in the map type was the arena, and identical maps should be from the same arena.
Not sure what this is fixing. FWIW, the map will become a cgo type and so private, where this change will have to be reverted.
More context on what you are attempting to fix here would be helpful.

gapis/api/api.go
```diff
@@ -106,3 +106,11 @@ func Register(api API) {
 func Find(id ID) API {
 	return apis[id]
 }
+
+// CloneContext is used to keep track of references when cloning API objects.
+type CloneContext map[interface{}]interface{}
+
+// NewCloneContext creates a new empty CloneContext.
+func NewCloneContext() CloneContext {
```
ben-clayton: Constructors for trivial default-initialized types are frowned upon in Go.
It would be more common to build these with: `CloneContext{}`

gapis/api/templates/api.go.tmpl
ben-clayton: > Make maps uncomparable, they can't be map keys

So this is the reason, right? How come they can't be map keys?

ben-clayton:
Merged the final CL with https://github.com/google/gapid/pull/1968.

gapis/api/templates/api.go.tmpl
pmuetschard: Done.

pmuetschard:
I've dropped the controversial change to the maps. They are (pointer-)comparable again.

PR #1966, opened 2018-06-07
Fix crashes with Vulkan sparse bindings

Qining:
Three issues fixed:

1) The opaque sparse memory binding info cached in Buffer and Image
object should not be fed to MustUnpackReadMap(), as they are actually
mapping from offsets to info, not a list of info.

2) IsFullyBound() does not work correctly.

3) Since the creation of ImageLevel.Data() slice has been moved from
vkCreateImage to vkBindImageMemory, for sparse binding images, the Data
slice is empty. For now we just check the size of the Data slice for
each time a sparse binding is to be submitted and if the Data slice is
empty, create the shadow memory Data slice for each image level.

### DIFF
--- a/gapis/api/vulkan/api/queued_command_tracking.api
+++ b/gapis/api/vulkan/api/queued_command_tracking.api
@@ -108,12 +108,75 @@ sub void execPendingCommands(VkQueue queue) {
                   bind.flags)
               }
             }
+            // Creates the shadow image data memory here if has not been done so.
+            // This is exactly the same to vkBindImageMemory except for linear
+            // and preinitialized images, we also create individual pool here.
+            // TODO: Redirect to bound memories for the data of linear preinitialized
+            // images.
+            for _, _, aspectBit in unpackImageAspectFlags(image.ImageAspect).Bits {
+              aspect := image.Aspects[aspectBit]
+              for lay in (0 .. image.Info.ArrayLayers) {
+                for lev in (0 .. image.Info.MipLevels) {
+                  level := aspect.Layers[lay].Levels[lev]
+                  if len(level.Data) == 0 {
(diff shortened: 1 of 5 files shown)


Qining:
@iburinoc Please take a look, hope this fixes the issue with sparse bindings for now.

iburinoc:
I checked this PR fixes [https://github.com/google/vulkan_test_applications/tree/master/application_sandbox/sparse_binding](https://github.com/google/vulkan_test_applications/tree/master/application_sandbox/sparse_binding) for me, lgtm

gapis/api/vulkan/api/queued_command_tracking.api
```diff
@@ -108,13 +108,76 @@ sub void execPendingCommands(VkQueue queue) {
...
+                      case VK_IMAGE_ASPECT_COLOR_BIT:
+                        elementAndTexelBlockSize.ElementSize
+                      case VK_IMAGE_ASPECT_DEPTH_BIT:
+                        depthElementSize
+                      case VK_IMAGE_ASPECT_STENCIL_BIT:
+                        // stencil element is always 1 byte wide
+                        as!u32(1)
+                    }
+                    size := widthInBlocks * heightInBlocks * level.Depth * elementSize
+                    level.Data = make!u8(size)
```
AWoloszyn: We will have to see what this looks like if/when we actually switch our pools over to actual byte slices. This will likely not be what we want.

Qining: I think we need to 2 steps
1) Get the texture data from replay side, so that nothing rely on the image level data pool any more
2) Move to use the device memory data pool.

Or as you said, we can move to the pool first, for the cases that the calculated size is smaller than the required size.

PR #1972, opened 2018-06-12
Vulkan: Fix support of sparse buffer

Qining:
1) Support sparse buffer in vkCmdCopyBuffer, vkCmdFillBuffer and
vkCmdCopyBufferToImage.

2) Support extract geometry data from sparse buffer

3) 'Read' the memory of sparse buffers

4) Implement doVkCmdFillBuffer

Tested with all the samples we have.

### DIFF
--- a/gapis/api/vulkan/api/coherent_memory.api
+++ b/gapis/api/vulkan/api/coherent_memory.api
@@ -60,12 +60,7 @@ sub void readMemoryInBuffer(ref!BufferObject buffer, VkDeviceSize readOffset, Vk
   memPieces := getBufferBoundMemoryPiecesInRange(buffer, readOffset, readSize)
   for _, _, mp in memPieces {
     readCoherentMemory(mp.DeviceMemory, mp.MemoryOffset, mp.Size)
-    // On the trace side, 'Data' pool won't actually be created, so its length
-    // will be zero. But on the server side, the pool will be created and the
-    // state block should contains the data in the buffer memory.
-    if len(mp.DeviceMemory.Data) != 0 {
-      read(mp.DeviceMemory.Data[mp.MemoryOffset:mp.MemoryOffset+mp.Size])
-    }
+    read(mp.DeviceMemory.Data[mp.MemoryOffset:mp.MemoryOffset+mp.Size])
   }
 }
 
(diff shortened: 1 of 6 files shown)


gapis/api/vulkan/api/coherent_memory.api
```diff
@@ -57,19 +57,14 @@ sub void readCoherentMemory(ref!DeviceMemoryObject memory, VkDeviceSize readOffs
...
-        readSize
-    }
-    readEnd := readBegin + actualReadSize
-    readCoherentMemory(mem, readBegin, actualReadSize)
-    if len(mem.Data) != 0 {
-      read(mem.Data[readBegin : readEnd])
+  memPieces := getBufferBoundMemoryPiecesInRange(buffer, readOffset, readSize)
+  for _, _, mp in memPieces {
+    readCoherentMemory(mp.DeviceMemory, mp.MemoryOffset, mp.Size)
+    // On the trace side, 'Data' pool won't actually be created, so its length
```
AWoloszyn: Instead of this way, since data is marked as "spy_disable" can we just not generate the read intrinsic for this case?

We already handle other cases like this, so I think cutting out the read intrinsic for spy_disabled is probably a cleaner approach than a check like this.

gapis/api/vulkan/api/copy_clear_commands.api
```diff
@@ -618,12 +772,12 @@ cmd void vkCmdClearAttachments(
 
 sub void dovkCmdResolveImage(ref!vkCmdResolveImageArgs args) {
   for _ , _ , r in args.ResolveRegions {
-    srcRange := VkImageSubresourceRange(r.srcSubresource.aspectMask,
-                                        r.srcSubresource.mipLevel, 1,
-                                        r.srcSubresource.baseArrayLayer, r.srcSubresource.layerCount)
-    dstRange := VkImageSubresourceRange(r.dstSubresource.aspectMask,
-                                        r.dstSubresource.mipLevel, 1,
-                                        r.dstSubresource.baseArrayLayer, r.dstSubresource.layerCount)
+    srcRange := VkImageSubresourceRange(               r.srcSubresource.aspectMask,
```
AWoloszyn: Something weird with the spacing here

Qining: wow, looks like the format tool goes crazy.

gapis/api/vulkan/api/coherent_memory.api
Qining: Fixed in 2a680ed

gapis/api/vulkan/api/copy_clear_commands.api
Qining: Fixed in 2a680ed

Qining:
Build passes on all three platforms, tested manually.

PR #1982, opened 2018-06-13
Fix all warnings. Treat new ones as errors

ben-clayton:


core/cc/id.h
```diff
@@ -51,7 +51,12 @@ namespace std {
 
 template <> struct hash<core::Id> {
     inline size_t operator()(const core::Id& id) const {
-        return *reinterpret_cast<const size_t*>(id.data);
+        union {
```
AWoloszyn: IIRC this breaks strict aliasing.
The only way to do this kind of cast "legally" is
```
static_assert(sizeof(core::Id*) == sizeof(size_t), "Incompatible types"); // Sanity check
size_t hash;
memcpy(&hash, &id, sizeof(size_t));
return hash;

All of the compilers SHOULD optimize the memcpy away, and this becomes a no-op anyway.

File: gapil/runtime/cc/string_test.cpp [hunk 2]
```diff
@@ -44,7 +44,7 @@ TEST(StringTest, constructors) {
         EXPECT_STREQ(strB.c_str(), abc123);
     }
     {
-        gapil::String str(std::move(gapil::String(&arena, abc123)));
+        gapil::String str(gapil::String(&arena, abc123));
```
AWoloszyn: Can this not be
```
gapil::String str(&arena, abc123);
```

core/cc/id.h
ben-clayton: Done. My memory is set on telling me that the `union` is the correct solution, regardless on how many times you correct me. :)

gapil/runtime/cc/string_test.cpp
AWoloszyn: Sorry my bad!

core/cc/id.h
AWoloszyn: Looks good!

ben-clayton:
Thanks. I'll be appending minor fixes until the build machines are green.

ben-clayton:
The windows build is having unrelated issues which I'm already looking into.
Submitting.

PR #1984, opened 2018-06-14
Add transformer that logs replay to capture file

sean-purcell:
Extra transformer and corresponding config option to help with debugging replays

### DIFF
--- a/gapis/api/transform/capture_log.go
+++ b/gapis/api/transform/capture_log.go
@@ -1,0 +1,76 @@
+// Copyright (C) 2018 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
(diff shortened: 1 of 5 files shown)


gapis/api/transform/capture_log.go
```diff
@@ -0,0 +1,77 @@
...
+		log.E(ctx, "Failed to create replay storage capture: %v", err)
+		return
+	}
+	c, err := capture.ResolveFromPath(ctx, capt)
+	if err != nil {
+		log.E(ctx, "Failed to resolve capture from path %v: %v", capt, err)
+		return
+	}
+	err = c.Export(ctx, t.file)
+	if err != nil {
```
ben-clayton: nit: candidate for putting the `err` [in the initializer of the if](https://golang.org/doc/effective_go.html?#if):

```if err := c.Export(ctx, t.file); err != nil {```

PR #1986, opened 2018-06-14
Get the string table to work with "bazel run ..."

pmuetschard:
Fixes the string table targets so they show up in runfiles and fixes the handling of the string table in the runfiles for gapic.


pmuetschard:
ping

PR #1989, opened 2018-06-15
Small fixes to memory breakdown

iburinoc:
Recommended changes by @ben-clayton in #1954
- Change Metrics proto to not be oneof, and change path.Metrics to have
  a flag per type instead of an enum
- Remove redudancy on flags ConstantSet index, and document better
- Add aspectList type with Format method
- Use text/tabwriter in cmd/gapit/memory

### DIFF
--- a/cmd/gapit/memory.go
+++ b/cmd/gapit/memory.go
@@ -108,9 +108,9 @@ func (verb *memoryVerb) Run(ctx context.Context, flags flag.FlagSet) error {
 
 	for _, alloc := range mem.Allocations {
 		fmt.Fprintln(w, "Name:", alloc.Name)
-		fmt.Fprintln(w, "\tDevice:      ", alloc.Device)
-		fmt.Fprintln(w, "\tMemory Type: ", alloc.MemoryType)
-		fmt.Fprintln(w, "\tSize:        ", alloc.Size)
+		fmt.Fprintf(w, "\tDevice: \t%v\n", alloc.Device)
+		fmt.Fprintf(w, "\tMemory Type: \t%v\n", alloc.MemoryType)
+		fmt.Fprintf(w, "\tSize: \t%v\n", alloc.Size)
 
 		if alloc.Flags != 0 && len(allocationFlags) != 0 {
 			fmt.Fprintln(w, "\tFlags:")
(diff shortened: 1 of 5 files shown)


cmd/gapit/memory.go
```diff
@@ -95,35 +100,37 @@ func (verb *memoryVerb) Run(ctx context.Context, flags flag.FlagSet) error {
...
 		return mem.Allocations[i].Handle < mem.Allocations[j].Handle
 	})
+
 	for _, alloc := range mem.Allocations {
-		fmt.Println("Name:", alloc.Name)
-		fmt.Println("\tDevice:      ", alloc.Device)
-		fmt.Println("\tMemory Type: ", alloc.MemoryType)
-		fmt.Println("\tSize:        ", alloc.Size)
+		fmt.Fprintln(w, "Name:", alloc.Name)
+		fmt.Fprintln(w, "\tDevice:      ", alloc.Device)
```
ben-clayton: The main reason I suggested the tab writer was that it allows you to align the fields here without having to have a load of whitespace indentations.

cmd/gapit/memory.go
```diff
@@ -142,81 +149,81 @@ func (verb *memoryVerb) Run(ctx context.Context, flags flag.FlagSet) error {
...
+	strs := make([]string, len(a))
+	for i, asp := range a {
+		var typ string
+		switch asp {
+		case api.AspectType_COLOR:
+			typ = "Color"
+		case api.AspectType_DEPTH:
+			typ = "Depth"
+		case api.AspectType_STENCIL:
+			typ = "Stencil"
```
ben-clayton: Please add an unknown case that includes the integer value

cmd/gapit/memory.go
```diff
@@ -142,81 +149,81 @@ func (verb *memoryVerb) Run(ctx context.Context, flags flag.FlagSet) error {
...
+type aspectList []api.AspectType
+
+func (a aspectList) Format(f fmt.State, c rune) {
+	if len(a) == 0 {
+		return
+	}
+	strs := make([]string, len(a))
+	for i, asp := range a {
+		var typ string
+		switch asp {
```
ben-clayton: This can be a format method on `api.AspectType`. I suspect then you don't need the `aspectList` type at all.

cmd/gapit/memory.go
iburinoc: `api.AspectType` is an enum, so I'm not sure how to include the integer value there.  I can add an unknown type to the enum though.

cmd/gapit/memory.go
iburinoc: Ok, and then just print out the slice?

cmd/gapit/memory.go
ben-clayton: Something like:

```go
func (a AspectType) Format(f fmt.State, c rune) {
	switch a {
	case api.AspectType_COLOR:
		fmt.Fprint(f, "Color")
	case api.AspectType_DEPTH:
		fmt.Fprint(f, "Depth")
	case api.AspectType_STENCIL:
		fmt.Fprint(f, "Stencil")
	default:
		fmt.Fprintf(f, "Unknown AspectType %d", (int)a)
	}
}
```

Of course protobuf already produces a `String()` method for you, but I'm guessing it's not what you want?

cmd/gapit/memory.go
ben-clayton: Yeah, you'd have to check it formats it like you want though

iburinoc: Right, yeah I think I'd want to drop the `[]` since the usual case is just one aspect

cmd/gapit/memory.go
iburinoc: Oh, I didn't realize protobuf enums could take values that weren't specified in the proto description, but I suppose it makes sense it would for version compatibility.

PR #1990, opened 2018-06-15
Lazily resolve Vulkan commands (actually all indirect commands)

Qining:
This is to avoid some Vulkan driver (actually some ICDs) crashing when resolving unsupported and never used commands.

This is probably the 'minimum changes' approach. It uses lambda
functions, which will take a copy of the handles used for command resolving.

If we really want to avoid that copy, we probably need to use the template
system to scan through all the commands first and  generate c++ template
specialization code. But I think the benefit may be not worth of the extra work?

### DIFF
--- a/gapir/cc/gfx_api.h
+++ b/gapir/cc/gfx_api.h
@@ -45,7 +45,7 @@ template <typename FuncPtr>
 class LazyResolved {
  public:
   LazyResolved() : resolve_(nullptr), ptr_(nullptr) {}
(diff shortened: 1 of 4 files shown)


Qining:
Reason for this CL just added in the description. Generally, we found, for some cases, the ICD may crash when resolving a command from unsupported Vulkan Extensions. Even though the application may never use the unsupported command, such a crash will also crash our replay.

PR #1995, opened 2018-06-18
Work around a protobuf/libc incompatability.

pmuetschard:
Fields named "major" and "minor" will cause problems in the future. See https://github.com/google/protobuf/issues/4654.

Also updated all proto files to follow the same naming style for fields (using all lowercase and `_` separated names, except for abbreviations for go's sake).
Finally, ran `clang-format` over all proto files.


core/os/device/device.proto
```diff
@@ -155,84 +155,84 @@ message Configuration {
...
+    Drivers drivers = 4;
 }
 
 // Instance represents a physical device.
 // An instance is persistable, and can be used to retain information about
 // offline devices, and reconnect to them.
 message Instance {
     // The unique identifier of the instance.
-    ID id = 1;
+    ID ID = 1;
```
ben-clayton: A little bit odd that `ID` breaks the lower case rules.

core/os/device/device.proto
```diff
@@ -155,84 +155,84 @@ message Configuration {
...
     // The hardware and software configuration of the device.
-    Configuration Configuration = 4;
+    Configuration configuration = 4;
 }
 
 // Drivers describes the drivers available on a device.
 message Drivers {
     // The OpenGL or OpenGL ES driver support.
-    OpenGLDriver OpenGL = 1;
+    OpenGLDriver open_gl = 1;
```
ben-clayton: nit: I'd probably call this `opengl` but that's just me.

pmuetschard: done.

PR #1998, opened 2018-06-18
Vulkan: Fix crash at VkDestroyDebugReportCallbackEXT

Qining:
If the application uses debug report callback, we should not use our
synthetic commands to process their callback handles.


Qining:
PTAL: @iburinoc Hopfully this should fix the hang issue.

iburinoc:
Confirmed that this fixes the hang, thanks!

PR #2008, opened 2018-06-20
Presubmit Checks

pmuetschard:
`kokoro/presubmit/build.sh` to be used by Kokoro to run the checks.
`kokoro/presubmit/presubmit.sh` can be run in a clean workspace as presubmit checks. If a presubmit check fails, the workspace will contain the delta to make that specific check pass.

### DIFF
--- a/core/cc/connection_test.cpp
+++ b/core/cc/connection_test.cpp
@@ -21,13 +21,13 @@
 
 #include <string>
 
+using ::testing::_;
 using ::testing::DoAll;
 using ::testing::ElementsAre;
 using ::testing::Return;
 using ::testing::ReturnArg;
 using ::testing::StrictMock;
 using ::testing::WithArg;
-using ::testing::_;
 
 namespace core {
 namespace test {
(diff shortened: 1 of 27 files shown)


core/cc/connection_test.cpp
```diff
@@ -21,13 +21,13 @@
 
 #include <string>
 
+using ::testing::_;
```
ben-clayton: I'm concerned that your clang-format is doing something different to my clang-format here!
Sure this was with the Google style?

pmuetschard: I got clang-format via `apt-get install clang-format`. It shows the following version:
```
$ clang-format -version
clang-format version 6.0.0-1ubuntu2 (tags/RELEASE_600/final)
```
I've also built it at head and it makes the same changes (I compared against running it on clean master). What's the version of your clang-format?

> Sure this was with the Google style?

Yes, without `-style=Google` it would make a whole ton more changes...

PR #2012, opened 2018-06-21
Fix a bug in current DCE

### LINKED ISSUE #2010: VulkanTerminator doesn't handle subcommand indices changing between Add and Transform
When using dead code elimination in vulkan replays, the subcommand index for the last subcommand in the replay request can change if commands in the buffer were removed.  This means the terminator can end up including extra commands in the buffer.

Qining:
For vkCmdWaitForEvent, the 'read' to the event handle and signals should
be carried out no matter whether there are memory barriers.

Fix #2010


Qining:
@iburinoc PTAL, this should fix the problem.

iburinoc:
Fixes it for me :+1:

PR #2013, opened 2018-06-22
Add support to BC4 and BC5 format

Qining:
So that BC4 and BC5 format images can be converted to RGBA8_UNORM format.


core/image/s3.go
```diff
@@ -160,36 +160,35 @@ func decodeAlphaDXT3(r binary.Reader, dst []pixel) {
 
 func decodeAlphaDXT5(r binary.Reader, dst []pixel) {
 	a0, a1, codes := int(r.Uint8()), int(r.Uint8()), uint64(r.Uint16())|(uint64(r.Uint32())<<16)
+	for i := 0; i < 16; i++ {
+		dst[i].a = mixAlphaDXT5(a0, a1, codes)
```
ben-clayton: Ignorable: I unfortunately suspect this might not inline, but I do side with readability / reuse than raw performance in this case.

Issue #1197, opened 2017-10-14
Add mid-execution capture support for GLES


dsrbecky:

dsrbecky:
Duplicate of #1043

AWoloszyn:
This is now working, closing this in favor of opening specific bugs when we have them.

PR #2016, opened 2018-06-26
GLES Mid-Execution Capture

pmuetschard:
Still missing lots of pieces, but works on some stuff.

For #1197


gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -440,17 +440,6 @@ public AndroidInput(
         traceTarget.addBoxListener(SWT.Modify, targetListener);
         targetListener.handleEvent(null);
 
-        Listener apiListener = e -> {
```
AWoloszyn: Are we confident enough marking this as "works" for GLES, or do we want to mark it as experimental?

AWoloszyn: Because as soon as the button works, people will start sending bugs.

gapis/api/gles/state_builder.go
```diff
@@ -201,6 +201,10 @@ func (sb *stateBuilder) contextObject(ctx context.Context, handle EGLContext, c
 	if !c.Other().Initialized() {
 		return
 	}
+	if c.Other().Destroyed() {
```
ben-clayton: Not too sure what the point is in creating a context just to delete it again.
Seems a bit excessive for the correctness argument.

gapis/api/gles/externs.go
```diff
@@ -61,15 +61,15 @@ func (e externs) unmapMemory(slice memory.Slice) {
 }
 
 func (e externs) GetEGLStaticContextState(EGLDisplay, EGLContext) StaticContextStateʳ {
-	return FindStaticContextState(e.s.Arena, e.cmd.Extras()).Clone(e.s.Arena, api.CloneContext{})
+	return FindStaticContextState(e.s.Arena, e.cmd.Extras())
```
ben-clayton: Not sure about this change. "No need to clone what is already a copy." Where is this deep copy of which you speak?

gapis/api/gles/api/buffer_objects.api
```diff
@@ -267,9 +267,9 @@ cmd void glBufferData(GLenum target, GLsizeiptr size, BufferDataPointer data, GL
   }
   CheckSizeGE!GLsizeiptr(size, 0)
 
-  b.Data = switch (data != null) {
-    case true:  clone(as!u8*(data)[0:size])
-    case false: make!u8(size)
+  b.Data = make!u8(size)
+  if (data != null) {
+    copy(b.Data, as!u8*(data)[0:size])
```
ben-clayton: What was the reason for this change?

pmuetschard: `@spy_disabled` and `clone` do not get along well, causing the observation to be dropped. `copy` on the other hand works fine.

ben-clayton: I'm working on this stuff with the new compiler. Care to elaborate a little? Maybe I can get this fixed.

pmuetschard: `@spy_disabled` is not really fully implemented. It is partially handled in a few places In the templates. One of the problems is that when evaluating an assignment, if the LHS is `@spy_disabled`, the RHS is not evaluated. This is why `foo = clone(bar...)` drops the observation on `bar` if `foo` is `@spy_disabled` as the clone is not evaluated. Simply evaluating the RHS will lead to other problems, as it would have to be evaluated in a `@spy_disabled` way, for example not issuing an error (maybe) if reading from another `@spy_disabled` source - say from `bar` in the `clone` - it's OK to copy from `@spy_disabled` to `@spy_disabled`, but not to copy from `@spy_disabled` to not-`@spy_disabled` since the data would not be available at trace time.

It seemed overly complex to me to get assignment to behave correctly, and too hacky to add special handling to "assignment with clone RHS", so I did what Vulkan does and dropped the clone shortcut. For tracing it doesn't matter, since the `b = make!u8(size)` gets dropped and the `copy` is only used to create the observation. I didn't think the `clone` is all that much faster during mutate than `make` followed by `copy`.

gapic/src/main/com/google/gapid/views/TracerDialog.java
pmuetschard: Done.

pmuetschard: Done.

gapis/api/gles/state_builder.go
pmuetschard: Done.

gapis/api/gles/externs.go
pmuetschard: Done.

gapis/api/gles/state_builder.go
```diff
@@ -276,9 +276,14 @@ func (sb *stateBuilder) contextObject(ctx context.Context, handle EGLContext, c
 		}
 	}
 	if objs := c.Objects().Programs(); sb.once(objs) {
+		// Get the largest used shader ID.
+		maxID := uint32(0)
+		if l := c.Objects().Shaders().Len(); l > 0 {
+			maxID = uint32(c.Objects().Shaders().Keys()[l-1])
```
ben-clayton: Some assumptions here about sequential keys which I'm not sure is guaranteed. Would iterating over all shaders to take the maximum id field really too slow?

gapis/api/gles/api/buffer_objects.api
ben-clayton: ack

gapis/api/gles/state_builder.go
```diff
@@ -63,6 +65,11 @@ func (s *State) RebuildState(ctx context.Context, oldState *api.GlobalState) ([]
 	representative := map[ShareListʳ]EGLContext{}
 	for i := ContextID(0); i < s.NextContextID(); i++ {
 		for handle, c := range s.EGLContexts().All() {
+			// Don't recreate destroyed or unitialized contexts.
```
ben-clayton: uninitialized

gapis/api/gles/state_builder.go
pmuetschard: Done. I knew keys are sorted, but thought `.Keys()` was constant time...

PR #2017, opened 2018-06-27
Vulkan: Allocation optimization for byte slice during state rebuilding

Qining:
For a typical Vulkan application, this saves ~200MB memory used for
storing the slice data.

### DIFF
--- a/gapis/api/vulkan/state_rebuilder.go
+++ b/gapis/api/vulkan/state_rebuilder.go
@@ -37,7 +37,6 @@ type stateBuilder struct {
 	readMemories          []*api.AllocResult
 	writeMemories         []*api.AllocResult
 	extraReadIDsAndRanges []idAndRng
-	reservedMemories      []*api.AllocResult
 	memoryIntervals       interval.U64RangeList
 	ta                    arena.Arena // temporary arena
 }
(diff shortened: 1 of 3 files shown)


gapis/api/vulkan/state_rebuilder.go
```diff
@@ -828,24 +859,71 @@ func memoryTypeIndexFor(memTypeBits uint32, props VkPhysicalDeviceMemoryProperti
 	return -1
 }
 
-func (sb *stateBuilder) allocAndFillScratchBuffer(device DeviceObjectʳ, data []uint8, usages ...VkBufferUsageFlagBits) (VkBuffer, VkDeviceMemory) {
-	buffer := VkBuffer(newUnusedID(true, func(x uint64) bool { return sb.s.Buffers().Contains(VkBuffer(x)) }))
-	deviceMemory := VkDeviceMemory(newUnusedID(true, func(x uint64) bool { return sb.s.DeviceMemories().Contains(VkDeviceMemory(x)) }))
+type bufSubRngFillInfo struct {
```
AWoloszyn: For class names, can we use the full words,  bufferSubRangeFillInfo.

Qining:
@AWoloszyn 
1. struct renamed
2. add `mustReadSlice()` to `stateBuilder` to take care of slice data: reserve, then read the slice at the reserved address.

Do not handle the data used in "image priming through imageStore", as
1) If the data get truncated into multiple batches of vkCmdDispatch, new ID and record in database will always be required. This is also the case for sparse resident images.
2) If truncation is not required, the data used for `AllocDataOrPanic()` will be exactly the same to the data slice already in the database so is the hash, so no new record will be created.

PR #2020, opened 2018-06-27
Make associativity explicit for copy_clear_commands

iburinoc:
GAPIL seems to parse `a - b + c` as `a - (b + c)` which doesn't appear to be the expected behaviour in some of the API code.  This fixes it for copy_clear_commands as it was causing issues with vkCmdCopyBufferToImage.


ben-clayton:
Yeah, gapil doesn't have any precedence rules. We have an ancient internal bug for this, but we should move this to Github.

PR #2022, opened 2018-06-28
Vulkan: Fix Device Lost error due to using wrong queue for querypools

Qining:
When rebuilding state, using a wrong queue to recreate query pools and their states may result into device lost error and crash or hang the whole replay.


Qining:
Manually tests for the builds on all three platforms, and all passed. (And this CL should not make any difference for builds). Submitted.

PR #2023, opened 2018-06-28
Create transformer that dumps handle mappings to file

iburinoc:
It dumps the mapping from the handles in the original trace to the
handles used in the replay.  Useful for debugging replays with
validation layers as they print the new handles, not the original ones.

Writes to "mappings.txt" and looks like:

```
<...>
VkPipelineLayout(230575652): 140240292132736
VkQueue(140256319542576): 140240287537760
VkRenderPass(140256326180576): 140240310809728
VkRenderPass(1632168291): 140240292132224
VkRenderPass(2945999679): 140240291907504
VkSampler(140256326179936): 140240310823840
VkSemaphore(140256326176336): 140240283299536
<...>
```

Doesn't work for GLES at the moment because GLES uses extra data in
their remapping keys making it hard to determine what the right size to
read is.


gapis/config/config.go
```diff
@@ -22,6 +22,7 @@ const (
 	DebugDeadCodeElimination   = false
 	LogExtrasInTransforms      = false // Logs all commands' extras together with transforms
 	LogMemoryInExtras          = false // Logs all commands' read/write memory observation together with extras
+	LogMappingsToFile          = false
```
AWoloszyn: We should at least have a comment here that
1) This only works for Vulkan
2) Describes exactly what this transform is printing.
   (E.g. the remapping state at the end of the program)
Since 2 handles can be re-used it only shows you what the final remapping table looks like, and won't hold onto intermediates.

PR #2026, opened 2018-06-29
Read shader source for programs from the link extra.

### LINKED ISSUE #398: Need a distinction between attached and linked program shaders
It is valid GL to detach (and even delete) a shader from a program and still keep using that program. The shader only gets "really detached" once the program is linked again.

Currently, if glDetachShader is called, GAPIS tells the UI that no shader is attached to the program and the UI cannot show the program's source. Instead, GAPIS needs to make a distinction between attached shaders and linked shaders.

As an extension of this, care needs to be taken not to reap deleted shaders that are still "linked" by a program, so the source can still be shown in the UI.

Also, it is allowed to change the source of a shader that is still attached to a program, re-compile it, attach it to another program, and link and use that program. Doing so has no effect on the original program the shader was/is attached to.

pmuetschard:
Use the link extra for the shader source for the program resource data, so that the source will still be shown, even if shaders have been detached at the given point in the trace.

Fixes #398


hevrard:
@pmuetschard Should we rebase and merge this one?

PR #2031, opened 2018-07-04
Implement draw calls per frame and add it to gapit/stats

iburinoc:
For Vulkan a dependency graph of the synchronization primitives is created and used to determine which draw calls are run when.  For other APIs, it's assumed that the commands are executed synchronously.  Since Vulkan annotates `vkQueueSubmit` with `@draw_call`, `@executed_draw` was added to indicate that a command is a draw call only when executed as a subcommand (e.g. for `vkCmdDraw`).


iburinoc:
Rebased

PR #2035, opened 2018-07-05
core/data/slice: Add Cast() function.

ben-clayton:
This is full of voodoo and will likely break in future releases of Golang. However, I believe the performance efficiency and usage convenience may justify the magic here.

### DIFF
--- a/core/data/slice/slice.go
+++ b/core/data/slice/slice.go
@@ -108,6 +108,24 @@ func Bytes(ptr unsafe.Pointer, size uint64) []byte {
 	return ((*[1 << 30]byte)(ptr))[:size]
 }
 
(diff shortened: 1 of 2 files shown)


core/data/slice/slice.go
```diff
@@ -108,6 +108,24 @@ func Bytes(ptr unsafe.Pointer, size uint64) []byte {
 	return ((*[1 << 30]byte)(ptr))[:size]
 }
 
+// Cast performs an unsafe cast of the slice at sli to the given slice type.
```
AWoloszyn: A bit scary, but I agree worth it. 
Especially if we can detect a failure.

PR #2041, opened 2018-07-06
Vulkan: Fix Crashes caused by not using NewXXX ctor

Qining:
Fix crash when rebuilding command buffer commands


gapis/api/vulkan/command_buffer_rebuilder.go
```diff
@@ -879,7 +879,7 @@ func rebuildVkCmdSetBlendConstants(
 	s *api.GlobalState,
 	d VkCmdSetBlendConstantsArgsʳ) (func(), api.Cmd, error) {
 
-	var constants F32ː4ᵃ
+	constants := NewF32ː4ᵃ(s.Arena)
```
ben-clayton: The constructor takes array values, so you can fold the next 4 lines into this one

Qining: Changed to use the constructor for array values.

PR #2044, opened 2018-07-09
Vulkan: State rebuilding optimization reduce the number of commands

Qining:
This CL does:

0) Lazily reserve 128MB device memory for the use of scratch buffers.

1) Batch buffer->image copy commands, defer the command submissions until
the reserved scratch memory is full. This reduces the number of
VkCommandBufferBegin/End, VkQueueSubmit, VkQueueWaitIdle calls.

2) Batch the commands for 'priming by rendering' and 'priming by imageStore' in the similar way.

3) Track image last bound queue at 'ImageLevel' level.

For now, on a typical Vulkan app, this reduce 30% of the initial
commands. For `gapit report`, saves ~10% of time.


Qining:
Ready for review

PR #2052, opened 2018-07-11
Fix encoding of slices.

### LINKED ISSUE #2051: Slice serialization is broken by 8fe0726f
Repro steps:

Trace calculator in MEC mode capturing 10 frames. When you start the capture, quickly mash calculator buttons (otherwise the only frames captured are for the textbox with the blinking cursor, which is fine, but not as interesting).

Do this after and before commit 8fe0726f.

Observe that before 8fe0726f you get a good trace.

After 8fe0726f:
Observe that the frames are empty.
Observe that the slices for GL buffers are not at offset 0, but at some high value.
Look at the slice in the memory view and observe that it's all `??`.
Goto->Memory Location and type the pool of the GL buffer and adress 0 to see the data of buffer.

pmuetschard:
In 8fe0726f, the meaning of a slice's base and root have changed. They are now offsets from the pool's base. Fix the serialization code to no longer attempt to adjust the slice's root and base, since they no longer are absolute addresses.

Fixes #2051

### DIFF
--- a/gapil/compiler/plugins/encoder/encoder.go
+++ b/gapil/compiler/plugins/encoder/encoder.go
@@ -523,21 +523,11 @@ func (e *encoder) encodeValue(s *compiler.S, ptr, buf *codegen.Value, ty semanti
 		return
 	case *semantic.Slice:
 		e.writeBlob(s, buf, func(buf *codegen.Value) {
-			root := s.LocalInit("root", ptr.Index(0, compiler.SliceRoot).Load().Cast(e.T.Size))
-			base := s.LocalInit("base", ptr.Index(0, compiler.SliceBase).Load().Cast(e.T.Size))
+			root := s.LocalInit("root", ptr.Index(0, compiler.SliceRoot).Load().Cast(e.T.Size)).Load()
+			base := s.LocalInit("base", ptr.Index(0, compiler.SliceBase).Load().Cast(e.T.Size)).Load()
 			size := ptr.Index(0, compiler.SliceSize).Load()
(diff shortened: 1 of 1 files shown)


pmuetschard:
Don't know if this breaks anything else or what other implications this has, but it did fix GLES MEC for me... at least for what I've tried.

gapil/compiler/plugins/encoder/encoder.go
```diff
@@ -523,22 +523,12 @@ func (e *encoder) encodeValue(s *compiler.S, ptr, buf *codegen.Value, ty semanti
 		return
 	case *semantic.Slice:
 		e.writeBlob(s, buf, func(buf *codegen.Value) {
-			root := s.LocalInit("root", ptr.Index(0, compiler.SliceRoot).Load().Cast(e.T.Size))
-			base := s.LocalInit("base", ptr.Index(0, compiler.SliceBase).Load().Cast(e.T.Size))
+			root := s.LocalInit("root", ptr.Index(0, compiler.SliceRoot).Load().Cast(e.T.Size)).Load()
```
ben-clayton: You don't need the `LocalInit` as that was there just to handle the branch of the if.

```
root := ptr.Index(0, compiler.SliceRoot).Load()
base := ptr.Index(0, compiler.SliceBase).Load()
size := ptr.Index(0, compiler.SliceSize).Load()
count := ptr.Index(0, compiler.SliceCount).Load()
pool := ptr.Index(0, compiler.SlicePool).Load()
```

I'll submit as-is, and I'll then make a follow up change.

PR #2058, opened 2018-07-12
gazelle: move prefix from rule to directive

### LINKED ISSUE #2054: Running gazelle trashes the BUILD.bazel files
What was:
`    importpath = "github.com/google/gapid/foo/bar"`
`    deps = ["//foo/bar:go_default_library"]`

now becomes:
`    importpath = "foo/bar"`
`    deps = ["//github.com/google/gapid/foo/bar:go_default_library"]`

For 206 files.

jayconrod:
The attributes passed to the gazelle rule are translated into command
line arguments, which may be replaced if additional arguments are
given on the command line. It's better to set options like "prefix" in
directives (special comments in build files) so that Gazelle will
always see them.

Fixes #2054

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -15,10 +15,11 @@
 load("@bazel_gazelle//:def.bzl", "gazelle")
 load("//tools/build:rules.bzl", "copy_to")
 
+# gazelle:prefix github.com/google/gapid
+
 gazelle(
     name = "gazelle",
     mode = "fix",
-    prefix = "github.com/google/gapid",
 )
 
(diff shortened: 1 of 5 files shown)


jayconrod:
cc @ben-clayton 

I also re-ran Gazelle as part of this PR. There are a couple minor changes, but nothing that seems to make a difference. I wasn't able to test *everything* without installing the Android SDK, but the modified targets seem to build okay.

ben-clayton:
Although the presubmits don't look happy. I'll take a look tomorrow.

ben-clayton:
Build failures are unrelated. Thank you!

PR #2063, opened 2018-07-13
Add ability to replace all resources of a type

apazylbe:
Currently only works for shaders
UpdateResourceBinary should read the resource data from
standard input and output the modified resource data to
standard output.
Also modifying replace_resource to consume spir-v
disassembly to replace a single shader for consistency

I know MultiResourceData is not the best of names, but couldn't come up with anything better.
Suggestions are welcome.


cmd/gapit/replace_resource.go
```diff
@@ -50,13 +54,13 @@ func (verb *replaceResourceVerb) Run(ctx context.Context, flags flag.FlagSet) er
 		return nil
 	}
 
-	if verb.Handle == "" {
-		app.Usage(ctx, "-handle argument is required")
+	if verb.Handle == verb.UpdateResourceBinary {
```
ben-clayton: Is this right?

cmd/gapit/replace_resource.go
```diff
@@ -90,36 +94,101 @@ func (verb *replaceResourceVerb) Run(ctx context.Context, flags flag.FlagSet) er
...
 	for _, types := range resources.GetTypes() {
 		if types.Type == api.ResourceType_ShaderResource {
-			var matchedResource *service.Resource
-			for _, v := range types.GetResources() {
-				if strings.Contains(v.GetHandle(), verb.Handle) {
-					if matchedResource != nil {
-						return fmt.Errorf("Multiple resources matched: %s, %s", matchedResource.GetHandle(), v.GetHandle())
+			switch {
+			case verb.Handle != "":
+				matchedResource, err := resources.FindSingle(func(t api.ResourceType, r service.Resource) bool {
```
ben-clayton: This doesn't need to be in a for loop now, right?

cmd/gapit/replace_resource.go
```diff
@@ -50,13 +54,13 @@ func (verb *replaceResourceVerb) Run(ctx context.Context, flags flag.FlagSet) er
 		return nil
 	}
 
-	if verb.Handle == "" {
-		app.Usage(ctx, "-handle argument is required")
+	if (verb.Handle != "" && verb.UpdateResourceBinary != "") || (verb.Handle == "" && verb.UpdateResourceBinary == "") {
```
ben-clayton: Which is equivalent to:
`if (verb.Handle == "") != (verb.UpdateResourceBinary == "")`

Can keep as-is if you find this clearer.

cmd/gapit/replace_resource.go
```diff
@@ -90,36 +94,100 @@ func (verb *replaceResourceVerb) Run(ctx context.Context, flags flag.FlagSet) er
...
-			}
-			resourcePath := capture.Command(uint64(verb.At)).ResourceAfter(matchedResource.ID)
-			newResourceBytes, err := ioutil.ReadFile(verb.ResourcePath)
+	var resourcePath *path.Any
+	var resourceData interface{}
+
+	switch {
+	case verb.Handle != "":
+		matchedResource, err := resources.FindSingle(func(t api.ResourceType, r service.Resource) bool {
+			return strings.Contains(r.GetHandle(), verb.Handle)
```
ben-clayton: I think you need to check type here too?

cmd/gapit/replace_resource.go
```diff
@@ -90,36 +94,100 @@ func (verb *replaceResourceVerb) Run(ctx context.Context, flags flag.FlagSet) er
...
+	err = client.SaveCapture(ctx, newCapture, newCaptureFilepath)
+
+	log.I(ctx, "Capture written to: %v", newCaptureFilepath)
+	return nil
+}
+
+// getNewResourceData runs the update resource binary on the old resource data
+// and returns the newly generated resource data
+func (verb *replaceResourceVerb) getNewResourceData(ctx context.Context, resourceData string) (string, error) {
+	updateCmd := exec.Command(verb.UpdateResourceBinary)
```
ben-clayton: Did my example code not work for you?

gapis/api/resource.go
```diff
@@ -100,3 +100,7 @@ func NewResourceData(data interface{}) *ResourceData {
 		panic(fmt.Errorf("%T is not a ResourceData type", data))
 	}
 }
+
+func NewMultiResourceData(resources []*ResourceData) *MultiResourceData {
```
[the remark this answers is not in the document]
ben-clayton: Still missing docs.

cmd/gapit/replace_resource.go
apazylbe: It does. Sorry, totally missed that comment.

cmd/gapit/replace_resource.go
apazylbe: Done

cmd/gapit/replace_resource.go
apazylbe: Sorry misread your original suggestion.
Updated, though I added : (verb.Handle == "") == (verb.UpdateResourceBinary == "")
since if condition should be true for error cases

ben-clayton: Yup, just testing 😉

PR #2064, opened 2018-07-13
core/assert map support

iburinoc:


core/assert/map.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// EqualsWithComparator asserts the array or map matches expected using a comparator function
+func (o OnMap) EqualsWithComparator(expected interface{}, same func(a, b interface{}) bool) bool {
+	return o.mapsEqual(expected, same)
+}
+
+// DeepEquals asserts the array or map matches expected using a deep-equal comparison.
+func (o OnMap) DeepEquals(expected interface{}) bool {
+	return o.mapsEqual(expected, reflect.DeepEqual)
```
ben-clayton: Any reason you're using `reflect.DeepEqual` and not `compare.DeepEqual`?

core/assert/map.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func isZero(v reflect.Value) bool {
```
ben-clayton: Is this used?

core/assert/map.go
```diff
@@ -0,0 +1,107 @@
...
+	switch v.Kind() {
+	case reflect.Array, reflect.String:
+		return v.Len() == 0
```
ben-clayton: Fairly sure you don't want to consider zero length arrays as zero. The size of the array is a property of the type, not the value.

core/assert/map.go
```diff
@@ -0,0 +1,107 @@
...
+		es := reflect.ValueOf(expected)
+		if gs.Len() < es.Len() {
+			o.Printf("\tShorter\tby\t%v\tkeys\n", es.Len()-gs.Len())
+			return false
+		}
+		equal := true
+		for _, k := range gs.MapKeys() {
+			gv := gs.MapIndex(k)
+			ev := es.MapIndex(k)
+			if ev == reflect.ValueOf(nil) {
```
ben-clayton: `if !ev.IsValid()`

core/assert/map.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	return false
+}
+
+func (o OnMap) mapsEqual(expected interface{}, same func(a, b interface{}) bool) bool {
+	return o.Test(func() bool {
+		gs := reflect.ValueOf(o.mp)
+		es := reflect.ValueOf(expected)
+		if gs.Len() < es.Len() {
+			o.Printf("\tShorter\tby\t%v\tkeys\n", es.Len()-gs.Len())
```
ben-clayton: Not sure I understand why these words are tab-delimited?

core/assert/map.go
iburinoc: Nope, forgot to delete

core/assert/map.go
iburinoc: Nope, I modeled this off of core/assert/slice.go and wasn't aware compare existed, will change.

core/assert/map.go
iburinoc: Yeah ignore this function

core/assert/map.go
iburinoc: I was basing it off the convention from the rest of core/assert, I can change it if it would be better to just be spaces.

core/assert/map.go
ben-clayton: I guessed, but I was confused as slice.go does use `compare.DeepEqual`. 😃

core/assert/map.go
ben-clayton: The only place I see this is in slice.go, where the tabs are used to align a side-by-side comparison of got vs expected. I don't see it used for whitespace in sentences.

iburinoc: It shows up in //core/assert/string.go as well for sentence whitespace, but I'll change this to spaces

core/assert/map.go
iburinoc: Oh.  Hmm.  I thought I copied from slice.go, weird.

core/assert/map.go
ben-clayton: > It shows up in //core/assert/string.go as well for sentence whitespace

That's just odd.

iburinoc: Fixed

core/assert/map.go
iburinoc: Fixed

PR #2077, opened 2018-07-23
 MEC GLES support cube maps.

pmuetschard:
And refactoring for readability.

### DIFF
--- a/gapii/cc/gles_mid_execution.cpp
+++ b/gapii/cc/gles_mid_execution.cpp
@@ -62,6 +62,7 @@ class Sampler {
 
   virtual std::string getExtensions() const = 0;
   virtual std::string getUniform() const = 0;
+  virtual std::string getFragmentPreamble() const { return ""; }
   virtual std::string getSamplingExpression() const = 0;
 
   void bindTexture(const GlesImports& imports, GLint texId) const {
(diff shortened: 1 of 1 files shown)


gapii/cc/gles_mid_execution.cpp
```diff
@@ -50,137 +50,6 @@ class TempObject {
   std::function<void()> mDeleteId;
 };
```
ben-clayton: Sorry for not noticing this in an earlier review, but I think this class is dangerous and currently depends on compiler-specific behavior.

The destructor deletes the ID, but you might have a temporary instance when returning the `TempObject`. In this case, you'll call `mDeleteId` multiple times (and before you expect). The compiler is free to remove this temporary if it is smart enough (which is why I think your code has been working), but I believe in your usage, it is allowed to use temporaries.

To ensure that this doesn't happen, you can declare a move constructor, and delete the implicit copy constructor and assignment operator:

```c++
class TempObject {
 public:
  TempObject(uint64_t id, const std::function<void()>& deleteId)
      : mId(id), mDeleteId(deleteId) {}
  TempObject(TempObject&& other) {
      mId = other.mId;
      mDeleteId = other.mDeleteId;
      other.mDeleteId = nullptr;
  }
  uint64_t id() { return mId; }
  ~TempObject() { 
      if (mDeleteId) {
          mDeleteId();
      }
  }
 private:
  TempObject(const TempObject&) = delete;
  TempObject& operator = (const TempObject&) = delete;
  uint64_t mId;
  std::function<void()> mDeleteId;
};
```

For more info, [take a look at this post](https://akrzemi1.wordpress.com/2011/08/11/move-constructor/).

pmuetschard: Done. Thanks, Ben! After first reading this, I had the same worry and have had this TODO in the back of my head to follow up with you/Andrew about this. As far as I had found out/understood, copy elision is mandatory for function return values, which is why it has been working. However, I feel much more confident about it now, disallowing copy construction.

PR #2079, opened 2018-07-23
Copy depth aspect by rendering in overdraw

iburinoc:
Specifically do this for the store end (i.e. new depth/stencil image to
old image), so that we don't have to add TRANSFER_DST to all depth
images.

Additionally, remove the conditions in gapii and always add the transfer_src flag to images, so that we can use it to copy from at the start of the overdraw renderpass.

### DIFF
--- a/gapis/api/vulkan/image_primer.go
+++ b/gapis/api/vulkan/image_primer.go
@@ -109,13 +109,32 @@ func (p *imagePrimer) primeByRendering(img ImageObjectʳ, opaqueBoundRanges []Vk
 	for _, aspect := range p.sb.imageAspectFlagBits(img.ImageAspect()) {
 		for layer := uint32(0); layer < img.Info().ArrayLayers(); layer++ {
 			for level := uint32(0); level < img.Info().MipLevels(); level++ {
+				inputImageObjects := copyJob.srcAspectsToDsts[aspect].dstImgs
+				inputImages := make([]ipRenderImage, len(inputImageObjects))
+				for i, iimg := range inputImageObjects {
+					layout := iimg.Aspects().Get(
+						VkImageAspectFlagBits_VK_IMAGE_ASPECT_COLOR_BIT).Layers().Get(
+						layer).Levels().Get(
+						level).Layout()
+					inputImages[i] = ipRenderImage{
+						image:         iimg,
+						aspect:        VkImageAspectFlagBits_VK_IMAGE_ASPECT_COLOR_BIT,
+						layer:         layer,
+						level:         level,
+						initialLayout: layout,
+						finalLayout:   layout,
+					}
+				}
 				renderJobs = append(renderJobs, &ipRenderJob{
-					inputAttachmentImages: copyJob.srcAspectsToDsts[aspect].dstImgs,
-					renderTarget:          img,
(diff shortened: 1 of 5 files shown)


gapis/api/vulkan/image_primer.go
```diff
@@ -1263,26 +1293,46 @@ func (h *ipRenderHandler) render(job *ipRenderJob, tsk *scratchTask) error {
...
-					0, // baseMipLevel
-					input.Info().MipLevels(), // levelCount
+					aspects, // aspectMask
+					0,       // baseMipLevel
+					input.image.Info().MipLevels(), // levelCount
+					0, // baseArrayLayer
+					input.image.Info().ArrayLayers(), // layerCount
+				),
+			))
+		dstBarriers = append(dstBarriers,
```
Qining: 1) Prefer to have a check to see if the `finalLayout` and the current layout is the same. Because for primeByRendering jobs, the layout of input attachments are guaranteed to be SHADER_READ_ONLY_OPTIMAL before the rendering (line 80), so this is not necessary for those cases, also the barriers for the input attachments in the `inputSrcBarriers` above.

2) When `render()` is called from places other than `primeByRendering` for aspects other than ONLY stencil, not used  when priming for COLOR or DEPTH aspects, so that layout transitions for input attachments can be missed.

Qining: Sorry, the 2) point is not valid, and for the 1) not necessary for the `inputSrcBarriers`, I misunderstood the code here.

gapis/api/vulkan/image_primer.go
```diff
@@ -1047,44 +1074,45 @@ func (h *ipRenderHandler) free() {
 	}
 }
 
+func imageBarrierAspectFlags(aspect VkImageAspectFlagBits, fmt VkFormat) VkImageAspectFlags {
```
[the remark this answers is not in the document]
iburinoc: Done

gapis/api/vulkan/image_primer.go
```diff
@@ -1461,41 +1513,41 @@ func (h *ipRenderHandler) render(job *ipRenderJob, tsk *scratchTask) error {
...
+		h.sb.write(h.sb.cb.VkCmdPipelineBarrier(
+			commandBuffer,
+			VkPipelineStageFlags(VkPipelineStageFlagBits_VK_PIPELINE_STAGE_ALL_COMMANDS_BIT),
+			VkPipelineStageFlags(VkPipelineStageFlagBits_VK_PIPELINE_STAGE_ALL_COMMANDS_BIT),
+			VkDependencyFlags(0),
+			0,
+			memory.Nullptr,
+			0,
+			memory.Nullptr,
+			1,
```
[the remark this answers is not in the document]
iburinoc: Done

gapis/api/vulkan/image_primer.go
iburinoc: Done

Qining:
`image_primer` code looks good to me. Defer the other parts to @AWoloszyn .

iburinoc:
Also adding a fix for an incorrect handling of clear values that led to the stencil aspect being cleared to the wrong value sometimes.

PR #2081, opened 2018-07-24
A few minor updates to the tracing interface.

AWoloszyn:
This allows you to specify environment variables in your
ssh config, (instead of having to add them every time).

It also exposes portWatcher for external use.

### DIFF
--- a/core/os/process/client.go
+++ b/core/os/process/client.go
@@ -29,7 +29,23 @@ import (
 
 var portPattern = regexp.MustCompile(`^Bound on port '(\d+)'$`)
 
-type portWatcher struct {
+// NewPortWatcher creates and returns an initialized port watcher
+func NewPortWatcher(portChan chan<- string, opts *StartOptions) *PortWatcher {
+	return &PortWatcher{
+		portChan: portChan,
+		stdout:   opts.Stdout,
+		portFile: opts.PortFile,
+		device:   opts.Device,
+	}
+}
+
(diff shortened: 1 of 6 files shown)


core/os/process/client.go
```diff
@@ -29,7 +29,17 @@ import (
 
 var portPattern = regexp.MustCompile(`^Bound on port '(\d+)'$`)
 
-type portWatcher struct {
+func NewPortWatcher(portChan chan<- string,
```
ben-clayton: Nit: No need to break this signature across two lines

core/os/process/client.go
```diff
@@ -29,7 +29,17 @@ import (
...
+	opts *StartOptions) *PortWatcher {
+	return &PortWatcher{
+		portChan: portChan,
+		stdout:   opts.Stdout,
+		portFile: opts.PortFile,
+		device:   opts.Device,
+	}
+}
+
+type PortWatcher struct {
```
ben-clayton: Godocs please

gapis/trace/desktop/trace.go
```diff
@@ -144,7 +141,7 @@ func (t *DesktopTracer) GetTraceTargetNode(ctx context.Context, uri string, icon
 		children[i] = filepath.Join(uri, children[i])
```
ben-clayton: Don't you need to deal with the target device slashes here too?

PR #2082, opened 2018-07-24
Add an option to GAPID to disable unknown extensions.

AWoloszyn:


gapii/cc/spy_base.h
```diff
@@ -185,6 +185,8 @@ class SpyBase {
   track_memory::MemoryTracker mMemoryTracker;
 #endif  // TARGET_OS
 
+  // This is true if we should not show disabled extensions to the application.
```
pmuetschard: nit: this comment was hard for me to parse :) I think it be better phrased as "If true, we will hide disabled extensions from the application."

AWoloszyn:
@ben-clayton  Thanks, I hadn't run the presubmits locally because this was done on my windows machine. I will pull on Linux and fix that.

gapii/cc/spy_base.h
AWoloszyn: Sure

AWoloszyn:
@pmuetschard  According to Ben, GLES doesn't advertise unknown extensions, so this should be a NO-op

I will open a bug to make the option... optional. (Similar to disablePCS)

PR #2087, opened 2018-07-24
core/cc: Add AsyncBuffer - a threaded ringbuffer queue.

ben-clayton:
This can be placed in between the `PackEncoder` and the `ConnectionStream` to buffer the writes and perform socket IO on a separate thread.

### DIFF
--- a/core/cc/async_buffer.cpp
+++ b/core/cc/async_buffer.cpp
@@ -1,0 +1,118 @@
+/*
+ * Copyright (C) 2018 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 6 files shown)


core/cc/async_buffer.cpp
```diff
@@ -0,0 +1,113 @@
...
+  while (true) {
+    int count = written_.acquire_all();
+    if (count < 0) {
+      return;  // Signal that AsyncBuffer is destructing.
+    }
+
+    size_t remaining = static_cast<size_t>(count);
+    while (remaining > 0) {
+      size_t chunk = std::min(remaining, size_ - read_head_);
+      size_t count = out_->write(&buffer_[read_head_], chunk);
```
ben-clayton: Note: if the socket is disconnected, this will likely return 0 and spin forever. Not sure how to best tackle this tonight.

AWoloszyn:
I am going to close this for now, if it is needed in the future we can take a look then,

PR #2090, opened 2018-07-25
A few more small fixes around images and descriptor sets.

AWoloszyn:
It is valid to call VkAllocateDescriptorSets even in the cases
where it will fail. (1025 sets in a 1024 set pool).

Clean up some warnings that the validation layers were throwing
from the state rebuilder.


gapis/api/vulkan/state_rebuilder.go
```diff
@@ -2405,11 +2405,12 @@ func (sb *stateBuilder) createDescriptorPoolAndAllocateDescriptorSets(dp Descrip
 	descSetHandles := make([]VkDescriptorSet, 0, dp.DescriptorSets().Len())
 	descSetLayoutHandles := make([]VkDescriptorSetLayout, 0, dp.DescriptorSets().Len())
 	for vkDescSet, descSetObj := range dp.DescriptorSets().All() {
-		if sb.s.DescriptorSets().Contains(vkDescSet) {
+		if vkDescSet != VkDescriptorSet(0) && sb.s.DescriptorSets().Contains(vkDescSet) && sb.s.DescriptorSetLayouts().Contains(descSetObj.Layout().VulkanHandle()) {
```
ben-clayton: Nit: no need to cast literals. `VkDescriptorSet(0)` becomes `0`

PR #2097, opened 2018-07-27
Overdraw: Check for format support and convert if necessary

iburinoc:
This also requires adding to the extras on vkEnumeratePhysicalDevices to check what formats the device supports.  I set it up to just get the format support for all formats, but if that is too expensive this feature only needs to know support for the stencil formats.


ben-clayton:
Am I right in thinking you're getting information about the trace device, storing this in the capture as an extra, and then using this to determine what you should do for replay? If so, we should chat.

ben-clayton:
Hi Sean.

I've filed #2101 to track my concerns here. I know this is your last week, so I won't block this.

iburinoc:
Thanks Ben, I agree it's not ideal.  If we were to loosen the "replay on original device" restriction at some point I don't think the modifications to this would be too complex.

PR #2098, opened 2018-07-27
Even more Compiler changes

ben-clayton:


gapil/compiler/compiler.go
```diff
@@ -438,6 +444,61 @@ func (c *C) CurrentExpression() semantic.Expression {
 	return c.expressionStack[len(c.expressionStack)-1]
 }
 
+// SourceLocation holds information about a line in the API source.
```
AWoloszyn: If it holds info about the line, why is there a Column in it?

PR #2099, opened 2018-07-27
GLES MEC: Support for more texture formats.

pmuetschard:


gapii/cc/gles_mid_execution.cpp
```diff
@@ -224,10 +204,54 @@ class SamplerCube : public Sampler {
   GLint mFace;
 };
 
+typedef struct {
```
ben-clayton: The use of `typedef struct` is only really used in C code. Just use `struct texture_state { ... }`

ben-clayton:
```
gapii/cc/gles_mid_execution.cpp:84:12: error: private field 'mTarget' is not used [-Werror,-Wunused-private-field]
  uint32_t mTarget;
```

PR #2102, opened 2018-07-30
Don't merge stdout/stderr on remote device-info.

AWoloszyn:
If the remote device-info produces stderr output, but does not
fail, we should log it, but not try to parse it as json.

### DIFF
--- a/core/os/device/remotessh/commands.go
+++ b/core/os/device/remotessh/commands.go
@@ -22,6 +22,7 @@ import (
 	"net"
 	"os"
 	"path"
+	"runtime"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


core/os/device/remotessh/commands.go
```diff
@@ -185,8 +186,16 @@ func (b binding) PushFile(ctx context.Context, source, dest string) error {
 	if err != nil {
 		return err
 	}
+	mode := permission.Mode()
+	// If we are on windows pushing to Linux, we lose the executable
+	// bit, geti t back.
```
ben-clayton: `get it back`

PR #2112, opened 2018-07-31
Add support for rendering during replay to the surface

iburinoc:
Currently only Android is supported, but it's easily extensible for other surface types (e.g. Xcb/Win32)

Creating/obtaining the create info for the surface to render to is handled on the gapir side.  On the virtual swapchain side, it takes the create info and makes a surface and swapchain, and then when an image is presented to the virtual swapchain it blits to the real swapchain and presents that as well.


core/vulkan/vk_virtual_swapchain/cc/base_swapchain.cpp
```diff
@@ -0,0 +1,287 @@
...
+    &swapchain_, // pSwapchains,
+    &index, // pImageIndices
+    nullptr, // pResults
+  };
+  if ((res = queue_functions->vkQueuePresentKHR(queue, &presentInfo)) !=
+      VK_SUCCESS) {
+    return res;
+  }
+
+  device_functions_->vkWaitForFences(device_, 1, &blit_fence_, VK_TRUE,
```
AWoloszyn: I am a bit leery about this vkWaitForFences. Since this is running on the main thread, we can end up blocking the user thread and synchronizing with the GPU.

iburinoc:
Odd that presubmit failed here with "Git workspace not clean", not sure what's causing that.  Presubmit succeeds locally.

AWoloszyn:
It could be a difference in clang-format versions.

iburinoc:
Ah "git workspace not clean" is just the way of saying presubmit changed things?  That makes sense

core/vulkan/vk_virtual_swapchain/cc/base_swapchain.cpp
```diff
@@ -0,0 +1,342 @@
...
+      nullptr,                                      // pInheritanceInfo
+  };
+  if ((res = device_functions_->vkBeginCommandBuffer(cmdbuf, &beginInfo)) !=
+      VK_SUCCESS) {
+    return res;
+  }
+
+  // The source image is already in VK_IMAGE_LAYOUT_TRANSFER_SRC, we need to
+  // transition our image between VK_IMAGE_LAYOUT_TRANSFER_DST and
+  // VK_IMAGE_LAYOUT_SHADER_PRESENT_KHR
```
apazylbe: s/SHADER/SHARED

PR #2113, opened 2018-07-31
gapis/service/path: Add ResolveConfig

ben-clayton:
Pass this to `Get()`, `Set()`, `Follow()` and `Find()`.
Propagate the ResolveConfig thoughout the resolve package.

This is used to specify the replay device for resolving state data pulled back from the GPU.

### DIFF
--- a/gapis/resolve/framebuffer_attachment.go
+++ b/gapis/resolve/framebuffer_attachment.go
@@ -39,7 +39,9 @@ func FramebufferAttachment(
 	attachment api.FramebufferAttachment,
 	settings *service.RenderSettings,
 	hints *service.UsageHints,
+	config *path.ResolveConfig,
 ) (*path.ImageInfo, error) {
+
 	if replaySettings.Device == nil {
 		devices, err := devices.ForReplay(ctx, after.Capture)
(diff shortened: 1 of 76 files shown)


gapis/resolve/framebuffer_attachment.go
```diff
@@ -39,7 +39,9 @@ func FramebufferAttachment(
 	attachment api.FramebufferAttachment,
 	settings *service.RenderSettings,
 	hints *service.UsageHints,
+	config *path.ResolveConfig,
```
pmuetschard: This a bit redundant now. Both the `ReplaySettings` and the `ResolveConfig` contain the replay device.

Should we also remove the default device selection below?

ben-clayton: I want to move this RPC into a Get. At that point, yes, we should certainly de-dupe this field.
I haven't done this yet as I wanted to make sure these changes worked first.

pmuetschard: SGTM

ben-clayton:
Merged in https://github.com/google/gapid/pull/2130

PR #2122, opened 2018-08-01
Pin mingw-gcc to version 7.3.0-2.

pmuetschard:
The current version, 8.2, fails to build.


pmuetschard:
Sorry, this was not yet meant to be reviewed. I want to check that the presubmit builds.

AWoloszyn:
It looksl ike this fixes the windows build (and the mac build is still flakey, but I cant see how this would affect that)

PR #2124, opened 2018-08-02
Add support for replaying to the surface on Windows and Linux

iburinoc:


iburinoc:
Rebased on top of fix from #2125

gapis/api/templates/vulkan_gfx_api_extras.tmpl
```diff
@@ -352,14 +356,57 @@ bool Vulkan::replayCreateVkDeviceImpl(Stack* stack, size_val physicalDevice,
...
           vsPNext = *pNext;
 
-	  void* data = SurfaceInfo();
+          gapir::SurfaceType target = gapir::SurfaceType::Unknown;
+	  switch (*(uint32_t*)pNext->surfaceCreateInfo) {
+          case VK_STRUCTURE_TYPE_ANDROID_SURFACE_CREATE_INFO_KHR:
+	    target = gapir::SurfaceType::Android;
+	    break;
+	  // These are all linux, so just make an Xcb surface
+          case VK_STRUCTURE_TYPE_MIR_SURFACE_CREATE_INFO_KHR:
```
AWoloszyn: This is not strictly true, but we haven't seen anything that violates this, but technically, if you are on linux you MAY be running a MIR or Wayland compositor where X would fail.

We can handle this if we ever see a system that doesn't allow X.

AWoloszyn:
I will merge when the builds go green

gapis/api/templates/vulkan_gfx_api_extras.tmpl
iburinoc: Ah ok, yeah if it's an issue adding more window types in the future shouldn't be challenging.

PR #2136, opened 2018-08-08
Fix a bug where trace times were completely wrong.

AWoloszyn:


cmd/gapit/trace.go
```diff
@@ -146,7 +146,7 @@ func (verb *traceVerb) Run(ctx context.Context, flags flag.FlagSet) error {
 		AdditionalCommandLineArgs: verb.AdditionalArgs,
 		Cwd:                   verb.WorkingDir,
 		Environment:           verb.Env,
-		Duration:              float32(verb.For * time.Second),
+		Duration:              float32(verb.For / time.Second),
```
ben-clayton: Suggest using the `Seconds` method: `float32(verb.For.Seconds())`

AWoloszyn:
Updated.

PR #2140, opened 2018-08-10
Use a more robust way to retrieve screen states

zhangyiwei:
The current implementation won't work on late Android Pie releases because dumpsys window policy removed the state flag for display and lockguard. This change proposed a more robust and consistent way for retrieving screen states using power manager and activity manager.

### DIFF
--- a/core/os/android/adb/adb_data_test.go
+++ b/core/os/android/adb/adb_data_test.go
@@ -52,6 +52,8 @@ push_device              device
 rooted_device            unauthorized
 run_device               unknown
 screen_off_device        offline
+screen_doze_device       offline
+screen_unready_device    offline
 screen_on_device         device
 `)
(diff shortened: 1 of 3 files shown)


ben-clayton:
Thank you for your PR @zhangyiwei.

The change looks good - we just need to update the tests. (You can run them all with `bazel test tests`).

[The failing test is here](https://github.com/google/gapid/blob/master/core/os/android/adb/adb_data_test.go#L110) and can be run with `bazel test //core/os/android/adb:go_default_test`.

zhangyiwei:
Thanks for your review, and sorry for the missing bit. Will send an update soon.

zhangyiwei:
Not sure how can I remove those useless "Merge*" commits...

zhangyiwei:
The update is combined in the 1 commit because of getting rid of false merge commits. The related tests have passed. Let me know if there's any other concerns.

PR #2141, opened 2018-08-10
Workaround an issue in golang 1.10.3 where a 4GB file is truncated.

AWoloszyn:
https://github.com/golang/go/issues/26923


AWoloszyn:
The golang was in the first comment. (As well as commit comment).
I can move this out.

PR #2157, opened 2018-08-22
Vulkan: Reduce DCE memory usage

Qining:
For a typical Vulkan application trace of size 1.8GB. This CL saves
~700MB of memory. Also this CL reserves space for slice/map whenever
possible for footprint building, to improve performance.


Qining:
@AWoloszyn This CL now only contains DCE related changes.

PR #2159, opened 2018-08-24
Add export_replay for standalone replay with gapir.

hysw:


hysw:
I can't see any build results, local linux build seems to be fine.

cmd/gapir/cc/main.cpp
```diff
@@ -210,9 +211,12 @@ int main(int argc, const char* argv[]) {
   const char* portArgStr = "0";
   const char* authTokenFile = nullptr;
   int idleTimeoutSec = 0;
+  const char* replayArchive = nullptr;
 
   for (int i = 1; i < argc; i++) {
-    if (strcmp(argv[i], "--auth-token-file") == 0) {
+    if (strcmp(argv[i], "--replay-archive") == 0) {
+      replayArchive = argv[++i];
```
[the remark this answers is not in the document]
Qining: And I think 'using archive' and 'connecting to gapis' are mutual exclusive, using both at the same time might be invalid (at least for this CL, and same to options like: port, idle-timeout-sec), so probably we should document this somewhere.

gapis/replay/export_replay.go
```diff
@@ -0,0 +1,150 @@
...
+	if generator == nil {
+		return log.Errf(ctx, nil, "Unable to find replay API.")
+	}
+
+	d := bind.GetRegistry(ctx).Device(pDevice.ID.ID())
+	if d == nil {
+		return log.Errf(ctx, nil, "Unknown device %v", pDevice.ID.ID())
+	}
+
+	// executeCounter.Increment()
```
AWoloszyn: Remove commented out code

gapir/cc/replay_archive.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+#include <grpc++/grpc++.h>
+#include <fstream>
+#include <memory>
+
+#include "gapir/replay_service/service.grpc.pb.h"
+
+namespace gapir {
+
+std::unique_ptr<ReplayArchive::Payload> ReplayArchive::getPayload() {
```
Qining: ReplayConnection::Payload? Not sure which one is better.

cmd/gapir/cc/main.cpp
```diff
@@ -283,6 +287,24 @@ int main(int argc, const char* argv[]) {
     core::Debugger::waitForAttach();
   }
 
+  if (replayArchive) {
```
AWoloszyn: Now that we have a big mode-switch at the start of the replay, I would split this out into 2 methods. One that starts an "archived replay" and one that opens a connection.

It is a bit weird to have a "start the program in an entirely different way" early-out in the middle of main().

gapir/cc/replay_archive.cpp
```diff
@@ -0,0 +1,50 @@
...
+  payload->ParseFromIstream(&input);
+  return std::unique_ptr<Payload>(new Payload(std::move(payload)));
+}
+
+std::unique_ptr<ReplayArchive::Resources> ReplayArchive::getResources(
+    std::unique_ptr<ResourceRequest> req) {
+  return nullptr;
+}
+bool ReplayArchive::sendReplayFinished() { return true; }
+bool ReplayArchive::sendCrashDump(const std::string& filepath,
```
AWoloszyn: We should be writing out the crash dump here.  (And logging it's location).

gapir/cc/replay_archive.cpp
```diff
@@ -0,0 +1,50 @@
...
+                                  const void* crash_data, uint32_t crash_size) {
+  return true;
+}
+bool ReplayArchive::sendPostData(std::unique_ptr<Posts> posts) { return true; }
+bool ReplayArchive::sendNotification(uint64_t id, uint32_t severity,
+                                     uint32_t api_index, uint64_t label,
+                                     const std::string& msg, const void* data,
+                                     uint32_t data_size) {
+  return true;
```
AWoloszyn: We should log any notifications that we got here.

hysw: GAPIR already write those to log.

core/archive/archive.go
```diff
@@ -0,0 +1,45 @@
...
+type Archive = *C.archive
+
+// New creates an archive.
+func New(name string) Archive {
+	cstr := C.CString(name)
+	defer C.free(unsafe.Pointer(cstr))
+	return C.archive_create(cstr)
+}
+
+// Dispose flush and close the underlying archive.
```
Qining: flushes and closes

gapis/api/vulkan/replay.go
```diff
@@ -781,3 +781,8 @@ func (a API) QueryIssues(
 	}
 	return res.([]replay.Issue), nil
 }
+
+// ExportReplayRequest returns request type for standalone replay.
+func (a API) ExportReplayRequest() replay.Request {
+	return issuesRequest{}
```
Qining: We really should have a new type of request, instead of using `issuesRequest`, eventually. And using `issuesRequest` won't have any framebuffer data postback, so can be hard to tell if the replay is correct or not.

AWoloszyn:
Build failure is a flake, merging.

gapis/replay/export_replay.go
```diff
@@ -0,0 +1,148 @@
...
+
+	_, ranges, err := initialcmds.InitialCommands(ctx, pCapture)
+
+	generatorReplayTimer.Time(func() {
+		err = generator.Replay(
+			ctx,
+			Intent{pDevice, pCapture},
+			Config(&struct{}{}),
+			[]RequestAndResult{{
+				Request: Request(generator.(interface{ ExportReplayRequest() Request }).ExportReplayRequest()),
```
ben-clayton: Instead of (ab)using the requests, could we not just add a `ForExport bool` field to the `Intent` struct? That way it is clearly documented, and not using magic interfaces.

PR #2160, opened 2018-08-24
Track MEC in analytics events.

pmuetschard:
Adds a "mec" label to the capture analytics event, if the capture is an MEC capture.

As a side effect:
  - reneables analytics and crash reporting
  - gets them building again
  - updates to a newer version of gazelle and rules_go (so we can use build tags)


pmuetschard:
:cry: i guess we need a new vesion of bazel as well....

pmuetschard:
ping

Issue #1668, opened 2018-02-28
Vulkan replays allocate way too much Volatile Memory


baldwinn860:
While debugging issues with vulkan replays it was discovered that far too much volatile memory was being allocated, as if something were not being freed.

ben-clayton:
Related to: #1316

AWoloszyn:
Fixed!

PR #2167, opened 2018-08-28
Fix to reduce memory usage of gapis video.

AWoloszyn:
It no longer grows with each image.


AWoloszyn:
Fixes https://github.com/google/gapid/issues/1668

Qining:
Looks like the presubmit is not passing.

PR #2172, opened 2018-08-29
Drop the raw bytes of loaded captures

Qining:
Once a capture is loaded, delete its raw bytes data from database. This
can save memory.


Qining:
Closed as replaced by #2179

PR #2173, opened 2018-08-30
Add Dense map to the API language.

AWoloszyn:
This allows us to create maps that exist in the 0->n where
n is small. This provides some performance boost.

TODO: Make the same optimization on the Golang side, this should
provide better performance there too.

This reduces trace overhead by ~10% for a particularly heavy
vulkan application.


gapil/runtime/cc/map.inc
```diff
@@ -98,13 +103,84 @@ inline void Map<K, V>::Allocation::release() {
...
-V* Map<K, V>::Allocation::index(K key, bool insert) {
+template<typename T, bool DENSE>
+typename std::enable_if<std::is_integral<T>::value, uint64_t>::type dense_hash(const T& t) {
+    return static_cast<uint64_t>(t);
+}
+
+
+template<typename T, bool DENSE>
+typename std::enable_if<!std::is_integral<T>::value, uint64_t>::type dense_hash(const T& t) {
+    return 0;
```
ben-clayton: Can we delete this implementation so that it won't compile if the key is non-integral (that or put some statement in here that doesn't compile with a comment)? Doesn't seem like you'd ever want this to compile...

AWoloszyn: We can once 
    // TODO(awoloszyn): Template tricks to move this to compile-time
is resolved. Until then we have a dynamic check, which means this will need to compile.

PR #2179, opened 2018-09-04
Do not cache the capture data in database if the capture is from a file

Qining:
If the capture data is stored in a trace file, do not cache the data in raw bytes again in the database.

This CL reduce memory usage, for a typical application, around 2-4GB.


Qining:
This is to replace #2172

Qining:
Close as #2186 did the same thing

PR #2181, opened 2018-09-05
 Pre-resolve dependency graph

Qining:
Resolve the dependency graph in parallel immediately once the capture
data is resolve. This speed up the GAPIS side a bit (~30 sec for a
typical Vulkan app, 1.8G trace file) for buliding the instructions to
get the framebuffer data from the replay device.

based on #2172

### DIFF
--- a/gapis/server/server.go
+++ b/gapis/server/server.go
@@ -27,7 +27,9 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/google/gapid/core/app/crash"
 	"github.com/google/gapid/core/app/crash/reporting"
+	"github.com/google/gapid/core/context/keys"
 
 	"github.com/google/gapid/core/app"
 	"github.com/google/gapid/core/app/analytics"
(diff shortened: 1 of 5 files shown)


gapis/server/server.go
```diff
@@ -230,6 +233,14 @@ func (s *server) LoadCapture(ctx context.Context, path string) (*path.Capture, e
 	if _, err = capture.ResolveFromPath(ctx, p); err != nil {
 		return nil, err
 	}
+	// Pre-resolve the dependency graph.
```
Qining: @AWoloszyn I'm not sure if this is a proper place to insert the code. This makes the function name: 'LoadCapture' not that accurate. If so, we also need to do the same for 'ImportCapture'.

Another place is the capture.go:Import(), just after the `path.Capture` is built. Then we will need to resolve the capture there first, then resolve dependency graph.

Any suggestions?

Qining:
Will drop the second commit: `Do not cache capture file data 6402de1` then submit. The capture raw data is handled in #2179 and #2186

PR #2182, opened 2018-09-06
Remove the link-time dependency on libXCB.

AWoloszyn:
We still require the headers, but not the library to be present.
This is useful if you want to run gapir on headless machines.


Qining:
Just need to clang format.

PR #2184, opened 2018-09-07
Close ssh tunnel when either end closes the connection

Qining:
So that when users tracing through SSH with a specific number of frames to capture, the connection will be closed after that number of frames are captured.


core/os/device/remotessh/commands.go
```diff
@@ -208,8 +208,31 @@ func (b binding) doTunnel(ctx context.Context, local net.Conn, remotePort int) e
...
+				}
+			}
+			if er != nil {
+				if er != io.EOF {
+					err = er
+				}
+				break
+			}
+		}
+		writer.Close()
```
AWoloszyn: Could we not just continue using io.Copy() and just close the writer then that finishes?

Qining:
@AWoloszyn You are right, the previous code is exactly the same to the io.Copy. This one check the error message to ignore the errors from normal exit.

PR #2186, opened 2018-09-07
Another way to drop raw bytes of the capture data

Qining:
This is another approach to do the same thing as #2179


Qining:
@ben-clayton Any suggestions about this CL?

Qining:
Yep, I always squash and rebase for CLs.

PR #2188, opened 2018-09-07
Use mmap-ed archive on Linux.

hysw:


core/cc/archive.cpp
```diff
@@ -17,18 +17,190 @@
...
+    GAPID_FATAL("Unable to open archive data file %s", dataFilename.c_str());
+  }
+
+  // Open or create the archive index file in binary read/write mode.
+  const std::string indexFilename(archiveName + ".index");
+  if (!(mIndexFile.open(indexFilename.c_str()))) {
+    GAPID_FATAL("Unable to open archive index file %s", indexFilename.c_str());
+  }
+
+  // Load the archive index in memory.
```
AWoloszyn: This  should probably be shared with the non-linux implementation.

core/cc/archive.cpp
```diff
@@ -17,18 +17,190 @@
...
+#include <sys/stat.h>
+#include <sys/types.h>
+namespace core {
+
+bool Archive::MappedFile::open(const std::string &filename) {
+  fd = ::open(filename.c_str(), O_RDWR | O_CREAT, S_IRWXU);
+  struct stat st;
+  ::fstat(fd, &st);
+  size = st.st_size;
+  // ::ftruncate(size);
```
AWoloszyn: Remove commented out code

core/cc/archive.cpp
```diff
@@ -37,14 +46,152 @@ void must_truncate(int fd, off_t length) {
...
+    requiredCapacity = end * 3 / 2;
+  }
+
+  if (::ftruncate(fd, requiredCapacity)) {
+    GAPID_FATAL("Unable to ftruncate(grow) archive file.");
+  }
+
+  capacity = requiredCapacity;
+
+  // Align to the next 4k boundary, not necessary but good to have.
```
AWoloszyn: This should be moved up.

core/cc/archive.cpp
```diff
@@ -29,6 +29,15 @@
...
 #endif
 
+#if defined GAPID_ARCHIVE_USE_MMAP
+
+#include <fcntl.h>
+#include <sys/mman.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+
+#endif  //  TARGET_OS == GAPID_OS_LINUX
```
ben-clayton: Doesn't match the `#if`

core/cc/archive.cpp
```diff
@@ -29,6 +29,15 @@
 #include <unistd.h>
 #endif
 
+#if defined GAPID_ARCHIVE_USE_MMAP
```
ben-clayton: We use `#ifdef` elsewhere in the codebase, but I personally prefer to use just `#if` and always define it to `0` or `1`. This way you can get the compiler to warn if the define is missing instead of just enabled / disabled (that said, I'm not sure if we have those compiler flags enabled)

core/cc/archive.cpp
```diff
@@ -76,35 +221,32 @@ Archive::Archive(const std::string& archiveName) {
...
   if (mIndexFile) fclose(mIndexFile);
 }
 
-bool Archive::contains(const std::string& id) const {
+bool Archive::contains(const std::string &id) const {
   return mRecords.find(id) != mRecords.end();
 }
 
-bool Archive::read(const std::string& id, void* buffer, uint32_t size) {
+bool Archive::read(const std::string &id, void *buffer, uint32_t size) {
```
ben-clayton: nit: I believe clang-format will position these based on the common usage in the file. We position them next to the type elsewhere in our codebase.

core/cc/archive.cpp
```diff
@@ -40,11 +49,147 @@ void must_truncate(int fd, off_t length) {
...
+
+bool Archive::RecordFile::map() {
+  if (base || capacity == 0) {
+    return true;  // Already mapped or don't need to map.
+  }
+  base = ::mmap(nullptr, capacity, PROT_READ | PROT_WRITE, MAP_SHARED, fd, 0);
+  return base != nullptr;
+}
+
+#else
```
ben-clayton: This is a large preprocessor block. Please add a comment describing what the condition is.

core/cc/archive.cpp
```diff
@@ -40,11 +49,147 @@ void must_truncate(int fd, off_t length) {
...
+uint64_t Archive::RecordFile::size() {
+  fseek(fp, 0, SEEK_END);
+  return ftell(fp);
+}
+
+bool Archive::RecordFile::resize(uint64_t size) {
+  must_truncate(fileno(fp), size);
+  return true;
+}
+#endif
```
ben-clayton: Ditto

core/cc/archive.cpp
```diff
@@ -40,11 +49,147 @@ void must_truncate(int fd, off_t length) {
...
+
+Archive::Archive(const std::string &archiveName) {
   // Open or create the archive data file in binary read/write mode.
   const std::string dataFilename(archiveName + ".data");
-  if (!(mDataFile = fopen(dataFilename.c_str(), "ab+"))) {
+  if (!(mDataFile.open(dataFilename))) {
```
ben-clayton: nit: superfluous brackets

core/cc/archive.cpp
hysw: Thanks.

PR #2189, opened 2018-09-10
Remove explicit write in doVkCmdFillBuffer.

AWoloszyn:
Treat this like any other GPU write.
This saves us a significant amount of performance since
we were tracking the writes byte-by-byte.


ben-clayton:
Can we at least have a bug to track the fact that none of these buffers have correct data?

PR #2191, opened 2018-09-11
Vulkan: Fix wrong descriptor set indices when binding descriptor sets

Qining:
This fixes the wrong calculation of descriptor set indices when binding
descriptor sets.

Also, for multi-layer maps (e.g. `map!(u32, map!(u32...`), we need to
explicitly assign the child map back to the parent map. This is because
the generated go/c++ code will not assign the child map back to the
parent map automactically (go: `Get()`, c++: `findOrZero`) when write
api file like: `parent[parent_key][child_key] = child_content`.

TODO: Similar bugs for other mutlti-layer maps

### DIFF
--- a/gapil/analysis/reference_value.go
+++ b/gapil/analysis/reference_value.go
@@ -171,6 +171,11 @@ func (v *ReferenceValue) Clone() Value {
 func (v *ReferenceValue) field(s *scope, name string) Value {
 	candidates := make([]Value, 0, len(v.Assignments))
 	for a := range v.Assignments {
+		// Check the assignment has an instance. No instance can happen when the
+		// assignment took place in a sibling block which has not been merged
+		// into the common scope yet.
+		// In this particular case, the instance is inaccessible to this block,
+		// so skipping it is the correct thing to do.
 		if fh, ok := s.getInstance(a).(fieldHolder); ok {
 			field := fh.field(s, name)
 			candidates = append(candidates, field)
@@ -186,6 +191,11 @@ func (v *ReferenceValue) field(s *scope, name string) Value {
 // in the scope's instances map.
 func (v *ReferenceValue) setField(s *scope, name string, val Value) Value {
 	for a := range v.Assignments {
+		// Check the assignment has an instance. No instance can happen when the
+		// assignment took place in a sibling block which has not been merged
+		// into the common scope yet.
+		// In this particular case, the instance is inaccessible to this block,
+		// so skipping it is the correct thing to do.
 		if fh, ok := s.getInstance(a).(fieldHolder); ok {
 			s.instances[a] = fh.setField(s, name, val)
 		}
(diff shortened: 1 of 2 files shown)


gapil/analysis/reference_value.go
```diff
@@ -171,8 +171,10 @@ func (v *ReferenceValue) Clone() Value {
 func (v *ReferenceValue) field(s *scope, name string) Value {
 	candidates := make([]Value, 0, len(v.Assignments))
 	for a := range v.Assignments {
-		field := s.getInstance(a).(fieldHolder).field(s, name)
-		candidates = append(candidates, field)
+		if fh, ok := s.getInstance(a).(fieldHolder); ok {
```
Qining: This is changed because of build failure with the original code.

The build fails with the original code if we drop the line 721 to line 745 in gapis/api/vulkan/api/descriptor.api, without any other changes. This is weird. 

The build does not happen before at commit: 7915ae323ed9dc1763b91088256ec9f996cd7728, which is the last commit that drops the line 721 to line 745 in gapis/api/vulkan/api/descriptor.api

gapil/analysis/reference_value.go
```diff
@@ -184,7 +186,9 @@ func (v *ReferenceValue) field(s *scope, name string) Value {
 // in the scope's instances map.
 func (v *ReferenceValue) setField(s *scope, name string, val Value) Value {
 	for a := range v.Assignments {
-		s.instances[a] = s.getInstance(a).(fieldHolder).setField(s, name, val)
+		if fh, ok := s.getInstance(a).(fieldHolder); ok {
```
Qining: Same problem as above

gapil/analysis/reference_value.go
ben-clayton: I'm going to try and reproduce this error - I don't think there should be a type panic here...

ben-clayton: Okay, after much beard scratching, I think I know what's going on here. See my comment below.

gapil/analysis/reference_value.go
ben-clayton: After much beard scratching, I think I know what the issue is. Your fix is likely fine, however please add the below comment so that some poor bugger doesn't have to work this out again.

I suggest adding this new utility function, and using it in the two places you've changed:

``` go

func (v *ReferenceValue) fhAssignments() []fieldHolder {
	out := make([]fieldHolder, 0, len(v.Assignments))
	for a := range v.Assignments {
		// Check the assignment has an instance. No instance can happen when the 
		// assignment took place in a sibling block which has not been merged 
		// into the common scope yet.
		// In this particular case, the instance is inaccessible to this block, 
		// so skipping it is the correct thing to do.
		if fh, ok := s.getInstance(a).(fieldHolder); ok {
			out = append(out, fh)
		}
	}
	return out
}
```

PR #2192, opened 2018-09-11
Drop the global lock around most of the Vulkan API commands

Qining:
Unlock before calling the underlying API command calls, and lock again after
return from the underlying API command calls.

This improves the tracing performance, especially when the application
builds Vulkan pipelines in multiple threads.

### DIFF
--- a/gapis/api/templates/api_spy.cpp.tmpl
+++ b/gapis/api/templates/api_spy.cpp.tmpl
@@ -229,8 +229,8 @@
           observer->observeTimestamp();
         }
 
-        {{if (GetAnnotation $ "blocking")}}
-          unlock();
+        {{if (GetAnnotation $ "threadsafe")}}
+        unlock();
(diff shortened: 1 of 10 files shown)


AWoloszyn:
We have looked at this on the Vulkan side at least, and I think? everything is probably in good shape, but I agree just making the change across every call may be dangerous.

Qining, what about changing the annotation from @blocking to @threadsafe (or something).
Since we know pipeline creation is one of the big ones, that can certainly help.

Qining:
Annotation: "threadsafe" has been added to some of the Vulkan commands. Now we use "threadsafe" instead of "blocking" to control the unlock and relock behavior.

ben-clayton:
*shrug* if you folks are sure that the rest of gapii is safe with this, then let's give it a try. I'm sure random crashes will alert us to any issues :)

PR #2196, opened 2018-09-11
Fixes and improvements to core/codegen

ben-clayton:


core/codegen/builder.go
```diff
@@ -66,7 +66,51 @@ func (b *Builder) CallIndirect(f *Value, args ...*Value) *Value {
...
+	if sig.Result == b.m.Types.Void {
+		name = ""
+	}
+	return b.val(sig.Result, b.llvm.CreateCall(fn, l, name))
+}
+
+// Invoke invokes the function f with the specified arguments.
+// If an exception is thrown while calling f, then cleanup will be called before
+// rethrowing the exception.
+func (b *Builder) Invoke(f *Function, cleanup func(), args ...*Value) *Value {
```
AWoloszyn: Are we using exceptions here to just plumb through errors? 
I.e. so we don't have to return true/false from every method (slice access etc).

AWoloszyn:
It looks like the MacOS bug is a flake. Unrelated to this MR.

core/codegen/builder.go
ben-clayton: Yes - these are for `abort`s and `assert`s. The exceptions here are 'zero cost', and increse performance by about 10% - 15% based on my measurements.

ben-clayton:
Thank you.

PR #2199, opened 2018-09-11
Vulkan dependency graph

bjoeris:
This is a new PR for the dependency graph code, updated for all the Gapid changes in the last few months.

The dependency graph code is turned off by default, and generally shouldn't change Gapid's execution. The major exception is that the Go version of the generated Mutate() functions for Vulkan now supports a "watcher", which can enables the dependency graph builder to hook into all the reads/writes. This watcher is generally nil, but nevertheless seems to be causing performance regressions that I'm currently investigating.

### DIFF
--- a/gapis/api/cmd.go
+++ b/gapis/api/cmd.go
@@ -58,7 +58,7 @@ type Cmd interface {
 
 	// Mutate mutates the State using the command. If the builder argument is
 	// not nil then it will call the replay function on the builder.
-	Mutate(context.Context, CmdID, *GlobalState, *builder.Builder) error
+	Mutate(context.Context, CmdID, *GlobalState, *builder.Builder, StateWatcher) error
 
 	// Clone makes a shallow copy of this command.
 	Clone(arena.Arena) Cmd
--- a/gapis/api/cmd.go
+++ b/gapis/api/cmd.go
@@ -62,6 +62,9 @@ type Cmd interface {
 
 	// Clone makes a shallow copy of this command.
 	Clone(arena.Arena) Cmd
+
+	// Alive returns true if this command should be marked alive for DCE
+	Alive() bool
 }
 
 const (
--- a/gapis/api/watcher.go
+++ b/gapis/api/watcher.go
@@ -1,0 +1,114 @@
+// Copyright (C) 2018 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package api
+
+import (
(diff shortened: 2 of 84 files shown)


gapis/resolve/dependencygraph2/dependency_graph.go
```diff
@@ -0,0 +1,426 @@
...
+			return obsNodeIDs.writeNodeIDStart + NodeID(index)
+		} else {
+			return obsNodeIDs.readNodeIDStart + NodeID(index)
+		}
+	}
+	return NodeNoID
+}
+
+// ForeachCmd iterates over all API calls in the graph
+func (g *dependencyGraph) ForeachCmd(ctx context.Context, cb func(context.Context, api.CmdID, api.Cmd) error) error {
```
ben-clayton: Why can't we use the `ForeachCmd` in `gapis/api`?

ben-clayton: Why can't we use the `ForeachCmd` in `gapis/api`?

gapis/api/cmd.go
```diff
@@ -58,6 +59,9 @@ type Cmd interface {
 	// Mutate mutates the State using the command. If the builder argument is
 	// not nil then it will call the replay function on the builder.
 	Mutate(context.Context, CmdID, *GlobalState, *builder.Builder) error
+
+	// Clone makes a shallow copy of this command.
+	CloneCmd(arena.Arena) Cmd
```
ben-clayton: In another branch I have a `clone()` that returns the concrete command type, and a `Clone()` that returns an `api.Cmd`. Would be nice if this matched.

gapis/api/watcher.go
```diff
@@ -0,0 +1,114 @@
...
+// StateWatcher provides callbacks to track state effects
+type StateWatcher interface {
+	// OnBeginCmd is called at the beginning of each API call
+	OnBeginCmd(ctx context.Context, cmdID CmdID, cmd Cmd)
+
+	// OnEndCmd is called at the end of each API call
+	OnEndCmd(ctx context.Context, cmdID CmdID, cmd Cmd)
+
+	// OnGet is called when a fragment of state (field, map key, array index) is read
+	OnGet(ctx context.Context, owner Reference, f Fragment, v Reference)
```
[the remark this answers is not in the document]
bjoeris: Yes, I'll fix this and check that each CL compiles.

bjoeris:
I have created separate PRs for the independent .api changes

gapis/resolve/dependencygraph2/dependency_graph.go
bjoeris: This version loops over both the initial commands as well as the captured commands, and assigns a derived CmdID to the initial commands.

I can refactor the implementation of dependencyGraph.ForeachCmd to use api.ForeachCmd, rather than duplicating that logic here.

I'll also make the comment more explicit.

gapis/api/watcher.go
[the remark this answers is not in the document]
bjoeris: I have verified that each CL compiles.

bjoeris:
Everything up to 0eb7422c0 is now in separate PRs. The remaining CLs are all directly related to the dependency graph.

bjoeris:
- Rebased with no conflict
- `bazel test tests` all pass

bjoeris:
Edited author date to fix github's ordering of the commits.

PR #2201, opened 2018-09-12
Yet more compiler improvements and fixes

ben-clayton:


core/codegen/debug.go
```diff
@@ -0,0 +1,276 @@
...
+func (d *dbg) builder() *llvm.DIBuilder {
+	if d.llvm != nil {
+		return d.llvm
+	}
+
+	b := llvm.NewDIBuilder(d.m.llvm)
+	d.cu = b.CreateCompileUnit(llvm.DICompileUnit{
+		Language:       0x0001, // DW_LANG_C,
+		File:           "apis",
+		Dir:            "/ssd/src/gapid", // TODO: HACK
```
AWoloszyn: do we want this here?

core/codegen/module.go
```diff
@@ -98,6 +103,13 @@ func NewModule(name string, target *device.ABI) *Module {
 	m.Types.emptyStructField = Field{"empty-struct", m.Types.Int8}
 	m.Types.m = m
 
+	if true {
```
AWoloszyn: if true { ? 

Can we get a comment as to why?

ben-clayton: Good spot - bad merge!

core/codegen/debug.go
ben-clayton: Debug generation shouldn't be enabled (except you spotted it is below). I haven't figured out a nice way to get the project root here due to bazel complications. I'd like to revisit this once I have performance in the bag.

core/codegen/module.go
ben-clayton: Fixed.

ben-clayton:
Switch away from my LLVM fork, now that https://github.com/llvm-mirror/llvm/commit/e562960fe303c0ffab6f3458fcdb1544b56fd81e has been merged.

PR #2203, opened 2018-09-12
Only take the slow-path in MutationCmdsFor if we really have to.

AWoloszyn:
This saves us a bunch of performance since for the most part
we never have to hit the slow-path. This is especially
true for GLES.


ben-clayton:
`gofmt` is moaning about your pointer placements, aside from that LGTM.

PR #2205, opened 2018-09-12
MEC: fix reading of color renderbuffers.

pmuetschard:
The read pixels path requires the framebuffer to be bound to GL_FRAMEBUFFER, binding to GL_READ_FRAMEBUFFER doesn't work. This appears to be a driver bug.

### DIFF
--- a/gapii/cc/gles_mid_execution.cpp
+++ b/gapii/cc/gles_mid_execution.cpp
@@ -647,7 +647,7 @@ ImageData Reader::ReadPixels(GLsizei w, GLsizei h) {
   img.width = w;
   img.height = h;
   img.data.reset(new std::vector<uint8_t>(size));
-  GAPID_DEBUG("ReadPixels: Reading %d bytes", size);
+  GAPID_DEBUG("ReadPixels: Reading %dx%d at %d bytes", w, h, size);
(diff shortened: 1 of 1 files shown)


gapii/cc/gles_mid_execution.cpp
```diff
@@ -975,8 +975,8 @@ ImageData Reader::ReadRenderbuffer(Renderbuffer* rb) {
       "MEC: Reading renderbuffer %d format 0x%x type 0x%x sized 0x%x at 0x%x",
       rb->mID, img->mDataFormat, img->mDataType, format, attach);
   if (attach == GL_COLOR_ATTACHMENT0) {
-    auto readFb = CreateAndBindFramebuffer(GL_READ_FRAMEBUFFER);
-    imports.glFramebufferRenderbuffer(GL_READ_FRAMEBUFFER, GL_COLOR_ATTACHMENT0,
+    auto readFb = CreateAndBindFramebuffer(GL_FRAMEBUFFER);
```
ben-clayton: Please add a comment explaining that this is a workaround for a potential driver bug. It looks like this is a mistake with your code, and someone might try to fix it.

pmuetschard: Done.

PR #2207, opened 2018-09-13
GLES ODR: Support reading depth buffers

pmuetschard:
Adds support for reading the depth buffer when doing on device (GLES) replays. This only works for GLES 3.0+ devices, because of the need for `glBlitFramebuffer`. Note that the shader is written in `#version 300 es` as `#version 100` defines `gl_FragColor` as `mediump` and would cause banding.


gapis/api/gles/read_framebuffer.go
```diff
@@ -305,12 +313,16 @@ func postFBData(ctx context.Context,
 
 // getReadPixelsFormat returns a unsized-format and type that is compatible with
 // glReadPixels() for the given framebuffer unsized-format and type.
+// The third return value is a unsized-format "to pretend" the data is and is
```
ben-clayton: `the data is`  what?

gapis/api/gles/tweaker.go
```diff
@@ -91,6 +91,14 @@ func (t *tweaker) glDisable(ctx context.Context, name GLenum) {
 	}
 }
 
+func (t *tweaker) glViewport(ctx context.Context, x, y GLint, w, h GLsizei) {
+	if o := t.c.Rasterization().Viewport(); o.X() != x || o.Y() != y || o.Width() != w || o.Height() != h {
```
ben-clayton: Nit: You could use [`Rect.EqualTo`](https://github.com/google/gapid/blob/master/gapis/api/gles/math.go#L18)

cmd/gapit/screenshot.go
```diff
@@ -214,15 +215,19 @@ func rescaleBytes(ctx context.Context, data []byte, max int) {
 }
 
 func (verb *screenshotVerb) getAttachment(ctx context.Context) (api.FramebufferAttachment, error) {
-	switch verb.Attachment {
-	case 0:
+	switch strings.ToLower(verb.Attachment) {
+	case "", "color", "color0", "0":
```
ben-clayton: Nit: Given the new `d` and `s`, I was expecting to see `c` here too.

PR #2213, opened 2018-09-14
gapii: Don't encode commands pre MEC

ben-clayton:


AWoloszyn:
Great! This alone pulls out up 15-20% of our pre-MEC frame time.

PR #2220, opened 2018-09-14
Be explicit about the ldscript location in the linkopts.

### LINKED ISSUE #2214: Bazel build fails with 0.17.1.
```
def cc_dynamic_library(name, exports = "", visibility = ["//visibility:private"], deps = [], linkopts = [], **kwargs):
   _symbol_exports(name, exports)

   # All but one of these will fail, but the select in the filegroup
   # will pick up the correct one.
   cc_stripped_binary(
       name = name + ".so",
       srcs = [":" + name + "_syms"],
       deps = deps + [name + ".ldscript"],
       linkopts = linkopts + ["-Wl,--version-script", name + ".ldscript"], << Cannot find file libVkLayer_VirtualSwapchain.so.ldscript
       linkshared = 1,
       visibility = ["//visibility:private"],
       **kwargs
   )
```
Even though it depends on  `deps = deps + [name + ".ldscript"]`
And the file exists

pmuetschard:
This is to fix building with bazel 0.17, where labels are no longer automatically expanded.

Fixes #2214

### DIFF
--- a/tools/build/rules/dynlib.bzl
+++ b/tools/build/rules/dynlib.bzl
@@ -55,7 +55,7 @@ def cc_dynamic_library(name, exports = "", visibility = ["//visibility:private"]
         name = name + ".so",
         srcs = [":" + name + "_syms"],
         deps = deps + [name + ".ldscript"],
-        linkopts = linkopts + ["-Wl,--version-script", name + ".ldscript"],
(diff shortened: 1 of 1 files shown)


AWoloszyn:
Thanks!!

PR #2232, opened 2018-09-19
Fix tests broken by #2186

ben-clayton:


ben-clayton:
Changes to `capture.go` were suggestions from https://github.com/google/gapid/pull/2186#discussion_r218521246

PR #2233, opened 2018-09-19
 Add framebuffer postback to export_replay.

hysw:
Usage:
Add `--outputframes=true` to `gapit export_replay` to include framebuffer postback in the exported trace.
Add `--postback-dir ${dir}` to `gapir` to store the framebuffer postback.

### DIFF
--- a/cmd/gapir/cc/main.cpp
+++ b/cmd/gapir/cc/main.cpp
@@ -207,12 +207,20 @@ struct Options {
   int logLevel = LOG_LEVEL;
   const char* logPath = "logs/gapir.log";
 
+  enum ReplayMode {
+    kUnknown = 0,    // Can't determine replay type from arguments yet.
+    kConflict,       // Impossible combination of command line arguments.
+    kReplayServer,   // Run gapir as a server.
+    kReplayArchive,  // Replay an exported archive.
+  };
+  ReplayMode mode = kUnknown;
   bool wait_for_debugger = false;
   const char* cachePath = nullptr;
   const char* portArgStr = "0";
   const char* authTokenFile = nullptr;
   int idleTimeoutSec = 0;
   const char* replayArchive = nullptr;
+  const char* postbackDirectory = "";
   bool version = false;
 
   static Options Parse(int argc, const char* argv[]) {
@@ -220,21 +228,31 @@ struct Options {
 
     for (int i = 1; i < argc; i++) {
       if (strcmp(argv[i], "--replay-archive") == 0) {
+        opts.SetMode(kReplayArchive);
         if (i + 1 >= argc) {
-          GAPID_FATAL("Usage: --replay-archive <archive-dir>");
+          GAPID_FATAL("Usage: --replay-archive <archive-directory>");
         }
         opts.replayArchive = argv[++i];
+      } else if (strcmp(argv[i], "--postback-dir") == 0) {
+        opts.SetMode(kReplayArchive);
+        if (i + 1 >= argc) {
+          GAPID_FATAL("Usage: --postback-dir <output-directory>");
+        }
+        opts.postbackDirectory = argv[++i];
       } else if (strcmp(argv[i], "--auth-token-file") == 0) {
+        opts.SetMode(kReplayServer);
         if (i + 1 >= argc) {
           GAPID_FATAL("Usage: --auth-token-file <token-string>");
         }
         opts.authTokenFile = argv[++i];
       } else if (strcmp(argv[i], "--cache") == 0) {
+        opts.SetMode(kReplayServer);
         if (i + 1 >= argc) {
           GAPID_FATAL("Usage: --cache <cache-directory>");
         }
         opts.cachePath = argv[++i];
       } else if (strcmp(argv[i], "--port") == 0) {
+        opts.SetMode(kReplayServer);
         if (i + 1 >= argc) {
           GAPID_FATAL("Usage: --port <port_num>");
         }
@@ -271,6 +289,7 @@ struct Options {
         }
         opts.logPath = argv[++i];
       } else if (strcmp(argv[i], "--idle-timeout-sec") == 0) {
+        opts.SetMode(kReplayServer);
         if (i + 1 >= argc) {
           GAPID_FATAL("Usage: --idle-timeout-sec <timeout in seconds>");
         }
@@ -285,6 +304,14 @@ struct Options {
     }
     return opts;
   }
+
+ private:
+  void SetMode(ReplayMode mode) {
+    if (this->mode != kUnknown && this->mode != mode) {
+      mode = kConflict;
+    }
+    this->mode = mode;
+  }
 };
 
 }  // namespace
@@ -295,7 +322,7 @@ static int replayArchive(Options opts) {
   GAPID_LOGGER_INIT(opts.logLevel, "gapir", opts.logPath);
   MemoryManager memoryManager(memorySizes);
   std::string payloadPath = std::string(opts.replayArchive) + "/payload.bin";
-  gapir::ReplayArchive conn(payloadPath);
+  gapir::ReplayArchive conn(payloadPath, opts.postbackDirectory);
   std::unique_ptr<ResourceProvider> resourceProvider =
       ResourceDiskCache::create(nullptr, opts.replayArchive);
   std::unique_ptr<Context> context = Context::create(
@@ -383,7 +410,10 @@ int main(int argc, const char* argv[]) {
   if (opts.version) {
     printf("GAPIR version " GAPID_VERSION_AND_BUILD "\n");
     return EXIT_SUCCESS;
-  } else if (opts.replayArchive) {
+  } else if (opts.mode == Options::kConflict) {
+    GAPID_ERROR("Argument conflicts.");
+    return EXIT_FAILURE;
+  } else if (opts.mode == Options::kReplayArchive) {
     return replayArchive(opts);
   } else {
     return startServer(opts);
--- a/gapir/cc/replay_archive.cpp
+++ b/gapir/cc/replay_archive.cpp
@@ -42,7 +42,21 @@ bool ReplayArchive::sendCrashDump(const std::string& filepath,
   GAPID_INFO("Crash dump saved at: %s", filepath.c_str());
   return true;
 }
-bool ReplayArchive::sendPostData(std::unique_ptr<Posts> posts) { return true; }
+bool ReplayArchive::sendPostData(std::unique_ptr<Posts> posts) {
+  if (mPostbackDir.empty()) {
+    return true;
+  }
+  std::unique_ptr<replay_service::PostData> postdata(posts->release_to_proto());
+  int nMessages = postdata->post_data_pieces_size();
+  for (int i = 0; i < nMessages; ++i) {
+    uint64_t id = postdata->post_data_pieces(i).id();
+    std::string data = postdata->post_data_pieces(i).data();
+    std::string path = mPostbackDir + "/" + std::to_string(id) + ".bin";
+    std::fstream output(path, std::ios::out | std::ios::binary);
+    output.write(data.data(), data.size());
+  }
+  return true;
+}
 bool ReplayArchive::sendNotification(uint64_t id, uint32_t severity,
                                      uint32_t api_index, uint64_t label,
                                      const std::string& msg, const void* data,
--- a/gapis/api/vulkan/replay.go
+++ b/gapis/api/vulkan/replay.go
@@ -502,10 +502,6 @@ type issuesRequest struct {
 	displayToSurface bool
 }
 
-// exportReplayRequest requests full trace to be produced.
-type exportReplayRequest struct {
-}
-
 func (a API) Replay(
 	ctx context.Context,
 	intent replay.Intent,
@@ -611,18 +607,6 @@ func (a API) Replay(
 			if req.displayToSurface {
 				doDisplayToSurface = true
 			}
-
-		case exportReplayRequest:
-			// TODO: Implement a transform specifically for export replay
-			if issues == nil {
-				n, err := expandCommands(false)
-				if err != nil {
-					return err
-				}
-				issues = newFindIssues(ctx, c, n)
-			}
-			issues.reportTo(rr.Result)
-			optimize = false
 
 		case framebufferRequest:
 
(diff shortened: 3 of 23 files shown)


cmd/gapir/cc/main.cpp
```diff
@@ -285,6 +305,14 @@ struct Options {
     }
     return opts;
   }
+
+ private:
+  void SetMode(ReplayMode mode) {
+    if (this->mode != kUnknown && this->mode != mode) {
+      mode = kConflict;
```
Qining: Should there be a 'return' after this?

gapir/cc/replay_archive.cpp
```diff
@@ -42,7 +42,21 @@ bool ReplayArchive::sendCrashDump(const std::string& filepath,
...
+bool ReplayArchive::sendPostData(std::unique_ptr<Posts> posts) {
+  if (mPostbackDir.empty()) {
+    return true;
+  }
+  std::unique_ptr<replay_service::PostData> postdata(posts->release_to_proto());
+  int nMessages = postdata->post_data_pieces_size();
+  for (int i = 0; i < nMessages; ++i) {
+    uint64_t id = postdata->post_data_pieces(i).id();
+    std::string data = postdata->post_data_pieces(i).data();
+    std::string path = mPostbackDir + "/" + std::to_string(id) + ".bin";
```
Qining: std::filesystem::path might be a better option?

cmd/gapir/cc/main.cpp
hysw: `mode` is public, so we don't need to return anything.

cmd/gapir/cc/main.cpp
```diff
@@ -383,7 +411,10 @@ int main(int argc, const char* argv[]) {
   if (opts.version) {
     printf("GAPIR version " GAPID_VERSION_AND_BUILD "\n");
     return EXIT_SUCCESS;
-  } else if (opts.replayArchive) {
+  } else if (opts.mode == Options::kConflict) {
+    GAPID_INFO("Argument conflicts.");
```
ben-clayton: `GAPID_ERROR`?

cmd/gapir/cc/main.cpp
```diff
@@ -207,34 +207,52 @@ struct Options {
   int logLevel = LOG_LEVEL;
   const char* logPath = "logs/gapir.log";
 
+  enum ReplayMode {
```
ben-clayton: Please can we have some comments on what these enum values represent?

gapis/replay/export_replay.go
```diff
@@ -86,43 +78,142 @@ func ExportReplay(ctx context.Context, pCapture *path.Capture, pDevice *path.Dev
...
+	if err != nil {
+		return err
+	}
+
+	mgr := &exportManager{}
+	for _, a := range c.APIs {
+		intent := Intent{pDevice, pCapture}
+		if a, ok := a.(QueryIssues); ok {
+			// If there are no frame query, force an issue query.
+			// Since otherwise the trace will not be generated.
```
ben-clayton: Is this a hack to generate a no-transform replay of the capture?
If so, please can we have a TODO to add a simple-replay mode here? GLES heavily transforms the capture for QueryIssues.

gapis/replay/export_replay.go
```diff
@@ -86,43 +78,142 @@ func ExportReplay(ctx context.Context, pCapture *path.Capture, pDevice *path.Dev
...
+}
+
+type ExportReplayFramebufferRequest struct {
+	Request          *service.GetFramebufferAttachmentRequest
+	Width, Height    uint32
+	FramebufferIndex uint32
+}
+
+// ExportReplay write replay commands and assets to path.
+func ExportReplay(ctx context.Context, pCapture *path.Capture, pDevice *path.Device, outDir string, fbs []ExportReplayFramebufferRequest) error {
```
ben-clayton: Instead of having the caller build and pass in `ExportReplayFramebufferRequest`s, how about:

```go

exporter := replay.Exporter() // Returns an instance of your new replay.Manager 

// Use the existing API methods for building replays. The code in gapis/resolve/framebuffer_attachment_data.go could be changed to support both cases.
query, ok := api.(replay.QueryFramebufferAttachment)

for _, fbr := range FramebufferAttachmentRequest {
	// Call the existing API's QueryXXX functions to add replay requests to the exporter.
	_, err := query.QueryFramebufferAttachment(ctx, intent, exporter, ...)
}

// Call mgr.generate, spit out the files...
err := exporter.Export(outDir)
```

Your `Exporter.Export` function then doesn't have to do all the translation to Query calls itself, and you'll automatically support any future Query functions.
Note in `Export` you'll probably want to check that all requests share the same `replay.Config`, otherwise you're going to be building multiple replay batches (which maybe something it would helpful to handle in the future?).

gapis/replay/manager.go
```diff
@@ -37,9 +37,24 @@ const (
 	defaultBatchDelay    = time.Millisecond * 100
 )
 
+type Manager interface {
```
ben-clayton: Please can we have some docs on what a `Manager` now is?

gapis/replay/export_replay.go
hysw: Essentially yes, however I have not find a place in GAPID that a non-transformed trace is used.

gapis/replay/export_replay.go
hysw: That's a much cleaner solution. Thanks.

cmd/gapir/cc/main.cpp
```diff
@@ -271,11 +289,13 @@ struct Options {
         }
         opts.logPath = argv[++i];
       } else if (strcmp(argv[i], "--idle-timeout-sec") == 0) {
+        opts.SetMode(kReplayServer);
         if (i + 1 >= argc) {
           GAPID_FATAL("Usage: --idle-timeout-sec <timeout in seconds>");
         }
         opts.idleTimeoutSec = atoi(argv[++i]);
       } else if (strcmp(argv[i], "--wait-for-debugger") == 0) {
+        opts.SetMode(kReplayServer);
```
AWoloszyn: We probably want to allow --wait-for-debugger with any other flags.

cmd/gapir/cc/main.cpp
```diff
@@ -207,34 +207,52 @@ struct Options {
   int logLevel = LOG_LEVEL;
   const char* logPath = "logs/gapir.log";
 
+  enum ReplayMode {
+    kUnknown = 0,    // Can't determine replay type from argumets yet.
```
ben-clayton: nit `arguments`

gapis/api/vulkan/replay.go
```diff
@@ -783,13 +767,16 @@ func (a API) QueryFramebufferAttachment(
 	if err != nil {
 		return nil, err
 	}
+	if _, ok := mgr.(replay.Exporter); ok {
```
ben-clayton: This is nasty. Can't we just fake the POST data (say with 0's) in the exporter so the API's don't need to care about the manager implementation?

gapis/replay/export_replay.go
```diff
@@ -16,65 +16,63 @@ package replay
...
 	"github.com/google/gapid/gapis/service/path"
 )
 
-// ExportReplay write replay commands and assets to path.
-func ExportReplay(ctx context.Context, pCapture *path.Capture, pDevice *path.Device, outDir string) error {
-	if pDevice == nil {
-		return log.Errf(ctx, nil, "Unable to produce replay on unknown device.")
-	}
+// Exporter stores the input replays and export them as gapir instruction.
+// TODO(hysw): Exporter.Replay have a different behaivor than Manager.Replay.
```
ben-clayton: See my comment above.

gapis/api/vulkan/replay.go
hysw: It's not really possible as the exporter need to be aware of the type of the POST data. 
It's possible to ignore type check as
```
data, _ :=res.(*image.Data)
return data, nil
```
however it only hide the problem.

ben-clayton: What do you mean by type? I don't think there's anything in the postback handlers that will explode if the buffer returned contained unexpected (0) data. In the case it did check, it should just return an error once parsed, which will be ignored by your new manager anyway.

What I'm suggesting is:

Postbacks are requested by the API using the [`replay.Builder.Post()`](https://github.com/google/gapid/blob/master/gapis/replay/builder/builder.go#L474) function.
This [appends a `postBackDecoder`](https://github.com/google/gapid/blob/master/gapis/replay/builder/builder.go#L482) to the [`replay.Builder.decoders` slice](https://github.com/google/gapid/blob/master/gapis/replay/builder/builder.go#L89).
If you added a new method to `replay.Builder` to return the list of expected messages, you could call [the returned `PostDataHandler`](https://github.com/google/gapid/blob/master/gapis/replay/builder/builder.go#L634-L656) with 0 data for each of the expected postbacks.

gapis/replay/builder/builder.go
```diff
@@ -556,6 +556,28 @@ func (b *Builder) RegisterNotificationReader(reader NotificationReader) {
...
+
+	payload, _, _, err := b.Build(ctx)
+	if err != nil {
+		return payload, err
+	}
+
+	// Send phony postbacks.
+	byteOrder := b.memoryLayout.GetEndian()
+	for _, decoder := range b.decoders {
+		r := endian.Reader(bytes.NewReader([]byte{}), byteOrder)
```
ben-clayton: Nit: might be preferable to pretend there's at least the expected number of bytes returned. How about instead of `bytes.NewReader`, you use a dummy zero reader:

```go
type zeroReader struct{}

func (z zeroReader) Read(b []byte) (n int, err error) {
	for i := range b {
		b[i] = 0
	}
	return len(b), nil
}
```

gapis/replay/export_replay.go
```diff
@@ -16,65 +16,66 @@ package replay
...
 
-// ExportReplay write replay commands and assets to path.
-func ExportReplay(ctx context.Context, pCapture *path.Capture, pDevice *path.Device, outDir string) error {
-	if pDevice == nil {
-		return log.Errf(ctx, nil, "Unable to produce replay on unknown device.")
-	}
+// Exporter stores the input replays and export them as gapir instruction.
+type Exporter interface {
+	Manager
+	Export(ctx context.Context, waitRequests int) (*gapir.Payload, error)
```
ben-clayton: Please document the `Export` function.

gapis/server/export_replay.go
```diff
@@ -0,0 +1,156 @@
...
+	err = os.MkdirAll(out, os.ModePerm)
+	if err != nil {
+		return log.Errf(ctx, err, "Failed to create output directory: %v", out)
+	}
+
+	payloadBytes, err := proto.Marshal(payload)
+	if err != nil {
+		return log.Errf(ctx, err, "Failed to serialize replay payload.")
+	}
+	err = ioutil.WriteFile(gopath.Join(out, "payload.bin"), payloadBytes, 0644)
```
ben-clayton: Would be nicer if the server handed back the data to the client so that we don't have to worry about  all the `enableLocalFiles` stuff.

hysw: There's multiple files to be written: payload.bin, archive.{index,data} so it is not that easy to stream the data back. The code produce those the archive is in C land at core/cc/archive.*.

gapis/replay/builder/builder.go
hysw: I'm thinking about just passing it an error that says the replay were never actually executed, it should handle that case gracefully anyway.

ben-clayton: SGTM

gapis/server/export_replay.go
ben-clayton: Is all the gapir changes just so you can use the archive C++ library?
It does strike me as odd that we're writing files from two separate processes here.

ben-clayton: Hold on, I'm talking nonsense. Let me just check out this PR and figure out what the hell I'm talking about.

ben-clayton: Right, ignore the craziness I said above.

So we're writing out 3 files - `payload.bin` (this line) and `resources.index`, `resource.data` (below).
I don't really get why this couldn't just be put into a filename-data map which is returned, and handed back over gRPC.

However, I'm not suggesting making these changes right now - I just want to know if it is possible.

hysw: In that case, we either need to build all the files in memory (resource.data can be many GBs) or use a streaming rpc. And either way we need to duplicate archive.* code in Go somehow. So it is possible, however I don't see a compelling reason to build it and maintain it.

gapis/replay/builder/builder.go
```diff
@@ -556,6 +560,28 @@ func (b *Builder) RegisterNotificationReader(reader NotificationReader) {
...
+
+	payload, _, _, err := b.Build(ctx)
+	if err != nil {
+		return payload, err
+	}
+
+	// Send phony postbacks.
+	// byteOrder := b.memoryLayout.GetEndian()
+	for _, decoder := range b.decoders {
+		// r := endian.Reader(bytes.NewReader([]byte{}), byteOrder)
```
AWoloszyn: Remove commented out code.

gapis/replay/builder/builder.go
```diff
@@ -556,6 +560,28 @@ func (b *Builder) RegisterNotificationReader(reader NotificationReader) {
...
+	defer status.Finish(ctx)
+	ctx = log.Enter(ctx, "Export")
+
+	payload, _, _, err := b.Build(ctx)
+	if err != nil {
+		return payload, err
+	}
+
+	// Send phony postbacks.
+	// byteOrder := b.memoryLayout.GetEndian()
```
AWoloszyn: Same here

gapis/replay/export_replay.go
```diff
@@ -16,65 +16,69 @@ package replay
...
-// ExportReplay write replay commands and assets to path.
-func ExportReplay(ctx context.Context, pCapture *path.Capture, pDevice *path.Device, outDir string) error {
-	if pDevice == nil {
-		return log.Errf(ctx, nil, "Unable to produce replay on unknown device.")
-	}
+// Exporter stores the input replays and export them as gapir instruction.
+type Exporter interface {
+	Manager
+	// Export wait for waitRequests replay requests to be sent,
+	// it then compiles the instructions for replay and trigger
```
AWoloszyn: and triggers all postback

AWoloszyn:
Looks good to me. as long as the bots are happy.

PR #2238, opened 2018-09-21
Compiler upstream

ben-clayton:


gapil/runtime/cc/replay/dataex.h
```diff
@@ -19,6 +19,17 @@
 #include "core/cc/interval_list.h"
 
 #include <unordered_map>
+#include <vector>
+
+namespace gapil {
+namespace runtime {
+namespace replay {
+
+// align returns val aligned up the the next multiple of by.
```
AWoloszyn: Just as a note, if these are always power-of-2, this can be simplified

return (val + (by-1))&(by-1);

If this is being run a lot idiv is not friendly on Arm.

ben-clayton: Ack will optimize if I see it show up in a profile.

gapil/runtime/cc/runtime.h
```diff
@@ -102,18 +104,81 @@ typedef struct ref_t {
...
+  gapil_char* type_name;     // name of the type.
+  void (*reference)(void*);  // increment the reference count for the type.
+  void (*release)(void*);    // decrement the reference count for the type.
+} gapil_rtti;
+
+typedef struct gapil_any_t {
+  uint32_t ref_count;  // number of owners of this any.
+  arena* arena;        // arena that owns this any allocation.
+  gapil_rtti* rtti;    // type information of the value.
+  void* value;  // pointer to the value. For boxed value-types, this should be
```
AWoloszyn: Just for my own sake, does this mean any<float> would "effectively" be
gapil_any_t {
     uint32_t ref_count;
     arena* arena;
     gapil_rtti* rtti;
     void* value;   (= &hidden_value);
     float _hidden_value;
}

ben-clayton: Yes. I considered unions, but these are a nightmare to deal with in LLVM.
Given that they are infrequently used (both in the API and in typical numbers created), I've decided to leave this as-is for now. If the double indirection ever shows up in a profile, we can optimize.

ben-clayton:
Tested MEC and non-MEC for GLES and Vulkan.

ben-clayton:
Great, thanks.

PR #2239, opened 2018-09-21
Fix linearize_trace

Qining:
Broken by 5bf4a56


Qining:
This CL should not bring any differences about build. Offline build for Linux succeeded, submit it now.

PR #2243, opened 2018-09-24
Linux: Be more aggressive about keeping the memory tracker on.

AWoloszyn:


core/memory_tracker/cc/posix/memory_tracker.inc
```diff
@@ -22,13 +22,18 @@ namespace track_memory {
...
-      signal_handler_registered_(false),
       orig_action_{0},
       handle_segfault_(segfault_function) {
     }
 
 
-  bool IsInstalled() const { return signal_handler_registered_; }
+  bool IsInstalled() const {
+    struct sigaction orig_action {
+      0
```
Qining: Nit: does clang-format allow us to make it one line?

ben-clayton: I think it does if you change this to `sigaction orig_action = {0}`.

PR #2245, opened 2018-09-25
Replay tests for static array parameters

ben-clayton:


gapis/api/templates/mutate.go.tmpl
```diff
@@ -266,29 +266,6 @@
 {{end}}
 
 
-{{/*
```
AWoloszyn: Have we actually been not using this anywhere?

ben-clayton: Nope. There's a similar function in `api.go.tmpl` that I think you added after this was written.

PR #2246, opened 2018-09-25
Simplify gapir resource cache and replay connection code

Qining:
Prepare for the comming changes on resource caching.

* `ResourceProvider` is changed to `ResourceLoader` as an interface.
* `ResourceCache` is removed from `ResourceProvider (now
ResourceLoader)` inheritance tree, now as an individual interface.
* `CachedResourceLoader` is an implementation of `ResourceLoader` that
owns a `ResourceCache` instance.
* `ReplayConnection` is splitted into `ReplaySerivce` interface and
`GrpcReplayService` implementation, `ArchiveReplayService` is another
implementation of `ReplayService`.

TODO:
- [x] Unit test
- [x] Test with samples
  - [x] Android
  - [x] Desktop
- [x] Clear redundant `#include`
- [x] Update comments


gapir/cc/cached_resource_loader.h
```diff
@@ -0,0 +1,76 @@
...
+class CachedResourceLoader : public ResourceLoader {
+ public:
+  static std::unique_ptr<CachedResourceLoader> create(
+      ResourceCache* cache, std::unique_ptr<ResourceLoader> fallbackLoader) {
+    return std::unique_ptr<CachedResourceLoader>(
+        new CachedResourceLoader(cache, std::move(fallbackLoader)));
+  }
+  virtual ~CachedResourceLoader() {}
+
+  // Loads first tries to find the resources in cache and loads the data from
```
AWoloszyn: load first tries

AWoloszyn: loads the data from the cache if a resource

gapir/cc/cached_resource_loader.h
```diff
@@ -0,0 +1,76 @@
...
+  // cached if a resource is found. If a resource is not found, use the fallback
+  // loader to fetch the resource data, then load the data to target and put the
+  // data to cache. If the net size of all the resources exceeds size, then
+  // false is returned.
+  virtual bool load(const Resource* resources, size_t count, void* target,
```
AWoloszyn: // Load first tries to find the resources in cache and loads from there
  // if they are found. If a resource is not found it uses the fallback
  // loader to fetch the resource data, loads the data to target and puts the
  // data to cache. If the net size of all the resources exceeds targetSize, then
  // false is returned.

gapir/cc/grpc_replay_service.cpp
```diff
@@ -0,0 +1,121 @@
...
+  mGrpcStream->Write(res);
+  std::unique_ptr<replay_service::ReplayRequest> req(
+      new replay_service::ReplayRequest());
+  if (!mGrpcStream->Read(req.get())) {
+    return nullptr;
+  }
+  if (req->req_case() != replay_service::ReplayRequest::kPayload) {
+    return nullptr;
+  }
+  return std::unique_ptr<ReplayService::Payload>(new ReplayService::Payload(
```
AWoloszyn: This looks like a good place for make_unique, (if it actually exists in all of our compilers)

gapir/cc/resource_cache.cpp
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017 Google Inc.
+ * Copyright (C) 2018 Google Inc.
```
AWoloszyn: We tend to not change the copyright date on files.

gapir/cc/resource_loader.h
```diff
@@ -0,0 +1,101 @@
...
+
+  // Fetch queries the specified resources and returns a
+  // ReplayService::Resources instance which contains the resources data.
+  virtual std::unique_ptr<ReplayService::Resources> fetch(
+      const Resource* resources, size_t count) = 0;
+};
+
+// SimpleResourceLoader implements the ResourceLoader interface. It pull
+// resources from a ReplayService instance for every resource loading request.
+class SimpleResourceLoader : public ResourceLoader {
```
AWoloszyn: I would prefer a name other than "Simple". That doesn't tell us how it's simple.

PassthroughResourceLoader may be appropriate.

AWoloszyn:
Overall no real issues, a couple of comments.

PR #2249, opened 2018-09-26
 Bring on-disk cache alive for GAPIR on Linux

Qining:
Nothing should be changed for Windows and Android, neither on the GAPIS
side.

With `gapid` UI, use by feeding following cmd arguments to `--gapir-args "..."`

new command line argument for gapir:
* `--enable-disk-cache`: To turn on on-disk cache. For now this means
in-memory cache will not be used, later we may want to chain these two
caches up.
* `--disk-cache-path`: User defined directory for cache file. The final cache
files will be located at:
`<disk-cache-path>/resource.(index|data)`.
If no user defined path is given, $TMPDIR will be used, final cache file
directory will be: `<$TMPDIR>/gapir-cache.XXXXXX/resource.(index|data)`.
XXXXX is generated by `mkdtemp()`.
* `--cleanup-disk-cach`: If the `--disk-cache-path` is given, this flag
means the cache files will be deleted when GAPIR process ends.

obsolete argument for gapir:
* `--cache`: This argument is ambigous.

Cache files will be deleted after the GAPIR process dies. This is
guaranteed by a `fork()`ed process whose only job is monitoring if the
GAPIR process is alive, and deleting the files and `gapir-cache.XXXXXX`
folder once GAPIR is died.

Sits on top of #2246


cmd/gapir/cc/main.cpp
```diff
@@ -63,20 +67,84 @@ std::vector<uint32_t> memorySizes {
...
+    cacheFilesPath = getOnDiskCacheFilesPath(diskCachePath);
+    if (cacheFilesPath.size() == 0) {
+      GAPID_WARNING(
+          "No disk cache path specified and no $TMPDIR environment variable "
+          "defined, fallback to disable on-disk cache.");
+      enableDiskCache = false;
+    }
   }
+  if (enableDiskCache) {
+    if (fork() == 0) {
```
ben-clayton: [Something relevant I was reading this morning](http://www.linuxprogrammingblog.com/threads-and-fork-think-twice-before-using-them). GAPIR does use threads - so we might need to be careful about `fork` here.

AWoloszyn: I think this specific case we are ok. This is called from startServer() which happens before we start up GRPC. So I don't think we will have any extra threads running.

PR #2254, opened 2018-09-26
Logging Improvements

pmuetschard:
- if gapis is logging to a file, don't stop logging to the file when a `GetLogStream` RPC is in play
 - track GLES state reconstruction tasks via `core/app/status`
 - minor changes


gapir/cc/context.cpp
```diff
@@ -305,10 +305,10 @@ void Context::registerCallbacks(Interpreter* interpreter) {
         }
 
         if (stack->isValid()) {
-          GAPID_DEBUG("[%u]replayChangeBackbuffer(%d, %d, 0x%x, 0x%x, 0x%x)",
-                      label, backbuffer.width, backbuffer.height,
-                      backbuffer.format.color, backbuffer.format.depth,
-                      backbuffer.format.stencil);
+          GAPID_INFO("[%u]replayChangeBackbuffer(%d, %d, 0x%x, 0x%x, 0x%x)",
```
ben-clayton: Do you mean to bump this severity? Might affect performance if there's lots of changes?

pmuetschard: Yes I did. I did it to match the other synthetic commands, which are logging at INFO. Since every change backbuffer is (currently) followed by a bind renderer, which already logs to INFO, I didn't think it was an issue.

PR #2259, opened 2018-10-01
Allow `gapit screenshot` to create multiple screenshots in one go

pmuetschard:
Makes the `--at` and `--frame` flags repeatable and adds a `--draws` flag to request a screenshot of every draw call.

### DIFF
--- a/core/app/flags/flags.go
+++ b/core/app/flags/flags.go
@@ -141,42 +141,50 @@ func (s *Set) Bind(name string, value interface{}, help string) {
 		return
 	}
 	rv := reflect.ValueOf(value)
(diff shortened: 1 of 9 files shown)


core/app/flags/flags.go
```diff
@@ -176,7 +176,15 @@ func (s *Set) Bind(name string, value interface{}, help string) {
 		default:
 			fullname = name + "-" + fname
 		}
-		s.Bind(fullname, field.Addr().Interface(), usage)
+
+		if tags.Get("repeated") == "true" {
```
ben-clayton: Do we need the `repeated` tag? Isn't having the type a slice enough info?

pmuetschard: Done. Changed it so that slices are automatically repeated flags.

PR #2260, opened 2018-10-01
Various GLES state reconstruction fixes

pmuetschard:
Fixes to make the state comparison quieter:
 - Don't compare arenas
 - Don't compare RefIDs
 - Fix the slice comparison to actually work
 - Fix the recreation of programs to use the same IDs as during trace
 - Correctly record what extensions the trace device supported (fixing c++ bool-map default values)
 - Fix handling of resetting the scissor and viewport

### DIFF
--- a/gapis/api/gles/state_builder.go
+++ b/gapis/api/gles/state_builder.go
@@ -105,18 +105,20 @@ func (API) RebuildState(ctx context.Context, oldState *api.GlobalState) ([]api.C
 	// Verify that the recreated state matches the original desired state.
 	diffs := 0
 	compare.Compare(s, GetState(sb.newState), func(d compare.Path) {
-		last := d[len(d)-1]
-		if oldSlice, ok := last.Value.(memory.Slice); ok {
-			if newSlice, ok := last.Value.(memory.Slice); ok {
-				old := AsU8ˢ(sb.tmpArena, oldSlice, sb.oldState.MemoryLayout)
-				new := AsU8ˢ(sb.tmpArena, newSlice, sb.newState.MemoryLayout)
-				if old.ResourceID(ctx, sb.oldState) == new.ResourceID(ctx, sb.newState) {
-					return // The pool IDs are different, but the resource IDs match exactly.
-				}
-				oldData := old.MustRead(ctx, nil, sb.oldState, nil)
-				newData := new.MustRead(ctx, nil, sb.newState, nil)
-				if reflect.DeepEqual(oldData, newData) {
-					return // The pool IDs are different, but the actual data matches exactly.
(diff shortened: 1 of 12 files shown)


pmuetschard:
Removing the `ResetViewportScissor` field from the `DynamicContextState` extra has the following effect, when loading "old" trace files:
 - `PreserveBuffersOnSwap` becomes `true` even if it was `false` during trace time
 - `RedSize` becomes 0 or 1 depending on the actual `PreserveBuffersOnSwap`. `RedSize` is not used anywhere.
 - The remaining color/depth/stencil sizes are "sheared" (e.g `GreenSize` becomes `RedSize`, `BlueSize` becomes `GreenSize`, etc. `StencilSize` becomes 0). None of them are actually used for anything.

Please advice on what do about this.

gapis/api/gles/state_builder.go
```diff
@@ -105,18 +105,20 @@ func (API) RebuildState(ctx context.Context, oldState *api.GlobalState) ([]api.C
...
-				new := AsU8ˢ(sb.tmpArena, newSlice, sb.newState.MemoryLayout)
-				if old.ResourceID(ctx, sb.oldState) == new.ResourceID(ctx, sb.newState) {
-					return // The pool IDs are different, but the resource IDs match exactly.
-				}
-				oldData := old.MustRead(ctx, nil, sb.oldState, nil)
-				newData := new.MustRead(ctx, nil, sb.newState, nil)
-				if reflect.DeepEqual(oldData, newData) {
-					return // The pool IDs are different, but the actual data matches exactly.
+		if l := len(d); l > 1 {
+			last := d[l-2]
```
ben-clayton: How come -2?

gapis/api/reference.go
```diff
@@ -45,3 +47,9 @@ type NilReference struct{}
 func (NilReference) RefID() RefID {
 	return NilRefID
 }
+
+func init() {
+	// Don't compare RefIDs.
+	compare.Register(func(c compare.Comparator, a, b RefID) {
```
ben-clayton: No need to change now, but we should probably avoid creating these global comparison functions - I've found they hid differences that you might want to find under different circumstances. The compare package needs a little bit of work to expose this, but we should probably be passing the custom rules down with each compare.

gapis/api/gles/state_builder.go
pmuetschard: The last element is the field in the struct that is different (usually the pool id), whereas the penultimate is the actual slice struct itself. I've added a comment.

gapis/api/reference.go
pmuetschard: Agreed.

PR #2262, opened 2018-10-02
Show a warning in the UI if the adb path is missing/invalid

### LINKED ISSUE #2055: Feature request: check the ADB path specified
I was a bit confused, because on first run I specified the path to the folder containing adb.exe (this is what other tools have trained me to do, specify path to SDK folders etc). 

Before I realized my mistake (and specified the path including the "adb.exe") the app simply didn't show any devices - no other error.

Could the preferences check that the user has specified something that looks a valid adb.exe, not (for example) a folder?

pmuetschard:
Fixes #2055


gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -368,6 +370,15 @@ protected void configureDialog(DirectoryDialog dialog) {
         pcsWarning.setForeground(getDisplay().getSystemColor(SWT.COLOR_DARK_YELLOW));
         pcsWarning.setVisible(!models.settings.traceDisablePcs);
 
+        createLabel(this, "");
+        Link adbWarning = withLayoutData(
+            createLink(this,
+                "Path to adb invalid/missing. Please fix it in the <a>preferences</a>.",
```
ben-clayton: This suggests that adb is required, when for desktop usage it is not. Maybe the message should be something like:
`Path to adb invalid or missing, Android support disabled`

AWoloszyn: Maybe we could put it in the device dropdown?
That way it only shows up when you try to do anything Android. (Not sure if its easy to jam stuff like this into the device dropdown)

pmuetschard: I've reworded the message.

PR #2263, opened 2018-10-02
gapir: log a message every 10,000 api calls.

pmuetschard:


pmuetschard:
This has been helpful for me to know that a long replay is still happening. But, I understand if you don't want this in.

ben-clayton:
I wouldn't mind this if the client wasn't so slow handling log messages. Would every 100,000 be too infrequent?

pau-baiget:
Closing this one as we're not spending time on this for now. We can resurrect this if needed.

PR #2264, opened 2018-10-02
Don't serialize the initial state in the replay log.

pmuetschard:
The replay log will already contain the commands to recreate the state, including the initial state as well, makes the resulting trace invalid.


ben-clayton:
Added @bjoeris as [he recently added this code](https://github.com/google/gapid/pull/2231).

PR #2267, opened 2018-10-04
Fix repeated flags causing `-help` to panic.

pmuetschard:


pmuetschard:
This should be cherry picked into the release. Sorry about the regression.

AWoloszyn:
I am going to merge this one early, so that the bots can get a build out.

PR #2275, opened 2018-10-09
Added a trailing frame-time measurement to the benchmark.

AWoloszyn:
This allows us to separate loading-time from run-time
in trace-time benchmarking.


AWoloszyn:
I have merged this, if there are any comments, I will happily re-work parts of it. I just needed this in for testing.

PR #2280, opened 2018-10-11
Enable on-disk cache by default on remote-ssh devices.

AWoloszyn:
This improves performance when our in-memory cache is not
large enough.

### DIFF
--- a/gapir/client/session.go
+++ b/gapir/client/session.go
@@ -99,11 +99,26 @@ func (s *session) newRemote(ctx context.Context, d remotessh.Device, abi *device
 		return err
 	}
 
+	forceEnableDiskCache := true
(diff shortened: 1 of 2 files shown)


gapir/client/session.go
```diff
@@ -99,11 +99,25 @@ func (s *session) newRemote(ctx context.Context, d remotessh.Device, abi *device
 		return err
 	}
 
+	forceEnableDiskCache := true
+
+	// Add disk cache
```
ben-clayton: This comment is confusing as the block only might set `forceEnableDiskCache` to false.

AWoloszyn: Will elaborate.

PR #2292, opened 2018-10-16
External Images Fixes and Support for On Device Replay

### LINKED ISSUE #2279: Empty GL_TEXTURE_EXTERNAL_OES image data in v1.2.0
Version information: GAPID v1.2.0 on macOS 10.14

When capturing an app which uses texture from camera, the captured frame is always black.  It seems that camera texture GL_TEXTURE_EXTERNAL_OES image data is empty.

GAPID v1.1.0 is working fine.

"Texture from Camera" in Grafika can be used to reproduce the problem. 

Grafika Apk link:
https://github.com/google/grafika/releases/download/v1.0-33/app-debug.apk

pmuetschard:
Fixes a dependency graph bug for external images backed by GL textures.
Adds support for external images in ODR:

Since we don't fudge the shaders for ODR, shaders using external samplers will expect external textures. Therefore, we cannot use the desktop handling of external textures, which coerces them to regular textures. Instead, we need to create EGLImages and map them to the external textures, the same as the application did. However, since we won't have the native buffers anymore, we convert all EGLImages to be backed by GL textures. Since no context needs to be bound when creating an EGLImage, but we need one to create backing texture, the creation of the texture and EGLImage is delayed until the sync point, where a context is bound.

Introduces a new synthetic replay command to create an external image. This is needed, because the `eglCreateImageKHR` needs the display and context pointer, which are not exposed to gapis. Also adds a custom mutate to `eglCreateImageKHR` command, which augments the mutate by replacing the command with the synthetic one on replay.

Fixes #2279

### DIFF
--- a/core/vulkan/vk_virtual_swapchain/cc/BUILD.bazel
+++ b/core/vulkan/vk_virtual_swapchain/cc/BUILD.bazel
@@ -37,7 +37,7 @@ cc_library(
         # Android
         "//conditions:default": [
             "-ldl",
-            "-lm"
+            "-lm",
         ],
     }),
     visibility = ["//visibility:public"],
--- a/gapir/cc/android/gles_renderer.cpp
+++ b/gapir/cc/android/gles_renderer.cpp
@@ -21,6 +21,7 @@
 #include "core/cc/log.h"
 
 #include <EGL/egl.h>
+#include <EGL/eglext.h>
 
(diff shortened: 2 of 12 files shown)


pmuetschard:
I pushed another commit to fix the MEC state reconstruction crash of EGL images.

PR #2301, opened 2018-10-17
Allow filtering `gapit devices` by OS kind.

pmuetschard:


core/os/device/os.go
```diff
@@ -41,3 +41,7 @@ func (o1 *OS) CompareVersions(o2 *OS) VersionMatch {
 		return CompleteMatch
 	}
 }
+
+func (o *OSKind) Choose(c interface{}) {
```
AWoloszyn: Is this used anywhere?

pmuetschard: Yup, bu the flag parsing. Don't you love go's implicit interfaces?!

PR #2310, opened 2018-10-22
Minor fix for the temporary directory for on-disk cache

Qining:
If and only if the $TMPDIR env is not set should we try to fallback to
/tmp as the temporary directory for on-disk cache files.


hysw:
@googlebot rescan

PR #2317, opened 2018-10-25
Allow `gapit unpack` to dump objects to JSON.

pmuetschard:
Allows extracting of individual groups and objects into JSON to inspect/use externally. This is especially useful for extracting resource data, such as texture images, from a capture.

### DIFF
--- a/cmd/gapit/unpack.go
+++ b/cmd/gapit/unpack.go
@@ -16,8 +16,10 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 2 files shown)


cmd/gapit/unpack.go
```diff
@@ -111,3 +134,74 @@ func (u *unpacker) msgString(msg proto.Message) string {
...
+	ids      map[uint64]bool
+	children bool
+	dir      string
+	found    int
+	count    int
+}
+
+func (u *dumper) BeginGroup(ctx context.Context, msg proto.Message, id uint64) error {
+	if _, ok := u.ids[id]; ok {
+		u.found = 1
```
ben-clayton: I believe groups can overlap in the stream, in which case i think this approach may break.

AWoloszyn: This is correct, we have this in Vulkan in a few places where we have API calls that block. We mark the start and end completely separately. You may have to buffer the begin, until you see the end, then print it out all at once.

PR #2323, opened 2018-10-29
A couple changes to help with tracing performance.

AWoloszyn:


gapis/api/templates/api.go.tmpl
```diff
@@ -415,6 +415,21 @@ import (
     }
   }
 
+func (m {{$name}}) DestroyHandlesʷ(ϟctx context.Context, ϟw ϟapi.StateWatcher) {
+    if m.Map != nil {
+      // Figure this out
```
AWoloszyn: @bjoeris I am not 100% sure what these are supposed to look like.
Any help would be appreciated.

gapil/format/format.go
```diff
@@ -162,6 +162,8 @@ func (p *printer) markup(n ast.Node) {
 			p.inject(n.Fields[0], beforePrefix, "•»")
 			p.inject(n.Fields[c-1], beforeSuffix, "«•")
 		}
+	case *ast.Clear:
+		//p.inject(n.Value, afterPrefix, "•")
```
AWoloszyn: @ben-clayton This I was not sure about, I can't quite wrap my head around how the formatter works.

core/memory/arena/cc/arena.cpp
```diff
@@ -143,7 +143,6 @@ void* Arena::allocate(uint32_t size, uint32_t align) {
 
     // posix_memalign states that the alignment must be a multiple
     // of sizeof(void*).
-    align = std::max(align, static_cast<uint32_t>(sizeof(void*)));
     align = std::max(align, page_size_);
```
ben-clayton: You mean to remove the alignment statement here? You still have the comment. If you removed it because `sizeof(void*) < page_size_`, how about a `static_assert`?

core/memory/arena/cc/arena.cpp
```diff
@@ -184,12 +188,10 @@ void* Arena::allocate(uint32_t size, uint32_t align) {
       // The first allocation is offset by 1 block.
       allocation = data.current_chunk + block_size;
       data.offset_of_next_allocation_in_chunk = 2 * block_size;
+      header = reinterpret_cast<chunk_header*>(data.current_chunk);
     }
+    header->num_allocations += 1;
```
ben-clayton: It looks like the `num_allocations` are now always guarded by the lock - does it still need to be atomic?

gapil/analysis/analyze.go
```diff
@@ -210,6 +210,8 @@ func (s *scope) traverse(n semantic.Node) {
 	case *semantic.MapRemove:
 		// TODO: Put an entry that says the map did not contain the key.
 
+	case *semantic.MapClear:
```
ben-clayton: We should clear the map's entries here - there should be plenty of example test code to add a new check for this.
At the very least, please add a TODO.

gapil/resolver/statement.go
```diff
@@ -172,6 +174,15 @@ func delete_(rv *resolver, in *ast.Delete) *semantic.MapRemove {
 	return &semantic.MapRemove{AST: in, Type: mty, Map: m, Key: k}
 }
 
+func clear_(rv *resolver, in *ast.Clear) *semantic.MapClear {
+	m := expression(rv, in.Map)
+	mty, ok := m.ExpressionType().(*semantic.Map)
+	if !ok {
+		rv.errorf(in.Map, "clear's argument must be a map, got %s", typename(m.ExpressionType()))
+	}
+	return &semantic.MapClear{AST: in, Type: mty, Map: m}
```
ben-clayton: You are missing a mapping between the `ast` and `semantic` nodes here:

```golang
out := &semantic.MapClear{AST: in, Type: mty, Map: m}
rv.mappings.Add(in, out)
return out
```

Without this, you won't get debug info generated for this statement.

gapil/runtime/cc/map.inc
```diff
@@ -317,7 +317,7 @@ void Map<K, V, DENSE>::Allocation::remove(K key) {
 template<typename K, typename V, bool DENSE>
 void Map<K, V, DENSE>::Allocation::clear() {
     auto elems = els();
-    for (uint64_t i = 0; i < capacity; ++i) {
+    for (uint64_t i = 0; i < capacity && count > 0; ++i) {
         switch(elems[i].used) {
```
ben-clayton: Bug not introduced in these changes, but shouldn't we be setting the `used` flag to `GAPIL_MAP_ELEMENT_EMPTY` for each element? Seems like we could easily double destruct with something like:

```
map[3] = obj_a;
map.clear();
map[5] = obj_b;
map.clear(); // obj_a destructed again
```
(note `5 < GAPIL_MIN_MAP_SIZE`)

Your `memset` in `clear_keep` looks like it would fix this.

AWoloszyn: clear() actually deletes the elements array.
However one of my previous optimizations breaks this though. (GAPIL_MIN_MAP_SIZE) I think we need an else() memset there.

ben-clayton: Yeah, I see that the elements are deleted if the count is greater than GAPIL_MIN_MAP_SIZE. The example is below this threshold and triggers the bug described.

ben-clayton:
Please also fix the presubmits.

gapis/api/templates/api.go.tmpl
```diff
@@ -415,6 +415,19 @@ import (
...
+        ϟw.OnSet(ϟctx, m, ϟapi.CompleteFragment{},
+          ϟapi.NilReference{}, ϟapi.NilReference{})
+        for key, _ := range (*m.Map) {
+          ϟw.CloseForwardDependency(ϟctx, key)
+        }
+      }
+      *m.Map = make(map[{{$key}}]{{$value}})
+    }
+  }
+
```
bjoeris: I think these should now be correct implementations of the dependency graph side of DestroyHandles and Clear.

However, there will be minor conflicts with some incoming dependency graph changes, depending on which PR goes in first.

AWoloszyn: Thanks!

gapis/api/vulkan/api/coherent_memory.api
```diff
@@ -57,25 +57,24 @@ sub void readCoherentMemory(ref!DeviceMemoryObject memory, VkDeviceSize readOffs
...
     if bufferBinding.Buffer != as!VkBuffer(0) {
-      if (bufferBinding.Buffer in Buffers) {
-        offset := switch as!u32(i) in bufferBindingOffsets {
-          case true:
-            bufferBindingOffsets[as!u32(i)]
-          case false:
-            bufferBinding.Offset
+      if !(bufferBinding.Buffer in ProcessedBuffers.val) {
+        if (bufferBinding.Buffer in Buffers) {
+          readMemoryInBuffer(Buffers[bufferBinding.Buffer], 0, 0xFFFFFFFFFFFFFFFF)
```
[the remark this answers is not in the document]
AWoloszyn: I have removed these changes. I think they will break the dependency graph in subtle ways, and I want to avoid that. Reads are only necessary in tracing if we are dealing with coherent memory. I will re-work part of this to see if we can do better without breaking dependency graph stuff.

PR #2324, opened 2018-10-29
Update our MESA workounds to fix the latest versions.

AWoloszyn:


core/cc/linux/get_gles_proc_address.cpp
```diff
@@ -34,9 +34,14 @@ void* getGlesProcAddress(const char* name) {
   static void* _dummy0 = dlopen("libLLVM-3.0.so.1", RTLD_LAZY | RTLD_DEEPBIND);
```
ben-clayton: There comes a point where a loop might be preferred :)

```c++
static bool preload_llvm = true;
if (preload_llvm) {
    preload_llvm = false;
    char name[512];
    for (int i = 3; i <= 7; i++) {
        snprintf(name, sizeof(name), "libLLVM-%d.0.so.1", i);
        dlopen("libLLVM-6.0.so.1", RTLD_LAZY | RTLD_DEEPBIND);
    }
}
```

AWoloszyn: Agreed, I will switch.

AWoloszyn: Race condition here (not that it matters) but I can rework it very slightly to remove.

PR #2326, opened 2018-10-30
Improve gofuse.

paulthomson:
* Add example to comments.
* Add option for copying .proto files.
* Add option to specify bazel-out/X directory name.
* Use shell CWD (bazel run doesn't seem to work for me otherwise).
* Include bazel-genfiles.
* Some extra "logging".

### DIFF
--- a/cmd/gofuse/main.go
+++ b/cmd/gofuse/main.go
@@ -17,7 +17,7 @@
 // gofuse will create a new 'fused' directory in the project root which
 // contains:
 //  • Symlinks to authored .go (and optionally, .proto) source files in the GAPID source tree.
-//  • Symlinks to bazel-generated .go files (bazel-out and bazel-genfiles).
+//  • Symlinks to bazel-generated .go files (bazel-out/[config]/{bin,genfiles}).
 //  • Symlinks to external 3rd-party .go files.
 // These symlinks are 'fused' into a single, common directory structure that
 // is expected by the typical GOPATH rules used by go tooling.
--- a/cmd/gofuse/main.go
+++ b/cmd/gofuse/main.go
@@ -16,15 +16,17 @@
 //
 // gofuse will create a new 'fused' directory in the project root which
 // contains:
-//  • Symlinks to authored .go source files in the GAPID source tree.
-//  • Symlinks to bazel-generated .go files.
+//  • Symlinks to authored .go (and optionally, .proto) source files in the GAPID source tree.
+//  • Symlinks to bazel-generated .go files (bazel-out and bazel-genfiles).
 //  • Symlinks to external 3rd-party .go files.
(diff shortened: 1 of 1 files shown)


cmd/gofuse/main.go
```diff
@@ -46,6 +48,11 @@ var externals = map[string]string{
 
 var (
 	fuseDir = flag.String("dir", "", "directory to use as the fuse root")
+	bazelOutDirectory = flag.String("bazelout", "k8-fastbuild",
```
ben-clayton: This is great. It would be 92% more awesome if it defaulted to the correct directory [based on the current OS](https://golang.org/pkg/runtime/#pkg-constants). :smiley:

paulthomson: I agree although I actually don't know what they will look like, except for darwin and k8!

paulthomson: Actually, maybe this should be a mandatory argument, as it is very easy to accidentally accept the default (e.g. k8-fastbuild), even though you might have only done a debug build.

ben-clayton: How about:

```golang
func bazelOutDirDefault() string {
    switch runtime.GOOS {
        case "linux":
            return "k8"
        case "darwin":
            return "darwin"
        default:
            return ""
    }
}

var bazelOutDirectory = flag.String("bazelout", bazelOutDirDefault(), "The bazel-out/X directory name. e.g. k8-fastbuild, darwin-fastbuild, k8-dbg, etc.")
```

Then error if `bazelOutDirectory` is empty string?

paulthomson: Yeah I had written something like that, but you have to add "-fastbuild", and I was also checking that the arch was "amd64" (not sure if that is necessary). But now I think requiring it might be best. Unless you think defaulting to "-fastbuild" is fine. I just keep forgetting that I am actually doing debug builds so I need "-dbg".

ben-clayton: Humm, fair point.
I wonder if you called this using the `bazel run` approach, whether there's any way to pick up the build flags to figure this out?

paulthomson: Yeah I searched for environment variables that might get set but it doesn't look like there are any (except for `BUILD_WORKSPACE_DIRECTORY` and `BUILD_WORKING_DIRECTORY`, and I will actually switch to using the former instead of the latter for picking the working directory). I suppose anybody willing to use this tool is probably willing to check what directory is in bazel-out/ :smile: Especially with some examples. There is actually `bazel-out/host`; not sure if that is good enough.

paulthomson: OK, now I guess the `bazel-out/X` directory, but there is a warning if guessing. Works well for me.

ben-clayton:
LGTM, thanks for the changes!

PR #2327, opened 2018-10-30
Add resource.created field.

paulthomson:
We already have resource.deleted, which gives the command index at which the resource was deleted.


ben-clayton:
Thanks Paul! Looks good!

paulthomson:
Oh I didn't know that. Yes, `accessed[0]` would be fine.

ben-clayton:
We talked about this F2F. Yes, it should be the same as `accessed[0]`, but this is clearly not obvious to new people. For the sake of a single integer field vs readability / symmetry with the `destroyed` field, I recommended adding this.

PR #2337, opened 2018-10-31
Most gapit verbs can now take a capture ID instead of a gfxtrace file.

paulthomson:
I would like to be able to pass a capture ID and a gapis-port to `gapit screenshot` so that it will take a screenshot of some capture that is already loaded. This avoids me having to wait 5 mins per screenshot to load the capture; instead, I can load it once from disk and then refer to it later via its capture ID. I would also like to modify (i.e. create mutated versions of) captures in gapis without writing them to disk, so this PR will be useful in this scenario.

Slightly weird: since the capture file is a positional/non-flag argument, I am attempting to parse it as a hex string (capture ID) and, if that fails, treating it as a file as before. If you have a filename that happens to be a hex string, you could use e.g. "./0000FFFFF". Could alternatively use a flag and then expect no positional args.

### DIFF
--- a/cmd/gapit/screenshot.go
+++ b/cmd/gapit/screenshot.go
@@ -146,7 +146,7 @@ func (verb *screenshotVerb) getSingleFrame(ctx context.Context, cmd *path.Comman
 	}
 	iip, err := client.GetFramebufferAttachment(ctx,
 		&service.ReplaySettings{
-			Device:                    device,
+			Device: device,
 			DisableReplayOptimization: verb.NoOpt,
 			DisplayToSurface:          verb.DisplayToSurface,
 		},
--- a/cmd/gapit/screenshot.go
+++ b/cmd/gapit/screenshot.go
@@ -21,13 +21,11 @@ import (
 	"image"
 	"image/png"
 	"os"
-	"path/filepath"
 	"strings"
 	"sync"
 
 	"github.com/google/gapid/core/app"
 	"github.com/google/gapid/core/app/flags"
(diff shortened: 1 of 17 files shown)


AWoloszyn:
You could add a boolean flag to gapit, (--capture-id) that just tells gapit that the given string is a capture-id instead of a file.

We have several boolean flags already, and this would make it more unambiguous. (-noopt for example).

cmd/gapit/screenshot.go
```diff
@@ -60,9 +61,17 @@ func (verb *screenshotVerb) Run(ctx context.Context, flags flag.FlagSet) error {
 		return nil
 	}
 
-	filepath, err := filepath.Abs(flags.Arg(0))
+	captureFilepath := flags.Arg(0)
+
+	// Try parsing captureFilepath as a hex string.
+	captureId, err := id.Parse(captureFilepath)
```
ben-clayton: `captureID` - govet will moan about this.
Also - it might be nice to wrap all the verbs that take a capture path argument with a helper function that does this logic.

paulthomson:
Talked with Ben. Made most gapit verbs be able to take a capture ID. If you add -captureid, the capture arg will be interpreted as a capture ID instead of a capture file path.

paulthomson:
I also added another change: `gapit replace_resource` can optionally not write out a capture file by passing a flag: `-skipoutput`. This allows me to mutate traces in gapis and then get screenshots from them, without having to write them to disk.

PR #2349, opened 2018-11-02
Added basic prototypes of 2 new extensions and fixed some const-correctness in EXT_debug_marker and case in VkExtent3D

aellem:


gapis/api/vulkan/extensions/amd_buffer_marker.api
```diff
@@ -0,0 +1,65 @@
...
+
+//////////////
+// Commands //
+//////////////
+
+@threadSafety("app")
+@indirect("VkCommandBuffer", "VkDevice")
+@extension("VK_AMD_buffer_marker")
+@override
+@no_replay
```
AWoloszyn: Do we really want to avoid replaying this command?

aellem: I don't recall why it has that tag, but I wasn't sure it was safe to replay a command that was just stubbed out and wasn't implemented properly.

aellem:
I fixed that I thought, maybe it wasn't in the pull request?  I'll do another.

PR #2355, opened 2018-11-06
Add a config setting to control mutate during reconstruction.

pmuetschard:
Turn mutate off by default during state reconstruction, but allow it to be enabled in the config for debugging.

### DIFF
--- a/gapis/config/config.go
+++ b/gapis/config/config.go
@@ -27,8 +27,9 @@ const (
 	// Logs all mappings at the end of the replay from original trace
 	// handles to replay client handles (if handles are reused in the trace
 	// it will only print the last mapping).  Only works for Vulkan.
-	LogMappingsToFile      = false
-	LogTransformsToFile    = false
-	LogTransformsToCapture = false
-	SeparateMutateStates   = false
+	LogMappingsToFile        = false
+	LogTransformsToFile      = false
+	LogTransformsToCapture   = false
+	SeparateMutateStates     = false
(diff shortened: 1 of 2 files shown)


gapis/config/config.go
```diff
@@ -27,8 +27,9 @@ const (
...
 	// it will only print the last mapping).  Only works for Vulkan.
-	LogMappingsToFile      = false
-	LogTransformsToFile    = false
-	LogTransformsToCapture = false
-	SeparateMutateStates   = false
+	LogMappingsToFile          = false
+	LogTransformsToFile        = false
+	LogTransformsToCapture     = false
+	SeparateMutateStates       = false
+	MutateDuringReconstruction = false
```
ben-clayton: Yes, it mutates the state, but this is just part of the work to verify that the commands correctly reconstruct the state. So how's about `CheckRebuiltStateMatches` or something, as this better describes the feature and also includes the costly call to `compare.Compare`?

pmuetschard: The idea was for Vulkan to also use this setting. @AWoloszyn, is that still something that has a chance of happening? any input on the name?

AWoloszyn: It can be done (technically) but it would be quite a non-trivial change to make it work. It is however something we should strive for. 
`CheckRebuiltStateMatches` is fine with me.

pmuetschard: Done. PTAL

PR #2360, opened 2018-11-09
Move the constants and opcodes out of the volatile memory region.

AWoloszyn:
This is a prerequisite for some upcomming replay changes to
allow us to pre-warm the VM.

### DIFF
--- a/gapir/cc/memory_manager.cpp
+++ b/gapir/cc/memory_manager.cpp
@@ -24,21 +24,20 @@
 
 //                 mMemory[0]
 //   ┏━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━┓
-//   ┃             ┃        in-memory       ┃
-//   ┃             ┃        resource        ┃
-//   ┃             ┃          cache         ┃
-//   ┃             ┣━━━━━━━━━━━━━━━━━━━━━━━━┨
+//   ┃             ┃                        ┃
+//   ┃             ┃                        ┃
 //   ┃             ┃                        ┃
 //   ┃             ┃         volatile       ┃
 //   ┃   mMemory   ┃          memory        ┃
 //   ┃             ┃                        ┃
-//   ┃             ┣━━━━━━━━━━━━┳━━━━━━━━━━━┨
-//   ┃             ┃            ┃  constant ┃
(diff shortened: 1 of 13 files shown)


gapir/cc/memory_manager.cpp
```diff
@@ -78,48 +80,15 @@ MemoryManager::MemoryManager(const std::vector<uint32_t>& sizeList)
...
-}
-
 bool MemoryManager::setVolatileMemory(uint32_t size) {
-  if (size > mSize - mReplayData.size) {
+  if (size > mSize) {
     return false;
   }
 
-  mVolatileMemory = {align(mReplayData.base - size), size};
+  mVolatileMemory = {align(&mMemory[0] + mSize - size), size};
```
ben-clayton: Any reason you're allocating from the top instead of the bottom?
Depending on the alignment of `mMemory`, it looks like this might be possible to underflow the buffer as the alignment happens after the bounds check.

AWoloszyn: This was for consistency, but I think for future changes volatile memory should just be at the bottom of the memory manager, (and the cache can live at the top). So I will make that change now, with only 2 memory regions this is a lot easier.

AWoloszyn:
Fixed comments.

gapir/cc/memory_manager.cpp
```diff
@@ -32,13 +32,12 @@
...
 //   ┃   mMemory   ┃          memory        ┃
 //   ┃             ┃                        ┃
-//   ┃             ┣━━━━━━━━━━━━┳━━━━━━━━━━━┨
-//   ┃             ┃            ┃  constant ┃
-//   ┃             ┃   replay   ┃   memory  ┃
-//   ┃             ┃    data    ┣━━━━━━━━━━━┨
-//   ┃             ┃            ┃   opcode  ┃
-//   ┃             ┃            ┃   memory  ┃
-//   ┗━━━━━━━━━━━━━┻━━━━━━━━━━━━┻━━━━━━━━━━━┛
+//   ┃             ┃                        ┃
```
Qining: Need to update this graph, as we swapped the position of these two segments.

AWoloszyn: Done

PR #2362, opened 2018-11-09
Update GAPIC's third party dependencies

pmuetschard:
- Download the dependencies during the build
 - Remove the checked-in jars
 - Update the dependencies to latest/newer releases


ben-clayton:
Presubmits are weirdly claiming all the tests should be stripped?

ben-clayton:
Rest of the changes look good to me.

pmuetschard:
PTAL. Windows build is actually green (there was an issue before, but I fixed it). The presubmit wanting to remove all the tests is fixed and committed in #2366.

PR #2364, opened 2018-11-09
 Handle `port:<pipename>:<abi>` URLs for Android tracing.

pmuetschard:
A URL of this pattern indicates tracing an Android app that is assumed to be already setup for tracing (i.e. have libgapii loaded) and is waiting for a connection from the host.

### DIFF
--- a/gapii/cc/spy.cpp
+++ b/gapii/cc/spy.cpp
@@ -137,7 +137,12 @@ Spy::Spy()
   // Use a "localabstract" pipe on Android to prevent depending on the traced
   // application having the INTERNET permission set, required for opening and
   // listening on a TCP socket.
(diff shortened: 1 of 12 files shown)


gapii/cc/spy.cpp
```diff
@@ -137,7 +137,12 @@ Spy::Spy()
   // Use a "localabstract" pipe on Android to prevent depending on the traced
   // application having the INTERNET permission set, required for opening and
   // listening on a TCP socket.
-  mConnection = ConnectionStream::listenPipe("gapii", true);
+  std::string pipe = "gapii";
+  char* envPipe = getenv("GAPII_PIPE_NAME");
```
AWoloszyn: I wish there were a nicer way to pass things in. But I haven't figured one out yet either. We use this for the port file as well.

PR #2369, opened 2018-11-15
Update coherent memory tracker to avoid hitting max_map_count

Qining:
By grouping multiple acutal memory pages into a Cell and manage the
'dirtiness' state at the Cell level instead of page level, we can reduce
the number of maps.

The updated tracker allow configuring the number of pages in a Cell.


core/memory_tracker/cc/memory_tracker.cpp
```diff
@@ -16,6 +16,8 @@
 
 #include "memory_tracker.h"
 
+#include "core/cc/log.h"
```
Qining: This include will be removed.

Qining:
Built on all three platforms and tested with samples we have. Presubmit is only complaining on third_party repo. Submit it now.

PR #2371, opened 2018-11-19
Gapis: Preload new dependency graph or old footprint based on config

bjoeris:
This change ensures that gapis obeys the config flags (`DisableDeadCodeElimination` and `NewDeadCodeElimination`) when pre-loading the dependency graph.


AWoloszyn:
hmm all the builds have failed however.
It might be good to rebase this on-top of latest master, that will kick a new build.

bjoeris:
I think this PR should compile now, but Kokoro seems to have stopped running entirely. I'll hold off on submitting anything until it comes back.

PR #2373, opened 2018-11-19
Vulkan Dependency Graph: Ensure `vkEndCommandBuffer` depends on `vkBeginCommandBuffer`

bjoeris:
This explicit read of `buff.Recording` makes `vkEndCommandBuffer` depend on `vkBeginCommandBuffer`'s setting of `buff.Recording`.

### DIFF
--- a/gapis/api/vulkan/api/command_buffer_control.api
+++ b/gapis/api/vulkan/api/command_buffer_control.api
@@ -421,7 +421,9 @@ cmd VkResult vkEndCommandBuffer(
     vkErrorInvalidCommandBuffer(commandBuffer)
   } else {
     buff := CommandBuffers[commandBuffer]
-    _ = buff.Recording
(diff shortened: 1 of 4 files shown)


gapis/api/vulkan/api/command_buffer_control.api
```diff
@@ -421,6 +421,7 @@ cmd VkResult vkEndCommandBuffer(
     vkErrorInvalidCommandBuffer(commandBuffer)
   } else {
     buff := CommandBuffers[commandBuffer]
+    _ = buff.Recording
```
AWoloszyn: Something that might be better (and provide users some use)
if (buff.Recording != RECORDING) {
   vkErrInvalidCommandBufferState()
}
That way we both get the dependency, and there is a clear reason for it.

PR #2374, opened 2018-11-19
Add missing dependencies related to renderpasses and pipelines

bjoeris:
These changes add the following dependencies:

- RenderPass commands (vkCmdDraw*, vkCmdClearAttachments, vkCmdNextSubpass) now form a dependency chain, so that the DCE handles the entire renderpass as a unit.
- Pipeline dynamic state (vkCmdSet*) is now implemented, and is depended upon by later draw calls
- Push constants are now tracked, and cause dependencies from draw calls to the relevant vkCmdPushConstants


gapis/api/vulkan/api/draw_commands.api
```diff
@@ -147,6 +147,9 @@ cmd void vkCmdBindVertexBuffers(
 
 
 sub void dovkCmdDraw(ref!vkCmdDrawArgs draw) {
+  readDrawState()
```
AWoloszyn: These are likely going to have a noticeable performance hit on trace time.
Can we mark them as @spy_disabled.

bjoeris:
I've added `@spy_disabled` to the subroutines called from inside draw calls.

PR #2375, opened 2018-11-19
Vulkan Dependency Graph: Add missing dependencies for pipeline barriers and events

bjoeris:
Barriers, arising from both vkCmdPipelineBarrier and vkCmdWaitEvents, now read and write the relevant images and memory, so that later commands which read those images or memory regions will depend on the barrier commands.

### DIFF
--- a/gapis/api/vulkan/api/synchronization.api
+++ b/gapis/api/vulkan/api/synchronization.api
@@ -395,12 +395,8 @@ sub void dovkCmdWaitEvents(ref!vkCmdWaitEventsArgs args) {
     }
   }
   if len(LastBoundQueue.PendingEvents) == 0 {
-    for _ , _ , b in args.ImageMemoryBarriers {
-      if !(b.image in Images) { vkErrorInvalidImage(b.image) } else {
-        image := Images[b.image]
-      transitionImageLayout(image, b.subresourceRange, b.oldLayout, b.newLayout)
-      }
-    }
+    processBarriers(args.SrcStageMask, args.DstStageMask,
+      args.MemoryBarriers, args.BufferMemoryBarriers, args.ImageMemoryBarriers)
   }
 }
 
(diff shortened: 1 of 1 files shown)


gapis/api/vulkan/api/synchronization.api
```diff
@@ -468,16 +463,9 @@ cmd void vkCmdWaitEvents(
...
-    if !(v.image in Images) { vkErrorInvalidImage(v.image) } else {
-      image := Images[v.image]
-      transitionImageLayout(image, v.subresourceRange, v.oldLayout, v.newLayout)
-      if v.oldLayout == VK_IMAGE_LAYOUT_UNDEFINED {
-        writeImageSubresource(image, v.subresourceRange)
-        updateImageQueue(image, v.subresourceRange)
-      }
-    }
-  }
+  // TODO: handle depenedncyFlags?
```
AWoloszyn: dependencyFlags

gapis/api/vulkan/api/synchronization.api
```diff
@@ -583,4 +571,52 @@ cmd void vkCmdPipelineBarrier(
...
+sub void processBarriers(VkPipelineStageFlags             srcStageMask        ,
+                         VkPipelineStageFlags             dstStageMask        ,
+                         map!(u32, VkMemoryBarrier)       memoryBarriers      ,
+                         map!(u32, VkBufferMemoryBarrier) bufferMemoryBarriers,
+                         map!(u32, VkImageMemoryBarrier)  imageMemoryBarriers ) {
+  // TODO: srcStageMask/dstStageMask
+
+  if len(memoryBarriers) > 0 {
+    // TODO: srcAccessMask/dstAccessMask
+    for _ , _ , img in Images {
```
AWoloszyn: I am somewhat worried about the performance of this loop. It may be fine, as memory barriers are RELATIVELY few and far between, but we should double-check that.

bjoeris: Would it make sense to move most of this code into a `@spy_disabled` subroutine? It looks like the image layout transition might be the only piece that is needed by gapii. Or are you also worried about `Mutate` performance hit? Do you know of any applications that might use more global memory barriers?

gapis/api/vulkan/api/synchronization.api
bjoeris: Done

Issue #2378, opened 2018-11-20
Presubmit failing due to: 


AWoloszyn:
https://github.com/google/gapid/pull/2361

RenfengLiu:
fixed in #2380

PR #2379, opened 2018-11-20
Fix presubmit failing

RenfengLiu:
Fix presubmit checking failing described at https://github.com/google/gapid/issues/2378


RenfengLiu:
worked at wrong branch

PR #2384, opened 2018-11-21
Vulkan: Lazy create descriptor bindings and array elements in state

Qining:
Some applications allocate a lot of descriptors a head but only update a
few of them a few times. Allocating those bindings and elements only
when the descriptors are updated can reduce the state tracking overhead
for such cases.


gapis/api/vulkan/api/descriptor.api
```diff
@@ -658,8 +635,19 @@ cmd void vkUpdateDescriptorSets(
...
     srcSet := DescriptorSets[c.SrcSet]
     dstSet := DescriptorSets[c.DstSet]
-    srcBinding := srcSet.Bindings[c.SrcBinding]
-    dstBinding := dstSet.Bindings[c.DstBinding]
+    //srcBinding := srcSet.Bindings[c.SrcBinding]
+    srcBinding := switch (c.SrcBinding in srcSet.Bindings) {
+      case false:
+        srcSet.Bindings[c.SrcBinding]
+      case true:
+        new!DescriptorBinding(BindingType: srcSet.Layout.Bindings[c.SrcBinding].Type)
```
bjoeris: When a new `srcBinding` is created here, should it be stored in `srcSet.Bindings`?

Qining: Not really I think, as there is no need to replace the null binding with an empty one. This is kept to avoid the need to use an 'if' branch to handle the 'clearing the dstBinding' case in the code.

Qining: But will this cause problem in the new dependency graph?

bjoeris: I think this CL should have no impact on the dependency graph. The deleted code in `vkAllocateDescriptorSets` only accesses local objects that have not yet been added to the global state, which means those are all ignored by the dependency graph builder. The new code here to safely look up the bindings should be equivalent from the dependency graph's perspective, since `c.SrcBinding in srcSet.Bindings` and `srcSet.Bindings[c.SrcBinding]` are both just reads of the `c.SrcBinding` fragment in `srcSet.Bindings`.

PR #2385, opened 2018-11-21
Split the replay into Initial state and replay content.

AWoloszyn:
This helps significantly with interaction as it allows us to asynchronously get the replay client into the right state while we are processing on the server. It also prevents us from having to re-mutate the initial commands in replay generation, which speeds that up as well.


gapis/config/config.go
```diff
@@ -18,7 +18,7 @@ package config
 const (
 	DebugReplay                = false
 	DebugReplayBuilder         = false
-	DisableDeadCodeElimination = false
+	DisableDeadCodeElimination = true
```
AWoloszyn: This needs to be fixed before merge.

gapis/replay/batch.go
```diff
@@ -100,6 +100,139 @@ func (m *manager) batch(ctx context.Context, e []scheduler.Executable, b schedul
 	}
 }
 
+var initialID string
```
AWoloszyn: Before I merge this I will move this into the database. Right now this will completely break with > 1 replay type.

gapis/replay/export_replay.go
```diff
@@ -88,7 +88,7 @@ func (m *exportManager) Export(ctx context.Context, waitRequests int) (*gapir.Pa
 	}
 	ctx = log.V{"replay target ABI": replayABI}.Bind(ctx)
 
-	b := builder.New(replayABI.MemoryLayout)
+	b := builder.New(replayABI.MemoryLayout, nil)
```
AWoloszyn: This probably won't work, since we no longer have the initial commands in Replay(.

Fix: Either generate / store a new capture that contains the initial commands, or detect this in the replay(.

AWoloszyn: This should be fixed by latest.

AWoloszyn:
Other than putting the result into the database, the only issue is that this does not work on Android for some reason. It looks like it may be a grpc bug. I am trying to upgrade GRPC now.

gapis/database/database.go
```diff
@@ -54,6 +59,26 @@ func Build(ctx context.Context, r Resolvable) (interface{}, error) {
...
+	}
+	if Get(ctx).IsResolved(ctx, id) {
+		return Get(ctx).Resolve(ctx, id)
+	}
+
+	newCtx := status.PutTask(keys.Clone(context.Background(), ctx), nil)
+	go func() {
+		cctx := status.PutTask(newCtx, nil)
+		Get(cctx).Resolve(cctx, id)
+	}()
```
Qining: Not quite sure if this background resolving is good to be here. This means this function `GetBuilt()` has side effect not reflect in its name and blurbs. But is it really necessary? Shouldn't we always rely on pre-resolved dependency graph (the only use of this function is trying to get resolved dependency graph)?

gapis/replay/builder/builder.go
```diff
@@ -698,10 +726,19 @@ func (b *Builder) assertResourceSizesAreAsExpected(ctx context.Context) {
 	}
 }
 
-func (b *Builder) layoutVolatileMemory(ctx context.Context, w binary.Writer) *volatileMemoryLayout {
+func (b *Builder) layoutVolatileMemory(ctx context.Context, volatileSpace uint64, w binary.Writer) *volatileMemoryLayout {
```
Qining: new argument: `volatileSpace` not used?

gapis/database/database.go
AWoloszyn: So in this case, we want to use the dependency graph if it is ready, but also NOT use it if it is not ready yet. This way we don't have to block un-necessarily.

It might be better to update the name of the function, 
GetOrBuild() maybe?

Qining: GetOrBuild() sounds a good idea.

Qining:
LGTM except the existing comments.

PR #2388, opened 2018-11-22
Mark the entire directroy as +x when trying to set up GAPIR.

### LINKED ISSUE #2387: Can not launch GAPIR on latest non-rooted Android 9
I have an idea of what is going on here, but it will require some testing before I put up a MR.

AWoloszyn:
See: https://android.googlesource.com/platform/ndk.git/+/ndk-release-r18/ndk-gdb.py#386
for more information.

Fixes #2387

### DIFF
--- a/gapir/client/session.go
+++ b/gapir/client/session.go
@@ -309,6 +309,16 @@ func (s *session) newADB(ctx context.Context, d adb.Device, abi *device.ABI) err
 	if err != nil {
 		return log.Errf(ctx, err, "Getting gapid.apk files directory")
 	}
+	appDir, err := apk.AppDir(ctx)
+	if err != nil {
+		return log.Errf(ctx, err, "Getting gapid.apk directory")
+	}
+
+	// Ignore the error returned from this. This is best-effort.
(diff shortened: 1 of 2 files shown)


gapir/client/session.go
```diff
@@ -309,6 +309,14 @@ func (s *session) newADB(ctx context.Context, d adb.Device, abi *device.ABI) err
 	if err != nil {
 		return log.Errf(ctx, err, "Getting gapid.apk files directory")
 	}
+	appDir, err := apk.AppDir(ctx)
+	if err != nil {
+		return log.Errf(ctx, err, "Getting gapid.apk directory")
+	}
+
+	// Ignore the error return from this. This is best-effort.
```
ben-clayton: `returned from this`. Please also link to the ndk hack in the comment.

pmuetschard:
This will also be required on rooted devices due to the type of socket we are using for gapir.

gapir/client/session.go
AWoloszyn: Done

PR #2403, opened 2018-11-26
Trim some stuff out of libinterceptor

pmuetschard:


pmuetschard:
Yup. Unused code and unnecessary dependencies. Should speed up the build a bit, too.

PR #2407, opened 2018-11-28
Dependency Graph: Added specialized maps to efficiently track fragment data

bjoeris:
This data structure will be used in an incoming dependency graph change to track the usage of memory ranges.


bjoeris:
Closing duplicate PR

PR #2408, opened 2018-11-28
Dependency Graph: Add specialized maps to efficiently track fragment data

bjoeris:
The dependency graph tracker needs to track data about "fragments" (e.g. fields in a struct, keys in a map). Previously these were all stored in maps. However, in many cases the structure of the fragments allows more efficient storage. For example, the fields in a struct are known at compile time, so instead of looking up the field in a map, we can just use the field index to lookup in a slice. A similar optimization is used for maps that are marked as "dense" (meaning the keys are 0..max).


bjoeris:
This PR will be used in an incoming dependency graph refactor.

Issue #2418, opened 2018-12-03
GL Android game trace slow in gapic and can't dump resources


paulthomson:
GAPID commit: 0d58a5a129158a02733becaa695c2160b0300c5b
Device: Pixel 3

I capture an OpenGL trace from a game on Android. This is not a mid-execution capture. It has about 80 frames: many of these are just a loading screen; the rest are a 3D menu screen. 

**I have shared the capture offline in a shared drive folder called "GapidTraces".** It contains `20181203_1416`.

I open the trace in gapid/gapic and the Android device is selected as the replay device:
* The commands show in the GUI, but the thumbnails take forever to load; I eventually saw some appear, but I have never seen them all load, even after e.g. 10 mins. 
* Clicking on commands is very slow during this time. If I click enough, the GUI becomes unresponsive for several seconds.
* The memory view does not seem to show; it shows a spinning circle, then nothing, then back to the spinning circle. 
* I can click on shaders, but the source doesn't seem to show. It seems to work for early frames.

**Could I disable the thumbnails from being loaded? I don't really need them.**

I close gapid/gapic, unplug the Android device, relaunch gapid/gapic, and open the trace. My desktop is selected as the replay device:
* The thumbnails still don't seem to load. 
* Clicking on commands still makes the GUI unresponsive for seconds.
* The memory view still does not seem to show; it shows a spinning circle and then returns to "Select a command...".
* I can see the shader sources for earlier frames; for later frames, I can't get the shaders sources to load, just the list of shader handles. 

**I can't get the thumbnails to load on my desktop either. This machine has a lot of RAM, in case that is relevant.**

I do `gapit screenshot capture.gfxtrace`. My desktop is chosen as the replay device. I see this warning:

`14:57:38.906 I: <gapis> Replay device does not have a memory layout matching device used to trace`

And eventually I see a lot of these errors:

```
14:56:46.680 I: <gapir> 14:56:46.138 E gapir: [gapir/cc/context.cpp:162] [252183]renderer: GL_INVALID_OPERATION error generated. <program> has not been linked, or is not a program object.
14:56:46.680 I: <gapir> 14:56:46.138 E gapir: [gapir/cc/context.cpp:162] [252183]renderer: GL_INVALID_OPERATION error generated. <program> object is not successfully linked, or is not a program object.
```
The screenshot is black. **I am not sure why these errors occur but I guess the different memory layout is problematic?**

[...]

PR #2419, opened 2018-12-03
dump_shaders: call getDevice, use ResolveConfig.

paulthomson:
It looks like `gapit dump_resources`/`dump_shaders` does not call `getDevice` like the other verbs. I think this can lead to the wrong device being selected.

Now it does call `getDevice` and it uses the device in a `ResolveConfig`.

### DIFF
--- a/cmd/gapit/dump_shaders.go
+++ b/cmd/gapit/dump_shaders.go
@@ -24,6 +24,7 @@ import (
 	"github.com/google/gapid/core/os/file"
 	"github.com/google/gapid/gapis/api"
 	"github.com/google/gapid/gapis/service"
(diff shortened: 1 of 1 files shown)


AWoloszyn:
Is there a case that you know about where a resource depends on the device? I can not think of any in Vulkan, as all resources just come out of the trace. For now I think? only shaders are returned as resources, so the device should not matter.

If we wanted to expand this in the future with more resources (images buffers etc) then this might be necessary.

AWoloszyn:
Never mind I see the linked bug now. Sounds good!

paulthomson:
Indeed, it doesn't seem like it should matter but I get errors otherwise in a GL trace. See #2418

PR #2421, opened 2018-12-03
Vulkan: Create forward dependency from vkBeginCommandBuffer to vkEndCommandBuffer

bjoeris:
This dependency is not usually needed for DCE, but is necessary when explicitly requesting the vkBeginCommandBuffer or individual vkCmd* calls, in which case the vkEndCommandBuffer call might be dropped by the DCE. The forward dependency from vkBeginCommandBuffer to vkEndCommandBuffer added by this patch ensures that, if any vkCmd* call is included by the DCE, then the corresponding vkBeginCommandBuffer and vkEndCommandBuffer calls are both also included by the DCE.

### DIFF
--- a/gapii/cc/vulkan_extras.cpp
+++ b/gapii/cc/vulkan_extras.cpp
@@ -467,6 +467,8 @@ void VulkanSpy::recordFenceReset(CallObserver*, uint64_t) {}
 void VulkanSpy::recordAcquireNextImage(CallObserver*, uint64_t, uint32_t) {}
 void VulkanSpy::recordPresentSwapchainImage(CallObserver*, uint64_t, uint32_t) {
 }
+void VulkanSpy::recordBeginCommandBuffer(CallObserver*, VkCommandBuffer) {}
+void VulkanSpy::recordEndCommandBuffer(CallObserver*, VkCommandBuffer) {}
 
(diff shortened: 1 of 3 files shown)


gapii/cc/vulkan_extras.cpp
```diff
@@ -467,6 +467,8 @@ void VulkanSpy::recordFenceReset(CallObserver*, uint64_t) {}
 void VulkanSpy::recordAcquireNextImage(CallObserver*, uint64_t, uint32_t) {}
 void VulkanSpy::recordPresentSwapchainImage(CallObserver*, uint64_t, uint32_t) {
 }
+void VulkanSpy::recordBeginCommandBuffer(CallObserver*, uint64_t) {}
```
AWoloszyn: We should start considering if there is a better way to handle forward dependencies without externs. 
In general, we should be trying to remove extens, but I understand in this case. It would be nice to not have to keep on adding them though

bjoeris: Is there a way to represent `interface{}` in the api language? I think that would allow all of the forward dependency externs to be reduced to a set of simple wrappers around the three `watcher.*ForwardDependency()` functions. 

Alternatively, we could add language primitives to wrap these three functions, but it doesn't really feel like these belong at the language level.

PR #2427, opened 2018-12-03
Vulkan: Record touched buffers/images on recording command buffers

Qining:
This is to avoid traversing through bound index, vertex, descriptor
buffers and images at VkQueueSubmit during trace. The goal is to avoid
the deep loop iterations to improve the state tracking performance during
tracing.

Also moved a bit the unlock/lock position, to allow more overlapping for 
multi-thread apps.

Also annotate as many spy_disabled as possible for memory related subroutines.

### DIFF
--- a/gapis/api/vulkan/api/coherent_memory.api
+++ b/gapis/api/vulkan/api/coherent_memory.api
@@ -36,6 +36,10 @@
 // TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
 // MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
 
+////////////
+// Memory //
+////////////
+
 sub void readCoherentMemory(ref!DeviceMemoryObject memory, VkDeviceSize readOffset, VkDeviceSize readSize) {
   if IsMemoryCoherent(memory) && (memory.MappedLocation != null) {
     offset_in_mapped := deviceMemoryOffsetToMappedSpace(memory, readOffset)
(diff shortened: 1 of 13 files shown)


gapis/api/vulkan/api/command_buffer_control.api
```diff
@@ -269,6 +269,14 @@ enum SemaphoreUpdate {
   @unused BufferCommands                  BufferCommands
   @unused ref!CommandBufferBegin          BeginInfo
   @unused ref!VulkanDebugMarkerInfo       DebugInfo
+
+  // Caching touched resources to help the state tracking performance
+  @hidden @unused @untracked @untrackedMap @nobox
```
AWoloszyn: WE really need to start documenting these annotations somewhere. They might be getting slightly out of  hand.

gapis/api/vulkan/api/descriptor.api
```diff
@@ -705,6 +717,79 @@ cmd void vkUpdateDescriptorSets(
     }
     dstSet.Bindings[c.DstBinding] = dstBinding
   }
+
+  // Then update the list of linked images and buffers for each modified
+  // descriptor sets. This is to avoid travering all the bindings and array
```
bjoeris: travering -> traveling

gapis/api/vulkan/api/descriptor.api
```diff
@@ -803,10 +889,7 @@ sub void dovkCmdBindDescriptorSets(ref!vkCmdBindDescriptorSetsArgs args) {
               }
 
               if !hasBeenRead.b {
-                oldSets := ProcessedDescriptorSets.val[set].bufferOffsets
-                ProcessedDescriptorSets.val[set] = 
-                  ProcessedDescriptorSet(oldSets, true)
-
+                LastBoundDescriptorSetsBufferInfo[set].shouldReadBuffers = true
```
bjoeris: Perhaps it is too much of a corner case to worry about, but if I bind descriptor sets, and then never make a draw/dispatch call that could use those descriptor sets, then I think the buffer shouldn't actually be read.

gapis/api/vulkan/api/coherent_memory.api
```diff
@@ -237,12 +333,12 @@ sub void readWriteMemoryInBoundDescriptorSets(
   for i in (0 .. layerCount) {
     descriptorSet := descriptorSets[as!u32(i)]
     if descriptorSet != null {
-      if !(descriptorSet.VulkanHandle in ProcessedDescriptorSets.val) {
+      if !(descriptorSet.VulkanHandle in LastBoundDescriptorSetsBufferInfo) {
```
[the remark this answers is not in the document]
Qining: The use of "LastBoundDescriptorSetsBufferInfo" is exactly to access descriptor sets only once in one execPendingCommands, to optimize the performance. But now we can drop it and revert it back to the original way: accessing the descriptors all the time (but not on the tracing side).

gapis/api/vulkan/api/descriptor.api
Qining: This 'shouldReadBuffers' means their memory will be read the 'next time' they are used, so if they are never used in draw/dispatch, they will not be read. Anyway, as this comes with the optimization to minimize the number of accesses to descriptor sets, they will be reverted all together.

Qining:
Fully tested, submit now.

PR #2429, opened 2018-12-03
Dependency Graph: Fix conflict with Clear and DestroyHandles

bjoeris:
The build is broken. @AWoloszyn, I think this is the correct way to update your code to match the changes I made to the other map accessor functions.

I also added `v.OnDestroy` calls to `DestroyHandles` match the behaviour in the singular `DestroyHandle`.


AWoloszyn:
Feel free to submit now.

PR #2431, opened 2018-12-04
Vulkan Dependency Graph: Added support for subcommands

bjoeris:
This required adding a `EncodedCmdID` field to CommandReference to track the CmdID of the `vkCmd*` call that recorded the CommandReference to the Commandbuffer. It is encoded so that the default 0 value corresponds to CmdNoID, for backward compatibility with captures that did not include this field. This field avoids many expensive `FlattenSubCmdIdx` calls which dominated the running time of subcommand dependency graph creation in early experiments.

Subcommand DCE is currently experimental and disabled by default.


bjoeris:
I have removed the CmdID from the CommandReference, and instead added a SubCmdIdxTrie to the sync data to efficiently look up the recording CmdID for a SubCmdIdx.

bjoeris:
Using the sync data to get the recording CmdID from a SubCmdIdx (as in my last force push here) unfortunately makes it impossible to generate dependency graphs that include the initial (state rebuilding) commands. Fixing this using sync data would require including the initial commands in the sync data, which seems like a rather substantial change that probably doesn't make sense in the other places sync data is being used. I'm looking into other ways to track this data.

bjoeris:
The latest change moves into the watcher the problem of associating subcommands with the recording commands. That is currently a separate commit to make that change clear, but would be squashed into the subcommand commit before finalizing the PR.

PR #2440, opened 2018-12-06
Vulkan: Fix the state rebuilder as we changed the states of query and query pools

Qining:


gapis/api/vulkan/api/query_pool.api
```diff
@@ -41,10 +41,11 @@
...
 enum QueryStatus {
-  QUERY_STATUS_INACTIVE      = 0
-  QUERY_STATUS_ACTIVE        = 1
-  QUERY_STATUS_COMPLETE      = 2
-  QUERY_STATUS_UNINITIALIZED = 4
+  QUERY_STATUS_CREATED = 0
+  QUERY_STATUS_RESET = 1
+  QUERY_STATUS_BEGAN = 2
+  QUERY_STATUS_ENDED = 3
+  QUERY_STATUS_TIMESTAMP_WRITTEN = 4
```
bjoeris: I think that re-arranging this enumeration will break compatibility with old traces, since the old state values will be interpreted incorrectly.

If we keep the previous enumeration, then the state from old captures (prior to adding UNINITIALIZED) will mostly be correct except:
- QUERY_STATUS_INACTIVE will become QUERY_STATUS_RESET, even if it hadn't actually been initialized, which may cause an extra reset. A quick look at the spec makes me think that extra, unnecessary resets are safe.
- For captures made after the change ta vkCmdWriteTimestamp, and before this change, timestamp queries may have been in state QUERY_STATUS_COMPLETE (now QUERY_STATUS_ENDED) instead of QUERY_STATUS_TIMESTAMP_WRITTEN.

gapis/api/vulkan/state_rebuilder.go
```diff
@@ -2778,18 +2778,20 @@ func (sb *stateBuilder) createQueryPool(qp QueryPoolObjectʳ) {
...
+			case QueryStatus_QUERY_STATUS_CREATED:
+				// do nothing
+			case QueryStatus_QUERY_STATUS_RESET:
+				sb.write(sb.cb.VkCmdResetQueryPool(commandBuffer, qp.VulkanHandle(), i, 1))
+			case QueryStatus_QUERY_STATUS_BEGAN:
+				sb.write(sb.cb.VkCmdBeginQuery(commandBuffer, qp.VulkanHandle(), i, VkQueryControlFlags(0)))
+			case QueryStatus_QUERY_STATUS_ENDED:
+				{
+					sb.write(sb.cb.VkCmdBeginQuery(commandBuffer, qp.VulkanHandle(), i, VkQueryControlFlags(0)))
+					sb.write(sb.cb.VkCmdEndQuery(commandBuffer, qp.VulkanHandle(), i))
```
bjoeris: As mentioned above, there could be captures made with a gapid version after my change to vkCmdWriteTimestamp, and before this change, in which case we may have a timestamp query pool in this state, in which case begin/end is invalid. So I think we need to check whether this query pool is for timestamps, and treat this as TIMESTAMP_WRITTEN instead. Once we've added that check, the TIMESTAMP_WRITTEN might just be redundant.

Qining:
replaced by #2442

PR #2447, opened 2018-12-10
Add a `@gapic_third_party` indirection to the gapic build.

pmuetschard:
Makes overriding of where the gapic dependencies come from a bit easier.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -20,8 +20,9 @@ gapid_dependencies()
 load("@gapid//tools/build:workspace_go.bzl", "gapid_go_dependencies")
 gapid_go_dependencies()
 
-load("@gapid//tools/build:workspace_gapic.bzl", "gapic_dependencies")
+load("@gapid//tools/build:workspace_gapic.bzl", "gapic_dependencies", "gapic_third_party")
 gapic_dependencies()
+gapic_third_party()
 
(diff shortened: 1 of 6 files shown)


pmuetschard:
Let me know if this helps with the maven deps.

Qining:
hmm, I don't know if this will be helpful. Generally what I have to do is manually downloading all the dependencies, then create `local_repository` for them. As all the dependencies in GAPID are wrapped with `maybe_repository`, those already existing dependencies won't be pulled again. This is also the approach we have always been using.

In this way, for most dependencies, the BUILD file and WORKSPACE file will also have their local copies, so the package structure won't change.

pmuetschard:
Well, the point of this change is that you no longer have to do that! You now have lots of options:
1) Create **one** extra repository called "gapic_third_party" that contains the 6 third party targets we need (`:grpc`, `:okhttp`, `:lwjgl`, `:guava`, `:jface` and `:swt`).
2) Check the jars into your main repo and then in the `WORKSPACE` file call the `gapic_third_party` function with the mappings pointing to them and it will generate the `gapic_third_party` repo BUILD file for you.
3) With the mappings you can have the jars almost anywhere.
4) ...

Qining:
Great, thanks, this definitely helps a lot.

PR #2455, opened 2018-12-13
Vulkan: Fix the attachment image layout tracking

Qining:
The loadAttachment and storeAttachment subroutine should not be labelled
with @spy_disabled, as we need to track the attachments' layout


gapis/api/vulkan/state_rebuilder.go
```diff
@@ -1482,8 +1482,7 @@ func (sb *stateBuilder) createImage(img ImageObjectʳ, imgPrimer *imagePrimer) {
 	attBits := VkImageUsageFlags(VkImageUsageFlagBits_VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT | VkImageUsageFlagBits_VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT)
 	storageBit := VkImageUsageFlags(VkImageUsageFlagBits_VK_IMAGE_USAGE_STORAGE_BIT)
 
-	isDepth := (img.Info().Usage() & VkImageUsageFlags(VkImageUsageFlagBits_VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT)) != 0
```
Qining: `isDepth` was used for debugging depth/stencil image priming by rendering. It does not imply depth stencil images with TRANSFER_DST_BIT set can not be primed by buffer copy.

PR #2458, opened 2018-12-13
Dependency Graph: add missing dependency vkCreateImageView->vkBindImageMemory

### LINKED ISSUE #2456: Dependency Graph: Missing dependency from vkCreateImageView to vkBindImageMemory
The Vulkan spec requires that image memory be bound before calling vkCreateImageView (under Valid Usage for VkImageViewCreateInfo):
> If image is non-sparse then it must be bound completely and contiguously to a single VkDeviceMemory object
However, the semantics of vkCreateImageView don't obviously use the image memory, so the current API model implementation of vkCreateImageView does not read image memory. This current implementation is therefore missing a dependency from vkCreateImageView to vkBindImageMemory.

bjoeris:
Fixes #2456


bjoeris:
Yes, I'm not sure why bazel didn't catch that problem on my workstation. Apparently changing the API files does not rebuild the C++ code.

PR #2459, opened 2018-12-13
DCE: Keep @alive commands in the initial commands.

bjoeris:
These are commands like vkQueueWaitIdle that should never be eliminated by the DCE.

### DIFF
--- a/gapis/resolve/dependencygraph2/dce.go
+++ b/gapis/resolve/dependencygraph2/dce.go
@@ -91,6 +91,17 @@ func NewDCEBuilder(graph DependencyGraph) *DCEBuilder {
 		origCmdIDs: make([]api.CmdID, 0, len(graph.Capture().Commands)),
 		liveCmdIDs: make(map[api.CmdID]api.CmdID),
 	}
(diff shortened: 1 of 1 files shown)


gapis/resolve/dependencygraph2/dce.go
```diff
@@ -91,6 +91,17 @@ func NewDCEBuilder(graph DependencyGraph) *DCEBuilder {
 		origCmdIDs: make([]api.CmdID, 0, len(graph.Capture().Commands)),
 		liveCmdIDs: make(map[api.CmdID]api.CmdID),
 	}
+	for i := 0; i < b.graph.NumInitialCommands(); i++ {
```
AWoloszyn: I do worry a bit about another loop of all of the initial commands.
That being said I am not sure of a better way.

bjoeris: Yes, I think it would be awkward to avoid having the DCE do this initial loop over all the commands. I haven't measured the overhead, but I think it should be very small. For each initial command it looks like a couple of function calls, a slice index, a handful of arithmetic operations and a branch.

PR #2463, opened 2018-12-13
Add extensions, fix some const issues and case issues

aellem:
Fix case in VkExtent3D
Fix const in debug labels
Add VK_AMD_buffer_marker extension definitions
Add VK_EXT_debug_utils extension


gapis/api/vulkan/command_buffer_rebuilder.go
```diff
@@ -1461,6 +1461,12 @@ func GetCommandFunction(cr *CommandReference) interface{} {
 		return subDovkCmdDrawIndirectCountAMD
 	case CommandType_cmd_vkCmdDrawIndexedIndirectCountAMD:
 		return subDovkCmdDrawIndexedIndirectCountAMD
+	case CommandType_cmd_vkCmdBeginDebugUtilsLabelEXT:
```
AWoloszyn: If you can squash this into a single commit, you are good to go.
However it would be good if you could update the other 2 big switch functions in this file with these 3 functions as well. That way everything should "just work" in gapid.
If not, I can do it post-merge.

AWoloszyn:
Closing this has been merged elsewhere

PR #2474, opened 2018-12-14
Step (1/Many) MultiGPU Support for Vulkan 1.1

AWoloszyn:
This is the first set of refactorings necessary to handle mGPU support
in Vulkan 1.1. This abstracts out VkDeviceMemory and VkPhysicalDevices
into multiple copies.

We still have to handle multiple Queue context / descriptors / etc.
We also have to still handle memory for images.


Qining:
Looks like a lot of format change and breaks the presubmit check, otherwise looks good to me.

AWoloszyn:
This is out of date, and we will likely have to re-do this work.

PR #2478, opened 2018-12-17
Vulkan: Correctly handle 3D images with 2D views

bjoeris:
The spec allows 2D `VkImageView`s of 3D `VkImage`s. The array layers of the 2D
view correspond to depth in the 3D image. For example, the valid usage for
VkImageViewCreateInfo specifies:

> If subresourceRange::layerCount is not VK_REMAINING_ARRAY_LAYERS, image is a
> 3D image created with VK_IMAGE_CREATE_2D_ARRAY_COMPATIBLE_BIT set, and
> viewType is VK_IMAGE_VIEW_TYPE_2D or VK_IMAGE_VIEW_TYPE_2D_ARRAY,
> subresourceRange::layerCount must be non-zero and
> subresourceRange::baseArrayLayer + subresourceRange::layerCount must be less
> than or equal to the extent.depth specified in VkImageCreateInfo when image
> was created

### DIFF
--- a/gapis/api/vulkan/api/coherent_memory.api
+++ b/gapis/api/vulkan/api/coherent_memory.api
@@ -180,16 +180,47 @@ sub void readCoherentMemoryInImage(ref!ImageObject image) {
 // VkImageSubresourceRange data
 
 @spy_disabled
-sub void readImageSubresource(ref!ImageObject image, VkImageSubresourceRange rng) {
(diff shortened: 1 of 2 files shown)


Qining:
Maybe change the name `accessImageSubresourceDepth` to `accessImageSubresourceSlices`, which looks more consistent to the spec. Otherwise looks good to me.

PR #2491, opened 2018-12-19
Added GetGraphVisualizationFile function

ecapia:
This is the first commit for graph-visualization


ecapia:
I signed it!

cmd/gapit/BUILD.bazel
```diff
@@ -28,6 +28,7 @@ go_library(
         "dump_replay.go",
         "dump_shaders.go",
         "export_replay.go",
+		"graph_visualization.go",
```
AWoloszyn: Whitespace here looks off.

cmd/gapit/flags.go
```diff
@@ -334,4 +334,10 @@ type (
 		Gapir GapirFlags
 		Out   string `help:"output file to save the profiling result"`
 	}
+
+	GraphVisualizationFlags struct {
+		Gapis         GapisFlags
+		Gapir         GapirFlags
```
AWoloszyn: Do we really need Gapir flags here? (I don't expect that we have to kick off a replay)

gapis/resolve/dependencygraph2/graph_visualization/graph_visualization.go
```diff
@@ -0,0 +1,21 @@
+package graph_visualization
```
bjoeris: Add copyright notice to all the new files. You can just copy it from one of the other go files, and make sure the year says 2018.

gapis/server/server.go
```diff
@@ -298,6 +299,20 @@ func (s *server) DCECapture(ctx context.Context, p *path.Capture, requested []*p
 	return trimmed, nil
 }
 
+
+func (s *server) GetGraphVisualizationFile(ctx context.Context, p *path.Capture) (string, error) {
+	ctx = log.Enter(ctx, "Inside GetGraphVisualizationFile function")
```
bjoeris: Can you make this match the other RPC calls in this file:
```
ctx = status.Start(ctx, "RPC GetGraphVisualizationFile")
defer status.Finish(ctx)
ctx = log.Enter(ctx, "GetGraphVisualizationFile")
```

cmd/gapit/graph_visualization.go
```diff
@@ -0,0 +1,72 @@
...
+	})
+}
+
+func (verb *graph_visualizationVerb) Run(ctx context.Context, flags flag.FlagSet) error {
+	if flags.NArg() != 1 {
+		app.Usage(ctx, "Exactly one gfx capture file expected, got %d", flags.NArg())
+		return nil
+	}
+
+	capturePath, err := filepath.Abs(flags.Arg(0))
```
AWoloszyn: I recommend you stick with the pattern we are using elsewhere for GAPIS.
https://github.com/google/gapid/blob/master/cmd/gapit/screenshot.go#L62

It lets you interact with an already-open gapis & trace.

cmd/gapit/graph_visualization.go
```diff
@@ -0,0 +1,72 @@
...
+	graphVisualizationFile, err := client.GetGraphVisualizationFile(ctx, capture)
+	if err != nil {
+		return log.Errf(ctx, err, "ExportCapture(%v)", capture)
+	}
+
+	graphVisualizationName := verb.Out
+	if graphVisualizationName == "" {
+		graphVisualizationName = "graph_visualizatione.dot"
+	}
+	if err := ioutil.WriteFile(graphVisualizationName, []byte(graphVisualizationFile), 0666); err != nil {
```
AWoloszyn: This will fail on Windows if the file is > 4GB, I recommend you create the file and then write it.

gapis/resolve/dependencygraph2/graph_visualization/graph_visualization.go
```diff
@@ -0,0 +1,21 @@
...
+	 "github.com/google/gapid/gapis/resolve/dependencygraph2"
+	 "github.com/google/gapid/core/math/interval"
+	 "github.com/google/gapid/gapis/api"
+	 "github.com/google/gapid/core/log"
+ )
+
+
+func GetGraphVisualizationFileFromCapture(ctx context.Context , p *capture.Capture) (string , error) {
+
+	log.I(ctx , "Working on GetGraphDotFileFromCapture")
```
AWoloszyn: It would be better to use the status package here. It will let us profile this. We should try to be as quiet as possible in the normal case.

gapis/server/BUILD.bazel
```diff
@@ -61,5 +62,6 @@ go_library(
         "@com_github_google_go_github//github:go_default_library",
         "@org_golang_google_grpc//:go_default_library",
         "@org_golang_x_net//context:go_default_library",
-    ],
+    	
+		],
```
AWoloszyn: More whitespace issues

gapis/server/grpc.go
```diff
@@ -397,6 +397,15 @@ func (s *grpcServer) DCECapture(ctx xctx.Context, req *service.DCECaptureRequest
 	return &service.DCECaptureResponse{Res: &service.DCECaptureResponse_Capture{Capture: capture}}, nil
 }
 
+func (s *grpcServer) GetGraphVisualizationFile(ctx xctx.Context, req *service.GraphVisualizationFileRequest) (*service.GraphVisualizationFileResponse, error) {
+	defer s.inRPC()()
+	graphVisualizationFile , err := s.handler.GetGraphVisualizationFile(s.bindCtx(ctx), req.Capture)
```
AWoloszyn: GetGraphVisualization would probably be better, since nothing is a file yet.

gapis/service/service.proto
```diff
@@ -316,6 +316,16 @@ message DCECaptureResponse {
   }
 }
 
+message GraphVisualizationFileRequest {
```
AWoloszyn: These RPCs will fail if the returned file exceeds 2GB. We should probably be careful to make sure that is a rare case.

If we think it will be larger, than as a follow-up CL, we should turn this into a streaming RPC. Does not have to be done now.

cmd/gapit/flags.go
```diff
@@ -334,4 +334,10 @@ type (
...
+		Out string `help:"path to Save Dot File"`
```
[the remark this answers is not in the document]
ecapia: Yes, I  fixed it.

cmd/gapit/graph_visualization.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	graphVisualizationFile, err := client.GetGraphVisualizationFile(ctx, capture)
+	if err != nil {
+		return log.Errf(ctx, err, "ExportCapture(%v)", capture)
+	}
+
+	graphVisualizationName := verb.Out
+	if graphVisualizationName == "" {
+		graphVisualizationName = "graph_visualizatione.dot"
```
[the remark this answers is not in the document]
ecapia: :+1:

gapis/client/client.go
```diff
@@ -484,3 +484,17 @@ func (c *client) GetTimestamps(ctx context.Context, capture *path.Capture, devic
 	}
 	return res, nil
 }
+
+func (c *client) GetGraphVisualizationFile(ctx context.Context, capture *path.Capture) (string , error) {
```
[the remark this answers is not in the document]
ecapia: :+1:

gapis/resolve/dependencygraph2/graph_visualization/graph_visualization.go
ecapia: I added it!! :+1:

gapis/server/server.go
ecapia: Done!!

cmd/gapit/BUILD.bazel
ecapia: I fixed all problems in regard to the format.

cmd/gapit/flags.go
ecapia: No, We do not really need. I am expecting also do not to kick off a replay. I removed the Gapir flag.
:+1:

gapis/resolve/dependencygraph2/graph_visualization/graph_visualization.go
ecapia: I added the status package on the server function that calls directly to this function GetGraphVisualizationFileFromCapture, since It also allows us to do the profiling.

gapis/server/BUILD.bazel
ecapia: Fixed!

gapis/server/grpc.go
ecapia: Yes. I did the changes.

cmd/gapit/graph_visualization.go
ecapia: I did not know about that issue on Windows. I create now first the file and then I write on this (existing file). I am curious about why in Windows fail if the file > 4GB ?, assuming that ioutil.WriteFile creates a new file if it does not exist.

gapis/service/service.proto
ecapia: I have push a new commit in response to changes requested . Please review .

gapis/server/server.go
ecapia: I have push a new commit in response to changes requested . Please review

ecapia:
I have push a new commit in response to changes requested . Please review

cmd/gapit/create_graph_visualization.go
```diff
@@ -0,0 +1,74 @@
...
+		return log.Errf(ctx, err, "ExportCapture(%v)", capture)
+	}
+
+	filePath := verb.Out
+	if filePath == "" {
+		filePath = "graph_visualization.dot"
+	}
+
+	_, err = os.Stat(filePath)
+	if os.IsNotExist(err) {
```
bjoeris: `os.Create` is safe to call even if the file already exists. I think the file will be replaced if it already exists.

cmd/gapit/create_graph_visualization.go
```diff
@@ -0,0 +1,74 @@
...
+		file, err := os.Create(filePath)
+		if err != nil {
+			log.Errf(ctx, err, "Creating file (%v)", filePath)
+		}
+		defer file.Close()
+	}
+
+	if err := ioutil.WriteFile(filePath, []byte(graphVisualization), 0666); err != nil {
```
bjoeris: I think @AWoloszyn's comment was about a file size limit of `WriteFile`, not about whether the file exists. You should be able to call `os.Create` to get a `File*`, and then write to that with `Write`. But I'm not familiar with the particular file size limitation @AWoloszyn is referring to, so I could also be misunderstanding.

However, it sounds like we'll hit the 2GB limit in GRPC first, which will require further changes to this code. Since we are currently well under these file limits, perhaps the server could just return an error if the file is >2GB.

ecapia: I added in the server side an error returned in case the file exceeds 2GB.

cmd/gapit/graph_visualization.go
AWoloszyn: There is a bug in ioutil.WriteFile that causes it to try and memory-map a large file. This fails on Windows, as there is a limit on the size of the file.

gapis/server/server.go
```diff
@@ -298,6 +303,24 @@ func (s *server) DCECapture(ctx context.Context, p *path.Capture, requested []*p
...
+	ctx = log.Enter(ctx, "GetGraphVisualization")
+	c, err := capture.ResolveFromPath(ctx, p)
+	if err != nil {
+		return "", err
+	}
+	graphVisualization, err := graph_visualization.GetGraphVisualizationFromCapture(ctx, c)
+	if err != nil {
+		return "", err
+	}
+	if len(graphVisualization) > FILE_SIZE_LIMIT_IN_BYTES {
```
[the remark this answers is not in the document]
ecapia: Definitely it would be better keep a []byte instead a string.
So I changed this to return []byte instead string in the entire API.

gapis/client/client.go
```diff
@@ -484,3 +484,16 @@ func (c *client) GetTimestamps(ctx context.Context, capture *path.Capture, devic
 	}
 	return res, nil
 }
+
+func (c *client) GetGraphVisualization(ctx context.Context, capture *path.Capture) (string, error) {
```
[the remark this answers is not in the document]
ecapia: Definitely it would be better keep a []byte instead a string.
So I changed this to return []byte instead string in the entire API.

cmd/gapit/create_graph_visualization.go
```diff
@@ -0,0 +1,71 @@
...
+
+	file, err := os.Create(filePath)
+	if err != nil {
+		return log.Errf(ctx, err, "Creating file (%v)", filePath)
+	}
+	defer file.Close()
+
+	bytesWritten, err := file.Write([]byte(graphVisualization))
+	if err != nil {
+		return log.Errf(ctx, err, "Writing file of size %d", bytesWritten)
```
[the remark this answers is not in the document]
ecapia: I was considering just in case to track the file size in case of some error, but I agree renaming the message.

ecapia:
I changed the entire API to return []byte instead string and also renamed error messages.

cmd/gapit/create_graph_visualization.go
AWoloszyn: This error will come from GRPC as well automatically, but having a nicer error is also good.

AWoloszyn:
I just tried this.

 ~/Development/gapid2/bazel-bin/pkg/gapit create_graph_visualization lin.gfxtrace
```
18:58:17.754 I: <gapit> Creating graph visualization file from capture id: a700558f1a3655c9e026e27e4bed10043bed4dc2
13:58:26.603 I: <gapis> Dependency Graph Stats:
13:58:26.603 I: <gapis>           NumCmdNodes: 75794     NumObsNodes: 88852
13:58:26.603 I: <gapis>         Accesses:
13:58:26.603 I: <gapis>           NumFragReads: 21930064  UniqueFragReads: 281367
13:58:26.603 I: <gapis>           NumFragWrites: 1094903  UniqueFragWrites: 118837
13:58:26.603 I: <gapis>           NumMemReads: 726120     UniqueMemReads: 107917
13:58:26.603 I: <gapis>           NumMemWrites: 299863    UniqueMemWrites: 105356
13:58:26.603 I: <gapis>           NumForwardDepOpens: 15370  NumForwardDepCloses: 7284  NumForwardDepDrops: 282
13:58:26.603 I: <gapis>         Deps:
13:58:26.603 I: <gapis>           NumDeps: 442345           UniqueDeps: 404850
13:58:26.603 I: <gapis>           NumFragDeps: 281367  NumCompleteFragDeps: 21    NumMemDeps: 145608
13:58:27.466 I: [GetGraphVisualization] <gapis> Dependency Graph Stats:
13:58:27.466 I: [GetGraphVisualization] <gapis>           NumCmdNodes: 112827    NumObsNodes: 88852
13:58:27.466 I: [GetGraphVisualization] <gapis>         Accesses:
13:58:27.466 I: [GetGraphVisualization] <gapis>           NumFragReads: 21930064  UniqueFragReads: 743472
```

But the resulting file is almost entirely empty.
```
➜  t cat graph_visualization.dot
OutputFile
```

PR #2492, opened 2018-12-20
Bunch 'O Shutdown Fixes

pmuetschard:
The `gapis` and `gapir` shutdown handling was, well, quite broken. This fixes things. The most noticeable thing this fixes is Android `gapir`: it no longer uses a full core while idling and the app will exit when the client exits.

`gapis` fixes:
 - Most of the "at exit" clean stuff wasn't running, because we ignored the SIGTERM signal, cancelled contexts would make things be ignored and because of deadlocks.
 - Close the log handler at the very end, so stuff still gets logged while cleaning up
 - When wanting to shutdown the gRPC server, cancel any log streaming RPCs, since they would cause the server to stay running indefinitely.
 - Send the `Shutdown` RPC to `gapir`.

`gapir` fixes:
 - Use a 1 second timeout in `ALooper_pollAll` call, because 0 causes it to return immediately and the tight loop causes a full core to spin.
 - Exit the app when `gapir` exits.
 - Actually monitor whether the gRPC server is still running in the main loop.
 - Correctly handle the `Shutdown` RPC request.


cmd/gapir/cc/main.cpp
```diff
@@ -143,6 +143,46 @@ std::unique_ptr<Server> Setup(const char* uri, const char* authToken,
 
 #if TARGET_OS == GAPID_OS_ANDROID
 
+namespace {
+
+#define JNI_CALL_O(obj, name, sig, ...)                           \
```
AWoloszyn: I would prefer to see this as a template, rather than a macro.

Pardon my bad pseudo-code and making up names of JNI api things:
```
template<typename... Args>
jobject jni_call_o(Env* env, clazz obj, const char* name, const char* sig, Args&&... args) {
   return env->CallObjectMethod(obj, env->GetMethodIf()......, std::forward<Args>(args)...);
}
```

PR #2499, opened 2018-12-20
Revert a change that inexplicably caused a large perf regression.

AWoloszyn:
Not entirely sure why this was causing a perf regression,
it should be faster if anything, but reverting until we
can investigate further.


AWoloszyn:
Updated to also fix the presubmit.

PR #2510, opened 2019-01-03
Use `SaveCapture` in favor of `ExportCapture`.

### LINKED ISSUE #403: The capture save action needs to either call a SaveCapture RPC or get the bytes as a stream
There's [a TODO](https://github.com/google/gapid/blob/master/gapic/src/main/com/google/gapid/models/Capture.java#L133) in the capture model.

pmuetschard:
Fixes #403


AWoloszyn:
Awesome thanks!!

PR #2511, opened 2019-01-03
Vulkan: Support for external memory/fence/semaphore capabilities

Qining:
This is to support just the capabilities extensions, not real support of external handles.

* VK_KHR_external_memory_capabilities
* VK_KHR_external_semaphore_capabilities
* VK_KHR_external_fence_capabilities

TODO: we may need to cache those capabilities/properties later when we
add real support for external handles.

PS: added several missing files also


gapis/api/vulkan/vulkan.api
```diff
@@ -193,6 +201,9 @@ sub ref!ExtensionSet supportedInstanceExtensions() {
   supported.ExtensionNames["VK_KHR_win32_surface"] = true
   supported.ExtensionNames["VK_KHR_get_physical_device_properties2"] = true
   supported.ExtensionNames["Vk_KHR_get_surface_capabilities2"] = true
```
AWoloszyn: I think these might be wrong?

V**k**_KHR_external

Qining: Fixed in 68a52c2

Issue #2435, opened 2018-12-05
Replace AllResourceDataResolvable with one per type.


AWoloszyn:
This should greatly help #2418

Instead of grabbing all resources all the time (when one is requested) optimize this so that we only process ones of the same type.

This will greatly reduce the amount of time required.

PR #2512, opened 2019-01-03
Make AllResourceData resolvable based on resource type.

AWoloszyn:
This means that dump_resources does not have to actually
kick off replays. It also means we don't have to store
as much data if we do not need all of it.

Fixes https://github.com/google/gapid/issues/2435

### DIFF
--- a/gapis/service/service.proto
+++ b/gapis/service/service.proto
@@ -845,6 +845,7 @@ message RenderSettings {
 // Resources contains the full list of resources used by a capture.
 message Resources {
   repeated ResourcesByType types = 1;
+  map<string, api.ResourceType> resourcesToTypes = 2;
 }
 
 // ResourcesByType contains all resources of a specific type.
--- a/gapis/resolve/resolvables.proto
+++ b/gapis/resolve/resolvables.proto
@@ -99,6 +99,7 @@ message ResourceDataResolvable {
 message AllResourceDataResolvable {
   path.Command after = 1;
   path.ResolveConfig config = 2;
+  api.ResourceType type = 3;
 }
 
 message ResourceMetaResolvable {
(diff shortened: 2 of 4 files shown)


gapis/service/service.proto
```diff
@@ -845,6 +845,7 @@ message RenderSettings {
 // Resources contains the full list of resources used by a capture.
 message Resources {
   repeated ResourcesByType types = 1;
+  map<string, api.ResourceType> resourceTypes = 2;
```
Qining: Probably we should have a better name for `resourceTypes` to indicate this is a map from resource id to resource types.

pmuetschard: why would we need this map? The resources are already returned by type.

AWoloszyn: This is so we can answer the question:
I have resourceID<foo> what type is that. We could get the same information out of the proto, but its backwards, given resourceID<Foo>

1) loop through all types
2) loop through all ids, to see if Foo is found.
3) Profit? (n^2 if we have to do this lookup repeatedly)

The reason it went into this message at all, was because it is free to compute along with the rest of the resources, otherwise we would have to re-mutate the entire world to generate this information.

AWoloszyn: Used above 
gapis/resolve/resource_data.go:141

AWoloszyn: FYI: The only other "efficient" way I could see of doing this was forcing the Client to add the type in the request. 
I.e. have the path be 
```
// Copy of resource_type due to who includes whom
// Or move it entirely into path.proto but that seems worse
enum ResourceType {
  // UnknownResource represents an unknown resource type
  UnknownResource = 0;
  // Texture represents the Texture resource type
  TextureResource = 1;
  // ShaderResource represents the Shader resource type
  ShaderResource = 2;
  // ProgramResource represents the Program resource type
  ProgramResource = 3;
  // PipelineResource respresents the Pipeline resource type
  PipelineResource = 4;
}

message ResourceData {
  ID ID = 1;
  Command after = 2;
  ResourceType type = 3;
}

```

pmuetschard: OK. Alright, I think I can live with it :). Here are my concerns:
 - this is serialized to the client and stays resident in memory there
 - not a big fan of the `ID.String()` part as keys

I think the only way around it is to have an intermediate value stored in the DB and convert from it for both calls, but I'm not sure it's worth it...

PR #2513, opened 2019-01-03
Simplify descriptor set function.

AWoloszyn:


Qining:
lgtm, defer to @bjoeris for concerns about dependency graph

gapis/api/vulkan/api/descriptor.api
```diff
@@ -729,26 +729,15 @@ emptyBufferOffsets {
...
-    if args.DescriptorSets[as!u32(i)] in DescriptorSets {
-      set := args.DescriptorSets[as!u32(i)]
-      setObj := DescriptorSets[set]
-      if currentDescriptorSets[args.FirstSet + as!u32(i)] != setObj {
-        currentDescriptorSets[args.FirstSet + as!u32(i)] = setObj
-      }
+  if (args.PipelineBindPoint == VK_PIPELINE_BIND_POINT_COMPUTE) {
+    computeInfo := lastComputeInfo()
+    for _, i, set in  args.DescriptorSets {
+      computeInfo.DescriptorSets[args.FirstSet + as!u32(i)] = DescriptorSets[set]
```
Qining: Probably better to add a guard around the DescriptorSets[set]:
`if set in DescriptorSets {`

AWoloszyn: I avoided that, since we will create a default object (nil) in this case if it doesn't exist.
I think putting NIL in the DescriptorSets, is more valid than leaving it what it used to be.

gapis/api/vulkan/api/descriptor.api
```diff
@@ -729,26 +729,15 @@ emptyBufferOffsets {
 }
 
 sub void updateDescriptorBufferBindingOffsets(ref!vkCmdBindDescriptorSetsArgs args) {
-  _ = PipelineLayouts[args.Layout]
```
bjoeris: This line was added to fix missing dependencies between `vkCmdBindDescriptorSets` and `vkCreatePipelineLayout`. Is there any reason to think it is no longer necessary?

AWoloszyn: Because it is also in trackDescriptorBufferBindingOffsets(), which this calls.

AWoloszyn:
@bjoeris Is that a sufficient explanation? If so can you approve?

Issue #2356, opened 2018-11-07
The gapis server has exited with an error code of: 2


y-novikov:
GAPID Version: 1.2.1:developer
OS: linux amd64

Happens when opening the attached trace of angle_end2end_tests on Nexus 5X.
[org.chromium.native_test_20181107_1249.gfxtrace.zip](https://github.com/google/gapid/files/2558581/org.chromium.native_test_20181107_1249.gfxtrace.zip)

panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: Panic at command 1139:vkDestroyImage(device: 514417378880, image: 176, pAllocator: 0):
runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: 
--- *resolve.FramebufferChangesResolvable ---
 capture:<ID:<data:"8\022\302\025tK\035`\037\030\342}9\202\377\301'>\3700" > > config:<replay_device:<ID:<data:"\255\344|\302\331Y\007\256\3365\341\204\270\276\000e2\342\000\223" > > > 
 Store():
  gapis/database/database.go:50  github.com/google/gapid/gapis/database.Build
  gapis/resolve/framebuffer_changes.go:34  github.com/google/gapid/gapis/resolve.FramebufferChanges
  gapis/resolve/framebuffer_attachment.go:79  github.com/google/gapid/gapis/resolve.(*FramebufferAttachmentResolvable).Resolve
  gapis/database/memory.go:130  github.com/google/gapid/gapis/database.(*record).resolve
  gapis/database/memory.go:222  github.com/google/gapid/gapis/database.(*memory).resolveLocked.func1
  core/app/crash/crash.go:65  github.com/google/gapid/core/app/crash.Go.func1

 Build() #0:
  gapis/database/database.go:54  github.com/google/gapid/gapis/database.Build
  gapis/resolve/framebuffer_changes.go:34  github.com/google/gapid/gapis/resolve.FramebufferChanges
  gapis/resolve/framebuffer_attachment.go:79  github.com/google/gapid/gapis/resolve.(*FramebufferAttachmentResolvable).Resolve
  gapis/database/memory.go:130  github.com/google/gapid/gapis/database.(*record).resolve
  gapis/database/memory.go:222  github.com/google/gapid/gapis/database.(*memory).resolveLocked.func1
  core/app/crash/crash.go:65  github.com/google/gapid/core/app/crash.Go.func1

--- *resolve.FramebufferAttachmentResolvable ---
[...]

PR #2518, opened 2019-01-04
Handle the case where we destroy an ImageView attached to a framebuffer.

AWoloszyn:


AWoloszyn:
Fixes https://github.com/google/gapid/issues/2356

PR #2524, opened 2019-01-08
Graph Visualization: Added hierarchies for commands and subcommands. Added flag to get the graph format in pbtxt , proto or dot.

ecapia:


gapis/resolve/dependencygraph2/graph_visualization/graph_structure.go
```diff
@@ -289,3 +321,7 @@ func (g *graph) getGraphInPbtxtFormat() string {
 	}
 	return output
 }
+
+func (g *graph) getGraphInProtoFormat() string {
+	return ""
```
bjoeris: I don't really see much point in including this skeleton definition. I would probably rip out all the pieces of this PR related to the "proto" format, and put them into a separate branch with the code that actually dumps the raw dependency graph. I think having both "pbtxt" and "proto" is kind of confusing, particularly since only one of them is actually related to "graph visualization".

cmd/gapit/flags.go
```diff
@@ -336,7 +336,8 @@ type (
 	}
 
 	CreateGraphVisualizationFlags struct {
-		Gapis GapisFlags
-		Out   string `help:"path to save graph visualization"`
+		Gapis  GapisFlags
+		Out    string `help:"path to save graph visualization"`
+		Format string `help:"output format of the graph (pbtxt, proto or dot)"`
```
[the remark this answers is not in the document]
ecapia: okay, I changed it !!

gapis/resolve/dependencygraph2/graph_visualization/graph_structure.go
ecapia: Yes, in fact the idea to consider 'proto' format was because we wanted some fast way to get the output for the graph visualization, but the 'proto' format does not look exactly like 'pbtxt' (Tensorboard can not read this), and also in a later commit using a fast concatenation (bytes.Buffer) is possible get a similar performance to get the 'pbtxt' format of the graph. So I removed all code in regard to 'proto' format.

AWoloszyn: For what it's worth, in general I would prefer this rpc to return a proto, and generate the pbtext/dot in a command that calls into this rpc. I understand that it is a huge refactor, and we can leave it to later, but putting external format-specific code into GAPIS is generally something to avoid.

bjoeris: That does seem like a better solution, once we have proper serialization of the dependency graph. This serialization is a bit messy since the dependency graph currently holds onto a reference to the capture, which is how we get information about the commands associated with the nodes.

AWoloszyn: So how I would probably implement this is the dependency graph would contain
path.Command (s) that can be used by the external tool, to get the command information if it wants it.

gapis/resolve/dependencygraph2/graph_visualization/graph_visualization.go
```diff
@@ -19,33 +19,62 @@ import (
...
+	idToBuilder := map[api.ID]api.GraphVisualizationBuilder{}
 
 	err := dependencyGraph.ForeachNode(
 		func(nodeId dependencygraph2.NodeID, dependencyNode dependencygraph2.Node) error {
 			if cmdNode, ok := dependencyNode.(dependencygraph2.CmdNode); ok {
 				cmdNodeId := cmdNode.Index[0]
 				command := dependencyGraph.GetCommand(api.CmdID(cmdNodeId))
-				commandName := command.CmdName()
+				isReal := api.CmdID(cmdNodeId).IsReal()
+				if !isReal {
```
bjoeris: Nit: no need for the `isReal` temporary variable.

gapis/resolve/dependencygraph2/graph_visualization/graph_visualization.go
```diff
@@ -19,33 +19,62 @@ import (
...
+						label = builder.GetCommandLabel(command, cmdNodeId)
+						name = command.CmdName()
+					} else if len(cmdNode.Index) > 1 {
+						commandName := fmt.Sprintf("%s_%d", command.CmdName(), cmdNodeId)
+						dependencyNodeAccesses := dependencyGraph.GetNodeAccesses(nodeId)
+						subCommandName := EMPTY
+						if len(dependencyNodeAccesses.InitCmdNodes) > 0 {
+							subCmdDependencyNode := dependencyGraph.GetNode(dependencyNodeAccesses.InitCmdNodes[0])
+							if subCmdNode , ok := subCmdDependencyNode.(dependencygraph2.CmdNode) ; ok {
+								subCmdId := subCmdNode.Index[0]
```
bjoeris: `subCmdNode.Index` *should* always have exactly one element, but it is still probably worth checking that, rather than potentially segfaulting.

gapis/resolve/dependencygraph2/graph_visualization/graph_visualization.go
```diff
@@ -19,33 +19,62 @@ import (
...
 func createGraphFromDependencyGraph(dependencyGraph dependencygraph2.DependencyGraph) (*graph, error) {
 
 	currentGraph := createGraph(0)
-	currentHierarchy := &api.Hierarchy{}
+	idToBuilder := map[api.ID]api.GraphVisualizationBuilder{}
 
 	err := dependencyGraph.ForeachNode(
 		func(nodeId dependencygraph2.NodeID, dependencyNode dependencygraph2.Node) error {
 			if cmdNode, ok := dependencyNode.(dependencygraph2.CmdNode); ok {
 				cmdNodeId := cmdNode.Index[0]
```
bjoeris: Check `len(cmdNode.Index)` (see my comment below)

cmd/gapit/flags.go
```diff
@@ -336,7 +336,8 @@ type (
 	}
 
 	CreateGraphVisualizationFlags struct {
-		Gapis GapisFlags
-		Out   string `help:"path to save graph visualization"`
+		Gapis  GapisFlags
+		Out    string `help:"path to save graph visualization"`
+		Format string `help:""output format of the graph: 'pbtxt' (Tensorboard) or 'dot' (Graphviz)""`
```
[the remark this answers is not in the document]
ecapia: Thank you, I fixed it.

gapis/resolve/dependencygraph2/graph_visualization/graph_visualization.go
ecapia: okay, I fixed it.

gapis/resolve/dependencygraph2/graph_visualization/graph_visualization.go
ecapia: Now I am checking if 'Cmd Node.index' contains exactly one element.

gapis/resolve/dependencygraph2/graph_visualization/graph_visualization.go
ecapia: Considering that all previous cmdNode.Index have been checked to contains exactly one element, I think for this case is safe,, but for unexpected segfaulting I am adding the checking anyway.

PR #2525, opened 2019-01-09
Rename server_disabled/spy_disabled for subroutines.

AWoloszyn:
This renames server_disabled/spy_disabled, to allow us to
define which subroutines need to be called for dependency
graph generation or not. This is really what they are trying
to describe.


Qining:
Lgtm, defer to @bjoeris.

Qining:
Need to fix presubmit

bjoeris:
I found some dce tests that are failing with these changes, but I don't know yet whether they were already broken.

bjoeris:
It looks like these errors (in the new dce) are older than this PR, maybe from #2482. So lgtm.

AWoloszyn:
This causes some interesting issues that I did not forsee. I will close this for now, and re-open once I work through it,

The performance benefits are not quite as significant as I had expected anyway.

PR #2527, opened 2019-01-11
Graph Visualization: Added hierarchy based on frames, faster function to get the 'pbtxt' and 'dot' formats, and implicit edges are added between commands and subcommands.

ecapia:


gapis/resolve/dependencygraph2/graph_visualization/graph_visualization.go
```diff
@@ -110,14 +123,15 @@ func GetGraphVisualizationFromCapture(ctx context.Context, p *path.Capture, form
 		return []byte{}, err
 	}
 
-	currentGraph, err := createGraphFromDependencyGraph(dependencyGraph)
-	currentGraph.removeNodesWithZeroDegree()
-	output := ""
+	currentGraph, err := createGraphFromDependencyGraph(ctx, dependencyGraph)
```
AWoloszyn: It would be more canonical go with:

```go
	currentGraph, err := createGraphFromDependencyGraph(ctx, dependencyGraph)
	if err != nil {
		return []byte{}, err
        }
	currentGraph.joinNodesByFrame()
	currentGraph.joinNodesWithZeroDegree()
	output := []byte{}
	if format == service.GraphFormat_PBTXT {
		output = currentGraph.getGraphInPbtxtFormat()
	} else if format == service.GraphFormat_DOT {
		output = currentGraph.getGraphInDotFormat()
	}

	return output, nil
```

gapis/resolve/dependencygraph2/graph_visualization/graph_structure.go
```diff
@@ -257,60 +265,111 @@ func (g *graph) makeStronglyConnectedComponentsByCommandTypeId() {
 	}
 }
 
-func (g *graph) getEdgesInDotFormat() string {
-	output := ""
-	for _, currentEdge := range g.edgeIdToEdge {
-		lines := strconv.Itoa(currentEdge.source.id) + " -> " + strconv.Itoa(currentEdge.sink.id) + ";\n"
-		output += lines
+func (g *graph) dfs(currentNode *node, visited []bool, visitedNodes *[]*node) {
```
bjoeris: This recursive implementation makes me a bit nervous that a long chain of dependencies could cause a stack overflow.

gapis/resolve/dependencygraph2/graph_visualization/graph_structure.go
```diff
@@ -257,60 +265,111 @@ func (g *graph) makeStronglyConnectedComponentsByCommandTypeId() {
...
-		neighbours := []*node{}
-		for neighbourId := range currentNode.inNeighbourIdToEdgeId {
-			neighbours = append(neighbours, g.nodeIdToNode[neighbourId])
+func (g *graph) getEdgesInDotFormat() []byte {
+	nodes := g.getSortedNodes()
+	var output bytes.Buffer
+	for _, currentNode := range nodes {
+		inNeighbours := g.getSortedNeighbours(currentNode.inNeighbourIdToEdgeId)
+		for _, neighbour := range inNeighbours {
+			output.WriteString(fmt.Sprintf("%d -> %d;\n", neighbour.id, currentNode.id))
```
bjoeris: 'fmt.Fprintf` will let you write directly to the `bytes.Buffer`, rather than creating an intermediate string.

gapis/resolve/dependencygraph2/graph_visualization/graph_visualization.go
```diff
@@ -49,16 +50,23 @@ func createGraphFromDependencyGraph(dependencyGraph dependencygraph2.DependencyG
...
 					}
 					builder := idToBuilder[command.API().ID()]
+					commandNameAndNodeId := fmt.Sprintf("%s_%d", command.CmdName(), cmdNodeId)
+					isSubCommand, parentNodeId := false, 0
 
 					label, name := "", ""
 					if len(cmdNode.Index) == 1 {
 						label = builder.GetCommandLabel(command, cmdNodeId)
 						name = command.CmdName()
+						commandNameAndNodeIdToNodeId[commandNameAndNodeId] = int(nodeId)
```
bjoeris: The naming here is kind of confusing. In "commandNameAndNodeIdToNodeId", the first "NodeId" seems to refer to the CmdID (`cmdNode.Index[0]`), while the second seems to refer to the `nodeId dependencygraph2.NodeID`. If I'm understanding this correctly, would "cmdNameAndIdToNodeId" be correct? 

"commandNameAndNodeId" should probably also be changed to match.

gapis/resolve/dependencygraph2/graph_visualization/graph_visualization.go
```diff
@@ -81,6 +89,11 @@ func createGraphFromDependencyGraph(dependencyGraph dependencygraph2.DependencyG
 					newNode := getNewNode(int(nodeId), label)
 					newNode.name = name
 					newNode.attributes = attributes
+					newNode.isEndOfFrame = command.CmdFlags(ctx, api.CmdID(cmdNodeId), &api.GlobalState{}).IsEndOfFrame()
```
bjoeris: Using an empty `GlobalState` actually works? 

@AWoloszyn, is this safe in general? And if so, why does `CmdFlags` have a `GlobalState` argument?

gapis/resolve/dependencygraph2/graph_visualization/graph_structure.go
```diff
@@ -257,60 +265,111 @@ func (g *graph) makeStronglyConnectedComponentsByCommandTypeId() {
...
-	output := ""
-	for _, currentNode := range g.nodeIdToNode {
-		lines := strconv.Itoa(currentNode.id) + "[label=" + currentNode.label + "]" + ";\n"
-		output += lines
+func (g *graph) joinNodesByFrame() {
+	visited := make([]bool, g.maxNodeId+1)
+	frameNumber := 1
+	nodes := g.getSortedNodes()
+	for _, currentNode := range nodes {
+		if !visited[currentNode.id] && currentNode.isEndOfFrame {
```
bjoeris: Just an observation: Forward dependencies break this strategy. For example, if an image is created in frame 1 and destroyed in frame 2, the creation has a dependency to the destruction. Following that edge in the DFS would cause some of the frame 2 commands to be incorrectly placed in frame 1.

However, I think the issues from this should be pretty small, since nodes with forward dependencies (like resource destruction) probably don't have a lot of other dependencies. I also don't see an obviously more correct way to handle forward dependencies, since just ignoring them in the DFS will leave lots of orphan nodes not associated with any frame.

gapis/resolve/dependencygraph2/graph_visualization/graph_visualization.go
AWoloszyn: This is not safe in general. I expect we would be ok in Vulkan, but it will likely segfault in GLES.

The way you would normally do this is get all of the EndOfFrame events from the trace (using a path.Events resolvable), and use those.
Alternatively you could insert those parameters into the dependency graph when creating it, which will have a valid global state.


`gapis/resolve/events.go`

gapis/resolve/dependencygraph2/graph_visualization/graph_structure.go
bjoeris: Ah, interesting!

gapis/resolve/dependencygraph2/graph_visualization/graph_visualization.go
ecapia: You are right, I added the error checking.

gapis/resolve/dependencygraph2/graph_visualization/graph_structure.go
```diff
@@ -257,60 +265,111 @@ func (g *graph) makeStronglyConnectedComponentsByCommandTypeId() {
...
+		output.WriteString(fmt.Sprintf("%d[label=%s];\n", currentNode.id, currentNode.label))
+	}
+	return output.Bytes()
+}
+
+func (g *graph) getGraphInDotFormat() []byte {
+	var output bytes.Buffer
+	output.WriteString("digraph g {\n")
+	output.Write(g.getNodesInDotFormat())
+	output.Write(g.getEdgesInDotFormat())
```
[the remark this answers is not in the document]
ecapia: You are right. I fixed it.

gapis/resolve/dependencygraph2/graph_visualization/graph_structure.go
ecapia: Yes, I added it.

gapis/resolve/dependencygraph2/graph_visualization/graph_structure.go
```diff
@@ -257,60 +265,111 @@ func (g *graph) makeStronglyConnectedComponentsByCommandTypeId() {
...
+	output.WriteString("}\n")
+	return output.Bytes()
+}
+
+func (g *graph) getGraphInPbtxtFormat() []byte {
+	nodes := g.getSortedNodes()
+	var output bytes.Buffer
+	for _, currentNode := range nodes {
+		output.WriteString("node {\n")
+		output.WriteString("name: \"" + currentNode.label + "\"\n")
```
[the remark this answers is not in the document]
ecapia: I did a benchmarking among: a) fmt.Fprintf, b) + (concatenation) and c) using individual concatenation with WriteString. The third option is more efficient, so I added the last one. Thank you.

gapis/resolve/dependencygraph2/graph_visualization/graph_visualization.go
ecapia: Yes, you are right. I changed it, in addition to this I renamed 'cmdNodeId' to 'cmdId'. I think now it is more clear.

gapis/resolve/dependencygraph2/graph_visualization/graph_structure.go
ecapia: Thank you, I was not expecting an infinite stack. However, I implemented an iterative version (using a breadth first search). I think in some point it could end up a unnecessary memory consumption issue for huge graphs (millions of nodes and hundred of millions of edges).

gapis/resolve/dependencygraph2/graph_visualization/graph_visualization.go
bjoeris: I will store the `CmdFlags` in the dependency graph command nodes, and switch this to get the `CmdFlags` from the dependency graph in a separate PR.

gapis/resolve/dependencygraph2/graph_visualization/graph_structure.go
```diff
@@ -257,60 +265,125 @@ func (g *graph) makeStronglyConnectedComponentsByCommandTypeId() {
 	}
 }
 
-func (g *graph) getEdgesInDotFormat() string {
-	output := ""
-	for _, currentEdge := range g.edgeIdToEdge {
-		lines := strconv.Itoa(currentEdge.source.id) + " -> " + strconv.Itoa(currentEdge.sink.id) + ";\n"
-		output += lines
+func (g *graph) bfs(sourceNode *node, visited []bool, visitedNodes *[]*node) {
+	head, tail := 0, 0
```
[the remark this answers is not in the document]
ecapia: No, I do not . I was using tail for better understanding of the algorithm, but I think it is okay remove this.

I am assuming that always visitedNodes is empty at the beginning, but could be useful for some cases set head := len(*visitedNodes) such as get nodes from two frames, so I changed it. Thank you.

gapis/resolve/dependencygraph2/graph_visualization/graph_structure.go
```diff
@@ -257,60 +265,125 @@ func (g *graph) makeStronglyConnectedComponentsByCommandTypeId() {
...
+	nodes := g.getSortedNodes()
+	var output bytes.Buffer
+	for _, currentNode := range nodes {
+		output.WriteString("node {\n")
+		output.WriteString("name: \"")
+		output.WriteString(currentNode.label)
+		output.WriteString("\"\n")
+		output.WriteString("op: \"")
+		output.WriteString(currentNode.label)
+		output.WriteString("\"\n")
```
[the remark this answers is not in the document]
ecapia: Yes, the performance is almost the same considering small lengths of strings. Based on local becnhmarking and https://gist.github.com/dtjm/c6ebc86abe7515c988ec , using concatenation is a little faster than fmt.Fprintf. So I added concatenation by line.

PR #2536, opened 2019-01-11
Added some missing dependency information.

AWoloszyn:


gapis/api/vulkan/api/descriptor.api
```diff
@@ -627,6 +627,11 @@ cmd void vkUpdateDescriptorSets(
         bufferBindings := setBinding.BufferBinding
         bufferBindings[arrayIndex] = w.BufferInfo
         setBinding.BufferBinding = bufferBindings
+        for _, _, b in setBinding.BufferBinding {
```
bjoeris: Just curious, why are buffers handled differently than images/buffer views (no `registerDescriptorUser` call)?

Qining: I believe this is because buffer info does not use buffer view. The registerXXX() is to register users of an object to that object, so that later when the app destroy the buffer view, we can use the registered users to get the user object and label them 'invalid'. Otherwise we need to traverse all the descriptor sets to find the ones using a buffer view.

Qining: But we currently do not keep this 'user' dependencies in the state for many objects, just image view and buffer view (not so sure that is all) to work around some issues found before. So the Descriptor set -uses-> buffer info -uses-> buffer, chain is not recorded in the state. Probably not need to do any of such anymore as we can rely on the new dependency graph?

PR #2541, opened 2019-01-14
Graph Visualization: Added a 'Label' structure, debug marks , order of commandBuffer and reestructuration of the code.

ecapia:


gapis/resolve/dependencygraph2/graph_visualization/graph_algorithms.go
```diff
@@ -0,0 +1,142 @@
...
+)
+
+const (
+	NO_VISITED       = 0
+	VISITED_AND_USED = -1
+	FRAME            = "FRAME"
+	UNUSED           = "UNUSED"
+)
+
+func (g *graph) traverseGraph(currentNode *node, visitTime, minVisitTime, idInStronglyConnectedComponents, visitedNodesId *[]int, currentId, currentTime *int) {
```
AWoloszyn: These are large enough functions, it would be good to get comments on exactly what the parameters are supposed to be.

gapis/api/graph_visualization.go
```diff
@@ -14,6 +14,72 @@
...
+	"bytes"
+	"fmt"
+)
+
+type Label struct {
+	LevelsName []string
+	LevelsID   []int
+}
+
+func (label *Label) GetSize() int {
```
AWoloszyn: Godocs needed for all public methods.

gapis/api/graph_visualization.go
```diff
@@ -14,6 +14,72 @@
 
 package api
 
+import (
+	"bytes"
+	"fmt"
+)
+
```
AWoloszyn: Godocs here as well.

gapis/api/vulkan/graph_visualization.go
```diff
@@ -66,60 +75,127 @@ func getSubCommandHierarchyNames() *api.HierarchyNames {
...
 		}
 	}
-	if _, ok := builder.labelToHierarchy[label]; !ok {
-		builder.labelToHierarchy[label] = &api.Hierarchy{}
+	temporalLabelAsAString := label.GetLabelAsAString()
+	if _, ok := builder.labelAsAStringToHierarchy[temporalLabelAsAString]; !ok {
+		builder.labelAsAStringToHierarchy[temporalLabelAsAString] = &api.Hierarchy{}
 	}
-	hierarchy := builder.labelToHierarchy[label]
+	hierarchy := builder.labelAsAStringToHierarchy[temporalLabelAsAString]
```
bjoeris: In general this pattern of inserting a default value into the map when the key is missing can be re-expressed as
```
hierarchy, ok := builder.labelAsAStringToHierarchy[temporalLabelAsAString]
if !ok {
	hierarchy = &api.Hierarchy{}
	builder.labelAsAStringToHierarchy[temporalLabelAsAString] = hierarchy
}		
```
This removes one of the map lookups.

gapis/resolve/dependencygraph2/graph_visualization/graph_structure.go
```diff
@@ -197,122 +189,6 @@ func (input nodeSorter) Less(i, j int) bool {
 	return input[i].id < input[j].id
 }
 
-func (g *graph) traverseGraph(currentNode *node, visitTime, minVisitTime, idInStronglyConnectedComponents, visitedNodesId *[]int, currentId, currentTime *int) {
```
bjoeris: When moving big chunks of code like this, it's usually preferred to do the move in a separate commit, since it is hard to see what changed. Just a comment, no need to change this.

gapis/api/vulkan/graph_visualization.go
```diff
@@ -66,60 +75,127 @@ func getSubCommandHierarchyNames() *api.HierarchyNames {
...
+				builder.labelsInsideDebugMarkers = append(builder.labelsInsideDebugMarkers, currentLabel)
+				builder.positionOfDebugMarkersBegin = append(builder.positionOfDebugMarkersBegin, len(builder.labelsInsideDebugMarkers)-1)
+				addDebugMarker(builder, positionOfLastDebugMarkerBegin, len(builder.labelsInsideDebugMarkers)-1)
+				builder.positionOfDebugMarkersBegin = builder.positionOfDebugMarkersBegin[:len(builder.positionOfDebugMarkersBegin)-1]
+			}
+		}
+
+	} else if len(builder.positionOfDebugMarkersBegin) > 0 {
+		if labelOfLastDebugMarkerBegin.GetSize() != currentLabel.GetSize() {
+			builder.positionOfDebugMarkersBegin = builder.positionOfDebugMarkersBegin[:0]
```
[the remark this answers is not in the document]
AWoloszyn: We have a ton of code in vulkan.go to handle nesting of debug markers and similar, (as used for displaying groups in the UI) It handles some of these cases (bad nestings etc).

Is there a chance we can leverage that? 
It may be a follow-up, but it may save some time in handling some of these cases.

gapis/resolve/dependencygraph2/graph_visualization/graph_algorithms.go
ecapia: I abstracted the parameters used to compute the Strongly Connected Components i a directed graph based on Tarjan algorithm. Thank you.

gapis/api/vulkan/graph_visualization.go
ecapia: Thank you. Yes , using this I am removing one look up in the map. I changed it for the two cases.

gapis/api/vulkan/graph_visualization.go
```diff
@@ -66,60 +75,127 @@ func getSubCommandHierarchyNames() *api.HierarchyNames {
...
+				label.PushBackLabel(hierarchyLabel)
 			} else {
-				label += fmt.Sprintf("/%d", index[i])
+				label.PushBack("", int(index[i]))
 			}
 		}
 	}
-	if _, ok := builder.labelToHierarchy[label]; !ok {
-		builder.labelToHierarchy[label] = &api.Hierarchy{}
+	temporalLabelAsAString := label.GetLabelAsAString()
```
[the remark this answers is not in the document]
ecapia: It is intentional, but I figured out that 'temporal' is not the correct word to describe something provisional , thank you. I changed it to temporary.

gapis/api/vulkan/graph_visualization.go
```diff
@@ -66,60 +75,127 @@ func getSubCommandHierarchyNames() *api.HierarchyNames {
...
+		}
+	}
+	return size
+}
+
+func addDebugMarker(builder *labelForVulkanCommands, from, to int) {
+	level := builder.labelsInsideDebugMarkers[len(builder.labelsInsideDebugMarkers)-1].GetSize() - 1
+	builder.numberOfDebugMarker++
+	for i := from; i <= to; i++ {
+		builder.labelsInsideDebugMarkers[i].Insert(level, VK_CMD_DEBUG_MARKER, builder.numberOfDebugMarker)
```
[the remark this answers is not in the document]
ecapia: In fact some labels could be affected by more than one debug marker (nested debug markers), in that case the algorithm add twice the debug marker level.

gapis/api/vulkan/graph_visualization.go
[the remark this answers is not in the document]
ecapia: Now I am adding debug markers as long as it does not destroy any level of hierarchy defined in commands. So, if exists some command between debug_marker_begin and debug_marker_end which has smaller Label size , adding the debug marker here will destroy the levels of hierarchy and for this case I just ignore the debug marker. It means I am preserving Labels with no-decreasing size. I have tested with several captures and it is working very nice. In addition to this, I am analizing the possibility to include code from vulkan.go to addres this problem as well, but if it is the case , I will add this in the last commit.

gapis/api/graph_visualization.go
ecapia: Thank you, I added the Godocs for all public methods.

gapis/resolve/dependencygraph2/graph_visualization/graph_structure.go
ecapia: I have abstracted this part of code to simplify the understanding, because it is essentially the Tarjan algorithm to compute Strongly Connected Components,

gapis/api/vulkan/graph_visualization.go
```diff
@@ -66,60 +75,129 @@ func getSubCommandHierarchyNames() *api.HierarchyNames {
...
-			return commandBuffer
+			return parameter.Get().(VkCommandBuffer), true
 		}
 	}
-	return ""
+	return 0, false
 }
 
-func (builder *labelForVulkanCommands) GetCommandLabel(command api.Cmd, commandNodeId uint64) string {
+func (builder *labelForVulkanCommands) GetCommandLabel(command api.Cmd, cmdId uint64) *api.Label {
```
bjoeris: These functions that implement the interfaces are also public, and should also have godocs. Those godocs are usually exactly the same as the godocs from the interface method they are implementing.

ecapia: Thank you, I will add the Godocs here as well.

PR #2549, opened 2019-01-16
 Vulkan: Prime storage image by storage buffer instead of texel uniform buffer

Qining:
Based on #2543 

This is the first step to generalize the interface of image primer.


Qining:
This turns out to be not necessary (at least for now), as copy the image data to another image for storage image on the GPU side can be fulfilled by `imageLoad` on the src image and then `imageStore` to the dst image directly (given the dst image has the exact same image format), there is generally no shared code between priming image data from host memory and priming image data from device memory for a storage image. So, changing to using a storage buffer is not necessary.

PR #2553, opened 2019-01-21
Vulkan: add 'primeableImageData' (prime device side image data step 1)

Qining:
Introduce 'primeableImageData' which can prime stored data to the
corresponding image in the new state in the state builder.


gapis/api/vulkan/image_primer.go
```diff
@@ -48,286 +48,30 @@ const (
...
-}
-
 func (p *imagePrimer) free() {
 	p.rh.free()
 	p.sh.free()
 }
 
 // internal functions of image primer
 
-func (p *imagePrimer) allocStagingImages(img ImageObjectʳ, aspect VkImageAspectFlagBits) ([]ImageObjectʳ, []DeviceMemoryObjectʳ, error) {
```
AWoloszyn: This function has gotten complicated enough, that it really deserves a more verbose comment.
At the least, what the returns are. I do not know what that func() is supposed to be.

Qining: Comments added in #2558, also this function got renamed to `create32BitUintColorStagingImagesForAspect()`

PR #2554, opened 2019-01-21
gapit dump_resources: include shader type in file extension

paulthomson:
It can be useful to know the type of dumped shaders. This simple change adds the type as an extension. E.g. "Shader<2><125>.Fragment".


paulthomson:
OK I have done this (although I did not modify protos)

PR #2558, opened 2019-01-23
 Vulkan: use imageLoad/Store to prime storage image data (prime device side image data step 2)

Qining:
This Cl uses staging images to prime storage image data.

Existing approach to priming storage image data:
Host Data --memory map--> texel uniform buffer --imageStore--> target image

The approach in this CL:
Host Data --memory map--> buffer --buffer image copy--> staging image --imageLoad/Store--> target image

Proposed approach to recover the device side storage image data
Image --imageLoad/Store--> staging image --imageLoad/Store--> target image

Note the "staging image --imageLoad/Store--> target image" part is
shared between priming host data and device data.


gapis/api/vulkan/image_primer.go
```diff
@@ -956,9 +832,9 @@ func (h *ipRenderHandler) render(job *ipRenderJob, tsk *scratchTask) error {
 			[]bufferSubRangeFillInfo{newBufferSubRangeFillInfoFromNewData(sbic.Bytes(), 0)},
 			VkBufferUsageFlagBits_VK_BUFFER_USAGE_UNIFORM_BUFFER_BIT,
 			VkBufferUsageFlagBits_VK_BUFFER_USAGE_TRANSFER_DST_BIT)
-		tsk.deferUntilExecuted(func() {
-			h.sb.write(h.sb.cb.VkDestroyBuffer(dev, stencilIndexBuf, memory.Nullptr))
-		})
+		// tsk.deferUntilExecuted(func() {
```
AWoloszyn: Is this right? (Also if it is, can we remove the commented out code?)

Qining: @AWoloszyn 
This is right for this CL because scratch buffers will be destroyed automatically after the commands in the task are done.

But I'm also wondering which way is better, 1) destroy automatically, 2) explicitly destroy scratch buffers. Any suggestions?

PR #2560, opened 2019-01-23
gles dep graph: textures accessed by pipeline during draw

paulthomson:


gapis/api/gles/dependency_graph_behaviour_provider.go
```diff
@@ -323,12 +323,51 @@ func clearBuffer(g *dependencygraph.DependencyGraph, b *dependencygraph.CmdBehav
 	b.Write(g, data)
 }
 
+func (pipeline Pipelineʳ) Uniforms() []Uniform {
+	uniforms := make([]Uniform, 0)
```
pmuetschard: Apparently, the preferred way is to do `uniforms := []Uniform{}`. I think, only use `make` if specifying an explicit size or capacity. (Also a few times below)

gapis/api/gles/dependency_graph_behaviour_provider.go
```diff
@@ -323,12 +323,51 @@ func clearBuffer(g *dependencygraph.DependencyGraph, b *dependencygraph.CmdBehav
...
+}
+
 func getAllUsedTextureData(ctx context.Context, cmd api.Cmd, id api.CmdID, s *api.GlobalState, c Contextʳ) (stateKeys []dependencygraph.StateKey) {
-	// Look for samplers used by the current program.
-	if c.Bound().Program().IsNil() {
+
+	// Look for samplers used by the current program/pipeline.
+
+	// Get the uniforms in use.
+	uniforms := make([]Uniform, 0)
```
pmuetschard: just `var uniforms []Uniform` here.

PR #2561, opened 2019-01-24
Vulkan: prime storage image data on the device side

Qining:
Sits on top of #2549 

Primeable image data now can be built from storage image whose data is
not available on the host side.

Staging image with same image create info will be created, the original
image data will be imageLoad/imageStore to the staging image when
building primeable image data.

Later, when the staging image data will be imageLoad/Store to the
original image when we want to prime the original data back to the
image.


Qining:
Close as already merged in #2572

PR #2570, opened 2019-01-30
 Use an explicit stop signal to stop a trace.

pmuetschard:
Instead of canceling the context, use an explicit signal to notify the tracer to stop the trace. This way tracers can perform actions after the stop to cleanup/finalize the trace that would otherwise fail on a canceled context.

### DIFF
--- a/gapis/server/server.go
+++ b/gapis/server/server.go
@@ -704,10 +704,10 @@ func (r *traceHandler) Initialize(ctx context.Context, opts *service.TraceOption
 		return nil, log.Errf(ctx, nil, "Error initialize a running trace")
 	}
 	r.initialized = true
-	ctx, stop := context.WithCancel(ctx)
-	r.stopFunc = func(ctx context.Context) error { stop(); return nil }
+	stopSignal, stopFunc := task.NewSignal()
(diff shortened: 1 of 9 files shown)


gapis/server/server.go
```diff
@@ -687,28 +687,29 @@ func (s *server) FindTraceTargets(ctx context.Context, req *service.FindTraceTar
...
-	bytesWritten   int64              // How many bytes have been written so far
-	startSignal    task.Signal        // If we are in MEC this signal will start the trace
-	startFunc      task.Task          // This is the function to go with the above signal.
-	stopFunc       context.CancelFunc // stopFunc can be used to stop/clean up the trace
-	doneSignal     task.Signal        // doneSignal can be waited on to make sure the trace is actually done
-	doneSignalFunc task.Task          // doneSignalFunc is called when tracing finished normally
-}
-
-func (r *traceHandler) Initialize(opts *service.TraceOptions) (*service.StatusResponse, error) {
+	initialized    bool        // Has the trace been initialized yet
```
AWoloszyn: Thanks for removing the context, it was definitely wrong.

Issue #2540, opened 2019-01-14
Need to improve benchmark result display


pau-baiget:
When running gapit benchmark it would be nice to get the results in a format that can be later easily copied into performance reports.
* All time information in ms.
* Comma-separated.

AWoloszyn:
I would prefer if we do this as an optional output type.
--output=json (or some such)

AWoloszyn:
or --output=csv

PR #2573, opened 2019-01-30
gapit: Add OutputCSV flag to benchmark (#2540).

pau-baiget:
This flag produces a single row of comma-separated data.
Also, all time information is expressed in milliseconds.

### DIFF
--- a/cmd/gapit/benchmark.go
+++ b/cmd/gapit/benchmark.go
@@ -512,9 +512,9 @@ func (verb *benchmarkVerb) Run(ctx context.Context, flags flag.FlagSet) error {
 		for _, types := range resources.GetTypes() {
 			for ii, v := range types.GetResources() {
 				if (types.Type == api.ResourceType_TextureResource ||
-					types.Type == api.ResourceType_ShaderResource ||
-					types.Type == api.ResourceType_ProgramResource) &&
-					ii < 30 {
+						types.Type == api.ResourceType_ShaderResource ||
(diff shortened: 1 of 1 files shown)


cmd/gapit/benchmark.go
```diff
@@ -585,34 +586,57 @@ func (verb *benchmarkVerb) Run(ctx context.Context, flags flag.FlagSet) error {
...
-			verb.gapisCachingDoneTime.Sub(verb.traceDoneTime),
-			verb.interactionDoneTime.Sub(verb.interactionStartTime),
-			traceMaxMemory,
-			time.Duration(preMecFramerate)*time.Nanosecond,
-			time.Duration(nonLoadingFrameTime)*time.Nanosecond,
-		)
-		w.Flush()
+		if verb.OutputCSV {
+			csvWriter := csv.NewWriter(os.Stdout)
+			defer csvWriter.Flush()
```
AWoloszyn: Defer is a bit funny (and you probably don't actually want it here) but defer actually defers until the end of the current function (NOT the scope-block).
How I typically handle this is:
```
if verb.OutputCSV { 
   func() {
      csvWriter....
      defer csvwriter.Flush()
      ....
      ....
   } () 
 } else {
      ...
```

pau-baiget: Ah! Ok, didn't know that. Will change it and resubmit.

AWoloszyn: It is super subtle, and goes completely against all of my past experience (thanks Go)

PR #2580, opened 2019-01-31
Fix protoconv error introduced by b2cefde6.

pmuetschard:


pmuetschard:
@hevrard this should fix .Set for you.

Issue #2590, opened 2019-02-07
gapis crashes when loading a capture


hevrard:
When opening a capture through the UI, gapis crashes, see log below.

$A0003

```
W20190207-103913000[server.ChildProcess$LoggingStringHandler][server.ChildProcess$LoggingStringHandler.lambda$new$0] gapis: terminate called after throwing an instance of 'spirv_cross::CompilerError'
W20190207-103913001[server.ChildProcess$LoggingStringHandler][server.ChildProcess$LoggingStringHandler.lambda$new$0] gapis:   what():  Over 3 compilation loops detected. Must be a bug!
```

hevrard:
This crash is due to spirv-cross crashing on a shader, this reveals three issues:

1. The GLES to GL conversion is called although we tried to replay on device, this conversion should not be triggered (TODO: invert condition on this line: https://github.com/google/gapid/blob/a64a5871249827ec6bb11c364ba7638783943a33/gapis/api/gles/find_issues.go#L173 )
2. gapis should be robust to spirv-cross crashing
3. spirv-cross should be updated

hevrard:
spirc-cross crashes are gracefully handled thanks to #2600

hevrard:
It looks like spirv-cross has been last updated May of last year with:
https://github.com/google/gapid/pull/1871/files

Should we update it now?

In general, should we add update of third-party tools as part of a new release process?

AWoloszyn:
So, we typically only update third_party tools as necessary. There is quite a bit of effort needed in updating, and usually very little gain. If we think there is a specific reason we should update spirv-tools/spirv-cross/etc then we should. But otherwise it may end up being wasted work.

Given the number of dependencies we have, it would take O(days) to update them all properly. (llvm being the biggest problem here).

AWoloszyn:
In this case, we probably should update spirv-cross however.

AWoloszyn:
Updated

PR #2591, opened 2019-02-07
Fix condition on shader cross-compilation

hevrard:
A GLES shader must be cross-compiled to a GL shader only when the target
is *not* a GLES target.


hevrard:
Fixes part of #2590

PR #2600, opened 2019-02-11
Wrap spirv-cross to catch exceptions.

AWoloszyn:
SPIRV-Cross sometimes returns errors as exceptions,
catch those. This way at least we won't crash
the server on bad input.

### DIFF
--- a/gapis/shadertools/cc/libmanager.cpp
+++ b/gapis/shadertools/cc/libmanager.cpp
@@ -317,23 +317,34 @@ code_with_debug_info_t* convertGlsl(const char* input, size_t length,
     result->disassembly_string = new char[tmp.length() + 1];
     strcpy(result->disassembly_string, tmp.c_str());
   }
+  // Sometimes SPIR-V cross throws exceptions. \o/
+  // This is a big hammer, but should handle all of the cases
+  try {
+    std::string source =
+        spirv2glsl(std::move(spirv_new), options->target_glsl_version,
+                   options->strip_optimizations);
 
-  std::string source =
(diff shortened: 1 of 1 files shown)


gapis/shadertools/cc/libmanager.cpp
```diff
@@ -317,24 +318,30 @@ code_with_debug_info_t* convertGlsl(const char* input, size_t length,
...
-
-  if (!err_msg.empty()) {
-    set_error_msg(result, "Failed to parse modified source code:\n" + err_msg);
-    return result;
+    if (!err_msg.empty()) {
+      set_error_msg(result, "Failed to parse modified source code:\n" + err_msg);
+      return result;
+    }
+  } catch (...) {
+      set_error_msg(result, "Exception thrown from spirv-cross:\n");
```
pmuetschard: I'm assuming something is missing to be shown after the colon? It would be nice to show something from the exception.

gapis/shadertools/cc/libmanager.cpp
```diff
@@ -317,24 +318,33 @@ code_with_debug_info_t* convertGlsl(const char* input, size_t length,
...
-    parseGlslang(result->source_code, nullptr, &err_msg, options->shader_type,
-                 OPENGL, false);
-  }
-
-  if (!err_msg.empty()) {
-    set_error_msg(result, "Failed to parse modified source code:\n" + err_msg);
-    return result;
+    if (!err_msg.empty()) {
+      set_error_msg(result, "Failed to parse modified source code:\n" + err_msg);
+      return result;
```
paulthomson: Thanks for adding the return here

PR #2601, opened 2019-02-11
Check the aspect against the image.

AWoloszyn:
We were not checking that the aspect existed before dereferencing it.

### DIFF
--- a/gapis/api/vulkan/api/coherent_memory.api
+++ b/gapis/api/vulkan/api/coherent_memory.api
@@ -232,6 +232,7 @@ sub void accessImageSubresourceSlice(ref!ImageObject image, VkImageSubresourceRa
       case VK_IMAGE_ASPECT_STENCIL_BIT:
         as!u64(1)
     }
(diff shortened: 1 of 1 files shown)


gapis/api/vulkan/api/coherent_memory.api
```diff
@@ -232,6 +232,7 @@ sub void accessImageSubresourceSlice(ref!ImageObject image, VkImageSubresourceRa
       case VK_IMAGE_ASPECT_STENCIL_BIT:
         as!u64(1)
     }
+    if !aspectBit in image.Aspects { vkErrorInvalidImageAspect(image.VulkanHandle, aspectBit) } else {
```
pmuetschard: Is this really a syntax format pattern we are using? or are you just trying to avoid lots of whitespace changes?

AWoloszyn: We have used this in a few places.
Since we don't have early-returns, we have been putting anything that behaves like an early-return like this (in the vulkan api files). Ideally I want 
```
if !aspectBit in image.Aspecs { return; }
```
Otherwise we would have insane amounts of nesting, that does not help readability at all

PR #2602, opened 2019-02-11
Allowed changes to the SSH config while gapis is running.

AWoloszyn:
This will periodically poll for SSH devices, and update the
file that contains that data as necessary.

### DIFF
--- a/core/os/device/remotessh/device.go
+++ b/core/os/device/remotessh/device.go
@@ -23,9 +23,12 @@ import (
 	"net"
 	"os"
 	"strings"
+	"sync"
+	"time"
 
 	"github.com/golang/protobuf/jsonpb"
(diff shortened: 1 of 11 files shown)


core/os/device/remotessh/device.go
```diff
@@ -124,24 +132,132 @@ func (b *binding) newPooledSession() (*pooledSession, error) {
...
+				return err
+			}
+			defer func() {
+				for _, rc := range rcs {
+					rc.Close()
+				}
+			}()
+			configs := []Configuration{}
+			for _, rc := range rcs {
+				configurations, err := ReadConfigurations(rc)
```
Qining: Nit: Looks like there are some shared code here and in `remotessh.Devices()`, they all read a list of configurations and then call `scanDevices`

Issue #2608, opened 2019-02-15
gapit trim: `panic: runtime error: index out of range`


hevrard:
Calling `gapit trim com.graphicsfuzz.vkworker.gfxtrace` leads to a panic.

[com.graphicsfuzz.vkworker.gfxtrace.zip](https://github.com/google/gapid/files/2869870/com.graphicsfuzz.vkworker.gfxtrace.zip)

```
panic: runtime error: index out of range [recovered]
	panic: runtime error: index out of range

goroutine 1 [running]:
github.com/google/gapid/core/app/crash.Crash(0xf6f6a0, 0x177be30)
	core/app/crash/crash.go:89 +0xe1
github.com/google/gapid/core/app.Run.func1()
	core/app/run.go:161 +0xec
panic(0xf6f6a0, 0x177be30)
	GOROOT/src/runtime/panic.go:513 +0x1b9
main.(*trimVerb).getDCERequest(0xc00021b560, 0xc00011d0c0, 0x5, 0x8, 0xc0001dc2d0, 0x0, 0x0, 0x0)
	cmd/gapit/trim.go:121 +0x70b
main.(*trimVerb).Run(0xc00021b560, 0x115b5a0, 0xc0002550b0, 0xc00027e220, 0x0, 0x0, 0x1, 0x0, 0xc000254db0, 0xc00027e210, ...)
	cmd/gapit/trim.go:58 +0x5cb
github.com/google/gapid/core/app.(*Verb).Invoke(0x17d7500, 0x115b5a0, 0xc0002550b0, 0xc00027e200, 0x2, 0x2, 0x0, 0x0)
	core/app/verbs.go:92 +0x742
github.com/google/gapid/core/app.VerbMain(0x115b5a0, 0xc0002550b0, 0x0, 0x0)
	core/app/verbs.go:123 +0x9a
github.com/google/gapid/core/app.Run(0x1097da0)
	core/app/run.go:253 +0xbc5
main.main()
	cmd/gapit/main.go:24 +0x7d
```

PR #2609, opened 2019-02-15
Add basic test script

hevrard:
This is a very basic test script which runs gapit commands that do *not* need a device plugged, so they can be ran on the CI.

One can drop several gfxtrace files into the `traces/` directory to try these commands on all the traces. The included trace is a small trace from the graphicsfuzz worker with the sample shaders.

This is very simple but already lead to catching a crash on the trim command, see #2608

There are many ways to improve on this in the future.


hevrard:
Agreed to rewrite for portability, a new `smoke_test` gapit verb sounds good.

hevrard:
See rewrite in go in #2619

PR #2611, opened 2019-02-16
GraphicsSpy Layer Fixes for Q and Beyond

pmuetschard:
Prepares us for a JDWP free tracing in Q and beyond.


core/os/device/deviceinfo/cc/android/query.cpp
```diff
@@ -271,48 +265,39 @@ bool createContext(void* platform_data) {
...
 
-  Class build(env, "android/os/Build");
-  CHECK(build.get_field("SUPPORTED_ABIS", gContext.mSupportedABIs));
-  CHECK(build.get_field("HOST", gContext.mHost));
-  CHECK(build.get_field("SERIAL", gContext.mSerial));
-  CHECK(build.get_field("MANUFACTURER", manufacturer));
-  CHECK(build.get_field("MODEL", model));
-  CHECK(build.get_field("HARDWARE", gContext.mHardware));
-  CHECK(build.get_field("DISPLAY", gContext.mOSBuild));
+  GET_STRING_LIST_PROP("ro.product.cpu.abilist", gContext.mSupportedABIs);
```
ben-clayton: Be sure to test this on older versions of Android and with 3rd party devices. IIRC, there was something preventing me from using system properties to get this info, hence the JNI.

core/os/device/deviceinfo/cc/android/query.cpp
```diff
@@ -271,48 +265,39 @@ bool createContext(void* platform_data) {
...
-    default:
-      snprintf(gContext.mError, sizeof(gContext.mError),
-               "Failed to get Java env. (%d)", res);
-      destroyContext();
-      return false;
-  }
+#define GET_PROP(name, trans)                            \
+  do {                                                   \
+    char _v[PROP_VALUE_MAX] = {0};                       \
+    if (__system_property_get(name, _v) == 0) {          \
```
AWoloszyn: IIRC this is typically a method we shouldn't get calling?

AWoloszyn: I just say ben's comment below, this is basically the same.

AWoloszyn:
Other than the question about __getproperty, looks ok to me.

core/os/device/deviceinfo/cc/android/query.cpp
pmuetschard: According to https://developer.android.com/about/versions/nougat/android-7.0-changes `__system_property_get` is the function to call. The warning there is simply that there is no standard specifying the names and values of the existing properties. However, the properties I'm querying here are the same as the ones queries by `android.os.Build`. BTW, the native code behind `android.os.Build` also uses the `__system_property_*` functions.

ben-clayton: I don't think the issue was with the property names, more that the function was simply not available on all versions of Android.
To be honest, I'm slightly surprised that this is accessible at all, as I thought there was an effort to classify this API as private (non-NDK-public). These release notes appear to suggest otherwise though.

pmuetschard: I think you are probably referring to this: https://stackoverflow.com/questions/28413530/api-to-get-android-system-properties-is-removed-in-arm64-platforms
As far as I can make out, it may at one point have been the idea to remove them, but that idea seems to have been overturned. There is a lot of Android code that depends on those functions to be present and accessible to apps. With those release notes, they are now guaranteed at the API level we build for.

AWoloszyn: from the 7.0 docs, I would agree with Pascal that these are now officially supported only because of the wording:

```
Use __system_property_get instead of the private property_get
```
Which implies that __system_property_get is public. So I think this should be fine.

PR #2619, opened 2019-02-21
Add new smoketests command

hevrard:
Gapit smoke_test runs a series of gapit commands on traces. It uses the
gapit command found in PATH, so it can be used to test any gapid build.


hevrard:
Added a small sample trace so one can try it with:
`gapit smoke_test /<path/to/gapid>/test/traces`

ben-clayton:
Should this really be part of `gapit`? Do we ever want the shipped binaries to contain this? If not, should it become its own executable under `cmd`?

hevrard:
> Should this really be part of `gapit`? Do we ever want the shipped binaries to contain this? If not, should it become its own executable under `cmd`?

I understand that this can be viewed as something external to gapit. On the other hand, adding this as a smoke_tests verb was @AWoloszyn advice, and gapit already has a `stress-test` verb.

Maybe we can put this here for now and move it somewhere else later if proved relevant?

ben-clayton:
As I'm responsible for `gapit`'s `stress-test` verb, I guess it's not fair for me to pick on this. 😄
I was hoping there was a flag on `app.Verb` that let you hide the verb by default, but there doesn't appear to be one.
Never mind.

hevrard:
Rebased and move code to a standalone new command 'smoketests'

cmd/smoketests/main.go
```diff
@@ -0,0 +1,205 @@
...
+		os.RemoveAll(tmpdir)
+	}
+	return nil
+}
+
+func testTrace(ctx context.Context, nbErr *int, tracepath string) error {
+	trace := filepath.Base(tracepath)
+	var err error
+	err = gapit(ctx, nbErr, "commands", tracepath)
+	if err != nil {
```
ben-clayton: nit: It is more idiomatic to keep the scopes of variables as short lived as possible.
This then becomes:

```
if err := gapit(ctx, nbErr, "commands", tracepath); err != nil {
```

cmd/smoketests/main.go
```diff
@@ -0,0 +1,205 @@
...
+
+		// Run smoke tests on trace under temporary directory
+		tracewd := filepath.Join(tmpdir, trace)
+		tracepath := filepath.Join(tracedir, trace)
+		if err := os.Mkdir(tracewd, 0777); err != nil {
+			return err
+		}
+		os.Chdir(tracewd)
+		err = testTrace(ctx, &nbErr, tracepath)
+		if err != nil {
```
ben-clayton: `if err := testTrace(ctx, &nbErr, tracepath); err != nil {`

cmd/smoketests/main.go
hevrard: Thanks, this pushed me to refactor testTrace to make it easier to understand which commands are run.

PR #2621, opened 2019-02-21
[gapid2apk] add standalone APK creation to export_replay

hevrard:
This is all the code needed for gapid2apk on the gapir side: detect replay resources in the apk assets, and if present use them to do the replay.

The missing part is the external command to actually create a new APK with replay resources: I have this implemented as a bash script, but we will want it as a regular command in Go for the sake or portability.

I open the PR now for reviews of the GAPIR side of things, and I will add the APK creation command soon.

### DIFF
--- a/cmd/gapir/cc/main.cpp
+++ b/cmd/gapir/cc/main.cpp
@@ -748,7 +748,8 @@ std::unique_ptr<ResourceCache> createCache(
       exit(0);
     }
   }
-  return std::move(onDiskCache);
+
+  return onDiskCache;
 #else   // TARGET_OS == GAPID_OS_LINUX || TARGET_OS == GAPID_OS_OSX
   if (onDiskCacheOpts.enabled) {
     GAPID_WARNING(
--- a/cmd/gapir/cc/main.cpp
+++ b/cmd/gapir/cc/main.cpp
@@ -44,6 +44,8 @@
 #if TARGET_OS == GAPID_OS_ANDROID
 #include <sys/stat.h>
 #include "android_native_app_glue.h"
+#include "gapir/cc/android/asset_replay_service.h"
+#include "gapir/cc/android/asset_resource_cache.h"
 #elif TARGET_OS == GAPID_OS_LINUX || TARGET_OS == GAPID_OS_OSX
 #include <dirent.h>
 #include <ftw.h>
@@ -54,6 +56,13 @@ using namespace core;
 using namespace gapir;
 
 namespace {
+
+enum ReplayMode {
+  kUnknown = 0,    // Can't determine replay type from arguments yet.
+  kConflict,       // Impossible combination of command line arguments.
+  kReplayServer,   // Run gapir as a server.
+  kReplayArchive,  // Replay an exported archive.
+};
 
 std::vector<uint32_t> memorySizes {
 // If we are on desktop, we can try more memory
@@ -264,11 +273,43 @@ std::unique_ptr<Server> Setup(const char* uri, const char* authToken,
       });
 }
 
+static int replayArchive(core::CrashHandler* crashHandler,
+                         std::unique_ptr<ResourceCache> resourceCache,
+                         gapir::ReplayService* replayArchiveService) {
+  // The directory consists an archive(resources.{index,data}) and payload.bin.
+  MemoryManager memoryManager(memorySizes);
+  std::unique_ptr<ResourceLoader> resLoader =
+      CachedResourceLoader::create(resourceCache.get(), nullptr);
+
+  std::unique_ptr<Context> context = Context::create(
+      replayArchiveService, *crashHandler, resLoader.get(), &memoryManager);
+
+  if (context->initialize("payload")) {
+    GAPID_DEBUG("Replay context initialized successfully");
+  } else {
+    GAPID_ERROR("Replay context initialization failed");
+    return EXIT_FAILURE;
+  }
+
+  GAPID_INFO("Replay started");
+  bool ok = context->interpret();
+  replayArchiveService->sendReplayFinished();
+  if (!context->cleanup()) {
+    GAPID_ERROR("Replay cleanup failed");
+    return EXIT_FAILURE;
+  }
+  GAPID_INFO("Replay %s", ok ? "finished successfully" : "failed");
+
+  return ok ? EXIT_SUCCESS : EXIT_FAILURE;
+}
+
 }  // anonymous namespace
 
 #if TARGET_OS == GAPID_OS_ANDROID
 
 namespace {
+
+const char* kReplayAssetToDetect = "replay_export/resources.index";
 
 template <typename... Args>
 jobject jni_call_o(JNIEnv* env, jobject obj, const char* name, const char* sig,
@@ -287,6 +328,7 @@ int jni_call_i(JNIEnv* env, jobject obj, const char* name, const char* sig,
 struct Options {
   int idleTimeoutSec = 0;
   std::string authToken = "";
+  ReplayMode mode = kUnknown;
 
   static Options Parse(struct android_app* app) {
     Options opts;
@@ -306,6 +348,20 @@ struct Options {
       const char* tmp = env->GetStringUTFChars((jstring)token, nullptr);
       opts.authToken = tmp;
       env->ReleaseStringUTFChars((jstring)token, tmp);
+    }
+
+    opts.mode = kReplayServer;
+
+    // Select replay archive mode if replay assets are detected
+    jobject j_asset_manager =
+        jni_call_o(env, app->activity->clazz, "getAssets",
+                   "()Landroid/content/res/AssetManager;");
+    AAssetManager* asset_manager = AAssetManager_fromJava(env, j_asset_manager);
+    AAsset* asset = AAssetManager_open(asset_manager, kReplayAssetToDetect,
+                                       AASSET_MODE_UNKNOWN);
+    if (asset != nullptr) {
+      opts.mode = kReplayArchive;
+      AAsset_close(asset);
     }
 
     app->activity->vm->DetachCurrentThread();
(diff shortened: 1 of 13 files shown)


ben-clayton:
Please ensure you squash these changes before submitting. Given that nobody has started reviewing yet, perhaps now would be a good time to squash to one or a few CLs?

cmd/gapit/trace2apk.go
```diff
@@ -0,0 +1,118 @@
...
+
+	return nil
+}
+
+func fileCopy(dest, src string) error {
+	s, err := os.Open(src)
+	if err != nil {
+		return err
+	}
+
```
ben-clayton: `defer s.Close()`

cmd/gapit/trace2apk.go
```diff
@@ -0,0 +1,118 @@
...
+	d, err := os.Create(dest)
+	if err != nil {
+		return err
+	}
+
```
ben-clayton: `defer d.Close()`

cmd/gapit/trace2apk.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		return err
+	}
+	defer client.Close()
+
+	if err := client.ExportReplay(ctx, capturePath, nil, "assets/export_replay", &service.ExportReplayOptions{}); err != nil {
+		return log.Err(ctx, err, "Failed to export replay")
+	}
+
+	refAPKPath, err := layout.GapidApk(ctx, device.AndroidARM64v8a)
```
AWoloszyn: We probably don't want to hard-code the ABI here

cmd/gapit/trace2apk.go
```diff
@@ -0,0 +1,118 @@
...
+	// TODO(hevrard): implement the addition of assets, manifest package renaming (still todo!), re-sign, re-zip in core/os/android/apk module
+
+	// aapt add assets
+	execCommand(ctx, "aapt", "add", "-v", newAPK, "assets/export_replay/resources.data", "assets/export_replay/resources.index", "assets/export_replay/payload.bin")
+	user, err := user.Current()
+	if err != nil {
+		return err
+	}
+
+	keystore := filepath.Join(user.HomeDir, ".android/debug.keystore")
```
AWoloszyn: What should we do if this doesn't exist?
We may want to consider creating our own keystore if this is the case.

cmd/gapit/trace2apk.go
```diff
@@ -0,0 +1,118 @@
...
+	execCommand(ctx, "apksigner", "sign", "--ks", keystore, "--ks-pass", "pass:android", newAPK)
+
+	return nil
+}
+
+func fileCopy(dest, src string) error {
```
AWoloszyn: It looks like we have a similar function in core/os/file.go

core/cc/archive.cpp
```diff
@@ -210,6 +210,10 @@ Archive::Archive(const std::string& archiveName)
     GAPID_FATAL("Unable to open archive index file %s", indexFilename.c_str());
   }
 
+  // Linux fopen() with mode "a" leads to reads from the beginning of file, but
+  // this is not true on Android, hence the explicit rewind() here
+  rewind(mIndexFile);
```
AWoloszyn: This seems counter to what people will expect, we may want to file a bug.

cmd/gapit/trace2apk.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/google/gapid/core/log"
+)
+
+type trace2apkVerb struct{ Trace2apkFlags }
+
+func init() {
+	verb := &trace2apkVerb{}
+	app.AddVerb(&app.Verb{
+		Name:      "trace2apk",
+		ShortHelp: "Create an stand-alone APK which replays a trace (off-screen)",
```
pau-baiget: Create an -> Create a

AWoloszyn:
FYI: This may be a future update, but I tried this, and the performance was rather abysmal. From what I could tell, almost all of the time was spend unzipping resources from the archive. (Which is somewhat expected).

cmd/gapit/trace2apk.go
```diff
@@ -0,0 +1,229 @@
...
+		return err
+	}
+
+	trace := filepath.Base(flags.Arg(0))
+	replayLabel := strings.TrimSuffix(trace, ".gfxtrace")
+	replayPackage := replayLabel + ".replay"
+	replayAPK := replayPackage + ".apk"
+	log.I(ctx, "Replay apk: %s", replayAPK)
+
+	// Extract reference APK
```
AWoloszyn: We should consider unzipping ourselves, rather than relying on the "unzip" command-line tool. Which won't exist on windows. Golang has this functionality more or less built in.
https://golangcode.com/unzip-files-in-go/

cmd/gapit/trace2apk.go
```diff
@@ -0,0 +1,229 @@
...
+	type ManifestEntries struct {
+		Label   string
+		Package string
+	}
+	var entries = ManifestEntries{replayLabel, replayPackage}
+	t.Execute(f, entries)
+	f.Close()
+
+	// Re-assemble new APK with the new manifest
+	if err := execCommand(ctx, "aapt2", "link", "-o", replayAPK, "--manifest", "NewManifest.xml", "-I", os.ExpandEnv("${ANDROID_SDK_HOME}/platforms/android-26/android.jar")); err != nil {
```
AWoloszyn: It may make sense to use the aapt2 from the Android SDK. That way more users will have it work.

AWoloszyn: `${ANDROID_SDK_HOME}/platforms/android-26/android.jar` This probably won't work on Windows.

We can `path.join(getenv(ANDROID_SDK_HOME),  ...)`
Also it would be a good idea allow the user specify an sdk directory that isn't ANDROID_SDK_HOME. Forcing env_vars is not nice.

cmd/gapit/trace2apk.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	// Add replay assets, uncompressed
+	assets, err := ioutil.ReadDir("assets/replay_export/")
+	if err != nil {
+		return err
+	}
+	for _, a := range assets {
+		// Arguments ("-0", "") ensure the asset will not be compressed
+		if err := execCommand(ctx, "aapt", "add", "-0", "", replayAPK, "assets/replay_export/"+a.Name()); err != nil {
```
AWoloszyn: Same as above with aapt.

cmd/gapit/trace2apk.go
```diff
@@ -0,0 +1,229 @@
...
+		}
+	}
+
+	// Add the replay libraries
+	// TODO(#2677): get rid of this workaround once ABI naming is sanitized
+	abi := capture.ABI.Name
+	if abi == "armeabi" {
+		abi = "armeabi-v7a"
+	}
+	if err := execCommand(ctx, "zip", "-u", replayAPK, "lib/"+abi+"/libgapir.so", "lib/"+abi+"/libVkLayer_VirtualSwapchain.so"); err != nil {
```
AWoloszyn: We should be able to zip directly from golang.

cmd/gapit/export_replay.go
```diff
@@ -98,5 +114,155 @@ func (verb *exportReplayVerb) Run(ctx context.Context, flags flag.FlagSet) error
...
+		}
+		capture := boxedCapture.(*service.Capture)
+
+		refAPKPath, err := layout.GapidApk(ctx, capture.ABI)
+		if err != nil {
+			return err
+		}
+
+		// Extract reference APK
+		if err := execCommand(ctx, "unzip", "-o", refAPKPath.String()); err != nil {
```
AWoloszyn: I will re-add these comments in case they got buried from the previous MRs.

We should avoid using command-line executables that we cannot guarantee exist. zip/unzip are not present on windows for example.

cmd/gapit/export_replay.go
```diff
@@ -98,5 +114,155 @@ func (verb *exportReplayVerb) Run(ctx context.Context, flags flag.FlagSet) error
...
+		}
+		type ManifestEntries struct {
+			Package string
+		}
+		var entries = ManifestEntries{replayPackage}
+		t.Execute(f, entries)
+		f.Close()
+
+		// Re-assemble new APK with the new manifest
+		if err := execCommand(ctx, "aapt2", "link", "-o", verb.Apk, "--manifest", "NewManifest.xml", "-I", os.ExpandEnv("${ANDROID_SDK_HOME}/platforms/android-26/android.jar")); err != nil {
```
AWoloszyn: We should use aapt2 from the ANDROID_SDK_HOME.
Also it would be good to be able to specify the android_sdk_home from the command-line as well.

`--sdk_root="foo"`.  If not set then use the env var.

AWoloszyn: Does this HAVE To be android-26/android.jar?
Can this be a different jar file? If it must be 26, that's fine I just want to make sure we are flexible if we can be.

cmd/gapit/export_replay.go
```diff
@@ -98,5 +114,155 @@ func (verb *exportReplayVerb) Run(ctx context.Context, flags flag.FlagSet) error
...
+		}
+
+		// Add replay assets, uncompressed
+		assets, err := ioutil.ReadDir("assets/replay_export/")
+		if err != nil {
+			return err
+		}
+		for _, a := range assets {
+			// Arguments ("-0", "") ensure the asset will not be compressed
+			if err := execCommand(ctx, "aapt", "add", "-0", "", replayAPK, "assets/replay_export/"+a.Name()); err != nil {
```
AWoloszyn: Same as above, we should use aapt2 from the SDK `build-tools/**/aapt2`

cmd/gapit/export_replay.go
```diff
@@ -98,5 +114,155 @@ func (verb *exportReplayVerb) Run(ctx context.Context, flags flag.FlagSet) error
...
+			}
+		}
+
+		// Add the replay libraries
+		// TODO(#2677): get rid of this workaround once ABI naming is sanitized
+		abi := capture.ABI.Name
+		if abi == "armeabi" {
+			abi = "armeabi-v7a"
+		}
+		if err := execCommand(ctx, "zip", "-u", replayAPK, "lib/"+abi+"/libgapir.so", "lib/"+abi+"/libVkLayer_VirtualSwapchain.so"); err != nil {
```
AWoloszyn: We can zip from golang itself.

core/cc/archive.cpp
hevrard: I filed a bug internally.

cmd/gapit/export_replay.go
```diff
@@ -98,5 +114,155 @@ func (verb *exportReplayVerb) Run(ctx context.Context, flags flag.FlagSet) error
...
+			return err
+		}
+
+		// Add the compiled Java code
+		if err := execCommand(ctx, "zip", "-uj", replayAPK, "classes.dex"); err != nil {
+			return err
+		}
+
+		// Sign the new APK
+		if err := execCommand(ctx, "apksigner", "sign", "--ks", os.ExpandEnv("${HOME}/.android/debug.keystore"), "--ks-pass", "pass:android", replayAPK); err != nil {
```
AWoloszyn: I dont think this env will work on Windows.
We should probably set up a dummy keystore if none exist.

AWoloszyn: I think we need to run zipalign on the zi pfile before apksigner as well?

cmd/gapit/export_replay.go
hevrard: It has to match the targetSdkVersion in `gapidapk/android/apk/AndroidManifest.xml.in`, which is currently 26. I have abstracted this in export_replay, but not across export_replay and the manifest template.

hevrard:
Thanks for the previous review, I tried to fix all remarks. This PR enables to create a stand-alone APK which replay a given trace. We are still facing issues when trying to trace the created APKs, but we can fix these issues later.

As it is a big change, we may think of splitting this in a few PRs, feel free to suggest such splitting.

cmd/gapit/export_replay.go
```diff
@@ -98,5 +127,293 @@ func (verb *exportReplayVerb) Run(ctx context.Context, flags flag.FlagSet) error
...
+		}
+		for _, a := range assets {
+			// Arguments ("-0", "") ensure the asset will not be compressed
+			if err := execCommand(ctx, aapt, "add", "-0", "", tmpAPK, path.Join(assetsPath, a.Name())); err != nil {
+				return err
+			}
+		}
+
+		// Add the replay libraries
+		// TODO(#2677): get rid of this workaround once ABI naming is sanitized
```
pmuetschard: this is done, no?

cmd/gapit/export_replay.go
```diff
@@ -98,5 +127,293 @@ func (verb *exportReplayVerb) Run(ctx context.Context, flags flag.FlagSet) error
...
+		if err := execCommand(ctx, apksigner, "sign", "--ks", keystorePath, "--ks-pass", "pass:android", replayAPKPath); err != nil {
+			return err
+		}
+
+	}
+
 	return nil
 }
+
+func execCommand(ctx context.Context, name string, args ...string) error {
```
pmuetschard: We have core/os/shell that provides a nice API to do this sort of thing.

cmd/gapit/export_replay.go
```diff
@@ -98,5 +127,293 @@ func (verb *exportReplayVerb) Run(ctx context.Context, flags flag.FlagSet) error
...
+	if err != nil {
+		return err
+	}
+	defer z.Close()
+
+	for _, f := range z.File {
+		destPath := filepath.Join(outputDir, f.Name)
+
+		// Protect from ZipSlip, see https://snyk.io/research/zip-slip-vulnerability
+		prefix := ""
```
pmuetschard: The prefix can be determined outside of the loop.

cmd/gapit/export_replay.go
```diff
@@ -98,5 +127,293 @@ func (verb *exportReplayVerb) Run(ctx context.Context, flags flag.FlagSet) error
...
+	out, err := cmd.CombinedOutput()
+	logger := log.From(ctx).Writer(log.Debug)
+	if err != nil {
+		logger = log.From(ctx).Writer(log.Error)
+	}
+	logger.Write(out)
+	return err
+}
+
+func unzip(ctx context.Context, archiveFilename string, outputDir string) error {
```
pmuetschard: move this to core/os/file/zip.go

cmd/gapit/flags.go
```diff
@@ -124,6 +124,8 @@ type (
 		OriginalDevice bool   `help:"export replay for the original device"`
 		Out            string `help:"output directory for commands and assets"`
 		OutputFrames   bool   `help:"generate trace that output frames(disable diagnostics)"`
+		Apk            string `help:"(experimental) create a stand-alone APK with that name which performs the replay"`
```
pmuetschard: Please add here that the apk name should have the format "<app_package>.apk"

gapis/api/gles/api/egl.api
```diff
@@ -357,14 +357,12 @@ cmd EGLBoolean eglSurfaceAttrib(EGLDisplay display,
   return ? // TODO
 }
 
-@no_replay
```
pmuetschard: Please move these API changes into their own commit with a detailed message of why we are making them.

AWoloszyn:
Other than the comments from @pmuetschard it looks good to me.

cmd/gapir/cc/main.cpp
```diff
@@ -341,35 +397,76 @@ void android_process(struct android_app* app, int32_t cmd) {
...
+
+  if (opts.mode == kReplayArchive) {
+    GAPID_INFO("Started Graphics API Replay from archive.");
+
+    waiting_thread = std::thread([&]() {
+      // It's important to use a different JNIEnv as it is a separate thread
+      JNIEnv* env;
+      app->activity->vm->AttachCurrentThread(&env, 0);
+
+      // Hugues: maybe we need to open assets in the main thread?
```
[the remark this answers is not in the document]
hevrard: I clarified the comment, and thanks for the feedback on TODO formats!

cmd/gapit/export_replay.go
hevrard: yep!

cmd/gapit/export_replay.go
hevrard: thanks for the pointer

hevrard:
Many thanks Pascal for the detailed feedback!
The EGL frame delimiters replay is now separated in #2708

hevrard:
The OSX build fails with `ERROR: Aborting VM command due to timeout of 10800 seconds`, have you seen that before? How can we allocate more time for the build?

PR #2630, opened 2019-02-27
Get tests building again.

AWoloszyn:
All tests except for depedencygraph2 tests are building.
I have 7 failures on my machine with these tests.
I have had to disable "postback_test.go" since it now relies
on internal APIs. It will fail until we fix it properly.


gapis/resolve/get_set_test.go
```diff
@@ -42,11 +42,15 @@ func newPathTest(ctx context.Context) *path.Capture {
...
 	}
-	p, err := capture.New(ctx, a, "test", h, nil, cmds)
+	p, err := capture.NewGraphicsCapture(ctx, a, "test", h, nil, cmds)
 	if err != nil {
 		log.F(ctx, true, "Couldn't create capture: %v", err)
 	}
-	return p
+	path, err := p.Path(ctx)
+	if err != nil {
+		log.F(ctx, true, "Couldn't create capture: %v", err)
```
pau-baiget: Do we want to have the exact error message as above?

PR #2631, opened 2019-02-28
Fix adb tests

AWoloszyn:


AWoloszyn:
This sits on top of #2630, only the top commit is new

PR #2632, opened 2019-02-28
Bapi decoder

AWoloszyn:


AWoloszyn:
This sits on top of #2631, only the top commit is new

PR #2634, opened 2019-02-28
Add smoketests flags -gapit -keep

hevrard:
-gapit <path/to/gapit>
-keep : force keeping the temporary directory, even if no errors were found

### DIFF
--- a/cmd/smoketests/main.go
+++ b/cmd/smoketests/main.go
@@ -30,8 +30,9 @@ import (
 )
 
 var (
-	tracesArg = flag.String("traces", "traces", "The directory containing traces to run smoke tests on")
 	gapitArg  = flag.String("gapit", "gapit", "Path to gapit executable")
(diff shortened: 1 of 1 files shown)


cmd/smoketests/main.go
```diff
@@ -101,20 +112,24 @@ func run(ctx context.Context) error {
...
 	}
 
-	if nbErr > 0 {
-		errStr := "error"
-		if nbErr > 1 {
-			errStr = "errors"
-		}
-		log.I(ctx, "%d %s found, see logs in %s", nbErr, errStr, tmpdir)
+	// Print number of errors
+	errStr := "error"
```
pau-baiget: You can do the same using "error(s)". In the case nbError == 0 you'll have "found 0 error" which is awkward.

hevrard: Good spot! I kept the logic but fixed the message in case of 0 errors.

PR #2636, opened 2019-02-28
Fix memory test

AWoloszyn:
Sits on top of #2635. The only interesting commit is the last one.


AWoloszyn:
Good catch, I have updated this to reflect that

PR #2642, opened 2019-02-28
Add headless build tag for the tests.

AWoloszyn:
This allows us to disable some tests if the target machine does not have a GPU.


pmuetschard:
I would prefer we use bazel's tagging feature for this (see https://docs.bazel.build/versions/master/be/common-definitions.html#common.tags and https://docs.bazel.build/versions/master/command-line-reference.html#flag--test_tag_filters).

I propose we add a `needs_gpu` and then we can run them with the `--test_tag_filters=-needs_gpu` flag on the bots. The advantage is that this will work always and not create empty test targets.

AWoloszyn:
Sounds good. I will make that change tomorrow.

PR #2645, opened 2019-03-01
Correctly set the target/min SDK API.

### LINKED ISSUE #2389: Tracing Failed   Cause: Could not start trace   Cause: loadGAPII   Cause: Exception raised calling: public static void java.lang.System.load(java.lang.String)java.lang.UnsatisfiedLinkError: dlopen failed: cannot locate symbol "__aeabi_memclr4" referenced by "/data/app/com.google.android.gapid.armeabi-1/lib/arm/libgapii.so"...
GAPID Version: 1.2.1:9acb3d602489c7389134f043d4a1e9c12b6e3ff5
OS: mac os x x86_64

Please provide detailed steps that led to the error and copy-paste the stack trace.
Extra details from the logs and the trace file would be extra helpful.

pmuetschard:
Sigh. For the NDK rule, the API level is really the minimum API level, since that's what will be linked against, while for the SDK, it's the target API. Our native code requires API level 21 (5.0 Lollipop), so updated the manifests as well.

Fixes #2389

### DIFF
--- a/tools/build/workspace.bzl
+++ b/tools/build/workspace.bzl
@@ -199,14 +199,14 @@ def gapid_dependencies(android = True, mingw = True, locals = {}):
             native.android_sdk_repository,
             name = "androidsdk",
             locals = locals,
-            api_level = 26,
+            api_level = 26, # This is the target API
         )
 
         maybe_repository(
             native.android_ndk_repository,
             name = "androidndk",
             locals = locals,
-            api_level = 26,
+            api_level = 21, # This is the minimum API
         )
 
         maybe_repository(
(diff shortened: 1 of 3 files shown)


tools/build/workspace.bzl
```diff
@@ -199,14 +199,14 @@ def gapid_dependencies(android = True, mingw = True, locals = {}):
             native.android_sdk_repository,
             name = "androidsdk",
             locals = locals,
-            api_level = 26,
+            api_level = 26, # This is the target API
```
AWoloszyn: These are delightfully consistent with each other :\

PR #2652, opened 2019-03-05
 Fix reading of depth textures/buffers in ODR.

### LINKED ISSUE #2584: Fails to display depth textures.
If I try to see the contents of a depth texture, gapid gets confused about the size of the image, and fails to display it.

`image data size (0x400000) did not match expected (0x100000) for dimensions 2048x2048x1`

It seems that it thinks that GL_UNSIGNED_INT has the size of 1 byte. (The format of the texture is GL_DEPTH_COMPONENT, GL_UNSIGNED_INT)

image

pmuetschard:
Depth textures, and framebuffers backed by depth textures, that where created with an unsized internal format require special handling when read back from the GPU. For example, when blitting a FB, the source and target formats will have to match and thus have to both be created with a sized or unsized internal format.

Fixes #2584

### DIFF
--- a/core/cc/gl/formats.cpp
+++ b/core/cc/gl/formats.cpp
@@ -93,7 +93,7 @@ bool getDepthStencilFormat(int d, int s, uint32_t& depth, uint32_t& stencil) {
           stencil = GL_DEPTH24_STENCIL8;
           return true;
         case 32:
-          depth = GL_DEPTH24_STENCIL8;
+          depth = GL_DEPTH32F_STENCIL8;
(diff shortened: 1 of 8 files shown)


pmuetschard:
Note that the field added in 60bef28 is only used in one place in this PR, but will need to be used in other places as well. For example, my test app for this doesn't replay correctly in MEC, due to the fact that you can use `GL_LINEAR` for depth textures with an unsized internal format, while for sized internal format textures you are required to use `GL_NEAREST`. MEC replay, currently, always re-creates the textures using a sized internal format. See #2653

PR #2657, opened 2019-03-06
Fix GAPIR Android app termination

hevrard:
After calling ANativeActivity_finish(), the native app should keep on
polling events and only terminate when app->destroyRequested becomes
true.

This ensure a proper termination of the GAPIR app on Android.

### DIFF
--- a/cmd/gapir/cc/main.cpp
+++ b/cmd/gapir/cc/main.cpp
@@ -374,6 +374,7 @@ void android_main(struct android_app* app) {
 
   app->onAppCmd = android_process;
 
+  bool finishing = false;
   bool alive = true;
   while (alive) {
     int fdesc;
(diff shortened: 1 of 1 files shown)


cmd/gapir/cc/main.cpp
```diff
@@ -393,12 +396,16 @@ void android_main(struct android_app* app) {
...
+      ANativeActivity_finish(app->activity);
+      // Note that we need to keep on polling events, eventually APP_CMD_DESTROY
+      // will pop-up after which app->destroyRequested will be true, enabling us
+      // to properly exit the main loop.
     }
   }
   waiting_thread.join();
-  ANativeActivity_finish(app->activity);
+  GAPID_INFO("End of Graphics API Replay");
+  return;
```
AWoloszyn: We probably do want to unlink the file if it is the server at least.

cmd/gapir/cc/main.cpp
```diff
@@ -393,12 +396,16 @@ void android_main(struct android_app* app) {
     }
 
     if (serverIsDone) {
-      unlink(socket_file_path.c_str());
-      alive = false;
+      // Start termination of the app
+      ANativeActivity_finish(app->activity);
```
pmuetschard: This now gets executed on every iteration, as soon as serverIsDone is true. Setting it back to false seems a bit dirty, but would work.

cmd/gapir/cc/main.cpp
pmuetschard: It is in the `app->destroyRequested` block. However, I think it would be better to call unlink after joining the waiting thread, i.e. just before exiting.

hevrard: Thanks, moved unliking to be after the waiting_thread.join()

cmd/gapir/cc/main.cpp
hevrard: Good point! I added a "finishing" boolean to avoid calling ANativeActivity_finish() several time.

pmuetschard:
Please squash when committing, thanks!

PR #2674, opened 2019-03-13
Added a -gapir-nofallback flag.

### LINKED ISSUE #2589: Return error when the requested gapir device is not found
One can request a specific device to replay on with e.g. `-gapir-device`, but currently if that device is not found, then there is a fallback to some other device. This is problematic because when a device becomes unreachable due to e.g. adb issues, gapid will happily go on doing replays on the desktop.

I think gapid should properly return an error when the requested device is not available.

@AWoloszyn we chatted about this yesterday and I think you may have a quick fix somewhere already?

AWoloszyn:
This will prevent GAPIR from falling back to a default
device if the given device could not be found.

Fixes #2589.

### DIFF
--- a/cmd/gapit/common.go
+++ b/cmd/gapit/common.go
@@ -269,6 +269,10 @@ func getDevice(ctx context.Context, client client.Client, capture *path.Capture,
 		return selected, nil
 	}
 
+	if flags.NoFallback {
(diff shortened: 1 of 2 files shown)


hevrard:
Thanks for this!

I would be more on the side of a `-gapir-fallback` flag, i.e. do not fallback by default, but only when told so.

What do you think?

PR #2676, opened 2019-03-13
Make gapit video behave like gapit screenshot.

### LINKED ISSUE #1863: gapit: inconsistent file output paths
When specifying relative paths:

`gapit screenshot` will write the png relative to the CWD
`gapit video` will write the mp4 relative to the trace file

AWoloszyn:
This outputs the resulting video relative to CWD if not
otherwise specified.

Fixes #1863

### DIFF
--- a/cmd/gapit/video.go
+++ b/cmd/gapit/video.go
@@ -25,6 +25,7 @@ import (
 	"image/png"
 	"io"
 	"os"
+	fp "path/filepath"
 	"strings"
 	"sync/atomic"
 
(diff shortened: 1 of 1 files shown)


cmd/gapit/video.go
```diff
@@ -25,6 +25,7 @@ import (
 	"image/png"
 	"io"
 	"os"
+	fp "path/filepath"
```
pmuetschard: why the alias? do we have any style guide when to use aliases?

AWoloszyn: Because we use "filepath" elsewhere as a variable name.

pmuetschard: Ack.

PR #2678, opened 2019-03-14
Add more smoketests for 'commands' verb

hevrard:


hevrard:
These additional test lead to discover #2679

PR #2680, opened 2019-03-14
Gracefully fail when context argument is out of range

### LINKED ISSUE #2679: Panic when gapit commands -context is too high
Calling `gapit commands -context 100 foo.gfxtrace` leads to a panic, while it should gracefully fail.

Example of panic:
```
panic: runtime error: index out of range [recovered]
	panic: runtime error: index out of range

goroutine 1 [running]:
github.com/google/gapid/core/app/crash.Crash(0xf6ecc0, 0x177be30)
	core/app/crash/crash.go:89 +0xe1
github.com/google/gapid/core/app.Run.func1()
	core/app/run.go:161 +0xec
panic(0xf6ecc0, 0x177be30)
	GOROOT/src/runtime/panic.go:513 +0x1b9
main.CommandFilterFlags.commandFilter(0x2a, 0x115b460, 0xc000257110, 0x7fe3fcd37048, 0xc0002a80c0, 0xc0002562d0, 0x0, 0x0, 0x0)
	cmd/gapit/common.go:52 +0x2f0
main.(*commandsVerb).Run(0xc000234500, 0x115b460, 0xc000257110, 0xc000284200, 0x0, 0x0, 0x1, 0xc000257200, 0xc000256990, 0xc00027dd30, ...)
	cmd/gapit/commands.go:61 +0x493
github.com/google/gapid/core/app.(*Verb).Invoke(0x17d7500, 0x115b460, 0xc000257110, 0xc00027dd00, 0x4, 0x4, 0x0, 0x0)
	core/app/verbs.go:92 +0x742
github.com/google/gapid/core/app.VerbMain(0x115b460, 0xc000257110, 0x0, 0x0)
	core/app/verbs.go:123 +0x9a
github.com/google/gapid/core/app.Run(0x1097750)
	core/app/run.go:253 +0xbc5
main.main()
	cmd/gapit/main.go:24 +0x7d
```

hevrard:
Fix #2679

### DIFF
--- a/cmd/gapit/common.go
+++ b/cmd/gapit/common.go
@@ -45,14 +45,15 @@ import (
 func (f CommandFilterFlags) commandFilter(ctx context.Context, client service.Service, p *path.Capture) (*path.CommandFilter, error) {
 	filter := &path.CommandFilter{}
 	if f.Context >= 0 {
-		contexts, err := client.Get(ctx, p.Contexts().Path(), nil)
+		boxedContexts, err := client.Get(ctx, p.Contexts().Path(), nil)
 		if err != nil {
 			return nil, log.Err(ctx, err, "Failed to load the contexts")
 		}
-		if n := len(contexts.(*service.Contexts).List); f.Context >= n {
+		contexts := boxedContexts.(*service.Contexts)
+		if n := len(contexts.List); f.Context >= n {
 			return nil, log.Errf(ctx, err, "Context %d is out of range [0..%d]", f.Context, n-1)
 		}
-		filter.Context = contexts.(*service.Contexts).List[f.Context].ID
+		filter.Context = contexts.List[f.Context].ID
 	}
 	return filter, nil
 }
--- a/cmd/gapit/common.go
+++ b/cmd/gapit/common.go
@@ -49,7 +49,14 @@ func (f CommandFilterFlags) commandFilter(ctx context.Context, client service.Se
 		if err != nil {
 			return nil, log.Err(ctx, err, "Failed to load the contexts")
 		}
-		filter.Context = contexts.(*service.Contexts).List[f.Context].ID
+		numContexts := len(contexts.(*service.Contexts).List)
+		// f.Context is an ID, so it must be strictly inferior to numContexts
+		if f.Context < numContexts {
+			filter.Context = contexts.(*service.Contexts).List[f.Context].ID
+		} else {
+			msg := fmt.Sprintf("Context %d is out of range [0..%d]", f.Context, numContexts-1)
+			return nil, log.Err(ctx, err, msg)
+		}
 	}
 	return filter, nil
 }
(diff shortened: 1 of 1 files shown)


cmd/gapit/common.go
```diff
@@ -49,7 +49,14 @@ func (f CommandFilterFlags) commandFilter(ctx context.Context, client service.Se
...
 			return nil, log.Err(ctx, err, "Failed to load the contexts")
 		}
-		filter.Context = contexts.(*service.Contexts).List[f.Context].ID
+		numContexts := len(contexts.(*service.Contexts).List)
+		// f.Context is an ID, so it must be strictly inferior to numContexts
+		if f.Context < numContexts {
+			filter.Context = contexts.(*service.Contexts).List[f.Context].ID
+		} else {
+			msg := fmt.Sprintf("Context %d is out of range [0..%d]", f.Context, numContexts-1)
+			return nil, log.Err(ctx, err, msg)
```
ben-clayton: Nit: You can use log.Errf()

cmd/gapit/common.go
```diff
@@ -49,7 +49,14 @@ func (f CommandFilterFlags) commandFilter(ctx context.Context, client service.Se
 		if err != nil {
 			return nil, log.Err(ctx, err, "Failed to load the contexts")
 		}
-		filter.Context = contexts.(*service.Contexts).List[f.Context].ID
+		numContexts := len(contexts.(*service.Contexts).List)
+		// f.Context is an ID, so it must be strictly inferior to numContexts
+		if f.Context < numContexts {
```
ben-clayton: More nit: It is more idiomatic to check for errors and return, but have the main code flow be error free. For example:

```go
if n := len(contexts.(*service.Contexts).List); f.Context < n {
    return nil, log.Errf(ctx, err, "Context %d is out of range [0..%d]", f.Context, n-1)
}
filter.Context = contexts.(*service.Contexts).List[f.Context].ID
```

hevrard: Thanks for the feedback, applied!

I understand how it feels more idiomatic, although to be honest I find it less readable as the actual test is in the middle of the long if condition line.

Nit: you got the test wrong ;-)

ben-clayton: Heh, I haven't had my coffee today. </rubbish excuse> :)

> I find it less readable as the actual test is in the middle of the long if condition line.

True, but it is only long because you are doing a cast as well as a field access. This is more a more common pattern:

```go
contextsBoxed, err := client.Get(ctx, p.Contexts().Path(), nil)
if err != nil {
	return nil, log.Err(ctx, err, "Failed to load the contexts")
}
contexts := contextsBoxed.(*service.Contexts)

if n := len(contexts.List); f.Context >= n {
    return nil, log.Errf(ctx, err, "Context %d is out of range [0..%d]", f.Context, n-1)
}
filter.Context = contexts.List[f.Context].ID
```

But that's really nitpicking.

hevrard: Looks better! Applied, with 'boxedContexts' as variable name as it is how most boxed variables are named.

PR #2681, opened 2019-03-14
Homogenize Arm ABI naming

### LINKED ISSUE #2677: Get rid of 'armeabi' ABI naming
The `armeabi` ABI is depreciated, all references should be replaced with `armeabi-v7a`, including naming of the `gapid-armeabi.apk`.

hevrard:
Fix #2677 

Naming should now be armeabi-v7a and arm64-v8a everywhere, unless in Android package names which do not tolerate hypens, thus package names are armeabiv7a and arm64v8a.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -86,8 +86,8 @@ copy_to(
         ],
     }),
     rename = {
-        "armeabi-v7a.apk": "gapid-armeabi.apk",
-        "arm64-v8a.apk": "gapid-aarch64.apk",
+        "armeabi-v7a.apk": "gapid-armeabi-v7a.apk",
+        "arm64-v8a.apk": "gapid-arm64-v8a.apk",
         "x86.apk": "gapid-x86.apk",
     },
     to = "pkg",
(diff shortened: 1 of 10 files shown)


BUILD.bazel
```diff
@@ -86,8 +86,8 @@ copy_to(
         ],
     }),
     rename = {
-        "armeabi-v7a.apk": "gapid-armeabi.apk",
-        "arm64-v8a.apk": "gapid-aarch64.apk",
+        "armeabi-v7a.apk": "gapid-armeabi-v7a.apk",
```
AWoloszyn: We might have to update the packaging scripts in the kokoro directory (I am not actually sure about this).

hevrard:
Good point, actually it looks like this CL reverts the naming to match the kokoro README:
```
$ find kokoro -type f | xargs grep -i arm
kokoro/README.md:├─ gapid-arm64-v8a.apk
kokoro/README.md:├─ gapid-armeabi-v7a.apk
kokoro/README.md:├─ gapid-arm64-v8a.apk
kokoro/README.md:├─ gapid-armeabi-v7a.apk
kokoro/README.md:├─ gapid-arm64-v8a.apk
kokoro/README.md:├─ gapid-armeabi-v7a.apk
```

No other mention of "arm" under kokoro, so I think we're safe on that side?

PR #2690, opened 2019-03-25
Fix ABI renaming: sanitize supported ABI list

hevrard:


hevrard:
Abandon this PR: `armeabi` should be just treated as an unsupported architecture.

PR #2691, opened 2019-03-26
First pass on adding multi-planar images.

AWoloszyn:
Still TODO: UI work. This will be done in a follow-up.


gapis/api/templates/mutate.go.tmpl
```diff
@@ -35,6 +35,7 @@
     "github.com/google/gapid/gapis/replay/value"
     "github.com/google/gapid/gapis/stringtable"
 
+
```
ben-clayton: Nit: remove.

PR #2694, opened 2019-03-27
Use logging function for error reporting in deviceinfo

hevrard:


hevrard:
I kept std::cout for regular output as the logging function pollute them with timestamps etc, which is not acceptable for this tool which is expected to output JSON.

PR #2696, opened 2019-03-28
Initial Perfetto Support in GAPIS

pmuetschard:
- adds the Perfetto repository, build rules to build the config proto and trace processor along with its dependency sqlite.
- adds some of the scaffolding for taking a Perfetto trace in GAPIS (not yet exposed to the GAPIS API, though)
- adds support for opening a Perfetto trace in GAPIS via the API
- adds a GAPIS API to run queries against a Perfetto trace


gapis/perfetto/cc/processor.cpp
```diff
@@ -0,0 +1,63 @@
...
+  ptp::Config config;
+  config.window_size_ns = window_size_ns;
+
+  return ptp::TraceProcessor::CreateInstance(config).release();
+}
+
+bool parse_data(processor processor, const void* data, size_t size) {
+  ptp::TraceProcessor* p = static_cast<ptp::TraceProcessor*>(processor);
+  // TODO: do we need this copy?
+  std::unique_ptr<uint8_t[]> buf(new uint8_t[size]);
```
AWoloszyn: Eww this is an ugly API.
```
virtual bool Parse(std::unique_ptr<uint8_t[]>, size_t) = 0;
```

You will need the copy unless we can allocate the data with `new` but I think this is meant to be exported to go, so the data is coming from "magic place"

gapis/perfetto/processor.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+type Result struct {
+	Ready  task.Signal
+	Result service.PerfettoQueryResult
+	done   task.Task
+}
+
+func NewProcessor(ctx context.Context, data []byte) (*Processor, error) {
+	p := C.new_processor(60 * 1000 * 1000 * 1000)
```
AWoloszyn: Can we name this magic constant?

gapis/perfetto/trace.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/google/gapid/core/event/task"
+	"github.com/google/gapid/core/log"
+	"github.com/google/gapid/core/os/android"
+	"github.com/google/gapid/core/os/android/adb"
+	"github.com/google/gapid/core/os/file"
+	"github.com/google/gapid/gapis/service"
+)
+
+const (
+	perfettoTraceFile = "/data/misc/perfetto-traces/gapis-trace"
```
AWoloszyn: Is this the path on the Android device where we are putting the trace?

gapis/service/service.proto
```diff
@@ -68,6 +68,30 @@ message Devices {
 message Events {
   repeated Event list = 1;
 }
+// This needs to be binary compatible with perfetto.protos.RawQueryResult.
```
AWoloszyn: Any way we can check this at build-time?

PR #2706, opened 2019-04-03
Fix smoketests and run them on kokoro

hevrard:


hevrard:
This change fixes the smoketests command and run it as part of CI builds.

Running smoketests takes less than 2 minutes on any platform, compiling smoketests may also add a little overhead which is certainly negligible compare to the whole build process.

PR #2709, opened 2019-04-05
Support for taking a Perfetto trace in GAPIS

pmuetschard:
Also includes the fixes asked for in #2696, which I accidentally dropped in excitement when the build finally went green.


gapis/perfetto/trace.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/google/gapid/core/os/android"
+	"github.com/google/gapid/core/os/android/adb"
+	"github.com/google/gapid/core/os/file"
+	"github.com/google/gapid/gapis/service"
+)
+
+const (
+	// perfettoTraceFile is the location on the device where we'll ask Perfetto
+	// to store the trace data while tracing.
+	perfettoTraceFile = "/data/misc/perfetto-traces/gapis-trace"
```
AWoloszyn: This should probably be attached to the device. Especially if we want to be able to take perfetto traces elsewhere.

pmuetschard: I've moved this into an `android` subdirectory. PTAL

AWoloszyn:
Awesome thanks!

PR #2713, opened 2019-04-10
Some Small UI Improvements

pmuetschard:


gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -703,6 +729,10 @@ private File getOutputFile() {
         String dir = directory.getText();
         return dir.isEmpty() ? new File(name) : new File(dir, name);
       }
+
+      private static boolean isPerfetto(TraceTypeCapabilities config) {
```
hevrard: This assumes that a trace is either Perfetto, or gfxtrace by default. To be future proof of more than two types of traces, maybe we can introduce an enumeration of possible types of traces?

gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -649,8 +672,11 @@ public TraceRequest getTraceRequest(Settings settings) {
           options.setDisablePcs(disablePcs.getSelection());
         }
 
-        if (config.getType() == TraceType.Perfetto) {
-          readPerfettoConfig(options);
+        if (isPerfetto(config)) {
+          int duration = frameCount.getSelection() * 1000;
+          // TODO: this isn't really unlimitted.
```
hevrard: Use the `TODO(username):` format?

gapic/src/main/com/google/gapid/views/TracerDialog.java
pmuetschard: Yes, this enum exists and this function is there to check that enum. You are of course correct that if we add another trace type, everywhere where `isPerfetto` is called would need to be changed. However, it would need to be changed anyways, so for now I'd rather just stick with this shorthand

gapic/src/main/com/google/gapid/views/TracerDialog.java
pmuetschard: The username portion is optional and we have mostly not been using it in GAPID (small team and all).

PR #2725, opened 2019-04-22
Initial Version of the Perfetto System Profiler UI

pmuetschard:


pmuetschard:
Will send another PR with an updated version of this... soonish...

PR #2744, opened 2019-05-09
Update to bazel 0.25.1

pmuetschard:


hevrard:
Linux and Windows builds still hit bazel errors apparently.

hevrard:
It's all green expect on Windows:
```
ERROR: C:/users/builder/_bazel_kbuilder/fsjplg7y/external/gapid/tools/build/rules/repository.bzl:100:9: no such package '@androidsdk//': C:\users\builder\_bazel_kbuilder\fsjplg7y\external\androidsdk and referenced by '//external:android/dx_jar_import'
ERROR: Analysis of target '//:pkg' failed; build aborted: Analysis failed
```
I can't spot reference to 'androidsdk' in your changes, so maybe it's just Kokoro being flaky?

I'll start a new kokoro build to see if it goes through.

BUILDING.md
```diff
@@ -1,6 +1,6 @@
 # Building GAPID
 
-GAPID uses the [Bazel build system](https://bazel.build/). The current minimum required version of bazel is 0.18.1.
+GAPID uses the [Bazel build system](https://bazel.build/). The current minimum required version of bazel is 0.25.1.
```
AWoloszyn: Gross big jump. But I assume its basically impossible to avoid.

pmuetschard: The reality is that we were building with 0.20 before, but yeah, we should keep up with it a bit more.

PR #2745, opened 2019-05-10
Add GLES dependency graph unit test template, and glClear() dependencies

hevrard:
Also introduce the template for dependency graph "unit tests"

### DIFF
--- a/gapis/api/gles/dependencygraph2_test.go
+++ b/gapis/api/gles/dependencygraph2_test.go
@@ -1,0 +1,219 @@
+// Copyright (C) 2019 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package gles_test
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/google/gapid/core/assert"
+	"github.com/google/gapid/core/log"
+	"github.com/google/gapid/core/memory/arena"
+	"github.com/google/gapid/core/os/device"
+	"github.com/google/gapid/core/os/device/bind"
+	"github.com/google/gapid/gapis/api"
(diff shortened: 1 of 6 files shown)


hevrard:
I'm mostly interested in the feedback on `dependencygraph2_test.go`. The idea is to have unit tests which build a small capture, get its dependency graph, and check that expected dependencies are indeed present.

Let me know if the current approach looks good or not, and how to improve it.

pmuetschard:
looks good to me.

hevrard:
Updated with dependencies for `glClear{Color,Depthf,Stencil}`

gapis/api/gles/dependencygraph2_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+	for dep, present := range mapToCheck {
+		assert.For(tc.ctx, fmt.Sprintf("Dependency: %v", dep)).ThatBoolean(present).IsTrue()
+	}
+}
+
+// Actual tests
+
+// glClear(GL_COLOR_BUFFER_BIT) depends on glClearColor()
+func TestDependencyGlClearColor(t *testing.T) {
```
AWoloszyn: One comment I would like to make is that a dependency graph that just returned "KEEP EVERYTHING" would also pass these tests. 

If we want to write unit tests for the dependency graph, we should also at least make sure that some of the edges are NOT present.

e.g. 
glClear(GL_COLOR_BUFFER_BIT) does not depend on glClearDepthf()

Obviously we can't be exhaustive, but we should at least attempt in some cases to make sure it is not a complete graph.

gapis/api/gles/dependencygraph2_test.go
```diff
@@ -0,0 +1,189 @@
...
+	"github.com/google/gapid/gapis/memory"
+	"github.com/google/gapid/gapis/resolve/dependencygraph2"
+)
+
+// Test helpers
+
+// testDependencyCmdName is used as a map key to easily express
+// expected dependencies
+type testDependencyCmdName struct {
+	sourceCmdName, targetCmdName string
```
bjoeris: It seems that identifying nodes based only on the command name will only work for very small tests (with just 1 call to each distinct command). But it does work for the tests here.

hevrard: Yes I also though about this, I feel we may have to enrich the way we identify a node later on. I feel like I need to face more complex captures to know what is best to do, so I'll leave this for a future change.

gapis/api/gles/dependencygraph2_test.go
hevrard: Very good point, I updated the unit test template to take two list of dependencies: expected and unexpected dependencies.

PR #2748, opened 2019-05-13
Make the vulkan.api file publicly visible

aellem:
This allows other Bazel builds to reference the file.

### DIFF
--- a/gapis/api/vulkan/BUILD.bazel
+++ b/gapis/api/vulkan/BUILD.bazel
@@ -29,6 +29,12 @@ filegroup(
     visibility = ["//visibility:public"],
 )
 
+filegroup(
+    name = "api_file",
+    srcs = ["vulkan.api"],
+    visibility = ["//visibility:public"],
+)
+
 api_library(
     name = "api",
     api = "vulkan.api",


AWoloszyn:
@pmuetschard Can you take a look at this? 
I think it will be ok, but would like to get a second opinion.

pmuetschard:
It's already part of `api_files`, why do we need this one?

aellem:
So I can access the main api file from outside GAPID, to use apic as part
of a Vulkan layer's build system.


*From: *Pascal Muetschard <notifications@github.com>
*Date: *Mon, May 13, 2019 at 4:38 PM
*To: *google/gapid
*Cc: *Andrew Ellem, Author

It's already part of api_files, why do we need this one?

pmuetschard:
If you are using apic anyways, just depend on the public `api_library` target.

aellem:
We don't depend on apic as a target from GAPID, we have use a pre-built binary of apic.  Depending on api_library brings in all of GAPID's other build dependencies which I'd like to avoid.

pmuetschard:
This sounds very hacky, but OK. Yet, you still will need all the other api files that are imported from the vulkan.api file, which makes me think that you should use the `api_files` target instead, which bundles all the necessary files.
What does the build rule look like where you want to use this?

aellem:
Alternative workaround - create a Bazel repo w/copy of .api files.

PR #2756, opened 2019-05-16
Assert that cache load is a success

### LINKED ISSUE #2729: Variable is only written into, never read
This variable is only write into:

https://github.com/google/gapid/blob/4f00997554b7d863349e0121ab03618785b81780/gapir/cc/cached_resource_loader.cpp#L68

Should revisit the logic and probably report as soon as there is a failure in loadCache

hevrard:
mCache->loadCache() should always succeed given that
mCache->hasCache() had just returned true.

Fix #2729

### DIFF
--- a/gapir/cc/cached_resource_loader.cpp
+++ b/gapir/cc/cached_resource_loader.cpp
@@ -17,6 +17,8 @@
 #include "cached_resource_loader.h"
 #include "resource.h"
 
+#include "core/cc/assert.h"
+
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


gapir/cc/cached_resource_loader.cpp
```diff
@@ -57,15 +59,14 @@ bool CachedResourceLoader::load(const Resource* resources, size_t count,
...
 
-  bool succeeded = true;
-
   ResourceLoadingBatch batch;
   uint8_t* dst = reinterpret_cast<uint8_t*>(target);
   for (size_t i = 0; i < count; i++) {
     const auto& r = resources[i];
     // Check cache first
     if (mCache->hasCache(r)) {
-      succeeded &= mCache->loadCache(r, static_cast<void*>(dst));
```
pau-baiget: There was no check on "succeeded" at the end? Strange.

AWoloszyn: This was because 
`
if (mCache->hasCache(r)) == true implies   mCache->loadCache(r, static_cast<void*>(dst)); == true
`

And we have no other option if we cant actually load from the cache anyway (if it is in there), so asserting is likely the way to go.

PR #2758, opened 2019-05-16
Change the RPC for querying timestamps to stream RPC

RenfengLiu:


AWoloszyn:
This changes the server interface, but does not actually update the logic to be streaming. We should make this change in the context of a larger change to update the entire profile command to have streamed responses. (I.e. send the timestamps back as we get them).

PR #2760, opened 2019-05-17
Include timestamps in export_replay.

hysw:
Also fix query_timestamps commandpool allocation.

### DIFF
--- a/cmd/gapit/export_replay.go
+++ b/cmd/gapit/export_replay.go
@@ -88,7 +88,11 @@ func (verb *exportReplayVerb) Run(ctx context.Context, flags flag.FlagSet) error
 	}
 
 	var fbreqs []*service.GetFramebufferAttachmentRequest
-	if verb.OutputFrames {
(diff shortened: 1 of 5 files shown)


gapis/api/templates/vulkan_gfx_api_extras.tmpl
```diff
@@ -689,6 +689,7 @@ bool Vulkan::replayDebugReportCallback(uint32_t flags, uint32_t objectType,
      << ", Object: " << object << ", LayerPrefix: " << pLayerPrefix
      << ", Layer Mesage Code: " << messageCode << ", Message: " << pMessage;
   ctx->onDebugMessage(sev, Vulkan::INDEX, ss.str().c_str());
+  GAPID_ERROR("Debug Message %s", ss.str().c_str());
```
AWoloszyn: We should probably pull this out.

hysw:
> Looks good thanks!
> Ideally we can split this into 2 commits, but no strong need, the second one is pretty straight-forward.

Done.

PR #2762, opened 2019-05-17
Allow unsized internalformat for GLES state reconstruction (#2653)

### LINKED ISSUE #2653: Textures may be re-created incorrectly in GLES MEC traces.
GLES state reconstruction currently always recreates textures using a sized internal format, even if the application had created them with an unsized internal format. Textures behave differently depending on the "sized-ness" of the internal format and so MEC replay can be incorrect.

For example, depth textures with an unsized internal format may be sampled with `GL_LINEAR` while depth textures with a sized internal format may only be sampled with `GL_NEAREST`.

mikaelpessa:
Fixes #2653

### DIFF
--- a/gapis/api/gles/image.go
+++ b/gapis/api/gles/image.go
@@ -26,11 +26,47 @@ import (
 	"github.com/google/gapid/gapis/api"
 )
 
+func isUnsizedFormat(fmt GLenum) bool {
+	switch fmt {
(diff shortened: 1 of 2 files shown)


mikaelpessa:
#2653

AWoloszyn:
I will merge this tomorrow if @pmuetschard has no issues with it.

PR #2775, opened 2019-05-24
Initial scaffolding for profiling replays.

pmuetschard:


gapis/api/gles/replay.go
```diff
@@ -234,6 +248,11 @@ func (a API) Replay(
 		transforms.Add(rf)
 	}
 
+	if profile != nil {
+		// Don't use DCE. TODO: should we?
```
AWoloszyn: I don't think we should be using DCE for profiling. It would be way too easy for it to give completely wrong answers.

gapis/api/gles/replay.go
```diff
@@ -331,6 +352,25 @@ func (a API) QueryFramebufferAttachment(
 	return res.(*image.Data), nil
 }
 
+func (a API) SupportsPerfetto(ctx context.Context, i *device.Instance) bool {
```
AWoloszyn: It seems like this should be a function of the device, rather than a function of the API.

gapis/api/vulkan/replay.go
```diff
@@ -1086,3 +1122,22 @@ func (a API) QueryTimestamps(
 	}
 	return res.([]replay.Timestamp), nil
 }
+
+func (a API) SupportsPerfetto(ctx context.Context, i *device.Instance) bool {
```
AWoloszyn: Same as above, I think we should add this to the device.

AWoloszyn:
It's going to take me a bit longer to get through all of the transforms

gapis/api/vulkan/minimize_textures.go
```diff
@@ -0,0 +1,113 @@
...
+				cmd.commandBuffer,
+				cmd.srcImage,
+				cmd.srcImageLayout,
+				cmd.dstBuffer,
+				regionCount,
+				bufferImageCopyData.Ptr(),
+			).AddRead(bufferImageCopyData.Data())
+
+			out.MutateAndWrite(ctx, id, newCmd)
+		default:
```
AWoloszyn: We should also handle `VkCmdBlitImage` `VkCmdCopyImage` 
If the image can/would be used as a framebuffer, this may cause problems as well. Likely we should check for `VkImageUsageFlagBits_VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT` and do nothing, since that would indicate this is a color attachment.

gapis/api/vulkan/minimize_textures.go
```diff
@@ -0,0 +1,113 @@
...
+		cmd.Extras().Observations().ApplyReads(s.Memory.ApplicationPool())
+		switch cmd := cmd.(type) {
+		case *VkCreateImage:
+			imageCreateInfo := cmd.PCreateInfo().MustRead(ctx, cmd, s, nil)
+			if 0 != (imageCreateInfo.Usage() & VkImageUsageFlags(VkImageUsageFlagBits_VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT)) {
+				out.MutateAndWrite(ctx, id, cmd)
+				break
+			}
+
+			imageCreateInfo.SetExtent(NewVkExtent3D(s.Arena, newTexWidth, newTexHeight, newTexDepth))
```
AWoloszyn: If this is a mipmapped image, we have to adjust mip-levels appropriately.
`mipLevels must be less than or equal to the number of levels in the complete mipmap chain based on extent.width, extent.height, and extent.depth.`

That means any image-view that is created from this image must also be adjusted so only ever talk about the first mip-level.

gapis/api/vulkan/minimize_textures.go
```diff
@@ -0,0 +1,113 @@
...
+				cmd.PImage(),
+				VkResult_VK_SUCCESS,
+			).AddRead(imageCreateInfoData.Data())
+
+			for _, w := range cmd.Extras().Observations().Writes {
+				newCmd.AddWrite(w.Range, w.ID)
+			}
+			out.MutateAndWrite(ctx, id, newCmd)
+		case *VkCmdCopyBufferToImage:
+			bufferImageCopy := cmd.PRegions().MustRead(ctx, cmd, s, nil)
```
AWoloszyn: Mip-levels also have to be handled here.

gapis/api/vulkan/minimize_textures.go
AWoloszyn: Not sure if we want to do anything about array-layers, I expect we want to at least leave them alone,

gapis/api/vulkan/minimize_textures.go
```diff
@@ -0,0 +1,113 @@
...
+			out.MutateAndWrite(ctx, id, newCmd)
+		case *VkCmdCopyImageToBuffer:
+			bufferImageCopy := cmd.PRegions().MustRead(ctx, cmd, s, nil)
+			bufferImageCopy.SetImageOffset(NewVkOffset3D(s.Arena, 0, 0, 0))
+			imageExtent := bufferImageCopy.ImageExtent()
+			width := u32.Min(imageExtent.Width(), newTexWidth)
+			height := u32.Min(imageExtent.Height(), newTexHeight)
+			depth := u32.Min(imageExtent.Depth(), newTexDepth)
+
+			bufferImageCopy.SetImageExtent(NewVkExtent3D(s.Arena, width, height, depth))
```
AWoloszyn: If we are setting the image depth to 1, we have to be careful with the 2-d image view into 3d image.

gapis/api/vulkan/minimize_viewport.go
```diff
@@ -0,0 +1,68 @@
...
+		id api.CmdID, cmd api.Cmd, out transform.Writer) {
+
+		const width = 1
+		const height = 1
+
+		s := out.State()
+		l := s.MemoryLayout
+		cb := CommandBuilder{Thread: cmd.Thread(), Arena: s.Arena}
+		switch cmd := cmd.(type) {
+		case *VkCmdSetViewport:
```
AWoloszyn: The viewport may also be specified in the pipeline creation as a non-dynamic piece of data. We should catch it there as well.
https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VkPipelineViewportStateCreateInfo

gapis/api/vulkan/minimize_viewport.go
```diff
@@ -0,0 +1,68 @@
...
+			newCmd := cb.VkCmdSetViewport(cmd.commandBuffer,
+				cmd.FirstViewport(),
+				uint32(viewportCount),
+				newViewportDatas.Ptr()).AddRead(newViewportDatas.Data())
+
+			for _, w := range cmd.Extras().Observations().Writes {
+				newCmd.AddWrite(w.Range, w.ID)
+			}
+			out.MutateAndWrite(ctx, id, newCmd)
+		default:
```
AWoloszyn: You may have to also adjust scissor and renderpasscreateinfo structs for this to remain valid.

https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VkRenderPassBeginInfo
`The application must ensure (using scissor if necessary) that all rendering is contained within the render area. The render area must be contained within the framebuffer dimensions.`

gapis/api/vulkan/set_vertex_count_to_one.go
```diff
@@ -0,0 +1,96 @@
...
+	return transform.Transform("setVertexCountToOne", func(ctx context.Context,
+		id api.CmdID, cmd api.Cmd, out transform.Writer) {
+
+		s := out.State()
+		cb := CommandBuilder{Thread: cmd.Thread(), Arena: s.Arena}
+		var newCmd api.Cmd
+		isIndirectDraw := false
+		var cmdBuf VkCommandBuffer
+		switch cmd := cmd.(type) {
+		case *VkCmdDraw:
```
AWoloszyn: My only worry about this transform is that drawing partial primitives is not well-defined. As in the draw call has to not crash, but the driver is free to simply drop the entire call (if ALL primitives are partial).

gapis/api/vulkan/simplify_fragment_shader.go
```diff
@@ -0,0 +1,114 @@
...
+func simplifyFragmentShader(ctx context.Context) transform.Transformer {
+	ctx = log.Enter(ctx, "simplifyFragmentShader")
+	return transform.Transform("simplifyFragmentShader", func(ctx context.Context,
+		id api.CmdID, cmd api.Cmd, out transform.Writer) {
+
+		s := out.State()
+		l := s.MemoryLayout
+		cb := CommandBuilder{Thread: cmd.Thread(), Arena: s.Arena}
+		switch cmd := cmd.(type) {
+		case *VkCreateShaderModule:
```
AWoloszyn: You are allowed to specify multiple shaders in a single module. So we have to be more clever about the re-writing that we will be doing here.

It might be easier to ignore all of the VkCreateShaderModule calls, and intercept at the VkCreateGraphicsPipeline Calls (we can create shaders as necessary there), that way we can only override the fragment shader.

gapis/replay/profile.go
```diff
@@ -0,0 +1,60 @@
...
+// ProfilePostBack is a transform that causes a post back at the end of the replay.
+type ProfilePostBack struct {
+	Res []Result
+}
+
+func (t *ProfilePostBack) Transform(ctx context.Context, id api.CmdID, cmd api.Cmd, out transform.Writer) {
+	out.MutateAndWrite(ctx, id, cmd)
+}
+
+func (t *ProfilePostBack) Flush(ctx context.Context, out transform.Writer) {
```
AWoloszyn: It may make sense sometime to unify all of these PostBack finalizing calls, so that we don't have to duplicate this all over the place.

gapis/service/path/path.proto
```diff
@@ -491,6 +491,14 @@ message Thumbnail {
   bool disable_optimization = 7;
 }
 
+message OverrideConfig {
```
AWoloszyn: It might be nice to make this more generic. Something more similar to having transforms be able to register them as "profiling overrides" with a name.

That way we don't have to change the public interface every time we add a what-if.

PR #2776, opened 2019-05-24
 Log API commands to atrace during replay.

pmuetschard:
atrace requires API level 23 (M) in the NDK, which means this would make M required for on-device-replay.

### DIFF
--- a/gapis/api/templates/specific_gfx_api.cpp.tmpl
+++ b/gapis/api/templates/specific_gfx_api.cpp.tmpl
@@ -40,6 +40,10 @@
 ¶
   #define __STDC_FORMAT_MACROS
   #include <inttypes.h>
+
+#if TARGET_OS == GAPID_OS_ANDROID
+#include <android/trace.h>
+#endif  // TARGET_OS == GAPID_OS_ANDROID
 ¶
   {{$api := Title (Global "API")}}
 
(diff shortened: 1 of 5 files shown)


tools/build/workspace.bzl
```diff
@@ -211,7 +211,7 @@ def gapid_dependencies(android = True, mingw = True, locals = {}):
             native.android_ndk_repository,
             name = "androidndk",
             locals = locals,
-            api_level = 21, # This is the minimum API
+            api_level = 23, # This is the minimum API
```
AWoloszyn: Did we run into issues with the _armeabi memclr? .... stuff when increasing this? I thought there was some problem on older devices.

gapis/api/templates/specific_gfx_api.cpp.tmpl
```diff
@@ -133,6 +137,10 @@
     if (stack->isValid()) {
       GAPID_DEBUG({{Template "C++.PrintfCommandCallWithLabel" $}});
 
+      #if TARGET_OS == GAPID_OS_ANDROID
+      ATrace_beginSection("{{$name}}");
```
AWoloszyn: It would be nice to be able to do this in a cross-platform way. 
For now this is fine, but it would be nice to figure out how to avoid duplicating this for Desktop.

tools/build/workspace.bzl
pmuetschard: I think the memclr was from a later version, but effectively, this would mean that, at least for on-device replay, the minimum required version would be M, whereas currently it may work on L.

gapis/api/templates/specific_gfx_api.cpp.tmpl
pmuetschard: Agreed.

pau-baiget: Please create an issue to track this TODO.

PR #2779, opened 2019-05-28
Initial scaffolding for profiling replays.

pmuetschard:


pmuetschard:
This is the same as #2775 minus the override transforms, but addressing the remaining review comments.

AWoloszyn:
Thanks!

PR #2782, opened 2019-05-29
Add missing read of sType in GetPhysicalDeviceProperties2

hevrard:


gapis/api/vulkan/api/properties_features_requirements.api
```diff
@@ -487,6 +487,7 @@ sub void GetPhysicalDeviceProperties2(
       next := MutableVoidPtr(as!void*(props.pNext))
       for i in (0 .. numPNext) {
         sType := as!const VkStructureType*(next.Ptr)[0:1][0]
```
AWoloszyn: You should't need another _ = sType, the read should be handled by 
as!const VkStructureType*(next.Ptr)[0:1]**[0]**

hevrard:
I'm confused, I used to have `gapit report` errors on deqp tests related to this, but I cannot repro on current master -- I may have been inadvertently using binaries of an old release when I though I was testing master. Thanks for the review, sorry for the distraction!

PR #2785, opened 2019-05-29
Main Perfetto UI for System Profiling

pmuetschard:
To be able to take a perfetto trace, provide the UI with the desired Perfetto config protobuf in text format in a file passed via the `--perfetto` flag.


gapic/src/main/com/google/gapid/models/Settings.java
```diff
@@ -243,6 +245,9 @@ private void updateFrom(Properties properties) {
     disableReplayOptimization =
         getBoolean(properties, "replay.disableOptimization", disableReplayOptimization);
     reportCrashes = getBoolean(properties, "crash.reporting", reportCrashes);
+    perfettoSplitterWeights =
```
AWoloszyn: We had some problems with negative splitter weights before (somehow...?) 
Any chance we can pre-empt that issue here ?

pmuetschard: the negative splitter weights were only an issues with the tabs due to the custom handling there to allow tabs moving around. Here and in the other places where we use the splitter widget directly, the weights are already sanitized.

PR #2788, opened 2019-05-30
Fix for descriptors if we cannot parse the SPIR-V module.

AWoloszyn:
We would fail to re-create the state in this case.


gapis/api/vulkan/state_rebuilder.go
```diff
@@ -2193,7 +2193,7 @@ func (sb *stateBuilder) createRenderPass(rp RenderPassObjectʳ) {
 }
 
 func (sb *stateBuilder) createShaderModule(sm ShaderModuleObjectʳ) {
-	sbExtra := sm.Descriptors().Get().Clone(sb.newState.Arena, api.CloneContext{})
+
```
pmuetschard: nit: remove the line completely?

PR #2790, opened 2019-05-31
Auto-generate docs for GAPIT

pau-baiget:
New file GAPIT.md added to the repo root folder and is referenced in README. The docs are auto-generated using existing functions in help.go.


core/app/help.go
```diff
@@ -23,6 +23,45 @@ import (
 	"github.com/google/gapid/core/app/flags"
 )
 
+// MakeDocs
+func MakeDoc(ctx context.Context, name string){
+	filename := name + ".md"
+	w, _ := os.Create(filename)
+	globalVerbs.selected = nil
+	fmt.Fprintln(w, "**Note**: Autogenerated documentation, do not modify!")
```
hevrard: Here I would add in the comment how this doc is generated, e.g.:

**Note**: Autogenerated documentation via `gapit make_doc`, do not modify by hand.

core/app/help.go
```diff
@@ -23,6 +23,45 @@ import (
...
+		fmt.Fprintf(w, "*%s*", verb.ShortHelp)
+		fmt.Fprintln(w)
+		fmt.Fprintln(w)
+		fmt.Fprint(w, "**Usage:** ")
+		verbUsage(w, &globalVerbs, true)
+		// Full flags
+		fmt.Fprintln(w, "```")
+		verbHelp(w, &globalVerbs, true)
+		fmt.Fprintln(w, "```")
+		fmt.Fprintln(w, backlink)
```
hevrard: So you have a backlink after the title and at the end of the full flag section, do you also want the one at the end? (I'm OK either way, just to point it out)

AWoloszyn:
This is one of the very rare times that I think we should actually pull this command OUT of gapit. (It can still share some files inside of gapit (flags etc) but I think it should be its own executable).

My main reason is that if we do that, we can add it to our presubmit checks. Otherwise we would have to build the entirety of gapit before we run our presubmits, or not run our presubmits at all.

hevrard:
@AWoloszyn is the idea to generate the GAPIT.md as part of presubmit? (and/or check that re-generating it leads to the same output, to make sure the doc is up to date?)

AWoloszyn:
The idea would not be to generate it during presubmit, that would be tricky, but simply failing the presubmit if you would have had to update it to keep it in sync.

AWoloszyn:
Note: This is how we deal with code-formatting & gazelle as well. If you SHOULD have run it but didn't, then you fail pre-submit.

hevrard:
So the presubmit would build the standalone "makedoc" command, and then run it, and check output diff?

The difference is, as far as I understand, the formatters and gazelle don't need to be built, am I right?

But a small standalone command that only depends on the necessary packages to access the help flags should be fast enough to build in presubmit? (as opposed to building gapit which pulls in much more dependencies?)

AWoloszyn:
That is correct. Wanting to keep the presubmit as light as possible, we can't build the entirety of gapit. But we should be able to build JUST the bits we need to parse all of the flags.

AWoloszyn:
All that being said, I think this may actually be infeasible anyway. It occurs to me that we may not have the global list of verbs available, without having to start building the rest of the tool. I would like to at least validate that this is in fact impractical before we dismiss it though.

pau-baiget:
Agreed that this should be a different command outside gapit. The only reason for doing it inside is that it's difficult to answer the question "does the doc needs to be updated?" without actually generating it. And, since all the help messages are in the flag structs, we'd need to parse them if we don't want to do the build. Any different options?

GAPIT.md
```diff
@@ -0,0 +1,1752 @@
+**Note**: Autogenerated documentation, do not modify!
```
pmuetschard: This file should not be checked into master, but the gh-pages branch, which powers gapid.dev. We should avoid checking in generated source into master. The only generated files left that are checked in are third party config files.

core/app/help.go
```diff
@@ -23,6 +23,45 @@ import (
 	"github.com/google/gapid/core/app/flags"
 )
 
+// MakeDocs
+func MakeDoc(ctx context.Context, name string) {
+	filename := name + ".md"
+	w, _ := os.Create(filename)
```
pmuetschard: This should just print to stdout so that it makes it easier to automate this and possibly wrap the generated contents in static content.

core/app/help.go
pmuetschard: This should be a comment not visible to the user. Is that possible with MD? If not, it should be moved to the bottom or at least after the title.

core/app/help.go
```diff
@@ -23,6 +23,45 @@ import (
...
+	globalVerbs.selected = nil
+	fmt.Fprintln(w, "**Note**: Autogenerated documentation, do not modify!")
+	fmt.Fprintln(w, "# GAPIT Help")
+
+	// Print list of verbs
+	fmt.Fprintln(w, "| Command | Short help |")
+	fmt.Fprintln(w, "| -- | -- |")
+	format := fmt.Sprintf("|[%%s](#%%s)|%%s|")
```
pmuetschard: this is just a constant string, no need to send it through sprintf.

core/app/help.go
pau-baiget: Fixed.

core/app/help.go
pau-baiget: Fixed.

core/app/help.go
pau-baiget: Fixed

PR #2803, opened 2019-06-04
Fix screen unlock: make no assumption about screen lock state change

hevrard:


hevrard:
Turns out some devices do weird transitions, e.g. from OFF_UNLOCKED --[wakeup]--> ON_LOCKED, so it's better to not assume any logic there.

This means getScreenState() could be reduced to matching exactly ON_UNLOCKED, and return a simple boolean (screen is ON_UNLOCKED, or not), but I reckon it still makes sense to keep matching a meaningful screen state string and report an error if no such match happens.

core/os/android/adb/screen.go
```diff
@@ -88,22 +79,22 @@ func (b *binding) UnlockScreen(ctx context.Context) (bool, error) {
...
+		// credentials required) key event sequence.
 		if err := b.KeyEvent(ctx, android.KeyCode_Wakeup); err != nil {
 			return false, err
 		}
-		fallthrough
-	case screenOnLocked:
 		if err := b.KeyEvent(ctx, android.KeyCode_Menu); err != nil {
 			return false, err
 		}
+		// KeyEvent() eventually calls adb shell commands, which are
```
pmuetschard: This part of the comment is false. b.KeyEvent will only return once the adb command has finished. See the `Run()` function `core/os/shell/command.go`

hevrard: Yep, I got confused: adb shell commands are synchronous, some other commands are not (e.g. `adb am start` returns before the app is actually started). I updated the comment accordingly. Thanks!

PR #2804, opened 2019-06-04
Add what if transforms

mikaelpessa:


gapis/api/vulkan/minimize_vertex_count.go
```diff
@@ -0,0 +1,103 @@
...
+
+// minimizeVertexCount returns a transform that sets the maximum
+// number of vertices per draw call to 3
+func minimizeVertexCount(ctx context.Context) transform.Transformer {
+	ctx = log.Enter(ctx, "setVertexCountToOne")
+
+	const vertexCount uint32 = 3
+	const instanceCount uint32 = 1
+
+	return transform.Transform("setVertexCountToOne", func(ctx context.Context,
```
AWoloszyn: This is no longer setting the vertex count to one.
I would also recommend setting the primitive count to one, since we may be drawing points, or lines, etc (possibly also Tessellation patches).

gapis/api/vulkan/minimize_textures.go
```diff
@@ -0,0 +1,185 @@
...
+			for _, w := range cmd.Extras().Observations().Writes {
+				newCmd.AddWrite(w.Range, w.ID)
+			}
+			out.MutateAndWrite(ctx, id, newCmd)
+		case *VkCmdCopyBufferToImage:
+			bufferImageCopy := cmd.PRegions().Slice(0, 1, l).MustRead(ctx, cmd, s, nil)[0]
+			imageSubresource := bufferImageCopy.ImageSubresource()
+			imageSubresource.SetMipLevel(0)
+			bufferImageCopy.SetImageSubresource(imageSubresource)
+			bufferImageCopy.SetImageExtent(NewVkExtent3D(a, newTexWidth, newTexHeight, bufferImageCopy.ImageExtent().Depth()))
```
AWoloszyn: I think this is valid but we should double check against the valid usage of VkBufferImageCopy
```

If the calling command’s VkImage parameter is a compressed image, or a single-plane, “_422” image format, imageExtent.width must be a multiple of the compressed texel block width or (imageExtent.width + imageOffset.x) must equal the image subresource width

If the calling command’s VkImage parameter is a compressed image, or a single-plane, “_422” image format, imageExtent.height must be a multiple of the compressed texel block height or (imageExtent.height + imageOffset.y) must equal the image subresource height
```

gapis/api/vulkan/minimize_vertex_count.go
mikaelpessa: Changed to using minimum number of primitives instead of vertices. Ignoring tessellation patches for now.

AWoloszyn: So this means that anything that uses tessellation will just be disabled as far as I can tell? (Any draw calls that use tessellation will be effectively discarded).

As a follow-up we should start documenting these transforms, because this could mislead users.

gapis/api/vulkan/set_primitive_count_to_one.go
```diff
@@ -0,0 +1,143 @@
...
+			return
+		}
+
+		// Replace indirect draw calls with direct ones.
+		// TODO: Replace with appropriate indirect calls instead.
+		if isIndirectDraw {
+			vertexCount := perPipelineVertexCount[cmdPipelineBindings[cmdBuf]]
+			newCmd := cb.VkCmdDraw(
+				cmdBuf,
+				vertexCount,
```
AWoloszyn: This should also be "min(vertexCount, ...)" since it is legal for the original drawindirect to be 0.

AWoloszyn:
One small change then LGTM.

ben-clayton:
Not to be treated as part of the code review - but just wanted to say this is impressive stuff. Well done!

gapis/api/vulkan/set_primitive_count_to_one.go
mikaelpessa: I was gonna do that, but those values are stored in a vkbuffer, and I have found no clear way to read back vkbuffers. It's the same reason all indirect calls were replaced with this direct one.

If there is a way to read vkbuffers I'll change it. Otherwise an option would be to just set the vertex count to 0 for all indirect calls. At least that way the calls will still always be valid.

AWoloszyn: There are WAYS to read back vkBuffers, but we certainly don't want ot go down that path, it is horribly complicated. We should then set them to 0. I think we should avoid doing anything that may cause invalid behavior, since it has a REAL  chance of causing the replay driver to fail. (And sometimes when they get in a bad state, they do not recover).

mikaelpessa: Done. Just wondering if that makes the results unclear to users, since the behavior is quite different based on whether you're using direct or indirect draw calls.

AWoloszyn: I agree, but crashing the replay would be even worse. We should absolutely document all of this. Alternatively, we can just leave indirect calls alone entirely.

PR #2806, opened 2019-06-04
Some fixes to the GAPID documentation.

pmuetschard:
- Fix the order in the sidebar.
- Rename GAPIT.md to cli.md to be more consistent with the title/URL and other files.
- Add a short intro to the GAPIT documentation.
- Add a link to the CLI documentation in the create video tutorial.
- Update the requirements to match the current minimums.


cli.md
```diff
@@ -6,6 +6,12 @@ permalink: /cli/
 order: 50
 ---
 
+The GAPID package contains a command line tool named `gapit` that  can be used
```
pau-baiget: Shouldn't this be autogenerated somehow? How are you going to update this doc in the future?

pmuetschard: I didn't touch the generated part. To re-generate we can simply prepend the auto generated portion with the header. This can be done manually when updating this file, or if we setup an automated way to do that, we could make the header an input to it. Feel free to make changes to the header, however, since this is visible now, I figured it is best to put something in place now, rather than wait for some automatic way.

PR #2833, opened 2019-06-12
Shader View improvement

stellama0208:


gapis/api/gles/resources.go
```diff
@@ -269,7 +269,7 @@ func (s Shaderʳ) IsResource() bool {
 
 // ResourceHandle returns the UI identity for the resource.
 func (s Shaderʳ) ResourceHandle() string {
-	return fmt.Sprintf("Shader<%d>", s.ID())
+	return fmt.Sprintf("Shader %d", s.ID())
```
hevrard: Warning: the "handles" are a specific string that can be used in `gapit replace_resource`, where I the "<" and ">" are needed. I think it's better to not have those character displayed in the UI, but we may want to homogenize the way a handle is represented as a string.

@pmuetschard what do you think?

stellama0208: Yeah, need some suggestions here. 
Will it be better if I leave the server-side string untouched, and only do some string transformation at the UI-side code?

PR #2840, opened 2019-06-13
Make Android Perfetto tracing work on Windows.

pmuetschard:
Passing the binary data via adb doesn't quite work on Windows, so encode the proto to base64 to make it go safely through adb.


AWoloszyn:
Fun!

PR #2841, opened 2019-06-14
Add description how to find annotation usage

hevrard:


ben-clayton:
[There are also places where the guts of the compiler (`gapil`) will use annotations](https://github.com/google/gapid/blob/d79a5232a77bbd07582eca63c15de429c4e52d5d/gapil/semantic/api.go#L86).
I've been wanting a way to document these annotations for a long time. Ideally, an annotation would have a definition in the language, which provides an obvious place to document what it means. This would also prevent accidental mis-spellings, highlight duplicates, etc.

pau-baiget:
@AWoloszyn @ben-clayton are you actually requesting changes to this PR (to be moved somewhere else)?

AWoloszyn:
@pau-baiget I think we should not put this in. We should more likely add this a piece of the developer FAQ and merge that into our  tree.

PR #2843, opened 2019-06-14
Stop outputting every file we copy from remote.

AWoloszyn:


AWoloszyn:
I am going to merge this early. I can revert if anyone ever wants to see files in their gapis.log.

PR #2845, opened 2019-06-18
Update imagePrimer to support prime by copying from device image data.

RenfengLiu:


gapis/api/vulkan/primeable_image_data.go
```diff
@@ -258,12 +300,12 @@ func (pi *ipPrimeableByPreinitialization) prime(sb *stateBuilder, srcLayout, dst
 // true, the image data will be collected from the shadow memory of the old
 // state image object, which is on the host accessible space. If fromHostData is
 // false, the image data will be collected from the device memory.
-func (p *imagePrimer) newPrimeableImageData(img VkImage, opaqueBoundRanges []VkImageSubresourceRange, fromHostData bool) (primeableImageData, error) {
+func (p *imagePrimer) newPrimeableImageData(srcImg, dstImg VkImage, opaqueBoundRanges []VkImageSubresourceRange, fromHostData bool) (primeableImageData, error) {
```
AWoloszyn: There is a awkward correlation between
`fromHostData` and `dstImg` which if `fromHostData == false` then `dstImg` is expected to be valid. 

Is there any nicer way we can structure this function for this purpose? The simplest way I can think of is to split it out into 3 functions 

newPrimeableImageDataFromHost(srcImg, ....., <no fromHostData>)
newPrimeableImageDataFromDevice(srcImage, dstImage, opaque, ....)
newPrimeableImageDataInternal(<what we have now>)

at least that way people can call one of the 2 named functions with more succinct parameterss

gapis/api/vulkan/primeable_image_data.go
```diff
@@ -564,163 +565,219 @@ func (p *imagePrimer) newPrimeableImageData(img VkImage, opaqueBoundRanges []VkI
...
+			return nil, log.Errf(p.sb.ctx, nilQueueErr, "[Building primeable image data that can be primed by preinitialization with host data, image: %v]", img)
 		}
+		return &ipPrimeableByPreinitialization{p: p, img: img, opaqueBoundRanges: opaqueBoundRanges, queue: queue.VulkanHandle()}, nil
+
 	}
 	return nil, log.Errf(p.sb.ctx, nil, "No way build primeable image data for image: %v", img)
 }
+
+// newPrimeableImageDataFromDevice builds primeable image data from the source image, which is on the device already.
+func (p *imagePrimer) newPrimeableImageDataFromDevice(srcImg, dstImg VkImage) (primeableImageData, error) {
```
AWoloszyn: We will need to handle sparse images here as well.
I suggest, for now, we simply check and fail if we come across one.

gapis/api/vulkan/primeable_image_data.go
```diff
@@ -564,163 +565,219 @@ func (p *imagePrimer) newPrimeableImageData(img VkImage, opaqueBoundRanges []VkI
...
+		kits, err := kb.BuildStoreKits(p.sb, recipes...)
+		if err != nil {
+			return nil, log.Errf(p.sb.ctx, err, "failed at build store kits for priming storage image")
+		}
+		primeable.kits = kits
+		return primeable, nil
+	}
+	primeByPreinitialization := (!primeByCopy) && (!primeByRendering) && (!primeByImageStore) && (srcImgObj.Info().Tiling() == VkImageTiling_VK_IMAGE_TILING_LINEAR) && (srcImgObj.Info().InitialLayout() == VkImageLayout_VK_IMAGE_LAYOUT_PREINITIALIZED)
+	if primeByPreinitialization {
+		// TODO:
```
AWoloszyn: We should probably return an error here, if we ever run into this case. 
That way we won't forget.

PR #2848, opened 2019-06-19
Android: Pass GAPID argument --gapir-args from gapis to gapir

### LINKED ISSUE #2839: GAPIR on Android cannot receive arguments
GAPIR can take some arguments on desktop, but there is currently no way to pass arguments on Android.

The solution needs to use JNI to retrieve arguments from the Android intent "extra" field.

Assigned to @purvisa-at-google-com as a first task.

purvisa-at-google-com:
Fix #2839 
This pull request also unifies the gapir logging for PC and Android and adds support for some arguments that were previously not implemented by gapir on Android.

### DIFF
--- a/cmd/gapir/cc/main.cpp
+++ b/cmd/gapir/cc/main.cpp
@@ -39,6 +39,7 @@
 #include <unistd.h>
 #include <memory>
 #include <mutex>
+#include <sstream>
 #include <thread>
 
 #if TARGET_OS == GAPID_OS_ANDROID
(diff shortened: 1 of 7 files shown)


cmd/gapir/cc/main.cpp
```diff
@@ -76,6 +77,235 @@ std::vector<uint32_t> memorySizes {
...
+      } else if (strcmp(argv[i], "-h") == 0 || strcmp(argv[i], "-help") == 0 ||
+                 strcmp(argv[i], "--help") == 0) {
+        opts.help = true;
+      } else {
+        GAPID_FATAL("Unknown argument: %s", argv[i]);
+      }
+    }
+
+#if TARGET_OS == GAPID_OS_ANDROID
+    if (opts.android) {
```
AWoloszyn: I am not sure what `--android` is supposed to represent.
It looks like it is supposed to be the equivalent to "--replay-archive" on desktop based on the next little bit of code.

core/cc/log.h
```diff
@@ -132,6 +98,14 @@ class Logger {
 
 }  // namespace core
 
+#if TARGET_OS == GAPID_OS_ANDROID
+#include <android/log.h>
+#else  // TARGET_OS == GAPID_OS_ANDROID
+#include <stdio.h>
+#endif  // TARGET_OS == GAPID_OS_ANDROID
```
ben-clayton: Can these includes now be moved to the .cpp?

gapir/client/client.go
```diff
@@ -30,7 +30,7 @@ type tyLaunchArgsKey string
 
 // LaunchArgsKey is the bind device property key used to control the command
 // line arguments when launching GAPIR. The property must be of type []string.
-const LaunchArgsKey tyLaunchArgsKey = "<gapir-launch-args>"
+const LaunchArgsKey tyLaunchArgsKey = "gapir-launch-args"
```
ben-clayton: What's the purpose of this change? This constant is used as a property key, and there's no documentation / any obvious relation to the actual command line parameter name.

gapir/client/session.go
```diff
@@ -295,10 +296,19 @@ func (s *session) newADB(ctx context.Context, d adb.Device, abi *device.ABI) err
 		return err
 	}
 
+	var completeLaunchArgs []string;
```
ben-clayton: no semicolons. Please use gofmt to format your code (it's a good idea to get your IDE to format-on-save).

Also, this can be initialized with all the fields in one declaration:

```go
completeLaunchArgs := []string{
	"--idle-timeout-sec", string(int(sessionTimeout /time.Second)),
	"--auth-token-file", string(s.auth),
	"--android",
}
```

gapir/client/client.go
purvisa-at-google-com: Android doesn't seem to like <> characters in intent names. I played around with escaping them with some success, but it seemed like an unnecessary risk supporting exotic characters for what seemed like a cosmetic choice.

gapir/client/session.go
purvisa-at-google-com: Whoops. I do this in python all the time. It's a bad habit from being a c++ centric programmer.

purvisa-at-google-com: Good tip with the single initialiser. I only started go yesterday so while obvious, this didn't cross my mind.

cmd/gapir/cc/main.cpp
purvisa-at-google-com: --replay-archive does not do the same as --android.

I tried to find a way to express what the old android Parse() function was doing via arguments to the (now unified) PC Parse() function. Unfortunately I didn't see a way of achieving an exact behaviour match without adding a new flag.

core/cc/log.h
purvisa-at-google-com: Not unless we move them to (potentially) every c++ that uses log.h. This is quite a bit more than just Log.cpp

gapir/client/client.go
ben-clayton: Sure - my question was more: why is this being passed down to adb at all? It's an internal key for `gapir/client`. It appears that you've had to change the value because you've started using this a string to pass to adb, which is now doing more than it was intended to.

purvisa-at-google-com: I can create a new key for the intent if you'd prefer to keep the <> characters. I just thought it was simpler to use fewer identifiers around the place.

cmd/gapir/cc/main.cpp
hevrard: What about passing a `struct Options *opts` to Parse(), such that you can pre-fill the options before parsing with different defaults depending on whether you're on host or device entry point?

gapir/client/client.go
ben-clayton: There are other keys out there, none of them are reused for other string purposes, so I'd recommend just using the `"gapir-launch-args"` string literal at the site where you pass that at a parameter to the adb command, like all the other parameters.
My objection here is that there's nothing obvious or documented on this variable to say it is being used as a command line parameter string. Someone could decide they want to change the token value of `tyLaunchArgsKey` and unintentionally break the adb commands.

hevrard: This contant is mirrored in cmd/gapir/cc/main.cpp:596
```
const char* intent_flag = "gapir-launch-args";
```
This is fragile, as this string needs to exactly match the tyLaunchArgsKey.

I recommend to have a single declaration of the intent "keywork" as a different constant than tyLaunchArgsKey, and reuse that constant in both GAPIS and GAPIR, to make sure its actual value can be changed in a single place.

ben-clayton: > This is fragile, as this string needs to exactly match the tyLaunchArgsKey.

No. It should not have anything to do with `LaunchArgsKey`. :D

`gapir/client/session.go` has a whole bunch of command line parameter literal strings in it. Why is this one special?

If you create a whole bunch of named constants for command line parameters, okay - make them all constants, place them in `session.go`, but don't conflate it with device keys. They are completely unrelated concepts.

ben-clayton: To stress my point some more. `tyLaunchArgsKey` is a private type that is used purely as a unique map key with its single instance `LaunchArgsKey`. The only reason the underlying type is a string is so that it is easier to debug.
It could be an `int`, `struct`, `float64` - anything that produces a globally unique value-and-type key for the `bind.Registry.properties` map.
That's its soul purpose. Not to represent anything else.

cmd/gapir/cc/main.cpp
purvisa-at-google-com: Done!

gapir/client/client.go
purvisa-at-google-com: Changed to a string literal as @hevrard suggested at my desk

cmd/gapir/cc/main.cpp
```diff
@@ -76,6 +77,235 @@ std::vector<uint32_t> memorySizes {
...
+      } else if (strcmp(argv[i], "--auth-token-file") == 0) {
+        opts.SetMode(kReplayServer);
+        if (i + 1 >= argc) {
+          GAPID_FATAL("Usage: --auth-token-file <token-string>");
+        }
+        opts.authTokenFile = argv[++i];
+      } else if (strcmp(argv[i], "--enable-disk-cache") == 0) {
+        ensureNotAndroid(
+            "--enable-disk-cache");  // Fatal error if running on android. This
+                                     // flag is not supported.
```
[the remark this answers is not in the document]
purvisa-at-google-com: Done

cmd/gapir/cc/main.cpp
```diff
@@ -76,6 +77,235 @@ std::vector<uint32_t> memorySizes {
...
+        if (i + 1 >= argc) {
+          GAPID_FATAL("Usage: --disk-cache-path <cache-directory>");
+        }
+        opts.onDiskCacheOptions.path = argv[++i];
+      } else if (strcmp(argv[i], "--cleanup-on-disk-cache") == 0) {
+        ensureNotAndroid(
+            "--cleanup-on-disk-cache");  // Fatal error if running on android.
+                                         // This flag is not supported.
+        opts.onDiskCacheOptions.cleanUp = true;
+      } else if (strcmp(argv[i], "--port") == 0) {  // check it works
```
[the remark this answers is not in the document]
purvisa-at-google-com: Done

cmd/gapir/cc/main.cpp
```diff
@@ -395,8 +581,112 @@ void android_process(struct android_app* app, int32_t cmd) {
 
 }  // namespace
 
+// Extract command line arguments from the extra of Android intent:
+//   adb shell am start -n <...> -e gapir-launch-args "'list of arguments to be
```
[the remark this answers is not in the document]
purvisa-at-google-com: @hevrard and I talked at my desk. "gapir" would get too many false positives in a grep of the source code; we settled on "gapir-intent-flag" in the end as this was believed to be unique enough to get only relevant hits.

gapir/client/client.go
hevrard: Thanks Ben for the clarifications!

cmd/gapir/cc/main.cpp
```diff
@@ -76,6 +77,190 @@ std::vector<uint32_t> memorySizes {
       128 * 1024 * 1024U,       // 128MB
 };
 
+struct Options {
```
AWoloszyn: This Change is probably not the place to fix this, since I think it is a holdover from an earlier time. But none of these variables follow the Google C++ style guide (which we are at least TRYING to follow).

AWoloszyn: https://google.github.io/styleguide/cppguide.html#General_Naming_Rules

cmd/gapir/cc/main.cpp
```diff
@@ -395,8 +536,134 @@ void android_process(struct android_app* app, int32_t cmd) {
 
 }  // namespace
 
+// Extract command line arguments from the extra of Android intent:
+//   adb shell am start -n <...> -e gapir-launch-args "'list of arguments to be
+//   extracted'"
+// Note the quoting: from host terminal adb command, we need to double-escape
```
AWoloszyn: We have run into some insane quoting things on windows so we should test there as well.

cmd/gapir/cc/main.cpp
purvisa-at-google-com: Yeah, I moved the code. The names are as they were when I encountered the source.

cmd/gapir/cc/main.cpp
```diff
@@ -395,8 +536,134 @@ void android_process(struct android_app* app, int32_t cmd) {
...
+  std::vector<std::string> args;
+  args.push_back(
+      "gapir");  // POSIX says the first argv should be the program being run.
+                 // Lets inject a placeholder here for compatibility.
+
+  // Split extra_string
+  std::stringstream ss(extra_string);
+  std::string arg;
+
+  while (std::getline(ss, arg, ' ')) {
```
AWoloszyn: This ignores quoted arguments. I am not sure if we want/need to support that, but things like spaces in filenames etc will break here.

cmd/gapir/cc/main.cpp
purvisa-at-google-com: This is copied from Graphics Fuzz.

cmd/gapir/cc/main.cpp
purvisa-at-google-com: I left this behaviour un-altered. The problem you highlight exists in the original code.

cmd/gapir/cc/main.cpp
```diff
@@ -395,8 +536,134 @@ void android_process(struct android_app* app, int32_t cmd) {
 
 }  // namespace
 
+// Extract command line arguments from the extra of Android intent:
+//   adb shell am start -n <...> -e gapir-launch-args "'list of arguments to be
+//   extracted'"
+// Note the quoting: from host terminal adb command, we need to double-escape
+// the extra args string, as it is first quoted by host terminal emulator
+// (e.g. bash), then it must be quoted for the on-device shell.
+void GetArgvFromIntents(struct android_app* state, int* argc, char*** argv) {
```
AWoloszyn: I won't hold up the CR for this, but I have a preference of making this slightly more C++-ey.
```
struct Args {
   int argc;
   char** argv;
   - Other stuff here?
   ~Args() { 
      - free argv here
   }
}
```

You can return by std::move(), so that we don't have to heap-alloc the struct itself.

AWoloszyn: This way no-one has to forget to "Free"

cmd/gapir/cc/main.cpp
```diff
@@ -821,7 +926,8 @@ static int startServer(core::CrashHandler* crashHandler, Options opts) {
 
 // Main function for PC
 int main(int argc, const char* argv[]) {
-  Options opts = Options::Parse(argc, argv);
+  Options opts;
```
AWoloszyn: Any particular reason to switch this to passing by ref, and filling in?

Passing by ref violates the style guide as well. If you go down this route, you should pass by pointer. Style guide says you should only ever pass things by const-ref (which obviously won't work in this case)

AWoloszyn:
```
Options opts = Options::Parse(argc, argv);
```
Should have RVO applied, so you shouldn't end up paying any cost for the "copy".

purvisa-at-google-com: The pass by reference is not for "performance". It allows me to fill in some default values on android (that are gleaned from the apk) before later parsing the flags and applying the appropriate state changes on top.

I agree that copy elision should kick in here. And anyway the structure is so small and the operation once off, so optimizing it at the expense of clarity makes no sense.

Happily, I prefer passing things by pointer than non-cost reference. I used a reference at @hevrard  proposal. Would you prefer a pointer here?

AWoloszyn: Understood!

We try our best to obey the style guide, so it would be better to make it a pointer.

cmd/gapir/cc/main.cpp
AWoloszyn: Sounds good. Feel free to leave this as-is.

cmd/gapir/cc/main.cpp
AWoloszyn: Alrighty. We should just make double-sure to test on windows before the next release.

core/cc/log.h
purvisa-at-google-com: Done!

cmd/gapir/cc/main.cpp
purvisa-at-google-com: Isn't this dangerous in its own right; that anyone returning Args who forgets to use std::move will return a pointer into freed memory?

purvisa-at-google-com: OK, changed this code so it uses a std::vector<std::string> for data storage and string::c_str() / std::vector<unsigned  char*> to create a presentation of that data which looks like old fashioned c strings.

I think this is the most "c++ey" way of solving the problem.

God knows why anyone's still using c strings in c++.

AWoloszyn: As for "anyone who forgets to use std::move" we can disable copy construction/copy assignment. Then they will just get a compile-error in this case.

purvisa-at-google-com: What's your opinion of the std::string proposal I uploaded in the last revision?

AWoloszyn: That being said how you restructured it is completely fine with me.

We typically use c strings in 2 cases.
1) When we don't really have a choice. (Interfacing with a C api. eg. `int main(....))`
2) When we want to avoid copies. std::string can be a pain for a number of reasons, and often we want to avoid copies. C++14 doesn't yet have string_view or equivalent, and we haven't pulled an implementation in yet.

But for most other cases raw cstrings should generally be avoided.

AWoloszyn:
LGTM, you should just run clang-format before submitting to make the presubmit bots happy.
IIRC we use clang-format-6 on the bots. You can use the ./kokoro/presubmit/presubmit.sh to make sure everything is right. 

This only works if you have the same go/clang-format versions as our bots however, otherwise you can get differences.

cmd/gapir/cc/main.cpp
AWoloszyn: Yep the last version works for me. Thanks!

PR #2849, opened 2019-06-20
GLES API: Init framebuffer image data with proper slices

hevrard:
This is a first PR for a series of GLES dependency graph related PRs.

### DIFF
--- a/gapis/api/gles/gles.api
+++ b/gapis/api/gles/gles.api
@@ -744,22 +744,31 @@ sub void ApplyDynamicContextState(ref!Context ctx, ref!DynamicContextState dynam
   if (dynamicState != null) {
     backbuffer := ctx.Objects.Framebuffers[0]
 
+    width := dynamicState.BackbufferWidth
+    height := dynamicState.BackbufferHeight
+
+    colorFmtInfo := GetSizedFormatInfo(dynamicState.BackbufferColorFmt)
     backbuffer.ColorAttachments[0].Renderbuffer.Image = new!Image(
       Width: dynamicState.BackbufferWidth,
       Height: dynamicState.BackbufferHeight,
       SizedFormat: dynamicState.BackbufferColorFmt,
+      Data: make!u8(uncompressedImageSize(width, height, colorFmtInfo.UnsizedFormat, colorFmtInfo.DataType)),
     )
 
+    depthFmtInfo := GetSizedFormatInfo(dynamicState.BackbufferDepthFmt)
     backbuffer.DepthAttachment.Renderbuffer.Image = new!Image(
       Width: dynamicState.BackbufferWidth,
       Height: dynamicState.BackbufferHeight,
       SizedFormat: dynamicState.BackbufferDepthFmt,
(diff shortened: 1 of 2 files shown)


gapis/api/gles/gles.api
```diff
@@ -744,22 +744,31 @@ sub void ApplyDynamicContextState(ref!Context ctx, ref!DynamicContextState dynam
...
 
+    width := dynamicState.BackbufferWidth
+    height := dynamicState.BackbufferHeight
+
+    colorFmtInfo := GetSizedFormatInfo(dynamicState.BackbufferColorFmt)
     backbuffer.ColorAttachments[0].Renderbuffer.Image = new!Image(
       Width: dynamicState.BackbufferWidth,
       Height: dynamicState.BackbufferHeight,
       SizedFormat: dynamicState.BackbufferColorFmt,
+      Data: make!u8(uncompressedImageSize(width, height, colorFmtInfo.UnsizedFormat, colorFmtInfo.DataType)),
```
pmuetschard: since we don't ever write to this buffer, do we really need to allocate that much memory for it? can we not just make it an empty slice or a slice of size 1?

AWoloszyn: I think the main problem was that we DID write to this buffer. And it was in the app pool.
Well we at least mark writes into this buffer. We typically don't actually allocate the memory, although this might be something that is missed by the @spy_disabled annotation.
If we want to be sure the spy_disabled annotation is working correctly, we should just assign the data on the next line.

backbuffer.ColorAttachments[0].Renderbuffer.Image.Data = make!u8(...)
The buffers then won't really exist anywhere, Atlhough maybe in a MEC ?

hevrard:
So I tried a zero-lengh slice: it doesn't work. We need a non-empty slice to have read/write taken into account.

Following @AWoloszyn advice, I assigned the slices after the image object creation -- I think this is important for `@spy_disabled` given the comments, although it's not 100% clear to me.

Let mw know what you think!

AWoloszyn:
It looks good to me. I will let @pmuetschard be the one to approve since he had the comments.

gapis/api/gles/gles.api
pmuetschard: We don't (currently) pull back the default FB's data. In fact, the default FB's renderbuffers are not linked to the renderbuffer objects, which we should either fix, or may be as intended, I would have to check the spec. IIRC, as it stands now, this change has no effect on gapii, since the template will drop the assignment statement due to `@spy_disabled`, correct?

Since it worked with an obviously too small size (1byte/pixel) I would think it works with a one-byte slice as well, however, I don't have any strong objections to allocating it at the correct size, so I leave it up to @hevrard to decide.

AWoloszyn: That is correct, this would have no effect on gapii at all. And since slices take no "real" space in gapis, this should be fine there as well.

hevrard: I forgot to mention that I also tried with a 1-byte slice, and yes it works.

I think it's best to keep real-size allocation: this sounds more future-proof, and thanks to `@spy_disabled` it should not affect gapii / gapis significantly.

PR #2856, opened 2019-06-21
Fix tracing of Vulkan replay from archive

hevrard:


hevrard:
> Please be a bit more specific in the PR description of what the change is. For example, in this case a sentence or two of what the problem/fix is. Thank you!

Will do! Thanks

PR #2865, opened 2019-06-24
Fix logging by using vprint() style log function

hevrard:
Also revert to using "GAPID" as the logcat tag for all logging on Android.

The va_list was passed to __android_log_print, whereas the __android_log_vprint (note the "v") must be used. The compiler did not complain as __android_log_print is a variadic function, which accepts any type for its variadic args.

### DIFF
--- a/core/cc/log.cpp
+++ b/core/cc/log.cpp
@@ -144,7 +144,7 @@ void Logger::vlogf(unsigned level, const char* src_file, unsigned src_line,
       // There is no va_list version of __android_log_assert(), hence the
       // message pre-formatting in this case.
       char buf[2048];
-      vsnprintf(buf, sizeof(buf), format, args);
+      vsnprintf(buf, sizeof(buf), ss.str().c_str(), args);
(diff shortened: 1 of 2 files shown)


core/cc/log.cpp
```diff
@@ -129,32 +129,39 @@ void Logger::vlogf(unsigned level, const char* src_file, unsigned src_line,
...
+  }
   ss << "[" << src_file << ":" << src_line << "] " << format;
 
   switch (level) {
     case LOG_LEVEL_FATAL:
-      __android_log_assert(nullptr, mInstance.mSystem, ss.str().c_str(), args);
+      // There is no va_list version of __android_log_assert(), hence the
+      // message pre-formatting in this case.
+      char buf[2048];
+      vsnprintf(buf, sizeof(buf), format, args);
```
purvisa-at-google-com: src_file and src_line will be absent from the message?

hevrard: Indeed! I pushed a fix. Thanks!

PR #2866, opened 2019-06-24
Update the notification RPC to map its ID to a notification reader.

RenfengLiu:


gapis/replay/builder/builder.go
```diff
@@ -70,6 +70,10 @@ type postBackDecoder struct {
 	decode       Postback
 }
 
+const (
+	IssuesNotificationID = uint64(0)
```
AWoloszyn: We should at the very least mark this as needing to be in sync with the replay.
(I.e. put a comment here mentioning the other file and vice-versa)

PR #2880, opened 2019-06-26
Add support for template generated go and cpp file

RenfengLiu:
gofuse now seems can not collect the apic compiled files, trying to update the gofuse to collect them.

### DIFF
--- a/cmd/gofuse/main.go
+++ b/cmd/gofuse/main.go
@@ -152,6 +152,7 @@ func run() error {
 		return filepath.Join(fusedRoot, "src", trimUpTo(rel(projectRoot, path), "github.com"))
 	})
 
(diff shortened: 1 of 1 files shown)


cmd/gofuse/main.go
```diff
@@ -152,6 +152,22 @@ func run() error {
 		return filepath.Join(fusedRoot, "src", trimUpTo(rel(projectRoot, path), "github.com"))
 	})
 
+	templateGenedGofiles := collect(binOut, always).ifTrue(and(
```
hevrard: Add comment to say which files are retrieved here, e.g.:
`// Get ".go" files generated from templates`

And similar a few lines below:
`// Get ".cpp" and ".h" files generated from templates`

PR #2881, opened 2019-06-26
Update the Command Tree for Vulkan.

AWoloszyn:
This slightly re-works the command tree to be significantly
more explicit about what is run on CPU-side and what is run on
the GPU.

### DIFF
--- a/cmd/gapit/benchmark.go
+++ b/cmd/gapit/benchmark.go
@@ -312,6 +312,7 @@ func (verb *benchmarkVerb) Run(ctx context.Context, flags flag.FlagSet) error {
 		treePath.GroupByDrawCall = true
 		treePath.GroupByFrame = true
 		treePath.GroupByUserMarkers = true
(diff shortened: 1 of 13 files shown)


gapic/src/main/com/google/gapid/widgets/ImagePanel.java
```diff
@@ -1279,7 +1279,7 @@ public void accept(double value) {
 
         for (int bin = 0; bin < numBins; bin++) {
           float value = data.histogram.get(channel, bin);
-          value = (float)Math.pow(value, 0.3); // Exaggerate small values.
```
pmuetschard: unrelated change

AWoloszyn: Thanks, I thought I had removed that change, but obviously not.

AWoloszyn: Fixed

PR #2887, opened 2019-06-28
Improve resource cache prefetch logging

purvisa-at-google-com:
Make it more obvious when the prefetching did not prefetch everything as
a result of some data already existing in the cache; vs not prefetching
everything because it would overflow the cache memory limit.


purvisa-at-google-com:
Please ignore this review. I forgot to set CLANG_FORMAT="clang-format-6.0" and the pretty printer made a mess of the codebase.

PR #2894, opened 2019-07-01
Fix memory tracker destructor logic

hevrard:
DisableMemoryTrackerImpl() used to loop over the tracking_ranges_ map
using an iterator, yet it calls UntrackRangeImpl() which itselft also
uses an iterator on tracking_ranges_ map to erase the element, and
this does not play well with the iteration in
DisableMemoryTrackerImpl() loop. Hence we avoid using iterators to
loop over tracking_ranges_ in DisableMemoryTrackerImpl().

### DIFF
--- a/core/memory_tracker/cc/memory_tracker.h
+++ b/core/memory_tracker/cc/memory_tracker.h
@@ -443,10 +443,24 @@ class MemoryTrackerImpl : public SpecificMemoryTracker {
       std::function<void(void* dirty_addr, size_t dirty_size)> handle_dirty);
 
   bool DisableMemoryTrackerImpl() {
+    // Loop over tracking_ranges_ but DO NOT use an iterator, as
+    // UntrackRangeImpl() itself uses an iterator over tracking_ranges_ to
+    // *delete* an element.
+
+    struct range {
+      uintptr_t start;
+      uintptr_t end;
+    };
+    std::vector<struct range> ranges;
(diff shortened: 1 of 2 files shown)


core/memory_tracker/cc/memory_tracker.h
```diff
@@ -443,10 +443,23 @@ class MemoryTrackerImpl : public SpecificMemoryTracker {
...
+    // elements.
+
+    std::vector<uintptr_t> map_keys;
     for (auto& r : tracking_ranges_) {
-      result &= UntrackRangeImpl(reinterpret_cast<void*>(r.second->start()),
-                                 r.second->end() - r.second->start());
+      map_keys.push_back(r.first);
+    }
+
+    bool result = true;
```
AWoloszyn: I would strongly prefer we avoid `4*log(n)` operations. If we are going to push everything from tracking_ranges into a vector, can we just push the things we care about into the vector.

```
struct range { uintptr_t start; uintptr_t end }
std::vector<range> ranges;
ranges.reserve(tracking_ranges.size());
for (const auto& r: tracking_ranges_) {
    ranges.push_back({r.second.start(), r.second.end()});
}

iterate over ranges
```

hevrard: This assumes that UntrackRangeImpl() only deletes the tracking_ranges_ element which is referred by its arguments... which is the case :)
Done.

PR #2896, opened 2019-07-01
Fix stacktrace test

hevrard:
It looks like what was supposed to be line "nested2" was an
unfortunate copy-paste if line "nested3".

### DIFF
--- a/core/assert/slice.go
+++ b/core/assert/slice.go
@@ -53,64 +53,85 @@ func (o OnSlice) IsLength(length int) bool {
 
 // Equals asserts the array or slice matches expected.
 func (o OnSlice) Equals(expected interface{}) bool {
-	return o.slicesEqual(expected, func(a, b interface{}) bool { return a == b })
+	return o.slicesEqualTest(expected, func(a, b interface{}) bool { return a == b })
+}
+
+// Equals asserts the array or slice matches one of the expected array or slice
+func (o OnSlice) EqualsOneOf(expecteds interface{}) bool {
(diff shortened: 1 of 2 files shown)


hevrard:
We had a look with @pmuetschard :
- when compiled with `-c dbg`, this PR fix is needed for the test to pass
- when compiled with fastbuild, this PR fix will actually break the test(!)

So let's pause this PR for now.

AWoloszyn:
We could make the test smart enouhg to handle both cases 
EITHER have nested2 or nested3.  Pass if either pass.

hevrard:
> We could make the test smart enouhg to handle both cases
> EITHER have nested2 or nested3. Pass if either pass.

You mean by adding an "OR" in assert package? Isn't that adding too much complexity on our side for what is apparently a golang issue?

Also, we tried to get a small Golang repro example with Pascal, without success.

AWoloszyn:
So I disagree that this is a golang issue. As soon as you are no longer trying to keep debug info perfectly correct, it is normal for problems like this to creep up. `-dbg` will keep stack traces correct, but `fastbuild` and `opt` are likely to allow for some code-motion, and inlining. Stack-traces may not be exactly correct. It would be good to be able to handle both of these cases in the test.

hevrard:
Makes sense!
I started working on this but will finish next week (desk move...).

I went down the line of adding an `onSlice.EqualsOneOf()` function to assert/slice.go, which takes a `[][]string` in our case (it takes an `interface{}` that we assume to be a slice of slices, and we iterate on it).

Let me know what you think of this approach.

PR #2897, opened 2019-07-01
Resources cleanup

pmuetschard:
- No longer put the command id into the handle
 - Fix a warning in the UI code
 - Make the "show shaders/programs form all contexts" on by default
 - Use a proper table in the shader/program view


ben-clayton:
> I believe we talked about this at the time of my change.

Yes. I think the best way to deal with this sort of thing though is to include more structured information for the resource, including which context it belongs to. That way the user can pass a flag to `gapit` to optionally specify the context, and disambiguate ones with the same id.

PR #2899, opened 2019-07-01
Cleanup/Fix some exceptions reported to the crash server

pmuetschard:
Fixes a couple of frequently reported issues and don't report some that shouldn't be.

### DIFF
--- a/gapic/src/main/com/google/gapid/models/Capture.java
+++ b/gapic/src/main/com/google/gapid/models/Capture.java
@@ -50,6 +50,10 @@ import java.util.logging.Logger;
 public class Capture extends ModelBase<Capture.Data, File, Loadable.Message, Capture.Listener> {
   protected static final Logger LOG = Logger.getLogger(Capture.class.getName());
 
+  // Don't try to open files with 16 or less bytes. An empty graphics trace, without the
+  // capture header is already 16 bytes.
+  private static final int MIN_FILE_SIZE = 16;
+
   private final Settings settings;
   private String name = "";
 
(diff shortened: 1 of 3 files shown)


gapic/src/main/com/google/gapid/models/Capture.java
```diff
@@ -50,6 +50,10 @@
 public class Capture extends ModelBase<Capture.Data, File, Loadable.Message, Capture.Listener> {
   protected static final Logger LOG = Logger.getLogger(Capture.class.getName());
 
+  // Don't try to open files with 16 or less bytes. An empty graphics trace, without the
+  // capture header is already 16 bytes.
```
hevrard: "is already 16 bytes" -> is already *more than* 16 bytes?

gapic/src/main/com/google/gapid/models/Capture.java
```diff
@@ -80,9 +84,13 @@ public boolean isPerfetto() {
...
       return Futures.immediateFailedFuture(
-          new Exception("Trace file does not exist or is not accessible"));
+          new BadCaptureException("Trace file does not exist or is not accessible!"));
     } else if (file.length() == 0) {
-      return Futures.immediateFailedFuture(new Exception("Trace file is empty!"));
+      return Futures.immediateFailedFuture(
+          new BadCaptureException("Trace file is empty!"));
+    } else if (file.length() <= MIN_FILE_SIZE) {
+      return Futures.immediateFailedFuture(new BadCaptureException(
+          "Trace file is empty! Try capturing with buffering disabled."));
```
hevrard: The trace file is /technically/ not empty, maybe replace the error message by:
`"Trace file is too small to be valid! Try capturing with buffering disabled."`

gapic/src/main/com/google/gapid/models/Capture.java
pmuetschard: They are exactly 16bytes.

Issue #2892, opened 2019-07-01
GAPIR: replayer crashes when compiled with `-c opt`


hevrard:
This issue is known to be reproducible on commit 72898f38, but it was likely introduced earlier.

The replayer seems to crash here, but only when compiled with `bazel build -c opt`, like in the Kokoro builds:
https://github.com/google/gapid/blob/72898f38d62aaabf6ef25bb8d362c82bf54cc308/cmd/gapir/cc/main.cpp#L566

A few things to do to clean up:
- remove any `printf` in Android code, printf does not make sense on Android, replace with log calls where relevant
- move `GetArgsFromIntents()` into the empty namespace, lowercase the name
- we attach and detach two times to the JVM for checking presence of replay archive, and to retrieve the intent. Merge this two things into a single JVM attach/detach.

@purvisa-at-google-com can you have a go at this?

PR #2900, opened 2019-07-01
Fix crash when building with -c opt

purvisa-at-google-com:
The crash was being caused by subsequent calls to
GetMethodID/CallObjectMethod. Changed this to use the jni_call_o
template helper, and modified the calling code to only attach/detatch
the java environment to the current thread once on start-up. This
prevents the crash on launch.


hevrard:
Fix #2892

Issue #2791, opened 2019-05-31
Opening a trace triggers replay without trying to unlock device screen


hevrard:
The UI may be stuck waiting for a replay, but until the device screen is unlocked, GAPIR cannot properly replay.

We should use UnlockScreen() in pending PR #2789 or something similar to try to unlock the device, and report to the user when the device screen cannot be unlocked.

hevrard:
I would like to keep this one as an opportunity for first-time contribution.

PR #2903, opened 2019-07-02
Unlock screen when starting a new replay session

yalcinmelihyasin:
This mirrors the screen unlocking logic used in other places.

### DIFF
--- a/gapir/client/session.go
+++ b/gapir/client/session.go
@@ -289,6 +289,14 @@ var socketNames = map[device.Architecture]string{
 func (s *session) newADB(ctx context.Context, d adb.Device, abi *device.ABI, launchArgs []string) error {
 	ctx = log.V{"abi": abi}.Bind(ctx)
 
+	log.I(ctx, "Unlocking device screen")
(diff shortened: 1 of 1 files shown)


yalcinmelihyasin:
fix #2791

hevrard:
Please wait for the CI builds to finish before merging :)

gapir/client/session.go
```diff
@@ -289,6 +289,14 @@ var socketNames = map[device.Architecture]string{
 func (s *session) newADB(ctx context.Context, d adb.Device, abi *device.ABI, launchArgs []string) error {
 	ctx = log.V{"abi": abi}.Bind(ctx)
 
+	log.I(ctx, "Unlocking device screen")
```
AWoloszyn: Not sure if we need a log message every time we try and unlock.

AWoloszyn:
We may also want to make sure that gapir keeps the screen on. I think we can do something to force that.

gapir/client/session.go
AWoloszyn: From internal discussion, you can leave this as-is.

PR #2906, opened 2019-07-02
Update read/write to have relevant dependency for textured cube app

hevrard:
This change updates read/write such that dead code elimintation keeps
the relevant commands for frames of a simple textured spinning cube
app.

### DIFF
--- a/gapis/api/gles/api/draw_commands.api
+++ b/gapis/api/gles/api/draw_commands.api
@@ -345,6 +345,7 @@ extern u8[] ReadGPUTextureData(ref!Texture texture, GLint level, GLint layer)
 //////////////////
 
 // ReadFramebufferAttachment creates a read on this FramebufferAttachment.
+@spy_disabled
 sub void ReadFramebufferAttachment(FramebufferAttachment a) {
   switch (a.Type) {
     case GL_FRAMEBUFFER_DEFAULT, GL_RENDERBUFFER: {
@@ -360,6 +361,7 @@ sub void ReadFramebufferAttachment(FramebufferAttachment a) {
 }
 
 // WriteFramebufferAttachment creates a write on this FramebufferAttachment.
+@spy_disabled
 sub void WriteFramebufferAttachment(FramebufferAttachment a) {
   switch (a.Type) {
     case GL_FRAMEBUFFER_DEFAULT, GL_RENDERBUFFER: {
@@ -403,6 +405,7 @@ sub void DependencyReadTexture(ref!Texture texture) {
 }
 
 // DrawCommandDependencies creates relevant dependencies of draw commands.
+@spy_disabled
 sub void DrawCommandDependencies(ref!Context ctx) {
   read_fb := ctx.Bound.ReadFramebuffer
   draw_fb := ctx.Bound.DrawFramebuffer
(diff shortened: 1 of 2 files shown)


gapis/api/gles/api/draw_commands.api
```diff
@@ -374,6 +374,31 @@ sub void WriteFramebufferAttachment(FramebufferAttachment a) {
...
+  _ = texture.WrapT
+  _ = texture.WrapR
+  _ = texture.MinLod
+  _ = texture.MaxLod
+  _ = texture.BaseLevel
+  _ = texture.MaxLevel
+  _ = texture.DepthStencilTextureMode
+  _ = texture.CompareMode
+  _ = texture.CompareFunc
+}
```
pmuetschard: missing are `MaxAnisotropy` and `DecodeSRGB`

gapis/api/gles/api/draw_commands.api
```diff
@@ -389,8 +414,13 @@ sub void DrawCommandDependencies(ref!Context ctx) {
     ReadFramebufferAttachment(read_fb.DepthAttachment)
   }
 
-  // The program execution read uniforms
   for _, _, u in ctx.Bound.Program.UniformLocations {
+    texture_target := GetTextureTargetFromSamplerType(u.Type)
+    if texture_target != GL_NONE {
+      DependencyReadTexture(ctx, GetBoundTextureForUnit(ctx.Bound.TextureUnit, texture_target))
```
pmuetschard: This is not correct. `ctx.Bound.TextureUnit` is the last "activated" texture unit using `glActiveTexture`. However, the shader will read from the texture unit referenced by the uniform's value. E.g. if the uniform has value '3' it will read from `ctx.TextureUnits[3]`.

This is going to get complicated with uniform buffers... Maybe will have to pretend as though the shader will read from every active texture unit... sigh. For now, though, use the uniform's value and we'll deal with uniform buffers later.

gapis/api/gles/api/draw_commands.api
hevrard: thanks! done

gapis/api/gles/api/draw_commands.api
hevrard: Many thanks for this clarification, done!

hevrard:
Rebased on master, and applied review comments.

gapis/api/gles/api/draw_commands.api
```diff
@@ -374,6 +374,33 @@ sub void WriteFramebufferAttachment(FramebufferAttachment a) {
   }
 }
 
+// DependencyReadTexture read all relevant fields of texture to create dependencies
+sub void DependencyReadTexture(ref!Texture texture) {
```
AWoloszyn: We should mark this as @spy_disabled, since there is nothing that I can see in this function that needs to be done during trace. It will save us some performance.

hevrard: Can we also mark `DrawCommandDependencies()` as @spy_disabled?
It is only read/writes for dependencies.

AWoloszyn: Anything that doesn't change what is actually captured by Gapii can be spy_disabled.

PR #2907, opened 2019-07-02
GLES API: Add missing update of texture ImmutableLevels field

hevrard:
Also notice some missing checks.

### DIFF
--- a/gapis/api/gles/api/textures_and_samplers.api
+++ b/gapis/api/gles/api/textures_and_samplers.api
@@ -441,8 +441,8 @@ sub void TexImage(TexImageFlags  flags,
   }
 
   // TODO(hevrard): check that if t.ImmutableFormat == GL_TRUE, we are
-  // not trying to attempt to modify the format of dimension of any
-  // level (see GLES spec '8.18 Immutable-Format Texture Images')
+  // not trying to attempt to modify the format or dimension of any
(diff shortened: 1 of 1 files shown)


gapis/api/gles/api/textures_and_samplers.api
```diff
@@ -440,9 +440,16 @@ sub void TexImage(TexImageFlags  flags,
     }
   }
 
+  // TODO(hevrard): check that if t.ImmutableFormat == GL_TRUE, we are
+  // not trying to attempt to modify the format of dimension of any
```
AWoloszyn: format or dimension

gapis/api/gles/api/textures_and_samplers.api
```diff
@@ -440,9 +440,16 @@ sub void TexImage(TexImageFlags  flags,
...
+  // level (see GLES spec '8.18 Immutable-Format Texture Images')
```
pmuetschard: you should say which version of the spec you are talking about, otherwise the section number may not match up. Typically we refer to the 3.2 spec.

gapis/api/gles/api/textures_and_samplers.api
hevrard: done

gapis/api/gles/api/textures_and_samplers.api
hevrard: done

PR #2908, opened 2019-07-02
[WIP] Loop the trace to get the timestamps

RenfengLiu:
- Add the -loopcount flag to the `profile` cmd
- Added preLoop and postLoop the transform interface
- Loop by reset changed images and buffers between the start and end cmds;

### DIFF
--- a/cmd/gapit/flags.go
+++ b/cmd/gapit/flags.go
@@ -369,9 +369,10 @@ type (
 		CaptureFileFlags
 	}
 	GetTimestampsFlags struct {
-		Gapis GapisFlags
-		Gapir GapirFlags
-		Out   string `help:"output file to save the profiling result"`
+		Gapis     GapisFlags
+		Gapir     GapirFlags
+		LoopCount int    `help:"The number of times to loop the trace."`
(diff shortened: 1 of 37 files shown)


cmd/gapit/flags.go
```diff
@@ -369,9 +369,10 @@ type (
 		CaptureFileFlags
 	}
 	GetTimestampsFlags struct {
-		Gapis GapisFlags
-		Gapir GapirFlags
-		Out   string `help:"output file to save the profiling result"`
+		Gapis     GapisFlags
+		Gapir     GapirFlags
+		LoopCount int    `help:"The number of times to loop the trace."`
```
AWoloszyn: We should probably hide this until it is no longer experimental. If you add a _ to the start of the help message it will be hidden by default.

`help:"_The number of times to loop the trace. (experimental)"`

AWoloszyn:
We should also split this out into multiple changes. There are a bunch of independent changes that are being made, and it would make reviewing significantly easier.
1) Adding the pre-loop and post-loop to the transforms
2) The changes to the image primer
3) Actually adding the frame_loop code.

PR #2949, opened 2019-07-10
Fix dependency between glGenVertexArrays() and glBindVertexArray()

hevrard:


gapis/api/gles/api/vertex_arrays.api
```diff
@@ -80,7 +80,10 @@ sub ref!VertexArray GetOrCreateVertexArray(VertexArrayId id) {
   ctx := GetContext()
   if (id != 0) && (ctx.Objects.VertexArrays[id] == null) {
     ctx.Objects.VertexArrays[id] = NewVertexArray(id)
-    ctx.Objects.GeneratedNames.VertexArrays[id] = true
```
AWoloszyn: For future thinking: I wonder if we want to update the dependency graph generator to ignore no-op writes. (I.e. if there is not ACTUALLY a change, then can we avoid inserting the write?)

hevrard: Created #2952 to capture this.

PR #2951, opened 2019-07-11
Add dependencies for blending

hevrard:


gapis/api/gles/api/draw_commands.api
```diff
@@ -420,6 +420,20 @@ sub void DrawCommandDependencies(ref!Context ctx) {
...
+    if draw_fb.DrawBuffer[i] != GL_NONE {
+      blendState := ctx.Pixel.Blend[as!DrawBufferIndex(i)]
+      if blendState.Enabled == GL_TRUE {
+        _ = blendState.SrcRgb
+        _ = blendState.SrcAlpha
+        _ = blendState.DstRgb
+        _ = blendState.DstAlpha
+        _ = blendState.EquationRgb
+        _ = blendState.EquationAlpha
+      }
```
hevrard: The intended logic is: as soon as there is one drawbuffer which has blending enabled, read the blend state. So I would like to `break` from the loop here, because we don't need to read the blend state for all drawbuffers where it is enabled. Is it possible to break a loop in API language?

I tried to asbtract the loop in a subroutine to be able to do an early `return` in place of a `break`, but it was rejected by the API compiler as well.

AWoloszyn: Early Return and Break are both disallowed in our API language. This is because allowing them makes the analysis much more difficult.

hevrard:
Kokoro linux-test is failing to compile due to OOM, we'll fix that separately.

PR #2954, opened 2019-07-11
Work around instance issues in a new way.

AWoloszyn:
Some drivers will fail if you have 2 instances that have
different ApplicationName/EngineName values attached to them.

To work around this (yet another) issue, we leak the first instance
that was ever created.

### DIFF
--- a/gapir/cc/context.cpp
+++ b/gapir/cc/context.cpp
@@ -506,6 +506,22 @@ void Context::registerCallbacks(Interpreter* interpreter) {
       });
 
   interpreter->registerBuiltin(
+      Vulkan::INDEX, Builtins::ReplayDestroyVkInstance,
+      [this](uint32_t label, Stack* stack, bool) {
+        GAPID_DEBUG("[%u]replayDestroyVkInstance()", label);
+        if (mVulkanRenderer != nullptr) {
+          auto* api = mVulkanRenderer->getApi<Vulkan>();
+          return api->replayDestroyVkInstance(stack);
+        } else {
+          GAPID_WARNING(
+              "[%u]replayDestroyVkInstance called without a bound Vulkan "
+              "renderer",
+              label);
+          return false;
+        }
+      });
+
+  interpreter->registerBuiltin(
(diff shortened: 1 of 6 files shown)


gapir/cc/context.cpp
```diff
@@ -505,6 +505,22 @@ void Context::registerCallbacks(Interpreter* interpreter) {
...
       });
 
+  interpreter->registerBuiltin(
+      Vulkan::INDEX, Builtins::ReplayDestroyVkInstance,
+      [this](uint32_t label, Stack* stack, bool) {
+        GAPID_DEBUG("[%u]replayDestroyVkInstance()", label);
+        if (mVulkanRenderer != nullptr) {
+          auto* api = mVulkanRenderer->getApi<Vulkan>();
+          return api->replayDestroyVkInstance(stack);
+        } else {
```
RenfengLiu: The `else` may be omitted

gapis/api/templates/vulkan_gfx_api_extras.tmpl
```diff
@@ -221,6 +228,19 @@ bool Vulkan::replayCreateVkDeviceImpl(Stack* stack, size_val physicalDevice,
       return false;
     }
   }
+¶
+  bool Vulkan::replayDestroyVkInstance(Stack* stack) {
+    auto pAllocator = stack->pop<VkAllocationCallbacks const*>();
+    auto instance = static_cast<size_val>(stack->pop<size_val>());
+    // Leak this first instance. This is to work around a driver bug.
```
RenfengLiu: Do we need to check if the stack is valid or not here?

gapir/cc/context.cpp
AWoloszyn: It can, but I wanted to keep it consistent with the rest.

gapis/api/templates/vulkan_gfx_api_extras.tmpl
AWoloszyn: Thanks. We should. I will update.

PR #2961, opened 2019-07-12
Types 2

AWoloszyn:
This is the second PR for getting types into the UI. The first change is the same as the 1st one of these changes, so ignore that. The second actually does the work to output pointees correctly.


gapis/resolve/memory.go
```diff
@@ -115,3 +124,91 @@ func Memory(ctx context.Context, p *path.Memory, rc *path.ResolveConfig) (*servi
...
+	}
+
+	defer analytics.SendTiming("resolve", "memory")(analytics.Count(len(cmds)))
+
+	s, err := capture.NewState(ctx)
+	if err != nil {
+		return nil, err
+	}
+	err = api.ForeachCmd(ctx, cmds, func(ctx context.Context, id api.CmdID, cmd api.Cmd) error {
+		cmd.Mutate(ctx, id, s, nil, nil)
```
RenfengLiu: do we need to check the result of the `Mutate` here?

gapis/service/path/path.proto
```diff
@@ -48,23 +48,24 @@ message Any {
...
-    Result result = 31;
-    Slice slice = 32;
-    State state = 33;
-    StateTree state_tree = 34;
-    StateTreeNode state_tree_node = 35;
-    StateTreeNodeForPath state_tree_node_for_path = 36;
-    Stats stats = 37;
-    Thumbnail thumbnail = 38;
-    Type type = 39;
+    MemoryAsType memoryAsType = 23;
```
RenfengLiu: So here we reused the tag id, but I guess it is ok since the the gapis and gapit are always build and delivered the same time.

gapis/resolve/memory.go
AWoloszyn: This was mimicing the above call in memory.go in which we do not check the return. Typically we don't expect mutate to fail in the case where we haven't run a transform.

gapis/service/path/path.proto
AWoloszyn: That is correct. If we decide in the future to decouple them, we will have to be more careful.

cmd/gapit/common.go
```diff
@@ -384,12 +386,34 @@ func getConstantSet(ctx context.Context, client service.Service, p *path.Constan
 	return out, nil
 }
 
+var typeCache = map[uint64]*types.Type{}
+
+func getType(ctx context.Context, client service.Service, t *path.Type) (*types.Type, error) {
+	if tp, ok := typeCache[t.TypeIndex]; ok {
+		return tp, nil
+	} else {
```
RenfengLiu: maybe no `else` is needed.

cmd/gapit/common.go
```diff
@@ -402,6 +404,130 @@ func getType(ctx context.Context, client service.Service, t *path.Type) (*types.
 	}
 }
 
+func printPodValue(ctx context.Context, v *pod.Value) error {
```
ben-clayton: Seems valuable in the `pod` package. While protobuf steals the `String` method from you, you can implement [`Format`](https://golang.org/pkg/fmt/#Formatter). There's plenty of examples knocking around in the codebase.

gapis/memory/pool.go
```diff
@@ -37,8 +37,8 @@ import (
 // data occur.
 type Pool struct {
 	writes  poolWriteList
-	OnRead  func(Range, uint64)
-	OnWrite func(Range, uint64)
+	OnRead  func(Range, uint64, uint64)
```
ben-clayton: Would be lovely to have some documentation on these parameters - even a name would go a long way.

PR #2962, opened 2019-07-13
Send replay progress data from gapir to gapis.

stellama0208:
Send replay progress data from gapir to gapis using notification.

### DIFF
--- a/gapir/cc/interpreter.cpp
+++ b/gapir/cc/interpreter.cpp
@@ -70,8 +70,7 @@ Interpreter::Interpreter(core::CrashHandler& crash_handler,
       mInstructionCount(0),
       mCurrentInstruction(0),
       mNextThread(0),
-      mLabel(0),
-      mNumApiCallsMade(0) {
+      mLabel(0) {
   registerBuiltin(GLOBAL_INDEX, PRINT_STACK_FUNCTION_ID,
                   [](uint32_t, Stack* stack, bool) {
                     stack->printStack();
--- a/gapir/cc/interpreter.cpp
+++ b/gapir/cc/interpreter.cpp
@@ -32,8 +32,6 @@
 #include <utility>
 #include <vector>
 
-#define LOG_EVERY_N_API_CALLS 10000
-
 namespace gapir {
 
 namespace {
@@ -62,14 +60,12 @@ inline bool sum(Stack& stack, uint32_t count) {
 
 Interpreter::Interpreter(core::CrashHandler& crash_handler,
                          const MemoryManager* memory_manager,
-                         uint32_t stack_depth,
-                         ReplayService* srv)
+                         uint32_t stack_depth)
     :
 
       mCrashHandler(crash_handler),
       mMemoryManager(memory_manager),
       mStack(stack_depth, mMemoryManager),
-      mSrv(srv),
       mInstructions(nullptr),
       mInstructionCount(0),
       mCurrentInstruction(0),
@@ -85,6 +81,11 @@ Interpreter::Interpreter(core::CrashHandler& crash_handler,
 
 void Interpreter::setApiRequestCallback(ApiRequestCallback callback) {
   apiRequestCallback = std::move(callback);
+}
+
+void Interpreter::setCheckReplayStatusCallback(
+    CheckReplayStatusCallback callback) {
+  checkReplayStatusCallback = std::move(callback);
 }
 
 void Interpreter::registerBuiltin(uint8_t api, FunctionTable::Id id,
@@ -171,9 +172,7 @@ Interpreter::Result Interpreter::call(uint32_t opcode) {
   auto api = (opcode & API_INDEX_MASK) >> API_BIT_SHIFT;
   auto func = mBuiltins[api].lookup(id);
   auto label = getLabel();
-  if ((mNumApiCallsMade++ % LOG_EVERY_N_API_CALLS) == 0) {
-    mSrv->sendReplayStatus(label, mInstructionCount, mCurrentInstruction);
-  }
+  checkReplayStatusCallback(label, mInstructionCount, mCurrentInstruction);
   if (func == nullptr) {
     if (mRendererFunctions.count(api) > 0) {
       func = mRendererFunctions[api]->lookup(id);
--- a/gapir/cc/interpreter.cpp
+++ b/gapir/cc/interpreter.cpp
@@ -32,6 +32,8 @@
 #include <utility>
 #include <vector>
 
+#define LOG_EVERY_N_API_CALLS 10000
+
 namespace gapir {
 
 namespace {
@@ -60,17 +62,20 @@ inline bool sum(Stack& stack, uint32_t count) {
 
 Interpreter::Interpreter(core::CrashHandler& crash_handler,
                          const MemoryManager* memory_manager,
-                         uint32_t stack_depth)
+                         uint32_t stack_depth,
+                         ReplayService* srv)
     :
 
       mCrashHandler(crash_handler),
       mMemoryManager(memory_manager),
       mStack(stack_depth, mMemoryManager),
+      mSrv(srv),
       mInstructions(nullptr),
       mInstructionCount(0),
       mCurrentInstruction(0),
       mNextThread(0),
-      mLabel(0) {
+      mLabel(0),
+      mNumApiCallsMade(0) {
   registerBuiltin(GLOBAL_INDEX, PRINT_STACK_FUNCTION_ID,
                   [](uint32_t, Stack* stack, bool) {
                     stack->printStack();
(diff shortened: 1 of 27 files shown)


gapir/replay_service/service.proto
```diff
@@ -127,7 +134,8 @@ message Notification {
   uint64 id = 1;
   oneof val {
     ErrorMessage error_msg = 2;
-    Data data = 3;
+    ReplayStatus replay_status = 3;
```
AWoloszyn: I would prefer we add another ReplayResponse, instead of overloading Notification more. Having special values for the ID of notifications will get confusing and harder to maintain.

gapir/cc/interpreter.h
```diff
@@ -69,7 +70,8 @@ class Interpreter {
   // Creates a new interpreter with the specified memory manager (for resolving
   // memory addresses) and with the specified maximum stack size
   Interpreter(core::CrashHandler& crash_handler,
-              const MemoryManager* memory_manager, uint32_t stack_depth);
+              const MemoryManager* memory_manager, uint32_t stack_depth,
+              ReplayService* srv);
```
AWoloszyn: We should probably just have an std::function or similar here, instead of passing the entire ReplayService object down, just so we can get at one method. The rest of our methods are handled through what amounts to callbacks, so we should do something similar here, instead of coupling further.

gapir/cc/interpreter.cpp
```diff
@@ -166,6 +171,9 @@ Interpreter::Result Interpreter::call(uint32_t opcode) {
   auto api = (opcode & API_INDEX_MASK) >> API_BIT_SHIFT;
   auto func = mBuiltins[api].lookup(id);
   auto label = getLabel();
+  if ((mNumApiCallsMade++ % LOG_EVERY_N_API_CALLS) == 0) {
```
AWoloszyn: I would prefer if we sent the status every X% rather than N api calls. That would mean more reliable data coming out. Also that should PROBABLY be a function of the ReplayRequest rather than a hard-coded value.

gapis/replay/builder/builder.go
```diff
@@ -650,6 +655,20 @@ func (b *Builder) RegisterNotificationReader(notificationID uint64, reader Notif
...
+// RegisterReplayStatusReader create and register a NotificationReader, which is used to decode and handle
+// replay status information sent from GAPIR.
+func (b *Builder) RegisterReplayStatusReader(ctx context.Context) error {
+	reader := func(n gapir.Notification) {
+		r := n.GetReplayStatus()
+		label := r.GetLabel()
+		total_instrs := r.GetTotalInstrs()
+		finished_instrs := r.GetFinishedInstrs()
+
+		log.I(ctx, "Replay status: Label: %v; Total instructions: %v; Finished instructions: %v.", label, total_instrs, finished_instrs)
```
AWoloszyn: We probably don't want this in the log by default (Which is what log.I will do). For large traces this will fill the log rather quickly. It would be better to log.D(ctx, ...) which will only show up with debug logging enabled. Any client should be able to get the data from the status package if it cares.

gapir/replay_service/service.proto
stellama0208: I see, make sense! Will try to refactor the code to address this problem.

gapir/cc/interpreter.h
stellama0208: Thanks! This is a much better code pattern! Have made another commit to address this problem.

gapir/replay_service/service.proto
stellama0208: Hey Andrew, as I really put my hands on this work, trying to make it a separate ReplayResponse, I found I may include some more code complexity in this way. Some of the observations and thoughts are given below. Would you mind help me see whether these make sense and shall I keep the structure this way?

1.There's already many handlers to deal with in our builder. 
I will have to create another handler and return it in builder.Build(), which will make our code heavy and complex.
2.The definition of Notification: "sending data from GAPIR to GAPIS" match my needs here exactly. 
It simply handle data transfer and I don't see ways to make it more concise. I feel I will write a new set of exactly the same code if I really do it.
3.The notification framework is written very well but not used much.
Currently the customized sendNotificationData() is not called in our codebase (from my observation), the only notification method in use is sendErrorMsg(). I feel it is a waste not putting it into use and generate another logic line.

AWoloszyn: So the notification is meant to be free-form data. Basically a bunch of Bytes. (Similar to how Post works) we extended it to work with error messages, I am just worried about adding more types of structured data to it.

That being said we could update it with a slightly different notification setup. I just dont like the idea of overloading ID with more special values. WE could move ID into the "Data" message, and leave the other notifications out. That would be fine with me.

AWoloszyn: For example:

```
// Data is a message contains data that a GAPIR device collected during the
// replay.
message Data {
  uint64 id = 1;
  uint64 label = 2;
  bytes data = 3;
}

// Notification is a message that GAPIR device wants to send to GAPIS. This
// message can be used to send back data such as error message or other data to
// GAPIS.
message Notification {
  oneof val {
    ErrorMessage error_msg = 1;
    ReplayStatus replay_status = 2;
    Data data = 3;
  }
}
```

That way we can switch on `val.(type)` rather than an arbitrary ID that means a special thing.

RenfengLiu: The original modification to add an `ID` to `Notification` is to map the `ID` to the `notificaitonReader`, if we switch on type, then we need to find a  way to map the notification with its reader.

stellama0208: I see... So if we move the ID into Data, and switch on val.(type):
The way we carry Notification Readers in Builder might transfer to 
`	notificationReaderReplayStatus NotificationReader
	notificationReaderErrorMessage NotificationReader
	notificationReadersData map[uint64]NotificationReader`
to still be able to pinpoint the reader without an ID.

It can transfer the complexity from constants to builder structure.
May I ask which practice will benefit more in the future?
And how do we plan to use the customized Notification.Data? (If we don't use it we can directly get away with the map)

stellama0208: Sounds reasonable. And maybe we can generate a new struct NotificationReaders to wrap around all the readers to organize things better. 

If Robin also agrees with it, I can do this according to our discussion. ^_^

gapis/server/server.go
```diff
@@ -599,11 +600,22 @@ func (l *statusListener) OnMemorySnapshot(ctx context.Context, stats runtime.Mem
 	})
 }
 
-func (s *server) Status(ctx context.Context, snapshotInterval, statusInterval time.Duration, f func(*service.TaskUpdate), m func(*service.MemoryStatus)) error {
+func (l *statusListener) OnReplayStatusUpdate(ctx context.Context, label uint64, total_instrs uint32, finished_instrs uint32) {
+	l.progressMutex.Lock()
```
stellama0208: Can I delete these two lines to send replay status update without mutex lock on? Want to avoid lock for efficiency, will do the progress non-decreasing check at client side.

ben-clayton: I would say yes - it looks like the mutex is only there to guard `lastProgressUpdate` and `progressUpdateFreq`.
There's a whole bunch of locking and unlocking in other functions that looks redundant and could be removed.
I'd suggest commenting the fields that they belong to the mutex, so this is less confusing in the future:

```go
	lastProgressUpdate map[*status.Task]time.Time // guarded by progressMutex
	progressUpdateFreq time.Duration // guarded by progressMutex
	progressMutex      sync.Mutex
```

stellama0208: Thank you Ben for answering! It's of great help! Will add the comments.

AWoloszyn:
My only question is: what would happen if we have 2 different devices replaying at the same time?

stellama0208:
> My only question is: what would happen if we have 2 different devices replaying at the same time?

Yeah it's worth paying attention to. The short answer would be: the updates will get mixed.

For the current implementation, things all start from Interpreter::call. When it's time to report, a message will be sent all the way to gapic. The message only contains progress info, but no device identifying info. So the UI will update the things undifferentiatedly. If two threads execute Interpreter::call at the same time, I think things will get messy. 

The current implementation is based on the assumption that only one replay at a time. I also thought it was the case for GAPID. So do we support parallel replay now? And how can I enable it and test it out? If so I plan to see how to add device identifying info to address this problem.

AWoloszyn:
So we absolutely can have 2 different devices playing at the same time. This currently is not the default, but we should probably consider how to handle this.

You can do this right now by starting gapis manually.

`./bazel-bin/pkg/gapis -enable-local-files -persist -rpc localhost:40000`
Then you can open 2 clients and connect to the same gapis
`./bazel-bin/pkg/gapid --gapis localhost:40000`

These can both schedule things ont othe same device, or separate devices. It would probably make sense to add append the path.Device to the status message. That way the UI Can choose the correct device for its own session.

We can do this by making the replay status have a path.Device, and we can pass that down into the builder. When the builder fills the message it can add the path.Device there.

stellama0208:
> So we absolutely can have 2 different devices playing at the same time. This currently is not the default, but we should probably consider how to handle this.
> 
> You can do this right now by starting gapis manually.
> 
> `./bazel-bin/pkg/gapis -enable-local-files -persist -rpc localhost:40000`
> Then you can open 2 clients and connect to the same gapis
> `./bazel-bin/pkg/gapid --gapis localhost:40000`
> 
> These can both schedule things ont othe same device, or separate devices. It would probably make sense to add append the path.Device to the status message. That way the UI Can choose the correct device for its own session.
> 
> We can do this by making the replay status have a path.Device, and we can pass that down into the builder. When the builder fills the message it can add the path.Device there.

Thanks Andrew for pointing this out! Just tested, things will indeed being updated across both of the clients. Will start solving things in this way.

And during the testing, I found the memory usage updates and server task updates are also shared across clients. Are these expected behavior? (I guess maybe the memory status should be shared but server task updates should be separate?) Anyway, I will do the solution for replay status update first, if things work out fine, I can also apply similar strategy for other updates if needed!

PR #2967, opened 2019-07-15
Rough work on getting advanced trace features in to the UI.

AWoloszyn:
Rough-out adding advanced trace options in the UI.

### DIFF
--- a/gapic/src/main/com/google/gapid/views/TracerDialog.java
+++ b/gapic/src/main/com/google/gapid/views/TracerDialog.java
@@ -290,6 +290,9 @@ public class TracerDialog {
       private final Button hideUnknownExtensions;
       private final Button clearCache;
       private final Button disablePcs;
+      private final Button advanced;
+      private Label ofl;
+      private Button observeFramebuffers;
       private final FileTextbox.Directory directory;
       protected final Text file;
       private final Label pcsWarning;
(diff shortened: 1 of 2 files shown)


gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -458,6 +461,41 @@ protected void configureDialog(DirectoryDialog dialog) {
...
+        observeFramebuffers = null;
+        ofl = null;
+        if (models.settings.traceAdvanced) {
+          ofl = createLabel(this, "");
+          observeFramebuffers = withLayoutData(
+              createCheckbox(this, "Observe Framebuffers", models.settings.traceObserveFramebuffers),
+              new GridData(SWT.FILL, SWT.FILL, true, false));
+        }
+
+        Listener advancedListener = e -> {
```
AWoloszyn: This is definitely not what we want. But I dont know swt enough to make anything cleaner.

gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -458,6 +461,41 @@ protected void configureDialog(DirectoryDialog dialog) {
...
+          } else {
+            ofl.dispose();
+            ofl = null;
+            observeFramebuffers.dispose();
+            observeFramebuffers = null;
+          }
+          layout(true, true);
+          parent.layout(true, true);
+
+          parent.getParent().getParent().layout(true, true);
```
AWoloszyn: This is obviously wrong, any suggestions.

gapic/src/main/com/google/gapid/views/TracerDialog.java
stellama0208: The SWT widgets' method .setVisible(boolean) might be helpful. You may use it to replace creating and disposing.

stellama0208: It's worth noting that after .setVisible(false), the widget will still occupy a space, it just goes transparent. So if there's other setting checkboxes following in the future, there would be a weird blank line in the middle. I think maybe you can:
1.Make the dependent checkbox staying on the same line, and use .setVisible(false) when unwanted. No blank line.
2.Give the dependent checkbox some indent, and use .setEnabled(false) when unwanted. The checkbox will go grey and the UX is still straightforward.

AWoloszyn:
I will close this, we need a proper UI for this.

PR #2969, opened 2019-07-16
Split test targets to work around Kokoro OOM

yalcinmelihyasin:
Separate tests into different logical groups to prevent to hit memory limit in CI.

### DIFF
--- a/kokoro/linux-test/test.sh
+++ b/kokoro/linux-test/test.sh
@@ -33,11 +33,21 @@ export CC=/usr/bin/gcc-7
 cd $SRC
 BUILD_SHA=${KOKORO_GITHUB_COMMIT:-$KOKORO_GITHUB_PULL_REQUEST_COMMIT}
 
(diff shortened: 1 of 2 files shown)


kokoro/linux-test/test.sh
```diff
@@ -33,11 +33,20 @@ export CC=/usr/bin/gcc-7
...
+        --output_base="${TMP}/bazel_out" \
+        test -c opt --config symbols \
+        --define GAPID_BUILD_NUMBER="$KOKORO_BUILD_NUMBER" \
+        --define GAPID_BUILD_SHA="$BUILD_SHA" \
+        --test_tag_filters=-needs_gpu \
+        $@
+    echo $(date): Tests completed.
+}
+
+test tests-general
```
hevrard: Add a comment along the lines of:
`# Running all the tests in one go leads to an out-of-memory error on Kokoro, hence the division in smaller test sets`

PR #2973, opened 2019-07-18
Make sure to touch all pages, in replay APK capture logic

hevrard:


gapir/cc/android/asset_resource_cache.cpp
```diff
@@ -48,13 +48,16 @@ bool asset_read(AAsset *asset, void *buf, size_t count) {
   return true;
 }
 
-// touch pages will touch all memory pages for a given memory span
+// touch_pages will touch all memory pages for a given memory span
```
AWoloszyn: We should either update this comment, or slightly change the logic here. 

"touch_pages will write at least one 0 onto every page in the given memory span"

Alternatively we should read and write a the same value, so that at least the contents of the pages are unchanged.

gapir/cc/android/asset_resource_cache.cpp
```diff
@@ -48,13 +48,16 @@ bool asset_read(AAsset *asset, void *buf, size_t count) {
...
 void touch_pages(void *addr, uint32_t size) {
-  long pagesize = sysconf(_SC_PAGESIZE);
+  static const long pagesize = sysconf(_SC_PAGESIZE);
   char *end = ((char *)addr) + size;
   for (char *p = (char *)addr; p < end; p += pagesize) {
     *p = '0';
   }
+  // Make sure the last page is touched, as the loop may exit without touching
+  // it when p lands in the last page to touch, but beyond end.
+  *(end - 1) = '0';
```
AWoloszyn: if (size == 0) this will write out of bounds.

gapir/cc/android/asset_resource_cache.cpp
```diff
@@ -148,6 +151,9 @@ bool AssetResourceCache::loadCache(const Resource &resource, void *data) {
         // directly touch all pages of the destination memory to get the memory
         // tracker in a good state, and retry the read(). But try this only
         // once.
+        GAPID_WARNING(
```
AWoloszyn: This will get very spammy, if this actually happens.
We should consider only having this output once, or only periodically.

hevrard: Removed.

gapir/cc/android/asset_resource_cache.cpp
hevrard: Yes I was assuming this function to be called only for with non-zero size resource as parameter, but we'd better check this indeed.

PR #2974, opened 2019-07-18
Make cmd/gapid prefer java in JAVA_HOME over PATH

### LINKED ISSUE #2878: GAPID should prefer the Java from `$JAVA_HOME` over `$PATH`
Running GAPID on gLinux fails with this error:

```
Exception in thread "main" java.lang.UnsatisfiedLinkError: Could not load SWT library. Reasons: 
	no swt-pi3-gtk-4919 in java.library.path
```

I tried this with any version from 1.2.1 to 1.6. It looks like the libraries are in `~/.swt/lib/linux/x86_64/`, but doing `LD_LIBRARY_PATH=$LD_LIBRARY_PATH:~/.swt/lib/linux/x86_64/ /opt/gapid/gapid` still fails:

```
Exception in thread "main" java.lang.UnsatisfiedLinkError: Could not load SWT library. Reasons: 
	~/.swt/lib/linux/x86_64/libswt-pi3-gtk-4919.so: /usr/grte/v4/lib64/libc.so.6: version `GLIBC_2.28' not found (required by /usr/lib/x86_64-linux-gnu/libgio-2.0.so.0)
```

hevrard:
The cmd/gapid command that starts gapic should prefer the java found in JAVA_HOME over the one in PATH.

Yet, setting JAVA_HOME to java 11 leads to Bazel using it for the build, which fails with:
```
/com/google/gapid/proto/service/GapidGrpc.java:23: error: package javax.annotation does not exist
```

Because javax.annotation package has been dropped in java 11.

So we need to conditionnally depend on javax.annotation:javax.annotation-api when compiling with java >= 11, see:
https://stackoverflow.com/questions/48204141/replacements-for-deprecated-jpms-modules-with-java-ee-apis/48204154#48204154

@pmuetschard is there a way to do a such conditional dependency based on java version in Bazel?

Aim to fix #2878

### DIFF
--- a/cmd/gapid/main.go
+++ b/cmd/gapid/main.go
@@ -182,16 +182,16 @@ func (c *config) locateVM() error {
 		return nil
 	}
 
-	if java, err := exec.LookPath(c.javaExecutable()); err == nil && checkVM(java, true) {
-		c.vm = java
-		return nil
-	}
-
 	if home := os.Getenv("JAVA_HOME"); home != "" {
 		if java := c.javaInHome(home); checkVM(java, true) {
 			c.vm = java
 			return nil
 		}
+	}
+
+	if java, err := exec.LookPath(c.javaExecutable()); err == nil && checkVM(java, true) {
+		c.vm = java
+		return nil
 	}
 
 	if runtime.GOOS == "linux" {


AWoloszyn:
I think we need a separate bug to update to a more recent version of java in general. However this is a bit of a tricky problem, since it affects more than just our build.

Issue #2952, opened 2019-07-11
Dependency graph: ignore writes which do not change the state?


hevrard:
Given:
```
bool state

cmd void A() {
   state = true
}

cmd void B() {
  state = true
}

cmd void C() {
  _ = state
}
```
and a trace `A, B, C`, the dependencies of C will be B.

Yet the when building the dependency graph, we could detect that the write in B does not change the state, and thus "ignore" this write, which would lead to the dependencies of C being A.

bjoeris:
Just a few thoughts. It is possible this could be a win overall, but I think there are some complications we should consider.

Suppose `A` is a big, expensive operation that just happens to set `state`--then this optimization will pull in `A` unnecessarily. Or, similarly, maybe there is a `D` that depends on `B` and `C`, so `B,C,D` would be a valid set of transitive dependencies for `D`, but it looks like this optimization would give us `A,B,C,D` instead. Of course, both of these examples could be turned around by just swapping the order of `A` and `B`, and give examples of wins for this optimization.

I think the "correct" answer is that `C` really depends on "`A` or `B`". This kind of higher-order dependency doesn't fit well in the dependency graph model, and isn't currently supported. But it might be worth thinking about whether there is a way to support dependencies like this, because they certainly come up; e.g. for image layout, there may be several commands in the trace that transition to the correct layout--we don't really depend on any particular layout transition command, as long as the image winds up in the correct layout (and as long as there are no other commands in between that transition back out of that layout). These also probably come up for synchronization dependencies.

hevrard:
I've been thinking about this: would it make sense to ignore commands that do not change the state if the piece of state they touch was last updated using the same command?

This comes out of #2976 , where I wanted to get rid of `glUseProgram()` duplicates. More generally, given:

```
cmd void A(i) {
   state = i
   [other expensive operations]
}

cmd void B(i) {
   _ = state
   state = i
}

cmd void C() {
  _ = state
}
```

And a trace:

```
A(42)
B(42)
B(42)
C()
```

As per @bjoeris remark, we would like to avoid C to depend on A, because A is expensive.
Yet, here it would make sense to depend on B, but to not depend on both B's, because only a single B call is enough.

How does that sound?

AWoloszyn:
If a command leads to zero change in state, we should probably drop it. That would be a simple and sensible optimization.

However, we should take a look to see WHAT we are trying to do with the dependency graph. It has 2 purposes.
[...]

PR #2976, opened 2019-07-18
Make dead-code-elimination remove glUseProgram() duplicates

hevrard:
When glUseProgram() is called with the same program id as the
currently bound one, do not write on ctx.Bound.Program to avoid an
unnecessary dependency.

### DIFF
--- a/gapis/api/gles/api/programs_and_shaders.api
+++ b/gapis/api/gles/api/programs_and_shaders.api
@@ -2380,9 +2380,11 @@ cmd void glUseProgram(ProgramId program) {
   } else {
     p := GetProgramOrError(program)
     if p.LinkStatus == GL_FALSE { glErrorInvalidOperation() }
-    AdjustProgramUseCount(p, +1)
-    AdjustProgramUseCount(ctx.Bound.Program, -1)
-    ctx.Bound.Program = p
+    if p != ctx.Bound.Program {
+      AdjustProgramUseCount(p, +1)
+      AdjustProgramUseCount(ctx.Bound.Program, -1)
+      ctx.Bound.Program = p
+    }
   }
 }
 


gapis/api/gles/api/programs_and_shaders.api
```diff
@@ -2380,9 +2380,11 @@ cmd void glUseProgram(ProgramId program) {
   } else {
     p := GetProgramOrError(program)
     if p.LinkStatus == GL_FALSE { glErrorInvalidOperation() }
-    AdjustProgramUseCount(p, +1)
-    AdjustProgramUseCount(ctx.Bound.Program, -1)
-    ctx.Bound.Program = p
+    if p != ctx.Bound.Program {
```
AWoloszyn: Doesn't this CREATE a dependency between glUseProgram calls?
Now we are reading ctx.Bound.Program, rather than just writing it.

hevrard: Good point! I think this changes is still valid, but my explanation was not clear.

This change makes sense because a draw call will only depend on the latest UseProgram that wrote to ctx.Bound.Program. Consider:
```
1. glUseProgram(42)
2. glUseProgram(42)
3. glDraw()
```

Without the change, glUseProgram() calls AdjustProgramUseCount which performs a read-write to program.UseCount. This creates a dependency between 1 and 2. The draw call in 3 depends on 2, because 2 is the latest who wrote to ctx.Bound.Program.

With the change, there is still a dependency between 1 and 2, as you pointed out. Yet, 2 does NOT write on ctx.Bound.Program. So the draw call 3 depends on **1** this time, and not on 2. Thanks to DCE, 2 will be removed.

Does that make sense?

I will make sure the commit message clarifies this.

pmuetschard: I think, though, the problem is that if:
```
1. glUseProgram(1)
2. glUseProgram(2)
3. glDraw()
```
It will now no longer eliminate 1.

Can we make the dependency tracking smarter and have it ignore writes of the same value? i.e. outside of the API files?

hevrard: Ah yes indeed, thanks for the clarification.

See #2952 for a discussion about whether the dependency graph ignoring writes that do not modify the state.

AWoloszyn: Is the count something that should be tracked for dependencies, or is the count only something that is used internally? If the latter, we should not track  it.

hevrard: The count is relevant to know whether the program can be deleted:
```
sub void AdjustProgramUseCount(ref!Program p, s32 delta) {
  ctx := GetContext()
  if p != null {
    p.UseCount += delta
    assert(p.UseCount >= 0)
    // Delete if it was marked for deletion and it is no longer used anywhere
    if p.DeleteStatus && (p.UseCount == 0) {
      for _ , _ , s in p.Shaders {
        AdjustShaderAttachCount(s, -1)
      }
      delete(ctx.Objects.Programs, p.ID)
    }
  }
}
```

hevrard:
Closing as the idea behind this PR actually breaks dead code elimination.

PR #2979, opened 2019-07-23
Read all levels and layers of texture to support e.g. mipmap

hevrard:


gapis/api/gles/api/draw_commands.api
```diff
@@ -379,7 +379,12 @@ sub void WriteFramebufferAttachment(FramebufferAttachment a) {
 // DependencyReadTexture read all relevant fields of texture to create dependencies
 @spy_disabled
 sub void DependencyReadTexture(ref!Texture texture) {
-  read(texture.Image.Data)
+  // Read all levels and layers
+  for _, _, level in texture.Levels {
```
pmuetschard: This should respect the textures min filter and ignore levels unless its mipmaped?
At the least it needs to respect `BaseLevel` and `MaxLevel` and possibly `MinLod` and `MaxLod`

I'm OK, though, if for now that's just a todo.

hevrard: Thanks for this! I added the check to BaseLevel and MaxLevel. I will do MinLod and MaxLod in a later PR.

PR #2984, opened 2019-07-24
Update to latest Perfetto and add GPU Render Stage UI

pmuetschard:
- updates to latest Perfetto in order to pick up new trace processor changes that include the GPU render stages and GPU counters processing
- small fix to handle traces that don't contain any sched data or no CPU frequency data
- UI for displaying GPU render stages slices


pmuetschard:
@rschiu: FYI

gapic/src/main/com/google/gapid/perfetto/models/Tracks.java
```diff
@@ -55,14 +57,26 @@ private Tracks() {
...
-          }
-        }, true));
-    return CpuTrack.enumerate(summary.getId(), data);
+    return transform(CpuTrack.enumerate(summary.getId(), data), configs -> {
+      boolean hasAnyFrequency = configs.stream().anyMatch(c -> c.hasFrequency);
+      Group.UiFactory ui = hasAnyFrequency ?
+          group(state -> new CpuSummaryPanel(state, summary), false, (group, filtered) -> {
+              for (int cpu = 0, track = 0; cpu < configs.size(); cpu++, track++) {
+                if (configs.get(cpu).hasFrequency) {
+                  track++;
```
pau-baiget: You're increasing ```track``` twice here. Is that the intended behavior?

pau-baiget: Ah, one track for the CPU, one track for the frequency, right?

pmuetschard: right.

PR #2991, opened 2019-07-25
Loop the trace to get timestamps 

RenfengLiu:
First step in frame looping. Loop the trace from the first command
to the last one. Follow up CLs will add code
to detect changed resources and reset them before next iteration of
loop.

### DIFF
--- a/gapis/api/vulkan/frame_loop.go
+++ b/gapis/api/vulkan/frame_loop.go
@@ -137,7 +137,7 @@ func (f *frameLoop) Transform(ctx context.Context, id api.CmdID, cmd api.Cmd, ou
 	ctx = log.Enter(ctx, "FrameLoop Transform")
 
 	if cmd == f.loopStartCmd {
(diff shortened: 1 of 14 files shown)


gapis/api/vulkan/frame_loop.go
```diff
@@ -0,0 +1,475 @@
...
+func (f *frameLoop) backupChangedImages(ctx context.Context, sb *stateBuilder) error {
+	imgPrimer := newImagePrimer(sb)
+	s := GetState(sb.oldState)
+	defer imgPrimer.Free()
+	for img := range f.imageChanged {
+		if _, present := f.imageCreated[img]; present {
+			log.I(ctx, "Image [%v] is created in between.", img)
+			continue
+		}
+		log.I(ctx, "Image [%v] changed during loop.", img)
```
AWoloszyn: We should probably remove some of the logging. Or at least make them log.D

PR #2998, opened 2019-07-28
Only print out the bit of a slice in gapit commands.

AWoloszyn:
Otherwise these slices can get very long, and clog the output.

### DIFF
--- a/cmd/gapit/common.go
+++ b/cmd/gapit/common.go
@@ -424,7 +424,12 @@ func printBoxValue(ctx context.Context, client service.Service, t *path.Type, v 
 		oldPrefix := prefix
 		prefix := prefix + "│   "
 		fmt.Printf("\n")
+
 		for i := 0; i < len(sliceData.Values); i++ {
+			if i > 9 {
+				fmt.Printf("%s└──  ... %v more\n", oldPrefix, len(sliceData.Values)-i)
+				break
+			}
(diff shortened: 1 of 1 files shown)


cmd/gapit/common.go
```diff
@@ -545,6 +550,10 @@ func printCommand(ctx context.Context, client service.Service, p *path.Command,
 				if sl == nil {
 					return fmt.Errorf("Observations are expected to be slices")
 				}
+				if (tm.Range.Size > 1024) {
```
hevrard: I understand the logic to be "if it's bigger than 1k, don't even bother displaying the first 10 values". Yet for the sake of predictability of `gapit commands` output, maybe we want to always display the first 10 values?

AWoloszyn: So agreed, but then we have to actually pull the data back from the server. Which can take a significant amount of time. This was to speed it up, in the cases where we don't even want the data. 

There is a new RPC which will give the size of the underlying object, so we could use that, but it means one more round-trip per request. 

Since this is mostly for testing, I am ok either way if you have a preference.

hevrard: Oh OK I didn't have the data transfer aspect in mind. Let's keep your approach then!

hevrard:
with presubmit to be fixed :)

PR #3000, opened 2019-07-29
Add dependency reads for vertex input data

hevrard:


gapis/api/gles/api/draw_commands.api
```diff
@@ -452,6 +452,31 @@ sub void DrawCommandDependencies(ref!Context ctx) {
...
 
+  // Vertex input
+  if !IsDefaultVertexArrayBound() {
+    // Read host-side vertex input data. glDraw*() typically calls ReadVertexArrays() which
+    // reads client-side vertex input data, the only necessary reads at capture time.
+    // Here the reads for host-side vertex input is done to create dependencies.
+    arrayBuffer := ctx.Bound.ArrayBuffer
+    read(arrayBuffer.Data)
+    for _, _, a in ctx.Bound.VertexArray.VertexAttributeArrays {
+      if a.Enabled == GL_TRUE {
```
AWoloszyn: In general are all of these read in all circumstances?
I don't know gles well but are a.Integer and a.Pointer both read every time?

gapis/api/gles/api/draw_commands.api
```diff
@@ -452,6 +452,31 @@ sub void DrawCommandDependencies(ref!Context ctx) {
...
+        _ = a.Type
+        _ = a.Normalized
+        _ = a.Integer
+        _ = a.Pointer
+        _ = a.RelativeOffset
+	_ = a.Binding.Id
+	_ = a.Binding.Buffer
+	_ = a.Binding.Offset
+	_ = a.Binding.Stride
+	_ = a.Binding.Divisor
```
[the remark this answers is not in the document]
hevrard: Yes, I went down all the way to add indice and instance count into consideration for draw call dependencies. Can you have another look now?

gapis/api/gles/api/draw_commands.api
```diff
@@ -183,23 +190,46 @@ sub void DrawElementsBaseVertex(GLenum draw_mode, GLsizei indices_count, GLenum
...
+  // }
+  //
+  // ## The cast as!void*() below fails:
+  // DrawElementsInstancedBaseVertex(draw_mode, as!GLsizei(command.count), indices_type, as!void*(command.firstIndex * as!u32(size_of_type)), as!GLsizei(command.instanceCount), as!GLint(command.baseVertex))
+
+  // TODO: once the cast issue above is fixed, we should be able to get rid of these three calls:
+  checkPrimitiveType(draw_mode)
   WriteGPUFramebufferData(ctx)
-  DrawCommandDependencies(ctx)
+  DrawCommandDependencies(ctx, as!DrawElementsIndirectCommand*(indirect)[0].count, as!DrawElementsIndirectCommand*(indirect)[0].instanceCount)
```
hevrard: Here we cannot abstract `as!DrawElementsIndirectCommand*(indirect)[0]` in a local variable because it triggers a compilation error down the line. The error is due to the fact that DrawCommandDependencies is marked as `@spy_disabled`, so in spy-related generated code, the local variable appears to be never used, which leads to a compilation error.

gapis/api/gles/api/draw_commands.api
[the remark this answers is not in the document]
hevrard: As chatted offline, yes divisor is read everytime

gapis/api/gles/api/draw_commands.api
hevrard: Yes, as far as I understand. The names may be misleading here, Integer is a boolean that indicates whether vertex attrib array has unconverted integers.

gapis/api/gles/api/draw_commands.api
AWoloszyn: Does this work?
```
foo := as!DrawElementsIndirectCommand*(indirect)[0].count
_ = foo
``

hevrard: I works, thanks!

Issue #2859, opened 2019-06-21
Make the Report View "More Lazy"


pmuetschard:
For large traces, where replay can take a while, the report view will grab the replayer well before anything else gets a chance to replay anything. This means that by the time the thumbnails request a replay, the report is already running and everything has to wait.

We should make the report be "more lazy", possibly even only run the report if the user clicks a "run report" button, so that we don't unnecessarily hog the replayer to do the report.

PR #3010, opened 2019-08-02
Report only generates when clicking Generate Report

hliatis:
This will make report more "lazy" and not hog replay


hliatis:
Fixes #2859

Issue #2847, opened 2019-06-19
GLES context ID mismatch between UI and gapit


hevrard:
On a simple cube app trace, the GLES context ID (0 or 1) in the UI seems to be understood as the inverse of what `gapit commands -context <id>` does. E.g., selecting context 0 in the UI list the commands obtained by passing `-context 1` in gapit, and vice-versa.

hevrard:
I would like to keep this one as a first time contribution opportunity.

pmuetschard:
The UI shows the name of the context as provided by the server (i.e. the `name` field in the `service.Context` proto message), whereas gapit treats it as in index into the context list. The fact that the server calls them "GLES Context <n>" is just a coincidence. There are cases where we give the contexts different names though.

We should fix it so that
- the name matches the `context.Identifier` field
- gapit looks them up in the same way

We should probably add the identifier into the `service.Context` message.

PR #3014, opened 2019-08-06
Have GAPIT display commands by context handle

hliatis:
Instead of "commands -context i" request the i'th
context in the list, have it request the context
with handle/name i.

### DIFF
--- a/cmd/gapit/common.go
+++ b/cmd/gapit/common.go
@@ -52,11 +52,21 @@ func (f CommandFilterFlags) commandFilter(ctx context.Context, client service.Se
 		if err != nil {
 			return nil, log.Err(ctx, err, "Failed to load the contexts")
 		}
-		contexts := boxedContexts.(*service.Contexts)
-		if n := len(contexts.List); f.Context >= n {
-			return nil, log.Errf(ctx, err, "Context %d is out of range [0..%d]", f.Context, n-1)
+		contexts := boxedContexts.(*service.Contexts).List
+
+		for i, c := range contexts {
+			boxedContext, err := client.Get(ctx, p.Context(c.ID.ID()).Path(), nil)
+			if err != nil {
+				return nil, log.Errf(ctx, err, "Failed to load context at index %d", i)
+			}
+			context := boxedContext.(*service.Context)
+			if f.Context == int(context.Handle) {
+				filter.Context = c.ID
+				return filter, nil
+			}
 		}
-		filter.Context = contexts.List[f.Context].ID
+
+		return nil, log.Errf(ctx, err, "Could not find context %d", f.Context)
 	}
 	return filter, nil
 }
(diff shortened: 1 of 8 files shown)


hliatis:
Fixes #2847

cmd/gapit/common.go
```diff
@@ -52,11 +52,21 @@ func (f CommandFilterFlags) commandFilter(ctx context.Context, client service.Se
...
-			return nil, log.Errf(ctx, err, "Context %d is out of range [0..%d]", f.Context, n-1)
+		contexts := boxedContexts.(*service.Contexts).List
+
+		for i, c := range contexts {
+			boxedContext, err := client.Get(ctx, p.Context(c.ID.ID()).Path(), nil)
+			if err != nil {
+				return nil, log.Errf(ctx, err, "Failed to load context at index %d", i)
+			}
+			context := boxedContext.(*service.Context)
+			if f.Context == int(context.Handle) {
```
hevrard: Would it be possible to not introduce a `Handle`, and just compare f.Context with context.ID with some proper casting / stringification if need be?

gapis/service/service.proto
```diff
@@ -979,6 +979,8 @@ message Context {
   path.API API = 2;
   // The estimated importance for the context. 0 = lowest priority.
   uint32 priority = 3;
+  // The context ID. 0 or 1 for GLES. 1 for Vulkan
```
hevrard: This comment is not correct, a GLES application can have more than two contexts (e.g. google maps have contexts 0, 1, 2).

cmd/gapit/common.go
hevrard: Having a deeper look, this is where a plain gles context identifier gets translated to a ContexID:

https://github.com/google/gapid/blob/46511520bc56e6053b1ecdb3f46476361f476f6a/gapis/api/gles/context.go#L38

Maybe this logic should be abstracted to an exported function, such that in `commands.go` you can re-compute the ContextID from the f.Context integer using the same logic, to be able to compare them, without introducing a handle.

AWoloszyn: We should not do that last bit. `commands.go` should not have any knowledge about GLES or Vulkan to speak of. This one of the main advantages of the RPC infrastructure.

If we want to do something similar, we should export the name of the context through a `path`, then we can get them in a single place.

AWoloszyn: Actually reading this a bit more, scratch that last comment. We already have a name in the service.Context.
I think you may run into problems with context.ID. What happens in GLES if you create/destroy/create a context. Are you guaranteed to get a new unique handle? I expect the spec makes no such requirement. If that is the case we should strongly consider just ordering the contexts in the list on the server-side. Alternatively (or additionally) we could specify the context by string if we want that.

--context_name "Context 2" which would match by name.

hliatis: I don't know if ordering will work since you can have a 1 context without a 0. I have traces like this.

AWoloszyn: Alright, then I would prefer to leave as-is if we have to, but add the --context_name feature.

Baking GL information into gapit is not something we really want to do. It also means we cannot change things more flexibly in the future.

Issue #3012, opened 2019-08-05
Integration tests are broken


hevrard:
Running `bazel test tests --test_tag_filters=-needs_gpu` leads to:

```
/[...]/gapid/test/integration/replay/postback_test.go:68:18: not enough arguments in call to builder.New
	have (*device.MemoryLayout, nil)
	want (*device.MemoryLayout, *builder.Builder, id.ID)
```

hevrard:
@stellama0208 the recent change to add a replay status bar introduce a new deviceID argument to builder.New()

https://github.com/google/gapid/commit/4df3d56a144af151e7bdd75ad87d5807d182a13f#diff-5f113e8936e8f77f101964a65294d6a9R130

This seems to break the tests. These tests are run on a machine that does not require a GPU. It would be great to fix the tests without having to require a GPU -- I am not sure whether a deviceID does require a GPU underneath.

stellama0208:
Hi Hugues, just opened a new PR to address this problem. Changed the signature back and do what I wanted to do in another way. I have tested it on my local machine and `bazel test tests --test_tag_filters=-needs_gpu` will no longer get these arguments errors, it will give the same result as what is like before my PR (some test still fails in //core/java/jdbg:go_default_test and .../api/test/api_slices.go).

And btw may I ask what are the tests that can be run locally before code submission? Is there anything other than 
`./kokoro/presubmit/presubmit.sh` and 
`bazel test tests --test_tag_filters=-needs_gpu` ?
Thanks for pointing out this test for me, it is always better to run possible tests before code submission!

hevrard:
Thanks for the nice fix!

On your local machine, you can use `bazel test tests` to run all tests before you submit.

On Kokoro, we restrict to `bazel test tests --test_tag_filters=-needs_gpu` because we cannot rely on Kokoro VMs to have any kind of GPU. This is done in the "CI-linux-test" entry of Kokoro.

Unfortunately this is currently failing for various reasons. You just fixed one, thanks! I'm going to chase the root cause of other test failures.

In general, we want to keep this "CI-linux-test" Kokoro entry green before submitting.

PR #3015, opened 2019-08-06
Fix broken Integration test introduced by un-updated function call(#3012).

stellama0208:
Last PR #2962 fails some integration tests because the constructor in
builder.go changed with an additional argument, but the constructor call
in some test files didn't got updated together. Fix it in this commit by
reverting the constructor arguments back and communicate the wanted
argument through in another way.


hevrard:
Fix #3012

PR #3020, opened 2019-08-06
Fix some of the failing integration tests.

AWoloszyn:


gapir/cc/context.cpp
```diff
@@ -140,7 +140,7 @@ bool Context::interpret(bool cleanup) {
         // progress at each command call rather than each instruction. Also
         // extra check is given here to make sure to send notification when
         // approaching the last cmd.
-        if (finished_instrs % (total_instrs / 100) == 0 ||
+        if (total_instrs < 200 || (finished_instrs % (total_instrs / 100) == 0) ||
```
hevrard: Edit the above comment to explain this change?

AWoloszyn: I actually just updated to `total_instrs < 100` which means the commit remains valid. There was a bug when total_instrs < 100. 
`(finished_instrs % (99/100)) === finished_instrs % 0 == div0`

hevrard: OK, but that means `mSrv->sendReplayStatus()` is called for *every* of the CALL opcodes in the first 100 CALL instructions, is that what we want?

And more generally, `checkReplayStatusCallback()` is called only in the interpretation of a `CALL` opcode, and it sends a replay status only if that `CALL` instruction index is a multiple of 1% of `total_instrs`. Does this mean a pathological payload where **no** CALL opcode is a multiple of 1% of `total_instrs` would never call `sendReplayStatus()`?
@stellama0208

stellama0208: For the question of whether there can be pathological payload that never call sendReplayStatus(), I think it's quite impossible and we can quite rest assured on that. 
For my past experiment samples, mostly 50 frame android app traces: the payload I've seen usually contains more than 100000 instructions, the calling frequency will be pretty stable and the sending function will be called about 30 times for each replay: I guess it's because each call has 3 instructions on average. So even for a small payload with only 100 instructions, the possibility of never calling sendReplayStatus() is like (2/3)^33.

For the question of calling 100 times of sendReplayStatus() in a short period of time, I am also kind of worried here. I haven't taken very diverse traces so far, is it common for the payload to have `total_instrs < 100`? Will it be a burden for the server to send out 100 response very fast? If so we can extend the logic here a little bit.

hevrard: > For the question of whether there can be pathological payload that never call sendReplayStatus(), I think it's quite impossible and we can quite rest assured on that.
> For my past experiment samples, mostly 50 frame android app traces: the payload I've seen usually contains more than 100000 instructions, the calling frequency will be pretty stable and the sending function will be called about 30 times for each replay: I guess it's because each call has 3 instructions on average. So even for a small payload with only 100 instructions, the possibility of never calling sendReplayStatus() is like (2/3)^33.

Thanks for the clarification, that looks good to me then!

> 
> For the question of calling 100 times of sendReplayStatus() in a short period of time, I am also kind of worried here. I haven't taken very diverse traces so far, is it common for the payload to have `total_instrs < 100`? Will it be a burden for the server to send out 100 response very fast? If so we can extend the logic here a little bit.

I reckon most replay payloads will have more than 100 instructions. And with the previous clarification, sendReplayStatus will be triggered only on CALL opcodes, so it is very probably not going to send an actual 100 notifications.

So it's a green light for me!

Issue #2919, opened 2019-07-04
Handle extremely large queries in the perfetto Query UI.


AWoloszyn:
Sometimes I accidently don't LIMIT the number of results for a query, this means that we get back >10000 rows for our perfetto query. This causes the UI to hang (sometimes basically indefinitely) as it tries to add all of the rows.

It would be nice to either 
a) paginate by forcing `LIMIT 100` `OFFSET XXX` to our queries, so that we can show all of the data
b) If we detect > X rows coming back, offer to output as a CSV instead of trying to display. or like 
"data was truncated, save as CSV?"

PR #3031, opened 2019-08-07
Limit table to 100 items but with paging

hliatis:
The Perfetto Query table hangs for large queries.
This change should free up the UI thread.


hliatis:
Fixes #2919

PR #3032, opened 2019-08-07
[frame loop] Reset events/fences/semaphores during loop

RenfengLiu:


gapis/api/vulkan/frame_loop.go
```diff
@@ -184,8 +194,27 @@ func (f *frameLoop) Transform(ctx context.Context, id api.CmdID, cmd api.Cmd, ou
 		return
 	}
 
-	if _, ok := cmd.(*VkQueuePresentKHR); ok {
+	// Skip the destroy calls for resources that are not created during loop
+	switch cmd.(type) {
```
AWoloszyn: WE have to be very careful about these.
Vulkan is allow re-using handles, so vkDestroyBuffer, vkCreateBuffer may return the same handle. This means we may end up remapping them to the same value. I expect weird things to happen here. I would remove these DestroyBuffer fixes for now, or even better, *if* we destroy a buffer, and then create a new one with the same handle, throw an error. That way we can see how often that happens (if ever)

gapis/api/vulkan/frame_loop.go
```diff
@@ -428,18 +500,108 @@ func (f *frameLoop) resetImages(ctx context.Context, sb *stateBuilder) error {
...
+		}
+	}
+
+	return nil
+}
+
+func (f *frameLoop) resetFences(ctx context.Context, sb *stateBuilder) error {
+	s := GetState(sb.newState)
+
+	for k := range f.fenceChanged {
```
AWoloszyn: So this is not correct.
If the fence has been signaled (from the APIs perspective) we cannot simply reset it. We have to wait until the work that signaled it is actually done.

So we should vkWaitForFences() to wait until the fences have been actually signaled by the GPU. Remember the state mutation does not happen at the same time as the GPU, it only records the earliest it COULD be done.

gapis/api/vulkan/frame_loop.go
```diff
@@ -428,18 +500,108 @@ func (f *frameLoop) resetImages(ctx context.Context, sb *stateBuilder) error {
...
+			sb.write(sb.cb.VkQueueWaitIdle(queue.VulkanHandle(), VkResult_VK_SUCCESS))
+		}
+	}
+	return nil
+}
+
+func (f *frameLoop) resetEvents(ctx context.Context, sb *stateBuilder) error {
+	s := GetState(sb.newState)
+
+	for k := range f.eventChanged {
```
AWoloszyn: Similarly here, we need to actually make sure the event has actually been Set before continuing. Meaning we need a looped vkGetEventStatus(). I think we have a replayGetEventStatus() that does the right thing here.

gapis/api/vulkan/frame_loop.go
```diff
@@ -428,18 +500,108 @@ func (f *frameLoop) resetImages(ctx context.Context, sb *stateBuilder) error {
...
 		}
-		log.D(ctx, "Prime image from [%v] to [%v] succeed", src, dst)
 	}
+	return nil
+}
 
+func (f *frameLoop) resetSemaphores(ctx context.Context, sb *stateBuilder) error {
+	s := GetState(sb.newState)
+
+	for k := range f.semaphoreChanged {
```
AWoloszyn: Similar to the others, we have to wait for the semaphore if it has been "signaled" since the signal is not complete until someone actually waits.

PR #3033, opened 2019-08-08
Fix replay status bug with vulkan desktop MEC trace. (Fix #3006)

### LINKED ISSUE #3006: Got  Error communication with gapir: Handling notification when replay some trace.
17:21:34.939 E: <gapis> Error communication with gapir: Handling notification
   Cause: No active replay connection for this returned data

Initial investigation shows that it is caused by the `GrpcReplayService::sendReplayStatus` called in the `PrewarmReplay` stage.

stellama0208:
Solve by not allowing sending replay status notification at prewarm replay time.

### DIFF
--- a/cmd/gapir/cc/main.cpp
+++ b/cmd/gapir/cc/main.cpp
@@ -345,14 +345,14 @@ std::unique_ptr<Server> Setup(const char* uri, const char* authToken,
           return;
         }
 
-        auto cleanup_state = [&]() {
+        auto cleanup_state = [&](bool isPrewarm) {
           if (!prewarm->prewarm_context->initialize(prewarm->cleanup_id)) {
             return false;
           }
           if (cache != nullptr) {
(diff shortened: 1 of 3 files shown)


stellama0208:
This PR's solution is to not allow sending replay status notification at prewarm replay time.

From my current understanding it's the best trade-off to solve this bug. The bug is introduced by the initialization order of things in func (m *manager) execute(...). The order is 
1.prewarm replay 
2.initialize builder and generate notification handler 
3.initialize executor with notification handler and handle replay communications through executor afterwards. 
So if we want to keep the code structure of making executor the single point of handling replay communification, we can't send notifications generated in prewarm replay. 
And by the way, I think it's OK to hide the prewarm replay updates from the users. 

But still need confirmation about whether it's really OK not sending replay status updates in prewarm stage. Appreciate for your comments! :)

Issue #2739, opened 2019-05-07
Doom capture with GAPID: The gapis server has exited with an error code of: 2


xooiamd:
GAPID Version: 1.5.0:b33134fa99c398ce81dffd642aa11ea53c660518
OS: windows 10 amd64

Tried to capture a trace with Doom game in Desktop with the following setting:
![image](https://user-images.githubusercontent.com/19152049/57312325-49270500-70bb-11e9-884d-ba4132f32e1a.png)

Here is the error details:
panic: Process returned error
   Cause: exit status 53

goroutine 317 [running]:
github.com/google/gapid/gapis/trace/desktop.(*DesktopTracer).SetupTrace(0xc0000ef5f0, 0x62f4120, 0xc000140900, 0xc000246b60, 0x0, 0x0, 0x0, 0x0, 0xd)
	gapis/trace/desktop/trace.go:230 +0x5e2
github.com/google/gapid/gapis/trace.Trace(0x62f4120, 0xc000140900, 0xc00030fa10, 0xc0000e25a0, 0xc0000e2720, 0xc000246b60, 0xc0002088d8, 0x0, 0x0)
	gapis/trace/trace.go:55 +0x83f
github.com/google/gapid/gapis/server.(*traceHandler).Initialize.func1(0x62f4120, 0xc000140900, 0xc000246b60, 0xc0002088c0, 0xc0000e2720)
	gapis/server/server.go:717 +0x73
created by github.com/google/gapid/gapis/server.(*traceHandler).Initialize
	gapis/server/server.go:716 +0xa0

Thanks!

AWoloszyn:
This is likely because you are trying to trace a game meant to run underneath steam. The process start-up flow for Steam is not standard. (Executable closes, runs steam, steam then opens the executable etc).

You can double-check this with process explorer. If you just run `DOOMx64vk.exe` I expect in process explorer it will show up underneath `steam.exe`

hevrard:
Re-opening this: GAPID should display a proper warning to explain that it does not support Steam executables, because of the Steam-specific process start-up.

@xooiamd are you able to get a Doom executable which does not trigger a Steam start-up?

xooiamd:
I am not able to find a Doom executable that does not trigger Steam process.

AWoloszyn:
I will investigate an environment variable to help work around this issue.

xooiamd:
[...]

PR #3040, opened 2019-08-09
Fixes for multi-process applications.

AWoloszyn:
Some applications launch child processes, and those child
processes are the ones we are actually trying to trace.

This works as long as the parent process does not use a graphics
api. However i nthe case where the parent does this caused
us to attach to the parent.

This change allows an environment variable to be set
that will force gapii to effectively stifle itself if
it is attached to the wrong process.

GAPID_CAPTURE_PROCESS_NAME=my_process_name

gapii is still attached to the process, but does not try
and connect to gapis, nor does it write anything out.

### DIFF
--- a/core/cc/BUILD.bazel
+++ b/core/cc/BUILD.bazel
@@ -49,7 +49,10 @@ cc_library(
     linkopts = select({
         "//tools/build:linux": ["-ldl"],
         "//tools/build:darwin": [],
-        "//tools/build:windows": ["-lws2_32"],
+        "//tools/build:windows": [
+            "-lws2_32",
+            "-lshlwapi",
+        ],
         # Android.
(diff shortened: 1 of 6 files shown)


hevrard:
Fix #2739

AWoloszyn:
So, for desktop this works. Environment variables are (by default) carried through to child processes. However if an application specifically cleans out its env for its child, this won't work. But that is getting to be too much work for us to work around. If someone deliberately wants to break our tooling, the certainly can. 

On Android, I tried this briefly and it does not work. (other than the fact that you cannot set env vars) how we set up layers does not work with child processes. We should take some time to figure this out, since I think it is technically possible, but might take some concerted effort. Specifically with the new layer system recently introduced, if we can make that work with Vulkan layers, this approach may work.

Issue #3003, opened 2019-07-30
Package the correct version of java with Linux builds


irataxy:
Packages contain the right version of java with the windows and mac binaries, but not the Linux ones.

hevrard:
The release archives for Linux do not contain Java at all. On Linux, a pre-install Java 1.8 is required, as documented in: https://gapid.dev/requirements/

If you have any issue with your current Java version, this may be a workaround:
```
# install a regular JRE, e.g. on debian-like distros:
apt-get install openjdk-8-jre
# Use JAVA_HOME to tell GAPID to use this JRE
export JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64/
```

Does this fix your issue?

PR #3049, opened 2019-08-13
Add JRE in Linux zip releases

hevrard:
This can fix issues for Linux users which have exotic JRE in their path.


hevrard:
Fix #3003

hevrard:
It looks like openJDK does not provide readymade JRE archives for Java11, but this site does:
https://adoptopenjdk.net/releases.html?variant=openjdk11&jvmVariant=hotspot#x64_linux

My latest change gets the JRE from there.

Issue #1821, opened 2018-04-19
Cannot edit shaders that were created before the start of Mid-Execution capture.


AWoloszyn:

purvisa-at-google-com:
Fixed by #3054

PR #3054, opened 2019-08-14
Cannot edit shaders that were created before the start of MEC (#1821)

purvisa-at-google-com:
When performing a MEC, state reconstruction "linearisation"
creates new resources. This means there are more than zero
resources in existence when the first "real" commands are issued.
GAPIS's ResourcesResolvable does a fake replay of the driver's shadow
state to get the creation, read, modification and deletion points of a
capture's various resources. This is used to discover which resource
should be edited when the user modifies a shader in the UI. The bug was
caused by an error in the bookkeeping performed while computing these
access points for the various resources, when the "fake replay" began
with more than zero resources in existence. A mapping of resource ->
int, indexing a flat array of resources was being updated with the wrong
integer for a freshly created resource. This commit changes the map
creation to insert correct indexes for newly created resources.


hevrard:
Also, rebase on master to get the latest CI-linux-test fixes.

PR #3063, opened 2019-08-15
Load the perfetto config file each time a capture is done

purvisa-at-google-com:
Changed PerfettoConfig.java to load the config from file on demand.
This function used to only read the config once and return cached
results. Please be aware that calling this function is now more
expensive than it used to be, as it is no longer caching values.


AWoloszyn:
Can we add a comment in the code saying this call is expensive?

hevrard:
Superseded by #3080

PR #3066, opened 2019-08-15
Disables the new warning added by GCC 8.

yalcinmelihyasin:
This is a workaround for GCC 8 update.

Summary: There is a new warning added with GCC 8 and it will break the build when gcc updated to lastest version. This is a workaround by disabling the particular warning. I believe that we should not merge this at all and make a proper fix about it because the new warning is actually a meaningful one(in my opinion). This pr is for if someones machine randomly updates as mine, they can just copy this pr and continue to be able to build until the issue fixed for real.

### DIFF
--- a/tools/build/third_party/glslang.BUILD
+++ b/tools/build/third_party/glslang.BUILD
@@ -44,7 +44,13 @@ cc_library(
     copts = cc_copts() + [
         "-DNV_EXTENSIONS",
         "-Wno-unused-variable",
-    ],
+    ] + select({
+        "@gapid//tools/build:linux": [
+            "-Wno-error=class-memaccess",  # TODO(#3100): Remove this when glslang fixes the bug
+            "-Wno-maybe-uninitialized",
+        ],
+        "//conditions:default": [],
+    }),
     include_prefix = "third_party/glslang",
     linkopts = select({
         "@gapid//tools/build:windows": [],
--- a/tools/build/third_party/perfetto/perfetto.BUILD
+++ b/tools/build/third_party/perfetto/perfetto.BUILD
@@ -25,6 +25,7 @@ _COPTS_BASE = cc_copts() + [
     "-DNDEBUG",
 ] + select({
     "@gapid//tools/build:windows": ["-D__STDC_FORMAT_MACROS"],
+    "@gapid//tools/build:linux": ["-Wno-error=class-memaccess"],  # TODO(#3099): Remove this when perfetto fixes the bug
     "//conditions:default": [],
 })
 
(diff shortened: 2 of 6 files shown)


gapii/cc/BUILD.bazel
```diff
@@ -128,6 +128,7 @@ cc_library(
     copts = cc_copts() + select({
         "//tools/build:windows": ["-Wa,-mbig-obj"],
         "//conditions:default": [],
+        "//tools/build:linux": ["-Wno-error=class-memaccess"],
```
AWoloszyn: Are these errors in the header files? It seems like we may be able to do without having it in the base bazel. If that is true, then we should only enable it in the other projects (and we can submit it there)

yalcinmelihyasin:
I have updated this PR with the one that can be merged after a review.

gapii/cc/BUILD.bazel
```diff
@@ -128,6 +128,7 @@ cc_library(
     copts = cc_copts() + select({
         "//tools/build:windows": ["-Wa,-mbig-obj"],
         "//conditions:default": [],
+        "//tools/build:linux": [],
```
hevrard: Maybe this is not needed now?

tools/build/third_party/perfetto/perfetto.BUILD
```diff
@@ -25,6 +25,7 @@ _COPTS_BASE = cc_copts() + [
     "-DNDEBUG",
 ] + select({
     "@gapid//tools/build:windows": ["-D__STDC_FORMAT_MACROS"],
+    "@gapid//tools/build:linux": ["-Wno-error=class-memaccess"],
```
hevrard: Add a TODO comment referring to issue:
`TODO(#3099): remove Wno-error=class-memaccess once #3099 is fixed`

tools/build/third_party/glslang.BUILD
```diff
@@ -44,7 +44,9 @@ cc_library(
     copts = cc_copts() + [
         "-DNV_EXTENSIONS",
         "-Wno-unused-variable",
-    ],
+    ] + select({
+        "@gapid//tools/build:linux": ["-Wno-error=class-memaccess"],
```
hevrard: Can you:
1. create a GAPID issue to track that glslang should fix their warnings
2. Add a TODO referring to this issue

hevrard:
In kokoro/linux/build.sh , try to replace gcc7 by gcc8:
```
# Get GCC 7  #### TODO: replace by 8
sudo rm /etc/apt/sources.list.d/cuda.list*
sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
sudo apt-get -q update
sudo apt-get -qy install gcc-7 g++-7
export CC=/usr/bin/gcc-7
```

PR #3071, opened 2019-08-15
Sort the dependency graph.

AWoloszyn:
This sorts the dependencies in the graph just for the
puposes of test. This removes the flake from the test.

### DIFF
--- a/gapis/resolve/dependencygraph2/dependency_graph_test.go
+++ b/gapis/resolve/dependencygraph2/dependency_graph_test.go
@@ -16,6 +16,7 @@ package dependencygraph2
 
 import (
 	"context"
+	"sort"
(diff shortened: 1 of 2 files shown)


gapis/resolve/dependencygraph2/dependency_graph_test.go
```diff
@@ -140,6 +141,10 @@ func TestBuilder(t *testing.T) {
 	addNode(2)
 	eg.setDependencies(getNodeID(2), []NodeID{getNodeID(0), getNodeID(1)})
 
+	for i := range b.graphBuilder.GetGraph().dependenciesFrom {
```
hevrard: Add a comment on why we have to sort here.

PR #3074, opened 2019-08-15
Allow user to specify which frame to start

hliatis:


AWoloszyn:
So there are 2 issues that I have seen.
1) The saved value appears to be flipped. I.e. Every time I take a new trace the default value for this field is swapped
2) There is no way any longer to start at the very beginning of the trace. (I.e. if you select frame 1, you actually get the second frame, since GAPID treats frame as 0-indexed).

hliatis:
Do you want the spinner to allow for 0 or for the code to subtract one from the initial frame number?

AWoloszyn:
For consistency with our old code, I would suggest having the spinner start at 0 (and default to 0).  We have had no complaints about that approach in the past.

PR #3075, opened 2019-08-15
 Don't use the Google custom JDKs to run GAPID.

pmuetschard:


cmd/gapid/main.go
```diff
@@ -229,23 +247,36 @@ func checkVM(java string, checkVersion bool) bool {
...
 	if err != nil {
+		c.logIfVerbose("Not using " + java + ": failed to get version info")
+		return false
+	}
+
+	versionStr := string(version)
+
+	// Blacklist the Google custom JDKs.
+	if p := strings.Index(versionStr, googleInfix); p >= 0 {
+		c.logIfVerbose("Not using " + java + ": is a Google JDK")
```
hevrard: I think in this case we should clearly log how to work around this issue:
"Not using Google JDK: please install openjdk version >= 8 and set its path in JAVA_HOME".

Otherwise, the average Googler user experience is: "Gapid refuses to start. With --verbose-startup, it just says it refuses to use my Google JDK".

cmd/gapid/main.go
```diff
@@ -149,53 +157,62 @@ func newConfig() *config {
...
 		c.vm = java
 		return nil
 	}
 
 	if runtime.GOOS == "linux" {
-		if java := "/usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java"; checkVM(java, true) {
+		if java := "/usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java"; c.checkVM(java, true) {
 			c.vm = java
 			return nil
 		}
```
hevrard: Here, we know we're on Linux and no suitable JVM has been found, this is maybe the best place to help the user and unconditionally log a help message like:
```
} else {
    log.I(ctx, "No suitable JVM found, please install openjdk-8-jre")
}
```

cmd/gapid/main.go
pmuetschard: This is not necessary, since the .deb package causes the JDK to be installed already. This changes just causes us to ignore the environment telling us to use a different JDK than the default.

I've added a go link that gets printed in the log message now.

cmd/gapid/main.go
pmuetschard: Already doing that above. Again, the openjdk-8-jre should already be installed since the .deb package depends on it.

hevrard:
Thanks for the comments, yet I'm still concerned by the users installing with the Linux Zip? These ones may not have openjdk8 found on their machines. I think we really want to print some help message to users for whom cmd/gapid cannot detect a suitable JVM.

PR #3078, opened 2019-08-16
Create a database key to fix name collusion on loading captures

yalcinmelihyasin:
http://b/139294646

### DIFF
--- a/gapis/capture/capture.go
+++ b/gapis/capture/capture.go
@@ -129,12 +129,14 @@ func ResolvePerfettoFromPath(ctx context.Context, p *path.Capture) (*PerfettoCap
 }
 
 // Import imports the capture by name and data, and stores it in the database.
-func Import(ctx context.Context, name string, src Source) (*path.Capture, error) {
+func Import(ctx context.Context, name string, key string, src Source) (*path.Capture, error) {
 	dataID, err := database.Store(ctx, src)
 	if err != nil {
 		return nil, fmt.Errorf("Unable to store capture data source: %v", err)
 	}
+
 	id, err := database.Store(ctx, &Record{
+		Key:  key,
 		Name: name,
 		Data: dataID[:],
 	})
--- a/gapis/capture/capture.proto
+++ b/gapis/capture/capture.proto
@@ -32,10 +32,12 @@ message File {
 
 // Record holds all the data for an entire capture.
 message Record {
-  // Name of the capture.
-  string name = 1;
+  // Key for storing the capture.
+  string key = 1;
   // Database identifier of the data.
   bytes data = 2;
+  // Name of the capture.
+  string name = 3;
 }
 
 // Header holds information about the capture that is generated when the trace
--- a/gapis/server/server.go
+++ b/gapis/server/server.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"path/filepath"
 	"runtime"
 	"runtime/pprof"
 	"sync"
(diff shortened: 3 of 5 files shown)


gapis/capture/capture.proto
```diff
@@ -33,9 +33,11 @@ message File {
 // Record holds all the data for an entire capture.
 message Record {
   // Name of the capture.
-  string name = 1;
+  string key = 1;
```
yalcinmelihyasin: This is introduced because name is being used for other reasons then being a database key as well.

hevrard: The comment above is not matching the code anymore, can you edit the comment to:
`Key to index the capture`

gapis/server/server.go
```diff
@@ -233,10 +233,18 @@ func (s *server) LoadCapture(ctx context.Context, path string) (*path.Capture, e
...
 	}
-	name := filepath.Base(path)
 
+	fileInfo, err := os.Stat(path)
+	if err != nil {
+		return nil, err
+	}
+
+	name := fileInfo.Name()
+	key := fmt.Sprintf("%v%v%v", name, fileInfo.Size(), fileInfo.ModTime().Unix())
```
hevrard: Add a comment saying something like:
```
// The key contains more information than the file name in order to disambiguate capture that have a similar file name.
```

gapis/server/server.go
```diff
@@ -233,10 +233,19 @@ func (s *server) LoadCapture(ctx context.Context, path string) (*path.Capture, e
...
 
+	fileInfo, err := os.Stat(path)
+	if err != nil {
+		return nil, err
+	}
+
+	name := fileInfo.Name()
+	// Create a key to prevent name collusion between traces.
+	key := fmt.Sprintf("%v%v%v", name, fileInfo.Size(), fileInfo.ModTime().Unix())
+	key = strings.ReplaceAll(key, " ", "")
```
pmuetschard: what's wrong with spaces?

gapis/capture/capture.proto
```diff
@@ -32,10 +32,12 @@ message File {
...
 // Record holds all the data for an entire capture.
 message Record {
-  // Name of the capture.
-  string name = 1;
+  // Key for storing the capture.
+  string key = 1;
   // Database identifier of the data.
   bytes data = 2;
+  // Name of the capture.
+  string capture_name = 3;
```
pmuetschard: Why rename the field?

gapis/capture/capture.go
```diff
@@ -129,14 +129,16 @@ func ResolvePerfettoFromPath(ctx context.Context, p *path.Capture) (*PerfettoCap
 }
 
 // Import imports the capture by name and data, and stores it in the database.
-func Import(ctx context.Context, name string, src Source) (*path.Capture, error) {
+func Import(ctx context.Context, name string, key string, src Source) (*path.Capture, error) {
```
ben-clayton: Why don't we just use a hash of the capture's data?

yalcinmelihyasin: Hashing can take long time depending on the capture size which can be in GBs.

gapis/server/server.go
yalcinmelihyasin: Done.

gapis/capture/capture.proto
yalcinmelihyasin: reverted back.

gapis/capture/capture.go
yalcinmelihyasin: Merging this version and if we decide to use hash we can update it in another PR.

PR #3085, opened 2019-08-16
Remove some randomness from replays

pmuetschard:
In a few locations, iterate over maps in a deterministic order, rather than go's default random order, in order to ensure the order of commands is always the same between replays.


AWoloszyn:
We should do the same for Vulkan state rebuilding. But this is awesome!

PR #3095, opened 2019-08-19
Don't crash on draw calls without a bound program.

pmuetschard:


hevrard:
My bad, thanks for the fix!

Issue #3083, opened 2019-08-16
The new prefetching breaks the on-disk cache.


AWoloszyn:
The on-disk cache no longer prefetches everything. This seems to be related to the new prefetching scheme.

AWoloszyn:
I have a workaround so this does not need to be fixed over this weekend or anything, but should be handled asap next week.

purvisa-at-google-com:
Fixed

purvisa-at-google-com:
Fixed by #3097

PR #3097, opened 2019-08-20
Fix for #3083. OnDiskResourceCache now prefetches all data up front.

purvisa-at-google-com:
Andrew: Does this do what you need from the disk cache? You didn't give me repro steps for bug 3083 and I'm not sure the disk cache is even used on Android where I have been testing GAPID so far.

Let me know if this works or if there is something I'm missing and still need to patch up.


AWoloszyn:
LGTM seems to be working as expected.

Issue #3099, opened 2019-08-20
Perfetto raises Warnings with GCC 8


hevrard:
1. Update perfetto to latest, see if the warnings are still there
2. If so, get them to fix it

I tried to compile perfetto stand-alone in debug mode, it does not raise any issue.
https://android.googlesource.com/platform/external/perfetto/
commit 5817e8e9b2880d4fe03117def2f4ed5e2d95ab58

hevrard:
They seem to have fixed this issues by their commit e0d8eab
https://android.googlesource.com/platform/external/perfetto/+/e0d8eab313374bb164a309d3ffb27688ef0413bc

lpy:
This is fixed by #3101, please feel free to reopen if this is still an issue

hevrard:
https://github.com/google/gapid/blob/2b4d1ac50eb06dffaa17f28fbd1837c009e573a2/tools/build/third_party/perfetto/perfetto.BUILD#L28

This should probably be removed now, @yalcinmelihyasin can you have a look?

lpy:
I think @pmuetschard has removed it

Issue #3086, opened 2019-08-16
Query device GPU profiling capabilities when building device info


lpy:
P0: Verify we do return something
P1: Verify the returned values are correct

lpy:
Consider this is done.

PR #3101, opened 2019-08-20
Upgrade perfetto version.

lpy:
Previously we have made changes to perfetto to combine more protos into a
merged proto file which can be imported properly. This patch upgrades the
perfetto version up until that patch.

BUG: #3086
Test: bazel build pkg


lpy:
This is currently blocked by #3066, running `bazel build pkg` locally works

ptal

hevrard:
Fix #3099

lpy:
Pascal said the CI-linux-test is flaky, so I am going to merge this change to unblock myself.

PR #3102, opened 2019-08-21
Replay Status Enhancements

pmuetschard:
Notifies clients when a replay is first queued, when it is started (which is just before the building starts), while it is executing, and then when it has completed. This allows clients to better report the status of queued, under construction, and executing replays. Replays are also given a status identifiers, so clients can handle multiple concurrent replays correctly.

### DIFF
--- a/gapic/src/main/com/google/gapid/util/StatusWatcher.java
+++ b/gapic/src/main/com/google/gapid/util/StatusWatcher.java
@@ -17,7 +17,6 @@ package com.google.gapid.util;
 
 import com.google.common.collect.Maps;
 import com.google.gapid.proto.service.Service;
-import com.google.gapid.proto.service.path.Path;
 import com.google.gapid.server.Client;
 
 import java.util.HashMap;
@@ -33,6 +32,7 @@ public class StatusWatcher {
   private final Listener listener;
   private final Task root = Task.newRoot();
   private final HashMap<Long, Task> tasks = Maps.newHashMap();
+  private final ReplaySummary replays = new ReplaySummary();
   private String shownSummary = "";
 
   public StatusWatcher(Client client, Listener listener) {
@@ -97,14 +97,15 @@ public class StatusWatcher {
   }
 
   private void onReplayUpdate(Service.ReplayUpdate update) {
-    listener.onReplayProgress(update.getLabel(), update.getTotalInstrs(),
-        update.getFinishedInstrs(), update.getDevice());
+    if (replays.update(update)) {
+      listener.onReplayProgress(replays.getSummary());
+    }
   }
 
   public static interface Listener {
     public void onStatus(String status);
     public void onHeap(long heap);
-    public void onReplayProgress(long label, int totalInstrs, int finishedInstrs, Path.Device device);
+    public void onReplayProgress(String status);
   }
 
   private static class Task {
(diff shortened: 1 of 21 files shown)


gapic/src/main/com/google/gapid/util/StatusWatcher.java
```diff
@@ -204,4 +205,146 @@ public String getStatusLabel() {
...
+      replay.status = Service.ReplayStatus.REPLAY_EXECUTING;
+      replay.doneInstr = done;
+      replay.totalInstr = total;
+      return true;
+    }
+
+    private static boolean finished(Replay replay) {
+      // TODO: we should clean these out after some time.
+      replay.status = Service.ReplayStatus.REPLAY_FINISHED;
+      return true;
```
stellama0208: I think we should do       
```doneInstr -= replay.doneInstr; totalInstr -= replay.totalInstr;```
here, otherwise the doneInstr and totalInstr will be non-decreasing, and after a few replays the status progress will start from a large number like 95% (1+100000/100+100000)

File: gapic/src/main/com/google/gapid/util/StatusWatcher.java [hunk 2]
```diff
@@ -204,4 +205,146 @@ public String getStatusLabel() {
...
+        doneInstr += done - replay.doneInstr;
+        totalInstr += total - replay.totalInstr;
+      }
+      replay.status = Service.ReplayStatus.REPLAY_EXECUTING;
+      replay.doneInstr = done;
+      replay.totalInstr = total;
+      return true;
+    }
+
+    private static boolean finished(Replay replay) {
```
stellama0208: Just curious but is static intended here?

gapic/src/main/com/google/gapid/util/StatusWatcher.java
pmuetschard: Done. Thanks!

gapic/src/main/com/google/gapid/util/StatusWatcher.java
pmuetschard: It was static because it didn't use any instance state.

ben-clayton:
> Currently, when a replay is in progress, the scheduler is blocked and
doesn't batch up any of the incoming scheduled jobs (the pending channel
is not read and will eventually block). This change adds another
goroutine that will handle the executing of replays freeing up the
batching goroutine to continue batching even while a replay is in progress.
>
> This is required in order to accurately keep track of queued replays.

It's not clear to me why/how this helps to "accurately keep track of queued replays".

If I understand correctly, this is just trying to avoid blocking on the call to `s.pending`? While I'm not sure I why this is an issue (`Scheduler.Schedule` will block until the task is completed anyway), couldn't you get a similar behaviour by increasing the buffer size of the `pending` chan? Your implementation is unbounded, but I'm sure there's a sensible value you could use which would give you similar behaviour without resorting to condition variables.

ben-clayton:
Okay - so am I right in thinking the reason you're modifying scheduler is 'just' to:

* call `status.ReplayQueued` when a task is scheduled.
* call `status.Start` when a replay is kicked.
* call `status.Finish` when a replay is finished.

?

If so, couldn't all of these call be made externally to the scheduler?

I'm slightly concerned by the additional complexity to this class. While it _looks_ like the code is sound, this was already complex system to start with. Using condition variables in golang is exceptionally rare (the idiomatic types would say all you need is a `chan`).

ben-clayton:
Scheduler looks good to me. Thanks for being patient with my questions. :)

pmuetschard:
Thank you @ben-clayton for taking a look.

PR #3103, opened 2019-08-21
Help user to get a suitable Java on Linux

hevrard:
As discussed in b/139287646


pau-baiget:
The issue seems to be fixed, right? Is this necessary still?

Issue #3065, opened 2019-08-15
Setup a dogfooding option that can auto-update on a dogfood release 


hevrard:

natduca:
Please add a description of the engineering plan to this, or link a design doc.

hevrard:
See go/gapid-dogfood

hevrard:
After discussion, the plan is:
1. Reuse the existing update mechanism to check the special "nightly" release on GitHub
   a. Let user decide to follow-up nightly release (UI)
   b. use the commit SHA of the currently running app and the one of the nightly release to detect a new nightly release
2. Have Kokoro update the nightly release on GitHub

hevrard:
Step 1.b is  #3104

hevrard:
#3104 and the subsequent #3226 setup a proper "include pre-release" option in updates.

PR #3104, opened 2019-08-21
Enable to check pre-release updates

hevrard:
Nightly is only checked if pre-releases are checked, which is
currently a hardcoded "false" in the UI. Next step is to enable
checking pre-releases from the UI.

### DIFF
--- a/core/app/run.go
+++ b/core/app/run.go
@@ -97,6 +97,11 @@ func (v VersionSpec) GreaterThan(o VersionSpec) bool {
 	}
 }
 
+// Equal returns true if v is equal to o, ignoring the Build field.
+func (v VersionSpec) Equal(o VersionSpec) bool {
+	return (v.Major == o.Major) && (v.Minor == o.Minor) && (v.Point == o.Point)
+}
+
 func (v VersionSpec) String() string {
 	return fmt.Sprintf("%v", v)
 }
--- a/gapic/src/main/com/google/gapid/models/Settings.java
+++ b/gapic/src/main/com/google/gapid/models/Settings.java
@@ -86,6 +86,7 @@ public class Settings {
   public String[] recentFiles = new String[0];
   public String adb = "";
   public boolean autoCheckForUpdates = true;
+  public boolean includePrereleases = false;
   public boolean updateAvailable = false;
   public long lastCheckForUpdates = 0; // milliseconds since midnight, January 1, 1970 UTC.
   public String analyticsClientId = ""; // Empty means do not track.
@@ -250,6 +251,7 @@ public class Settings {
     recentFiles = getStringList(properties, "open.recent", recentFiles);
     adb = tryFindAdb(properties.getProperty("adb.path", ""));
     autoCheckForUpdates = getBoolean(properties, "updates.autoCheck", autoCheckForUpdates);
+    includePrereleases = getBoolean(properties, "updates.includePrereleases", includePrereleases);
     lastCheckForUpdates = getLong(properties, "updates.lastCheck", 0);
     updateAvailable = getBoolean(properties, "updates.available", updateAvailable);
     analyticsClientId = properties.getProperty("analytics.clientId", "");
@@ -306,6 +308,7 @@ public class Settings {
     setStringList(properties, "open.recent", recentFiles);
     properties.setProperty("adb.path", adb);
     properties.setProperty("updates.autoCheck", Boolean.toString(autoCheckForUpdates));
+    properties.setProperty("updates.includePrereleases", Boolean.toString(includePrereleases));
     properties.setProperty("updates.lastCheck", Long.toString(lastCheckForUpdates));
     properties.setProperty("updates.available", Boolean.toString(updateAvailable));
     properties.setProperty("analytics.clientId", analyticsClientId);
--- a/gapic/src/main/com/google/gapid/util/Messages.java
+++ b/gapic/src/main/com/google/gapid/util/Messages.java
@@ -63,7 +63,8 @@ public interface Messages {
       "Help improve GAPID by sending usage statistics to Google";
   public static final String CRASH_REPORTING_OPTION =
       "Help GAPID identify issues by sending crash reports to Google";
-  public static final String UPDATE_CHECK_OPTION = "Automatically check for GAPID updates";
+  public static final String UPDATE_CHECK_OPTION = "Automatically check for GAPID updates (please restart GAPID to force an update check)";
+  public static final String UPDATE_CHECK_PRERELEASE_OPTION = "Include developer preview releases";
   public static final String PRIVACY_POLICY =
       "Google's <a href=\"TOS\">APIs Terms of Service</a> and <a href=\"PP\">Privacy Policy</a>" +
       " govern your use of this application.";
--- a/gapic/src/main/com/google/gapid/views/SettingsDialog.java
+++ b/gapic/src/main/com/google/gapid/views/SettingsDialog.java
@@ -108,6 +108,7 @@ public class SettingsDialog extends DialogBase {
     private final Button allowAnalytics;
     private final Button allowCrashReports;
     private final Button allowUpdateChecks;
+    private final Button includePrerelease;
 
     public SettingsFormBase(Models models, Composite parent) {
       this(models, parent, 5, 5);
@@ -129,14 +130,17 @@ public class SettingsDialog extends DialogBase {
       beforeAnalytics();
 
       allowAnalytics = withLayoutData(
-          createCheckbox(this, Messages.ANALYTICS_OPTION, true),
+          createCheckbox(this, Messages.ANALYTICS_OPTION, models.settings.analyticsEnabled()),
           withSpans(new GridData(SWT.LEFT, SWT.TOP, false, false), 2, 1));
       allowCrashReports = withLayoutData(
-          createCheckbox(this, Messages.CRASH_REPORTING_OPTION, true),
+          createCheckbox(this, Messages.CRASH_REPORTING_OPTION, models.settings.reportCrashes),
           withSpans(new GridData(SWT.LEFT, SWT.TOP, false, false), 2, 1));
       allowUpdateChecks = withLayoutData(
-          createCheckbox(this, Messages.UPDATE_CHECK_OPTION, true),
+          createCheckbox(this, Messages.UPDATE_CHECK_OPTION, models.settings.autoCheckForUpdates),
           withSpans(new GridData(SWT.LEFT, SWT.TOP, false, false), 2, 1));
+      includePrerelease = withLayoutData(
+          createCheckbox(this, Messages.UPDATE_CHECK_PRERELEASE_OPTION, models.settings.includePrereleases),
+          withIndents(new GridData(SWT.LEFT, SWT.TOP, false, false), 20, 0));
       Label adbWarning = withLayoutData(createLabel(this, ""),
           withSpans(new GridData(SWT.FILL, SWT.FILL, true, false), 2, 1));
       adbWarning.setForeground(getDisplay().getSystemColor(SWT.COLOR_DARK_RED));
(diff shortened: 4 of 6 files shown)


AWoloszyn:
Until we actually have the builds coming out, we should probably hold on submitting this. It does look fine, but it will be dead code until the builds are coming out on a regular basis.

hevrard:
I revisited this based on feedback.

> * "nightly" has a specific meaning and makes a promise we are not going to keep

Now they are just "preview release"

> * tags should be static and not move

Fixed, with version tags now supporting an optional "preview" field which is expected to be the date in YYYYMMDD format
`v<major>.<minor>.<point>-<preview>`, with `-<preview>` being optional.
e.g. `v1.7.0.20190831`

> * once a release is public, I don't think it should ever change, even if it's marked as a pre-release

So now the tags will not move. We will just clean-up old pre-releases (assets etc) from GitHub once per quarter. Tags will stay around.

@pmuetschard I would need your advice on how best to trigger a UpdateWatcher.doCheck()  whenever the settings are modified?
Right now updates seem to be checked only every 8 hours, or at the very start if no `~/.gapic` file is found. It does not even wait for the WelcomeDialog to be completed before checking if there is an update. When a user goes to setting where he may change the update settings, I feel like we should trigger an update check. Yet I'm a bit confused on where to put that logic -- I don't have much experience in UI programming. Any advice much appreciated!

AWoloszyn:
I left a comment elsewhere, but v1.7.0-20190831 is the wrong format for the tags. 
Specifically because that is the tag BEFORE we release 1.7.0.  That is very misleading.

core/app/run.go
```diff
@@ -70,6 +70,8 @@ type VersionSpec struct {
 	Minor int
 	// Point version, not used if <0
 	Point int
+	// Dev version, not used if <0
+	Dev int
```
pmuetschard: Shouldn't this just be part of the build string?

gapic/src/main/com/google/gapid/views/SettingsDialog.java
```diff
@@ -156,6 +162,15 @@ protected void configureDialog(FileDialog dialog) {
       };
       adbPath.addBoxListener(SWT.Modify, adbListener);
       adbListener.handleEvent(null);
+
+      allowUpdateChecks.addSelectionListener(new SelectionAdapter()
+        {
```
pmuetschard: braces don't get their own lines in the Google Java Style guide.

core/app/run.go
hevrard: The point here is to be able to use this field in `func (v VersionSpec) GreaterThan(o VersionSpec) bool`

If we don't store this "dev" part of the version as a field, then the whole version comparison logic becomes convoluted as we would use `GreaterThan()` only for the regular part of the version, and then compare the "dev" version on the side.

Hence the addition of the "dev" version as a field.

gapic/src/main/com/google/gapid/views/SettingsDialog.java
hevrard: thanks, done

core/app/run.go
hevrard: I got rid adding the "Dev" field to the Version

core/app/run.go
```diff
@@ -97,6 +97,11 @@ func (v VersionSpec) GreaterThan(o VersionSpec) bool {
 	}
 }
 
+// Equal returns true if v is equal to o.
```
pmuetschard: Since this ignores the Build field, the comment should be updated to point that out.

gapis/server/server.go
```diff
@@ -129,13 +129,27 @@ func (s *server) CheckForUpdates(ctx context.Context, includePrereleases bool) (
...
+		// Dev pre-release: v<major>.<minor>.<point>-dev-<dev>
 		var version app.VersionSpec
-		fmt.Sscanf(release.GetTagName(), "v%d.%d.%d", &version.Major, &version.Minor, &version.Point)
-		if version.GreaterThan(mostRecentVersion) {
+		tagName := release.GetTagName()
+		devVersion := -1
+		numFields, err := fmt.Sscanf(tagName, "v%d.%d.%d-dev-%d", &version.Major, &version.Minor, &version.Point, &devVersion)
+		if err != nil && numFields < 3 {
+			// some non-release tags have other format, e.g. libinterceptor-v1.0
+			log.W(ctx, "Ignoring tag %s", tagName)
```
pmuetschard: This should not be a warning.

gapis/server/server.go
```diff
@@ -129,13 +129,27 @@ func (s *server) CheckForUpdates(ctx context.Context, includePrereleases bool) (
...
+			continue
+		}
+
+		if version.GreaterThan(mostRecentVersion) || (version.Equal(mostRecentVersion) && devVersion > mostRecentDevVersion) {
```
pmuetschard: Aren't dev releases going to be using the next version's major.minor? If that is the case, this is incorrect: If we currently are on M.N, and the dev release is M.(N+1), when we release M.(N+1), it will not be considered newer than the last M.(N+1) dev release.

gapic/src/main/com/google/gapid/models/Settings.java
```diff
@@ -88,6 +88,7 @@
   public String[] recentFiles = new String[0];
   public String adb = "";
   public boolean autoCheckForUpdates = true;
+  public boolean autoCheckForUpdatePrereleases = false;
```
pmuetschard: autoCheckForUpdatePrereleases -> onDevTrack
updates.autoCheckPrerelease -> updates.devTrack

gapic/src/main/com/google/gapid/util/Messages.java
```diff
@@ -60,6 +60,7 @@
   public static final String CRASH_REPORTING_OPTION =
       "Help GAPID identify issues by sending crash reports to Google";
   public static final String UPDATE_CHECK_OPTION = "Automatically check for GAPID updates";
+  public static final String UPDATE_CHECK_PRERELEASE_OPTION = "Automatically check for GAPID pre-release updates";
```
pmuetschard: "Include preview releases"

gapic/src/main/com/google/gapid/views/SettingsDialog.java
```diff
@@ -137,6 +141,9 @@ protected void configureDialog(FileDialog dialog) {
       allowUpdateChecks = withLayoutData(
           createCheckbox(this, Messages.UPDATE_CHECK_OPTION, true),
           withSpans(new GridData(SWT.LEFT, SWT.TOP, false, false), 2, 1));
+      allowUpdateCheckPrereleases = withLayoutData(
+          createCheckbox(this, Messages.UPDATE_CHECK_PRERELEASE_OPTION, false),
```
pmuetschard: Indent this checkbox, so it's clear that it's a dependent option.

gapic/src/main/com/google/gapid/views/SettingsDialog.java
```diff
@@ -108,6 +111,7 @@ private void update() {
     private final Button allowAnalytics;
     private final Button allowCrashReports;
     private final Button allowUpdateChecks;
+    private final Button allowUpdateCheckPrereleases;
```
pmuetschard: maybe find a shorter name for this?

gapic/src/main/com/google/gapid/views/SettingsDialog.java
```diff
@@ -156,6 +163,14 @@ protected void configureDialog(FileDialog dialog) {
       };
       adbPath.addBoxListener(SWT.Modify, adbListener);
       adbListener.handleEvent(null);
+
+      SelectionListener allowUpdateChecksListener = new SelectionAdapter() {
```
pmuetschard:
```
allowUpdateChecks.addListener(SWT.Selection,
    e -> allowUpdateCheckPrereleases.setEnabled(allowUpdateChecks.getSelection()));
```

Also consider clearing the checkbox if the auto-update checkbox is cleared.

core/app/run.go
hevrard: done

gapic/src/main/com/google/gapid/models/Settings.java
hevrard: I renamed to "includePrerelease" to keep the "prerelease" naming because:
- it reflects the GitHub naming
- it is already used in other parts of GAPIC, e.g. in function `checkForUpdates(boolean includePrereleases)`

gapis/server/server.go
hevrard: Changed to an Info

gapis/server/server.go
hevrard: Fixed

gapic/src/main/com/google/gapid/util/Messages.java
hevrard: done

gapic/src/main/com/google/gapid/views/SettingsDialog.java
hevrard: done

gapic/src/main/com/google/gapid/views/SettingsDialog.java
hevrard: also used "includePrereleases"

gapic/src/main/com/google/gapid/views/SettingsDialog.java
hevrard: done

hevrard:
Thanks @pmuetschard for the feedback, I addressed the comments.

Now I would like GAPIC to call `watchForUpdates()` when the update settings are modified, or maybe more easily whenever the settings dialog is closed. Yet, watchForUpdate needs a GAPIS Client, etc, which is available in the MainWindow, but not in Settings.

What would be the best way to trigger an update check if people change the Update settings?

core/app/run.go
```diff
@@ -97,6 +97,11 @@ func (v VersionSpec) GreaterThan(o VersionSpec) bool {
 	}
 }
 
+// Equal returns true if v is equal to o, ignoring the Build field
```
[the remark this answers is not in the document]
hevrard: done

gapic/src/main/com/google/gapid/views/SettingsDialog.java
```diff
@@ -34,6 +34,9 @@
 
 import org.eclipse.jface.dialogs.IDialogConstants;
 import org.eclipse.swt.SWT;
+import org.eclipse.swt.events.SelectionAdapter;
+import org.eclipse.swt.events.SelectionEvent;
+import org.eclipse.swt.events.SelectionListener;
```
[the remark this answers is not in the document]
hevrard: removed

gapic/src/main/com/google/gapid/views/SettingsDialog.java
```diff
@@ -156,6 +163,14 @@ protected void configureDialog(FileDialog dialog) {
       };
       adbPath.addBoxListener(SWT.Modify, adbListener);
       adbListener.handleEvent(null);
+
+      allowUpdateChecks.addListener(SWT.Selection,
+          e -> {
```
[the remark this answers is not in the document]
hevrard: fixed

gapis/server/server.go
```diff
@@ -129,13 +129,29 @@ func (s *server) CheckForUpdates(ctx context.Context, includePrereleases bool) (
...
+		numFields, err := fmt.Sscanf(tagName, "v%d.%d.%d-dev-%d", &version.Major, &version.Minor, &version.Point, &devVersion)
+		if err != nil && numFields < 3 {
+			// some non-release tags have other format, e.g. libinterceptor-v1.0
+			log.I(ctx, "Ignoring tag %s", tagName)
+			continue
+		}
+
+		// dev-releases are previews of the next release, so e.g. 1.2.3 is more recent than 1.2.3-dev-456
+		if version.GreaterThan(mostRecentVersion) ||
+			(version.Equal(mostRecentVersion) && (devVersion == -1 || devVersion > mostRecentDevVersion)) {
```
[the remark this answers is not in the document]
hevrard: Yes indeed, thanks! Fixed.

hevrard:
Fix #3065

PR #3110, opened 2019-08-22
Query GPU profiling capabilities from perfetto.

lpy:
Query all registered producers from perfetto, parse the output and construct
all GPU profiling capabilities.

BUG: #3086
Test: manually verified with log.E

### DIFF
--- a/core/os/android/adb/commands.go
+++ b/core/os/android/adb/commands.go
@@ -18,14 +18,18 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"encoding/base64"
 	"fmt"
 	"strings"
 	"time"
 
+	"github.com/golang/protobuf/proto"
 	"github.com/google/gapid/core/fault"
 	"github.com/google/gapid/core/log"
 	"github.com/google/gapid/core/os/android"
+	"github.com/google/gapid/core/os/device"
 	"github.com/google/gapid/core/os/shell"
+	perfetto_pb "perfetto/config"
 )
 
 const (
@@ -34,7 +38,8 @@ const (
 	ErrDeviceNotRooted = fault.Const("Device is not rooted")
 	ErrRootFailed      = fault.Const("Device failed to switch to root")
 
-	maxRootAttempts = 5
+	maxRootAttempts                            = 5
+	gpuRenderingStagesDataSourceDescriptorName = "gpu.renderstages"
 )
 
 func isRootSuccessful(line string) bool {
@@ -254,6 +259,51 @@ func (b *binding) GetEnv(ctx context.Context) (*shell.Env, error) {
 func (b *binding) SupportsPerfetto(ctx context.Context) bool {
 	os := b.Instance().GetConfiguration().GetOS()
 	return os.GetAPIVersion() >= 28
+}
+
+func (b *binding) QueryPerfettoServiceState(ctx context.Context) (*device.PerfettoCapability, error) {
+	if b.Instance().GetConfiguration().GetOS().GetAPIVersion() <= 29 {
+		return &device.PerfettoCapability{}, log.Errf(ctx, nil, "Querying perfetto capability requires Android API > 29")
+	}
+	res, err := b.Shell("perfetto", "--query-raw", "|", "base64").Call(ctx)
+	if err != nil {
+		return &device.PerfettoCapability{}, log.Errf(ctx, err, "adb shell perfetto returned error: %s", res)
+	}
+	decoded, _ := base64.StdEncoding.DecodeString(res)
+	tracingServiceState := &perfetto_pb.TracingServiceState{}
+	if err = proto.Unmarshal(decoded, tracingServiceState); err != nil {
+		return &device.PerfettoCapability{}, log.Errf(ctx, err, "Unmarshal returned error")
+	}
+	perfettoCapability := &device.PerfettoCapability{
+		GpuProfiling: &device.GPUProfiling{},
+	}
+	dataSources := tracingServiceState.GetDataSources()
+	for _, dataSource := range dataSources {
+		dataSourceDescriptor := dataSource.GetDsDescriptor()
+		if dataSourceDescriptor.GetName() == gpuRenderingStagesDataSourceDescriptorName {
+			perfettoCapability.GpuProfiling.HasRenderStage = true
+			continue
+		}
+		gpuCounterDescriptor := dataSourceDescriptor.GetGpuCounterDescriptor()
+		specs := gpuCounterDescriptor.GetSpecs()
+		if len(specs) == 0 || perfettoCapability.GpuProfiling.GpuCounterDescriptor != nil {
+			continue
+		}
+
+		// We mirror the Perfetto GpuCounterDescriptor proto into GAPID, hence
+		// they are binary format compatible.
+		data, err := proto.Marshal(gpuCounterDescriptor)
+		if err != nil {
+			continue
+		}
+		deviceGpuCounterDescriptor := &device.GpuCounterDescriptor{}
+		if err = proto.Unmarshal(data, deviceGpuCounterDescriptor); err != nil {
+			continue
+		}
+		perfettoCapability.GpuProfiling.GpuCounterDescriptor = deviceGpuCounterDescriptor
+	}
+
+	return perfettoCapability, nil
 }
 
 func extrasFlags(extras []android.ActionExtra) []string {
--- a/core/os/android/adb/device.go
+++ b/core/os/android/adb/device.go
@@ -145,6 +145,11 @@ func newDevice(ctx context.Context, serial string, status bind.Status) (*binding
 		d.To.Configuration.OS.Build = strings.TrimSpace(description)
 	}
 
+	// Query device Perfetto service state
+	if perfettoCapability, err := d.QueryPerfettoServiceState(ctx); err == nil {
+		d.To.Configuration.PerfettoCapability = perfettoCapability
+	}
+
 	// Check which abis the device says it supports
 	d.To.Configuration.ABIs = d.To.Configuration.ABIs[:0]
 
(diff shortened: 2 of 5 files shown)


lpy:
I am confused about how to convert from one protobuf to another protobuf which has exactly the same structure :/

We can't use gpu_counter proto directly since it's using LITE_RUNTIME, so I mirror it here

lpy:
clang format check failed 🤷‍♂️

AWoloszyn:
I would like to expand the proto somewhat.
We intend on exposing some of the possible counters via Vulkan layers, and we would need to describe which layers need to be enabled in order to get what counters, so that the server can do the right thing.

core/os/device/gpu_counter_descriptor.proto
```diff
@@ -0,0 +1,183 @@
+// Copyright (C) 2019 Google Inc.
```
AWoloszyn: Isn't it possible for a counter to have none of those units? Or combined units (In fact I expect MANY of our counters to have non-standard units).

#Vertices #Fragments MB/s ... etc?

lpy: maybe, unit here is optional
In fact the current returned unit from prototyped producer is completely wrong. If we need more we can create a request to perfetto and then sync from perfetto.

pmuetschard: We need to fix this in Perfetto now. I really, really don't want to handle all these units in the UI.

lpy: I am confused, which part are you referring to?

pmuetschard: This enum is copied from Perfetto, no? We should fix the Perfetto proto and drop all the non-nonsensical units and add units as Andrew mentioned (Vertices, Fragments, etc.)

lpy: aosp/1107601

lpy:
@pmuetschard As discussed on chat, I have moved the protobuf parsing into command.go and made sure we only query for R.

core/os/android/adb/commands.go
```diff
@@ -256,6 +295,47 @@ func (b *binding) SupportsPerfetto(ctx context.Context) bool {
 	return os.GetAPIVersion() >= 28
 }
 
+func (b *binding) QueryPerfettoServiceState(ctx context.Context) (*device.PerfettoCapability, error) {
+	if b.Instance().GetConfiguration().GetOS().GetAPIVersion() <= 29 {
+		return &device.PerfettoCapability{}, log.Errf(ctx, nil, "Querying perfetto capability requires Android API > 29\n")
```
pmuetschard: Drop the '\n'

core/os/android/adb/commands.go
```diff
@@ -256,6 +295,47 @@ func (b *binding) SupportsPerfetto(ctx context.Context) bool {
...
+	}
+	res, err := b.Shell("perfetto", "--query-raw", "|", "base64").Call(ctx)
+	if err != nil {
+		return &device.PerfettoCapability{}, log.Errf(ctx, nil, "adb shell perfetto returned error: \n%s", res)
+	}
+	decoded, _ := base64.StdEncoding.DecodeString(res)
+	tracingServiceState := &perfetto_pb.TracingServiceState{}
+	err = proto.Unmarshal(decoded, tracingServiceState)
+	if err != nil {
+		return &device.PerfettoCapability{}, log.Errf(ctx, nil, "Unmarshal returned error: \n")
```
pmuetschard: Drop the '\n' and pass the err along to `log.Errf`.

core/os/android/adb/commands.go
```diff
@@ -256,6 +295,47 @@ func (b *binding) SupportsPerfetto(ctx context.Context) bool {
 	return os.GetAPIVersion() >= 28
 }
 
+func (b *binding) QueryPerfettoServiceState(ctx context.Context) (*device.PerfettoCapability, error) {
+	if b.Instance().GetConfiguration().GetOS().GetAPIVersion() <= 29 {
+		return &device.PerfettoCapability{}, log.Errf(ctx, nil, "Querying perfetto capability requires Android API > 29\n")
+	}
+	res, err := b.Shell("perfetto", "--query-raw", "|", "base64").Call(ctx)
+	if err != nil {
+		return &device.PerfettoCapability{}, log.Errf(ctx, nil, "adb shell perfetto returned error: \n%s", res)
```
pmuetschard: Drop the '\n' and pass the err along to `log.Errf`.

core/os/android/adb/commands.go
pmuetschard: You can combine this into
```
if err := proto.Unmarshal(...); err != nil {
   ...
}
```

core/os/android/adb/device.go
```diff
@@ -145,6 +145,11 @@ func newDevice(ctx context.Context, serial string, status bind.Status) (*binding
 		d.To.Configuration.OS.Build = strings.TrimSpace(description)
 	}
 
+	// Query device Perfetto service state
+	if perfettoCapability, err := d.QueryPerfettoServiceState(ctx); err == nil {
```
pmuetschard: This drops the error. We should at least log it.

core/os/android/adb/commands.go
```diff
@@ -256,6 +295,47 @@ func (b *binding) SupportsPerfetto(ctx context.Context) bool {
...
+	}
+	dataSources := tracingServiceState.GetDataSources()
+	for _, dataSource := range dataSources {
+		dataSourceDescriptor := dataSource.GetDsDescriptor()
+		if dataSourceDescriptor.GetName() == "gpu.renderingstages" {
+			perfettoCapability.GpuProfiling.HasRenderStage = true
+			continue
+		}
+		gpuCounterDescriptor := dataSourceDescriptor.GetGpuCounterDescriptor()
+		if gpuCounterDescriptor == nil {
```
pmuetschard: Checking for nil with protos goes against established patterns and is not necessary. All the `Get*` functions hide the "nilness". You can simply remove this check and `gpuCounterDescriptor.GetSpecs()` will return an empty list if `gpuCounterDescriptor` is nil.

core/os/device/gpu_counter_descriptor.proto
```diff
@@ -0,0 +1,95 @@
...
+syntax = "proto3";
+
+package device;
+option java_package = "com.google.gapid.proto.device";
+option java_outer_classname = "GpuProfiling";
+option go_package = "github.com/google/gapid/core/os/device";
+// cc_package
+// js_path = "proto";
+
+// This file is a mirror of the gpu_counter_descriptor.proto from
```
pmuetschard: If it is a mirror, can you just deserialize it into the new proto rather than the old and then translate? If they are the same than they should be binary compatible.

core/os/android/adb/device.go
lpy: Wait, but the errors above are all dropped?

pmuetschard: Alright. OK.

lpy:
PTAL, I have addressed all comments

core/os/android/adb/commands.go
```diff
@@ -34,7 +38,8 @@ const (
 	ErrDeviceNotRooted = fault.Const("Device is not rooted")
 	ErrRootFailed      = fault.Const("Device failed to switch to root")
 
-	maxRootAttempts = 5
+	maxRootAttempts                            = 5
+	GPURENDERINGSTAGESDATASOURCEDESCRIPTORNAME = "gpu.renderstages"
```
pmuetschard: Use camel-case and I think it should start with a lowercase letter (means that it's private).

core/os/android/adb/commands.go
```diff
@@ -256,6 +261,51 @@ func (b *binding) SupportsPerfetto(ctx context.Context) bool {
...
+	}
+	decoded, _ := base64.StdEncoding.DecodeString(res)
+	tracingServiceState := &perfetto_pb.TracingServiceState{}
+	if err = proto.Unmarshal(decoded, tracingServiceState); err != nil {
+		return &device.PerfettoCapability{}, log.Errf(ctx, err, "Unmarshal returned error")
+	}
+	perfettoCapability := &device.PerfettoCapability{
+		GpuProfiling: &device.GPUProfiling{
+			HasRenderStage:       false,
+			GpuCounterDescriptor: nil,
```
pmuetschard: Remote these two lines, they are defaults.

core/os/android/adb/commands.go
```diff
@@ -256,6 +261,51 @@ func (b *binding) SupportsPerfetto(ctx context.Context) bool {
...
+	dataSources := tracingServiceState.GetDataSources()
+	for _, dataSource := range dataSources {
+		dataSourceDescriptor := dataSource.GetDsDescriptor()
+		if dataSourceDescriptor.GetName() == GPURENDERINGSTAGESDATASOURCEDESCRIPTORNAME {
+			perfettoCapability.GpuProfiling.HasRenderStage = true
+			continue
+		}
+		gpuCounterDescriptor := dataSourceDescriptor.GetGpuCounterDescriptor()
+		specs := gpuCounterDescriptor.GetSpecs()
+		if len(specs) == 0 || perfettoCapability.GpuProfiling.GpuCounterDescriptor != nil {
```
pmuetschard: Instead of ignoring the counter specs if we've already seen some, shouldn't we just append the new ones?

core/os/android/adb/commands.go
```diff
@@ -256,6 +261,51 @@ func (b *binding) SupportsPerfetto(ctx context.Context) bool {
...
+			continue
+		}
+		data, err := proto.Marshal(gpuCounterDescriptor)
```
pmuetschard: Please add a comment here that we're going via the binary form, because we expect them to be the same.

core/os/android/adb/commands.go
lpy: No we shouldn't. There should only be one counter producers

lpy:
Done, ptal!

PR #3118, opened 2019-08-24
Vulkan Struct UI.

stellama0208:


stellama0208:
Last commit is to solve a merge conflict with the current master.

pmuetschard:
Since you need to rebase in order to take care of the conflicts, I would suggest squashing before rebasing.

PR #3128, opened 2019-08-28
 Add the capability to tracks to dispaly a tooltip.

pmuetschard:
The tooltip is shown when the track title is hovered.


AWoloszyn:
s/dispaly/display

PR #3149, opened 2019-09-06
Updating CSS for title bar, there was a clickability issue

carlthomasiv:
While playing with the updated site I realized the top header (has the GAPID logo) is clickable anywhere. This fixes that oddity and makes it so the GAPID logo is what is clickable, not the entire header.


hevrard:
One can argue that being able to click anywhere on that banner is actually a *good feature*, as one doesn't have to target the pointer exactly within the logo/title

I'll let you make the call on what is best.

carlthomasiv:
Ya standard websites dont do the whole banner. They just do the logo as clickable. Plus it leaves it open in the future to allow other things to possible go into the header in the future.

PR #3154, opened 2019-09-09
Allow to query perfetto capability on Q build.

lpy:
A requirement to make sure perfetto work properly for GPU profiling is to
cherry-pick certain patches to Q build such that the capabilities of perfetto
can be queried, hence loosing the requirement here. If a Q build device doesn't
support this functionality, that's fine as well becasue it will fail naturally
when attempting to decode the returned string.

BUG: #3086
Test: Tried on Q and R devices.


lpy:
ptal

PR #3156, opened 2019-09-10
Expose Perfetto capability from gapis.

lpy:
To build up proper selection UI, we need to expose Perfetto capability from
gapis. Currently we only advertize render stages and GPU counters capability.

BUG: #3086
Test: manually verified with counter producer


lpy:
ptal

AWoloszyn:
Isn't this already part of the Device proto that the client can already ask for?

lpy:
ah I see :D

PR #3157, opened 2019-09-11
Check if build is userdebug before attempting adb root

hevrard:
Fix b/140612762


core/os/android/adb/commands.go
```diff
@@ -90,6 +91,15 @@ retry:
 	return log.Err(ctx, ErrRootFailed, buf.String())
 }
 
+// UserdebugBuild returns true if the device runs an Android userdebug build
+func (b *binding) UserdebugBuild(ctx context.Context) (bool, error) {
+	output, err := b.Command("shell", "getprop", "ro.build.type").Call(ctx)
+	if err != nil {
+		return false, err
+	}
+	return (output == "userdebug"), nil
```
pau-baiget: Is it always going to be "userdebug"? Isn't "userdebug" a substring of the actual build type?

gapidapk/packageinfo.go
```diff
@@ -46,7 +46,15 @@ func PackageList(ctx context.Context, d adb.Device, includeIcons bool, iconDensi
 		return nil, log.Err(ctx, nil, "Service intent was not found")
 	}
 
-	onlyDebug := d.Root(ctx) == adb.ErrDeviceNotRooted
+	// Can only capture debug application unless Android build is userdebug and adb root works
+	onlyDebug := true
+	userdebugBuild, err := d.UserdebugBuild(ctx)
+	if err != nil {
+		return nil, err
```
pau-baiget: Shouldn't this be 
```return nil, log.Err(ctx, err, "Starting service")```
as in code blocks below?

File: core/os/android/adb/commands.go (hunk 1 above)
Comment (hevrard) [reply core/os/android/adb/commands.go]: The complete build string is in property `ro.build.fingerprint`, e.g. 
"google/walleye/walleye:ABC123:userdebug/dev-keys"

The `ro.build.type` is only "userdebug" or "user" (or "eng", etc)

So we should be good to go by checking `ro.build.type`

File: gapidapk/packageinfo.go (hunk 2 above)
Comment (hevrard) [reply gapidapk/packageinfo.go]: Yes indeed! fixed.

File: core/os/android/adb/commands.go [hunk 3]
```diff
@@ -90,6 +91,15 @@ retry:
 	return log.Err(ctx, ErrRootFailed, buf.String())
 }
 
+// UserdebugBuild returns true if the device runs an Android userdebug build
+func (b *binding) UserdebugBuild(ctx context.Context) (bool, error) {
```
lpy: IsUserdebugBuild

gapidapk/packageinfo.go
```diff
@@ -46,7 +46,15 @@ func PackageList(ctx context.Context, d adb.Device, includeIcons bool, iconDensi
 		return nil, log.Err(ctx, nil, "Service intent was not found")
 	}
 
-	onlyDebug := d.Root(ctx) == adb.ErrDeviceNotRooted
+	// Can only capture debug application unless Android build is userdebug and adb root works
+	onlyDebug := true
+	userdebugBuild, err := d.UserdebugBuild(ctx)
```
lpy: isUserdebugBuild

core/os/android/adb/commands.go
hevrard: done

gapidapk/packageinfo.go
hevrard: done

PR #3159, opened 2019-09-11
Added an option to gapit to disable the coherent memory tracker.

AWoloszyn:
This lets us specify at runtime to disable the coherent memory
tracker. This makes debugging gapii simpler, because the
coherent memory tracker interferes with gdb a lot.

It is an optimization, so turning it off is not a problem (functionally)


cmd/gapit/trace.go
```diff
@@ -227,25 +227,26 @@ func (verb *traceVerb) Run(ctx context.Context, flags flag.FlagSet) error {
 	}
 
 	options := &service.TraceOptions{
-		Type:                      api.traceType,
```
hevrard: Many lines flagged as different by github, but I can't see the diff? maybe whitespaces?

AWoloszyn: It is whitespace, gofmt aligns everything and apparently github doesn't highlight whitespace changes.

PR #3164, opened 2019-09-12
Fix the MEC crash related to unopened forward dependency

yalcinmelihyasin:
During MEC we capture commands that are dependent on other commands(e.g VkUnmapMemory)
Normally they are discarded during DCE. But this may cause crash on the next command that
operates on the same resource. As the original command run during state reconstruction
the corresponding dependent command should also be kept alive and run


yalcinmelihyasin:
b/140542424

PR #3167, opened 2019-09-16
Clean up frame looping code and add stubs for DescriptorSetLayouts

purvisa-at-google-com:
THIS IS NOT A PROPOSAL FOR SUBMISSION

Putting this review up so that you guys can see what I'm proposing in terms of renaming variables, using more whitespace, removing aliased/shadowed pointers, comments and other (mostly cosmetic) readability improvements to the code going forward. This is for discussion, feel free to support or resist any of the changes you see here.

I have also added some stub work for DescriptorSetLayouts. @AWoloszyn  if you could take a look at this and tell me if I'm pointed in the right direction, this would be great. Right now I'm trying to find a way I can find where I can get the arguments needed for the commandBuilder in resetDescriptorSetLayouts().

### DIFF
--- a/gapis/api/vulkan/frame_loop.go
+++ b/gapis/api/vulkan/frame_loop.go
@@ -33,53 +33,39 @@ type stateWatcher struct {
 	memoryWrites map[memory.PoolID]*interval.U64SpanList
 }
 
-func (b *stateWatcher) OnBeginCmd(ctx context.Context, cmdID api.CmdID, cmd api.Cmd) {
-}
-
-func (b *stateWatcher) OnEndCmd(ctx context.Context, cmdID api.CmdID, cmd api.Cmd) {
-}
+func (b *stateWatcher) OnBeginCmd(ctx context.Context, cmdID api.CmdID, cmd api.Cmd) {}
+func (b *stateWatcher) OnEndCmd(ctx context.Context, cmdID api.CmdID, cmd api.Cmd)   {}
 func (b *stateWatcher) OnBeginSubCmd(ctx context.Context, subIdx api.SubCmdIdx, recordIdx api.RecordIdx) {
 }
-func (b *stateWatcher) OnRecordSubCmd(ctx context.Context, recordIdx api.RecordIdx) {
-}
-func (b *stateWatcher) OnEndSubCmd(ctx context.Context) {
-}
+func (b *stateWatcher) OnRecordSubCmd(ctx context.Context, recordIdx api.RecordIdx) {}
+func (b *stateWatcher) OnEndSubCmd(ctx context.Context)                             {}
 func (b *stateWatcher) OnReadFrag(ctx context.Context, owner api.RefObject, frag api.Fragment, valueRef api.RefObject, track bool) {
 }
-
 func (b *stateWatcher) OnWriteFrag(ctx context.Context, owner api.RefObject, frag api.Fragment, oldValueRef api.RefObject, newValueRef api.RefObject, track bool) {
-
 }
 
 func (b *stateWatcher) OnWriteSlice(ctx context.Context, slice memory.Slice) {
+
 	span := interval.U64Span{
 		Start: slice.Base(),
 		End:   slice.Base() + slice.Size(),
 	}
+
 	poolID := slice.Pool()
+
 	if _, ok := b.memoryWrites[poolID]; !ok {
 		b.memoryWrites[poolID] = &interval.U64SpanList{}
 	}
+
 	interval.Merge(b.memoryWrites[poolID], span, true)
 }
 
-func (b *stateWatcher) OnReadSlice(ctx context.Context, slice memory.Slice) {
-}
-
-func (b *stateWatcher) OnWriteObs(ctx context.Context, obs []api.CmdObservation) {
-}
-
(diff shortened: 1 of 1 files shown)


gapis/api/vulkan/frame_loop.go
```diff
@@ -33,62 +33,48 @@ type stateWatcher struct {
...
 
 // Transfrom
 type frameLoop struct {
 	capture        *capture.GraphicsCapture
 	cmds           []api.Cmd
 	numInitialCmds int
 	loopCount      int32
-	loopStartCmd   api.Cmd
-	loopEndCmd     api.Cmd
+	loopStartIdx   int
```
AWoloszyn: I think we want to avoid using indices here. Other transforms may insert commands so indices can get out of scope.

gapis/api/vulkan/frame_loop.go
```diff
@@ -102,17 +88,28 @@ type frameLoop struct {
 	imageDestroyed map[VkImage]bool
 	imageToBackup  map[VkImage]VkImage
 
+	descriptorSetLayoutCreated map[VkDescriptorSetLayout]bool
```
AWoloszyn: descriptorSetLayoutsCreated

gapis/api/vulkan/frame_loop.go
```diff
@@ -297,175 +378,244 @@ func (f *frameLoop) detectChangedResource(ctx context.Context, startState *api.G
...
 
 		primeable, err := imgPrimer.newPrimeableImageDataFromDevice(src, dst)
 		if err != nil {
 			return log.Errf(ctx, err, "Create primeable image data for image %v", dst)
 		}
-		defer primeable.free(sb)
-		err = primeable.prime(sb, useSpecifiedLayout(dstObj.Info().InitialLayout()), sameLayoutsOfImage(dstObj))
+		defer primeable.free(stateBuilder)
+
+		err = primeable.prime(stateBuilder, useSpecifiedLayout(dstObj.Info().InitialLayout()), sameLayoutsOfImage(dstObj))
```
AWoloszyn: More typical go would be 

if err = primable.prime(....); err != nil {
}

gapis/api/vulkan/frame_loop.go
```diff
@@ -126,129 +123,198 @@ func newFrameLoop(ctx context.Context, c *capture.GraphicsCapture, numInitialCmd
...
-	} else if cmd == f.loopEndCmd {
-		log.D(ctx, "Loop: last frame is %v id %v, cmd is %v.", f.frameNum, id, cmd)
-		st := GetState(out.State())
-		sb := st.newStateBuilder(ctx, newTransformerOutput(out))
-		defer sb.ta.Dispose()
-		out.MutateAndWrite(ctx, id, cmd)
+	} else if cmd == f.cmds[f.loopEndIdx] {
+
+		log.D(ctx, "Loop: last frame is %v cmdId %v, cmd is %v.", f.frameNum, cmdId, cmd)
+
```
pau-baiget: What's the value in having so empty blank lines in this body?

gapis/api/vulkan/frame_loop.go
```diff
@@ -126,129 +123,198 @@ func newFrameLoop(ctx context.Context, c *capture.GraphicsCapture, numInitialCmd
...
+			// vkCmd := cmd.(*VkDestroyDescriptorPool)
+
+		// DescriptorSet(s)
+		case *VkAllocateDescriptorSets:
+			// vkCmd := cmd.(*VkAllocateDescriptorSets)
+
+		case *VkFreeDescriptorSets:
+			// vkCmd := cmd.(*VkFreeDescriptorSets)
+
+		case *VkUpdateDescriptorSets:
```
AWoloszyn: Not sure you should be doing anything on UpdateDescriptorSets. as discussed, we should diff all existing descriptor sets at the end, and then just update the ones that are different.

gapis/api/vulkan/frame_loop.go
```diff
@@ -258,35 +324,50 @@ func (f *frameLoop) detectChangedResource(ctx context.Context, startState *api.G
 		log.E(ctx, "Mutate error: [%v].", err)
 	}
 
-	st := GetState(f.backupState)
+	apiState := GetState(f.backupState)
+
 	// Find out changed buffers.
-	vkBuffers := st.Buffers().All()
-	for k, buffer := range vkBuffers {
+	for bufferKey, buffer := range apiState.Buffers().All() {
```
AWoloszyn: I think this will be unfortunately non-deterministic. (I think the previous was also non-determinstic) this will create a pain for validating when you change things.

gapis/api/vulkan/frame_loop.go
purvisa-at-google-com: I've tried to use white space to group things functionally across the whole file.  The code in this block mostly kicks off a series of high level operations via single functions and thus many of the groups are a single line. Though you will still see some grouping like lines 172/173, and 180-183. There is no semantic meaning to this, but I find white space makes code look a lot less intimidating and easier on the eyes. Also when a block boils down to kicking off a series of high level operations I find it draws the eyes to the fact quite dramatically.

PR #3168, opened 2019-09-16
Fixes to the default device selection in the trace dialog.

pmuetschard:
The priorities are now:
1) The selected device from the most recent trace
2) If there's only one device, pick it
3) If there's only two devices, where one is Android and the other is not, select the Android device.

### DIFF
--- a/gapic/src/main/com/google/gapid/views/TracerDialog.java
+++ b/gapic/src/main/com/google/gapid/views/TracerDialog.java
@@ -597,18 +597,34 @@ public class TracerDialog {
         if (device != null && devices != null) {
           deviceLoader.stopLoading();
           device.setInput(devices);
-          if (!settings.traceDevice.isEmpty()) {
-            Optional<DeviceCaptureInfo> deflt = devices.stream()
-                .filter(dev -> settings.traceDevice.equals(dev.device.getSerial()))
-                .findAny();
-            if (deflt.isPresent()) {
-              device.setSelection(new StructuredSelection(deflt.get()));
+          DeviceCaptureInfo deflt = getPreviouslySelectedDevice(settings).orElseGet(() -> {
+            if (devices.size() == 1) {
+              return devices.get(0);
+            } else if (devices.size() == 2) {
+              // If there are exactly two devices and exactly one of them is an Android device,
+              // select the Android device. It is a fair assumption that a developer that has an
+              // Android device connected wants to trace Android, not desktop.
+              boolean firstIsAndroid = devices.get(0).isAndroid();
+              boolean secondIsAndroid = devices.get(1).isAndroid();
+              if (firstIsAndroid != secondIsAndroid) {
+                return firstIsAndroid ? devices.get(0) : devices.get(1);
+              }
             }
+            return null;
+          });
+          if (deflt != null) {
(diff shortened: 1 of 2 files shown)


gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -597,20 +597,36 @@ private void updateDevicesDropDown(Settings settings) {
...
           device.setInput(devices);
-          if (!settings.traceDevice.isEmpty()) {
-            Optional<DeviceCaptureInfo> deflt = devices.stream()
-                .filter(dev -> settings.traceDevice.equals(dev.device.getSerial()))
-                .findAny();
-            if (deflt.isPresent()) {
-              device.setSelection(new StructuredSelection(deflt.get()));
+          DeviceCaptureInfo deflt = getPreviouslySelectedDevice(settings).orElseGet(() -> {
+            if (devices.size() == 1) {
+              devices.get(0);
```
hevrard: I guess you meant: `return devices.get(0)`

gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -597,20 +597,36 @@ private void updateDevicesDropDown(Settings settings) {
...
+            } else if (devices.size() == 2) {
```
hevrard: This approach will not set any device by default if e.g. 3 devices are listed.

What about:
- returning the first android device found, if any
- otherwise defaulting to the first device listed

e.g.:

```
for (DeviceCaptureInfo dev: devices) {
  if (dev.isAndroid()) {
    return dev
  }
}
if (devices.size() > 0) {
  // default to first listed device
  return devices.get(0);
}
return null;
```

AWoloszyn:
There is one maybe slightly weird edge-case.
1) I have an android device attached that "at some point" had GAPID on it.
2) My current build of GAPID does not have android support. (ie.. there are no .apks around in the package)
3) We will end up showing the android devices. I think if we could filter out devices if we dont have available apks for any supported ABI, that would fix this problem.

gapic/src/main/com/google/gapid/views/TracerDialog.java
pmuetschard: Done.

gapic/src/main/com/google/gapid/views/TracerDialog.java
pmuetschard: No, If there are more than one Android device, selecting the wrong one is no better than selecting nothing. The reality is that 1) more than one device is uncommon, and 2) selecting the previous selected device will happen 99% of the time.

Also, device come in arbitrary order, so selecting "the first one" is actually meaningless.

pmuetschard:
> There is one maybe slightly weird edge-case.

@AWoloszyn that sounds like something to be fixed in gapis, not the UI.

gapic/src/main/com/google/gapid/views/TracerDialog.java
hevrard: Personally, when no previously selected devices are present, I would prefer the UI to pick up any device rather than showing a blank field. But I'm OK with your current implem also.

PR #3169, opened 2019-09-16
Reset command buffers that are changed during loop

RenfengLiu:
Reset command buffers that are changed during loop:
1. Re-allocate if freed during loop;
2. Free if allocated during loop;
3. Re-record it if it was submitted and then being re-recorded during
loop.

### DIFF
--- a/gapis/api/vulkan/frame_loop.go
+++ b/gapis/api/vulkan/frame_loop.go
@@ -150,8 +150,8 @@ type frameLoop struct {
 	renderPassToDestroy map[VkRenderPass]bool
 	renderPassToCreate  map[VkRenderPass]bool
 
-	commandBufferAllocated   map[VkCommandBuffer]bool
-	commandBufferFreed       map[VkCommandBuffer]bool
+	commandBufferToFree      map[VkCommandBuffer]bool
+	commandBufferToAllocate  map[VkCommandBuffer]bool
 	commandBufferChanged     map[VkCommandBuffer]bool
 	commandBufferRecorded    map[VkCommandBuffer]bool // Command buffers recorded during loop.
 	commandBufferNotRecorded map[VkCommandBuffer]bool // Command buffers submitted but not recorded during loop.
@@ -249,8 +249,8 @@ func newFrameLoop(ctx context.Context, graphicsCapture *capture.GraphicsCapture,
 		renderPassToDestroy: make(map[VkRenderPass]bool),
 		renderPassToCreate:  make(map[VkRenderPass]bool),
 
-		commandBufferAllocated:   make(map[VkCommandBuffer]bool),
-		commandBufferFreed:       make(map[VkCommandBuffer]bool),
+		commandBufferToFree:      make(map[VkCommandBuffer]bool),
+		commandBufferToAllocate:  make(map[VkCommandBuffer]bool),
 		commandBufferChanged:     make(map[VkCommandBuffer]bool),
 		commandBufferRecorded:    make(map[VkCommandBuffer]bool),
 		commandBufferNotRecorded: make(map[VkCommandBuffer]bool),
@@ -772,32 +772,29 @@ func (f *frameLoop) buildStartEndStates(ctx context.Context, startState *api.Glo
 		case *VkAllocateCommandBuffers:
 			vkCmd := cmd.(*VkAllocateCommandBuffers)
 			cmdBuffers := vkCmd.PCommandBuffers().MustRead(ctx, vkCmd, currentState, nil)
-			f.commandBufferAllocated[cmdBuffers] = true
+			f.commandBufferToFree[cmdBuffers] = true
 			log.D(ctx, "Command buffer %v allcoated.", cmdBuffers)
 		case *VkFreeCommandBuffers:
 			vkCmd := cmd.(*VkFreeCommandBuffers)
 			cmdBufCount := vkCmd.CommandBufferCount()
-			if cmdBufCount == 0 {
-				break
-			}
 			cmdBufs := vkCmd.PCommandBuffers().Slice(0, uint64(cmdBufCount), currentState.MemoryLayout).MustRead(ctx, cmd, currentState, nil)
 			for _, cmdBuf := range cmdBufs {
 				// The command buffer deleted in this call was created during loop, no action needed.
 				log.D(ctx, "Command buffer %v freed", cmdBufs)
-				if _, ok := f.commandBufferAllocated[cmdBuf]; ok {
-					delete(f.commandBufferAllocated, cmdBuf)
+				if _, ok := f.commandBufferToFree[cmdBuf]; ok {
+					delete(f.commandBufferToFree, cmdBuf)
 				} else { // The command buffer deleted in this call was not created during loop, need to back up it
-					f.commandBufferFreed[cmdBuf] = true
+					f.commandBufferToAllocate[cmdBuf] = true
 				}
 			}
 		case *VkBeginCommandBuffer:
 			vkCmd := cmd.(*VkBeginCommandBuffer)
 			cmdBuf := vkCmd.CommandBuffer()
 			f.commandBufferRecorded[cmdBuf] = true
-			log.D(ctx, "Command buffer %v beginned", cmdBuf)
+			log.D(ctx, "Command buffer %v began", cmdBuf)
 
 			// If this command buffer is allocated during loop then it is reset by the re-allocation step.
-			if _, ok := f.commandBufferAllocated[cmdBuf]; ok {
+			if _, ok := f.commandBufferToFree[cmdBuf]; ok {
 				break
 			}
 			// If this command buffer has not been submitted before no action needed.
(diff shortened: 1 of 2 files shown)


gapis/api/vulkan/frame_loop.go
```diff
@@ -196,6 +233,62 @@ func (f *frameLoop) Flush(ctx context.Context, out transform.Writer)    {}
...
+		si := submitInfos[i]
+		cmdBuffers := si.PCommandBuffers().Slice(0, uint64(si.CommandBufferCount()), s.MemoryLayout).MustRead(ctx, cmd, s, nil)
+		cmdCount := si.CommandBufferCount()
+
+		newCmdBuffers := make([]VkCommandBuffer, cmdCount)
+		for j := uint32(0); j < cmdCount; j++ {
+			newCmdBuffers[j] = cmdBuffers[j]
+			// re-record the command buffer if it's not beginned in this frame.
+			if _, ok := f.commandBeginned[cmdBuffers[j]]; !ok {
+				commandbuffer := f.recreateCommandBuffer(ctx, sb, cmdBuffers[j])
```
AWoloszyn: I dont think we want to re-record the command-buffer just before the quuee submit, we should be re-recording it when we are resetting all of the rest of the resources.

RenfengLiu: So the the re-recording only happened:
-  at the resetting step;
-  and only when the resource used in the command buffer has changed;
- and it was not re-recorded in the original application.

Now the questions is how to check if resource in a command buffer has been change.

AWoloszyn: I think I answered this offline but I would stick with 
 - At the resetting step
 - Only if it was submitted during the loop
 - Only if it was not re-recorded before the first submission in the loop.

RenfengLiu: The updated code implements this logic. And also allocate/free the freed/allocated commandbuffer during loop.

gapis/api/vulkan/frame_loop.go
```diff
@@ -105,6 +105,12 @@ type frameLoop struct {
 	descriptorSetLayoutCreated   map[VkDescriptorSetLayout]bool
 	descriptorSetLayoutDestroyed map[VkDescriptorSetLayout]bool
 
+	commandBufferAllocated   map[VkCommandBuffer]bool
+	commandBufferFreed       map[VkCommandBuffer]bool
+	commandBufferChanged     map[VkCommandBuffer]bool
+	commandBufferRecorded    map[VkCommandBuffer]bool // Command buffers recorded during loop.
+	commandBufferNotRecorded map[VkCommandBuffer]bool // Command buffers submitted but not recorded during loop.
```
purvisa-at-google-com: I have changed these to objectToDelete instead of objectCreated in the latest code. This is clearer than the current naming scheme once you have objects getting added to the maps because of object invalidation. Ie, things that need to be destroyed and recreated not because they were created or destroyed inside the loop, but because something else was.

gapis/api/vulkan/frame_loop.go
```diff
@@ -404,14 +416,82 @@ func (f *frameLoop) buildStartEndStates(ctx context.Context, startState *api.Glo
...
+			vkCmd := cmd.(*VkAllocateCommandBuffers)
+			cmdBuffers := vkCmd.PCommandBuffers().MustRead(ctx, vkCmd, currentState, nil)
+			f.commandBufferAllocated[cmdBuffers] = true
+			log.D(ctx, "Command buffer %v allcoated.", cmdBuffers)
+		case *VkFreeCommandBuffers:
+			vkCmd := cmd.(*VkFreeCommandBuffers)
+			cmdBufCount := vkCmd.CommandBufferCount()
+			if cmdBufCount == 0 {
+				break
+			}
```
purvisa-at-google-com: Do we need these lines or will the below code just return and empty slice and iterate over nothing?

gapis/api/vulkan/frame_loop.go
```diff
@@ -404,14 +416,82 @@ func (f *frameLoop) buildStartEndStates(ctx context.Context, startState *api.Glo
...
+					delete(f.commandBufferAllocated, cmdBuf)
+				} else { // The command buffer deleted in this call was not created during loop, need to back up it
+					f.commandBufferFreed[cmdBuf] = true
+				}
+			}
+		case *VkBeginCommandBuffer:
+			vkCmd := cmd.(*VkBeginCommandBuffer)
+			cmdBuf := vkCmd.CommandBuffer()
+			f.commandBufferRecorded[cmdBuf] = true
+			log.D(ctx, "Command buffer %v beginned", cmdBuf)
```
purvisa-at-google-com: beginned -> began

gapis/api/vulkan/frame_loop.go
```diff
@@ -404,14 +416,82 @@ func (f *frameLoop) buildStartEndStates(ctx context.Context, startState *api.Glo
...
+			if _, ok := f.commandBufferChanged[cmdBuf]; !ok {
+				f.commandBufferChanged[cmdBuf] = true
+			}
 
-		if err := cmd.Mutate(ctx, cmdId, currentState, nil, f.watcher); err != nil {
-			return fmt.Errorf("%v: %v: %v", cmdId, cmd, err)
+		case *VkResetCommandBuffer:
+			vkCmd := cmd.(*VkResetCommandBuffer)
+			cmdBuf := vkCmd.CommandBuffer()
+			log.D(ctx, "Command buffer %v resetted", cmdBuf)
```
purvisa-at-google-com: resetted -> reset

gapis/api/vulkan/frame_loop.go
```diff
@@ -404,14 +416,82 @@ func (f *frameLoop) buildStartEndStates(ctx context.Context, startState *api.Glo
...
+			if _, ok := f.commandBufferNotRecorded[cmdBuf]; !ok {
+				break
+			}
+			// Only backup the fist time it changes
+			if _, ok := f.commandBufferChanged[cmdBuf]; !ok {
+				f.commandBufferChanged[cmdBuf] = true
+			}
+		case *VkQueueSubmit:
+			vkCmd := cmd.(*VkQueueSubmit)
+			vkCmd.Extras().Observations().ApplyReads(currentState.Memory.ApplicationPool())
```
purvisa-at-google-com: I think this is done outside the loop in the current code? So no need to reapply the observations.

gapis/api/vulkan/frame_loop.go
```diff
@@ -404,14 +416,82 @@ func (f *frameLoop) buildStartEndStates(ctx context.Context, startState *api.Glo
...
+			submitCount := vkCmd.SubmitCount()
+			submitInfos := vkCmd.pSubmits.Slice(0, uint64(submitCount), currentState.MemoryLayout).MustRead(ctx, cmd, currentState, nil)
+			for i := uint32(0); i < submitCount; i++ {
```
purvisa-at-google-com: Do these need to be integer based loops? can we use range loops?

gapis/api/vulkan/frame_loop.go
```diff
@@ -450,6 +530,11 @@ func (f *frameLoop) detectChangedResources() {
 			continue
 		}
 
+		// Skip multi-sampled images.
+		if image.Info().Samples() != VkSampleCountFlagBits_VK_SAMPLE_COUNT_1_BIT {
+			continue
```
purvisa-at-google-com: Should we be warning here?

gapis/api/vulkan/frame_loop.go
```diff
@@ -709,3 +798,109 @@ func (f *frameLoop) resetDescriptorSetLayouts(ctx context.Context, stateBuilder
...
+		log.D(ctx, "Command buffer %v changed during loop, re-record it.", cmdBuf)
+		if err := f.reRecordCommandBuffer(ctx, stateBuilder, cmdBufObj, cmdBuf); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (f *frameLoop) reRecordCommandBuffer(ctx context.Context, stateBuilder *stateBuilder, refCmdBuf CommandBufferObjectʳ, commandBufferID VkCommandBuffer) error {
```
purvisa-at-google-com: Is there nothing in StateBuilder.go that can help here? It seems like something that would be required by MEC?

gapis/api/vulkan/frame_loop.go
RenfengLiu: yes, I saw your changes, and I can update this once your code is submitted.

gapis/api/vulkan/frame_loop.go
RenfengLiu: yes you're right, just want to return earlier here, it can be can removed

purvisa-at-google-com: If the code has no functional effect, it just serves to confuse a reader and use screen space.

gapis/api/vulkan/frame_loop.go
RenfengLiu: Moved to another PR #3194

PR #3171, opened 2019-09-17
Small Android binary that extracts and loads the Perfetto producer.

pmuetschard:
Adds a small Android native binary, packaged as part of gapid.apk, which will look for the driver APK and load the GPU Perfetto data producer from it.

### DIFF
--- a/cmd/launch_producer/BUILD.bazel
+++ b/cmd/launch_producer/BUILD.bazel
@@ -1,0 +1,24 @@
+# Copyright (C) 2019 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("//tools/build:rules.bzl", "android_native_binary")
+
+android_native_binary(
+    name = "launch_producer",
+    srcs = ["main.cpp"],
+    linkopts = [
+        "-llog",
+        "-ldl",
+    ],
+)
--- a/cmd/launch_producer/main.cpp
+++ b/cmd/launch_producer/main.cpp
@@ -1,0 +1,156 @@
+/*
+ * Copyright (C) 2019 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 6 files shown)


core/os/android/profiling/cc/launch_producer.cc
```diff
@@ -0,0 +1,121 @@
...
+      std::abort();                                 \
+    }                                               \
+    trans;                                          \
+  } while (0)
+#define GET_STRING_PROP(n, t) GET_PROP(n, t = _v)
+
+typedef void (*FN_PTR)(void);
+
+const char* kDriverProperty = "ro.gfx.driver.1";
+const char* kProducerSO = "/assets/libgpudataproducer.so";
```
lpy: We can't hardcode this, eventually this .so file is in lib/{ABI}/libgpudataproducer.so. Can a binary tell the ABI of itself?

pmuetschard: Done.

core/os/android/profiling/cc/launch_producer.cc
```diff
@@ -0,0 +1,154 @@
...
+
+  GET_STRING_PROP(kDriverProperty, driver);
+  LOG_INFO("Driver package: %s", driver.c_str());
+  if (driver.empty()) {
+    LOG_ERR("No driver package set");
+    std::abort();
+  }
+
+  for (char const& c : driver) {
+    if (c == ' ' || c == '"' || c == '\'') {
```
AWoloszyn: Could we maybe get a comment saying "WHY" these characters are a problem?

core/os/android/profiling/cc/launch_producer.cc
```diff
@@ -0,0 +1,154 @@
...
+
+  std::string so = path + "!" + lib;
+  LOG_INFO("Trying %s", so.c_str());
+  void *handle = dlopen(so.c_str(), RTLD_GLOBAL);
+  if ((error = dlerror()) != nullptr || handle == nullptr) {
+    LOG_WARN("Error loading lib: %s", error);
+    return nullptr;
+  }
+
+  FN_PTR startFunc = (FN_PTR)dlsym(handle, "start");
```
AWoloszyn: Is the starting function actually called "start". That seems ... uninformative.

core/os/android/profiling/cc/launch_producer.cc
```diff
@@ -0,0 +1,154 @@
...
+  std::string path = getApkPath(driver);
+
+  dlerror();
+  FN_PTR startFunc = nullptr;
+  for (int i = 0; startFunc == nullptr && i < NELEM(kProducerPaths); i++) {
+    startFunc = loadLibrary(path, kProducerPaths[i]);
+  }
+
+  if (startFunc == nullptr) {
+    LOG_ERR("Did not found the producer library");
```
AWoloszyn: Did not find the producer library.

core/os/android/profiling/cc/launch_producer.cc
pmuetschard: My comment exactly. Waiting on a decision in the design doc. If it gets changed, will update the code accordingly.

core/os/android/profiling/cc/launch_producer.cc
pmuetschard: Done. I've also improved the check to be more strict.

lpy:
Quick update: So far we will stick with `start` as the entry point name.

PR #3174, opened 2019-09-18
Clean up frame looping code and make it work when not first transform

purvisa-at-google-com:
This is a real code proposal for cleaning up the frame looping code and making it work when the transform is not first in the chain.

I am fairly new to this code as well as the code base, so it is very possible I have made some bad assumptions, especially about the more complex parts of the system. Please take special care when reviewing this to balance my inexperience!

### DIFF
--- a/gapis/api/vulkan/frame_loop.go
+++ b/gapis/api/vulkan/frame_loop.go
@@ -35,7 +35,6 @@ type stateWatcher struct {
 
 func (b *stateWatcher) OnBeginCmd(ctx context.Context, cmdID api.CmdID, cmd api.Cmd) {
 }
-
 func (b *stateWatcher) OnEndCmd(ctx context.Context, cmdID api.CmdID, cmd api.Cmd) {
 }
 func (b *stateWatcher) OnBeginSubCmd(ctx context.Context, subIdx api.SubCmdIdx, recordIdx api.RecordIdx) {
@@ -46,51 +45,51 @@ func (b *stateWatcher) OnEndSubCmd(ctx context.Context) {
 }
 func (b *stateWatcher) OnReadFrag(ctx context.Context, owner api.RefObject, frag api.Fragment, valueRef api.RefObject, track bool) {
 }
-
 func (b *stateWatcher) OnWriteFrag(ctx context.Context, owner api.RefObject, frag api.Fragment, oldValueRef api.RefObject, newValueRef api.RefObject, track bool) {
-
 }
 
 func (b *stateWatcher) OnWriteSlice(ctx context.Context, slice memory.Slice) {
+
 	span := interval.U64Span{
 		Start: slice.Base(),
 		End:   slice.Base() + slice.Size(),
 	}
+
 	poolID := slice.Pool()
+
 	if _, ok := b.memoryWrites[poolID]; !ok {
 		b.memoryWrites[poolID] = &interval.U64SpanList{}
 	}
+
 	interval.Merge(b.memoryWrites[poolID], span, true)
 }
 
 func (b *stateWatcher) OnReadSlice(ctx context.Context, slice memory.Slice) {
 }
-
-func (b *stateWatcher) OnWriteObs(ctx context.Context, obs []api.CmdObservation) {
+func (b *stateWatcher) OnWriteObs(ctx context.Context, observations []api.CmdObservation) {
 }
-
-func (b *stateWatcher) OnReadObs(ctx context.Context, obs []api.CmdObservation) {
+func (b *stateWatcher) OnReadObs(ctx context.Context, observations []api.CmdObservation) {
 }
-
 func (b *stateWatcher) OpenForwardDependency(ctx context.Context, dependencyID interface{}) {
 }
-
 func (b *stateWatcher) CloseForwardDependency(ctx context.Context, dependencyID interface{}) {
 }
-
 func (b *stateWatcher) DropForwardDependency(ctx context.Context, dependencyID interface{}) {
 }
 
 // Transfrom
 type frameLoop struct {
-	capture        *capture.GraphicsCapture
-	cmds           []api.Cmd
-	numInitialCmds int
-	loopCount      int32
-	loopStartCmd   api.Cmd
-	loopEndCmd     api.Cmd
-	backupState    *api.GlobalState
-	watcher        *stateWatcher
+	capture   *capture.GraphicsCapture
+	loopCount int32
+
+	loopStartIdx api.CmdID
+	loopEndIdx   api.CmdID
+
+	capturedLoopCmds   []api.Cmd
+	capturedLoopCmdIds []api.CmdID
+
+	backupState *api.GlobalState
+	watcher     *stateWatcher
 
 	bufferCreated   map[VkBuffer]bool
 	bufferChanged   map[VkBuffer]bool
@@ -105,14 +104,34 @@ type frameLoop struct {
 	loopCountPtr value.Pointer
 
 	frameNum uint32
+
+	loopTerminated      bool
+	lastObservedCommand api.CmdID
 }
 
-func newFrameLoop(ctx context.Context, c *capture.GraphicsCapture, numInitialCmds int, Cmds []api.Cmd, loopCount int32) *frameLoop {
-	f := &frameLoop{
-		capture:        c,
-		cmds:           Cmds,
-		numInitialCmds: numInitialCmds,
-		loopCount:      loopCount,
+func newFrameLoop(ctx context.Context, graphicsCapture *capture.GraphicsCapture, loopStart api.CmdID, loopEnd api.CmdID, loopCount int32) *frameLoop {
+
+	if api.CmdID.Real(loopStart) >= api.CmdID.Real(loopEnd) {
+		log.F(ctx, true, "FrameLoop: Cannot create FrameLoop for zero or negative length loop")
+		return nil
+	}
+
+	if loopStart == api.CmdNoID || loopEnd == api.CmdNoID {
+		log.F(ctx, true, "FrameLoop: Cannot create FrameLoop that starts or ends on api.CmdNoID")
+		return nil
+	}
+
+	return &frameLoop{
+
+		capture:   graphicsCapture,
+		loopCount: loopCount,
+
+		loopStartIdx: api.CmdID.Real(loopStart),
+		loopEndIdx:   api.CmdID.Real(loopEnd),
+
+		capturedLoopCmds:   make([]api.Cmd, 0),
+		capturedLoopCmdIds: make([]api.CmdID, 0),
+
 		watcher: &stateWatcher{
 			memoryWrites: make(map[memory.PoolID]*interval.U64SpanList),
 		},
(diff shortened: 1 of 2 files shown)


gapis/api/vulkan/frame_loop.go
```diff
@@ -126,129 +139,237 @@ func newFrameLoop(ctx context.Context, c *capture.GraphicsCapture, numInitialCmd
...
+			}
+
+			if len(f.capturedLoopCmdIds) != len(f.capturedLoopCmds) {
+				log.F(ctx, true, "FrameLoop: Control flow error: Somehow, the number of captured commands and commandIds are not equal.")
+			}
+
+			f.loopTerminated = true
+			log.D(ctx, "FrameLoop: last frame is %v cmdId %v, cmd is %v.", f.frameNum, cmdId, cmd)
+
+			// Do start loop stuff.
```
RenfengLiu: not sure how this is done inside the `if api.CmdID.Real(cmdId) >= f.loopEndIdx {` branch

gapis/api/vulkan/frame_loop.go
```diff
@@ -126,129 +139,237 @@ func newFrameLoop(ctx context.Context, c *capture.GraphicsCapture, numInitialCmd
...
+	}
 
-		st := GetState(out.State())
-		sb := st.newStateBuilder(ctx, newTransformerOutput(out))
-		defer sb.ta.Dispose()
+	// Are we before the loop or just at the start of it?
+	if lastObservedCommand == api.CmdNoID {
+
+		// This is the start of the loop.
+		if api.CmdID.Real(cmdId) == f.loopStartIdx {
```
RenfengLiu: If this the point where loop starts, then probably we need to backup the state here before further mutations.

gapis/api/vulkan/frame_loop.go
```diff
@@ -105,14 +104,28 @@ type frameLoop struct {
...
-	f := &frameLoop{
-		capture:        c,
-		cmds:           Cmds,
-		numInitialCmds: numInitialCmds,
-		loopCount:      loopCount,
+func newFrameLoop(ctx context.Context, graphicsCapture *capture.GraphicsCapture, loopStart api.CmdID, loopEnd api.CmdID, loopCount int32) *frameLoop {
+
+	if api.CmdID.Real(loopStart) >= api.CmdID.Real(loopEnd) {
+		log.F(ctx, true, "FrameLoop: Cannot create FrameLoop for zero or negative length loop")
+	}
```
AWoloszyn: You should return nil, or even more specifically, have this return (*frameLoop, error)

gapis/api/vulkan/frame_loop.go
```diff
@@ -126,129 +139,237 @@ func newFrameLoop(ctx context.Context, c *capture.GraphicsCapture, numInitialCmd
...
+
+	// Lets capture and update the last observed frame from f. From this point on use the local lastObservedCommand variable.
+	lastObservedCommand := f.lastObservedCommand
+	f.lastObservedCommand = cmdId
 
-	if cmd == f.loopStartCmd {
-		log.D(ctx, "Loop: start loop at frame %v, id %v, cmd %v.", f.frameNum, id, cmd)
-		f.detectChangedResource(ctx, out.State())
+	if lastObservedCommand != api.CmdNoID && lastObservedCommand > api.CmdID.Real(cmdId) {
+		log.F(ctx, true, "FrameLoop: expected next observed command ID to be >= last observed command ID")
```
AWoloszyn: How should we handle this error case?

gapis/api/vulkan/frame_loop.go
AWoloszyn: I expect this to be `api.CmdID.Real(cmdId) >= f.loopStartIdx || loopStarted` or something similar.

AWoloszyn: This is queuing up commands, and then only mutating them in detectChangedResource. That is where the backup is created.

gapis/api/vulkan/frame_loop.go
AWoloszyn: This is the "stop queuing things, and start actually running commands"case. So when we are at the end, then we deal with queued commands.

gapis/api/vulkan/frame_loop.go
```diff
@@ -126,129 +139,237 @@ func newFrameLoop(ctx context.Context, c *capture.GraphicsCapture, numInitialCmd
...
 	}
 
-	if _, ok := cmd.(*VkQueuePresentKHR); ok {
-		f.frameNum++
-	}
-
-	out.MutateAndWrite(ctx, id, cmd)
-
+	// Should have early out-ed before this point.
+	log.F(ctx, true, "FrameLoop: Internal control flow error: Should not be possible to reach this statement.")
```
AWoloszyn: We should panic if we ever hit this case.

gapis/api/vulkan/frame_loop.go
```diff
@@ -126,129 +139,237 @@ func newFrameLoop(ctx context.Context, c *capture.GraphicsCapture, numInitialCmd
...
+
 		switch cmd.(type) {
+
 		// Buffers.
 		case *VkCreateBuffer:
 			vkCmd := cmd.(*VkCreateBuffer)
-			vkCmd.Extras().Observations().ApplyWrites(s.Memory.ApplicationPool())
-			buffer := vkCmd.PBuffer().MustRead(ctx, vkCmd, s, nil)
+			buffer := vkCmd.PBuffer().MustRead(ctx, vkCmd, f.backupState, nil)
+			log.D(ctx, "Buffer %v created.", buffer)
```
AWoloszyn: We probably don't want to log everything here

gapis/api/vulkan/frame_loop.go
```diff
@@ -126,129 +139,237 @@ func newFrameLoop(ctx context.Context, c *capture.GraphicsCapture, numInitialCmd
...
 			log.D(ctx, "Buffer %v destroyed.", buffer)
 			f.bufferDestroyed[buffer] = true
 
 		// Images
 		case *VkCreateImage:
 			vkCmd := cmd.(*VkCreateImage)
-			vkCmd.Extras().Observations().ApplyWrites(s.Memory.ApplicationPool())
-			img := vkCmd.PImage().MustRead(ctx, vkCmd, s, nil)
+			img := vkCmd.PImage().MustRead(ctx, vkCmd, f.backupState, nil)
+			log.D(ctx, "Image %v created", img)
```
AWoloszyn: We probably don't want to log everything here

gapis/api/vulkan/frame_loop.go
```diff
@@ -126,129 +139,237 @@ func newFrameLoop(ctx context.Context, c *capture.GraphicsCapture, numInitialCmd
...
 			img := vkCmd.Image()
 			log.D(ctx, "Image %v destroyed", img)
 			f.imageDestroyed[img] = true
 
-		// TODO: Recreate destroyed resources.
-		default:
-			if err := cmd.Mutate(ctx, id, f.backupState, nil, f.watcher); err != nil {
-				return fmt.Errorf("%v: %v: %v", id, cmd, err)
-			}
+			// TODO: Recreate destroyed resources.
```
AWoloszyn: Do we need this TODO again, this is basically what this function is doing, it basically says 
// TODO: finish this.

gapis/api/vulkan/frame_loop.go
RenfengLiu: then the queueing action probably need to be done outside of the `Transfer` method. I think when you reached the `loopEndIdx`, the mutation from later transfer already done?

purvisa-at-google-com: @AWoloszyn any follow up on Robin's comment? I think this is correct, but to be honest I have not followed the whole Transformer system though end to end.

gapis/api/vulkan/frame_loop.go
purvisa-at-google-com: @AWoloszyn is this issue void?

gapis/api/vulkan/frame_loop.go
purvisa-at-google-com: log.F doesn't cause a panic?

gapis/api/vulkan/frame_loop.go
purvisa-at-google-com: done

gapis/api/vulkan/frame_loop.go
purvisa-at-google-com: I wanted to write the loops so no logical terms were contained in more than one "(else) if" in the branch structure. I also wanted to avoid representing loopStarted as a bool as it opens up the possibility of having loopEnded = true, loopStarted = false, which is a logically inconsistent state. 

Do you find the control flow hard to read? I guess ideally the blocks would be in order of before/start/inside/end/after, but that wasn't possible without contravening the above.

gapis/api/vulkan/frame_loop.go
purvisa-at-google-com: done

gapis/api/vulkan/frame_loop.go
purvisa-at-google-com: I was thinking we could strip the logging out as the code became more mature. For now while the code is under very heavy development, it's nice to have extremely verbose logging.

Can I leave it in for now and strip it out when we have more maturity?

gapis/api/vulkan/frame_loop.go
purvisa-at-google-com: See above

gapis/api/vulkan/frame_loop.go
purvisa-at-google-com: It was there before. I guess I saw no harm in leaving it there for now as the code is under heavy development. Like the extra logging I had imagined we'd do a beauty pass once the code is closer to finished.

gapis/api/vulkan/frame_loop.go
AWoloszyn: @RenfengLiu The transform from later transforms is only done when you call "out.MutateAndWrite", meaning I think we are safe here. This just  queues up commands and doesn't call out.MutateAndWrite.

gapis/api/vulkan/frame_loop.go
AWoloszyn: I think that is correct, from what I can tell this code is as expected. (Since nothing mutates)

gapis/api/vulkan/frame_loop.go
AWoloszyn: Oh, I think you are right, F is fatal, sorry.

gapis/api/vulkan/frame_loop.go
AWoloszyn: My main concern was what happens if f.loopStartIdx does not end up being any of the actual commands. Unless I am missing something, this seems possible?

gapis/api/vulkan/frame_loop.go
AWoloszyn: Sure, since it is log.D it won't show up anywhere by default.

gapis/api/vulkan/frame_loop.go
AWoloszyn: See above!

gapis/api/vulkan/frame_loop.go
AWoloszyn: We can, I wont hold this up either way, but it just seems a bit confusing.

gapis/api/vulkan/frame_loop.go
purvisa-at-google-com: Whoops! Fixed with >=

Really good catch!

PR #3177, opened 2019-09-19
Update gpu counter descriptor proto.

lpy:
BUG: #3086


lpy:
ptal

lpy:
Fix clang format :)

PR #3179, opened 2019-09-20
Split large data into chunks to copy and send

yalcinmelihyasin:
Copying and sending large memory resources causes crashes due to lowmemorykiller. Split large resources into smaller chunks to process for preventing it.

b/139421285

### DIFF
--- a/gapii/cc/vulkan_mid_execution.cpp
+++ b/gapii/cc/vulkan_mid_execution.cpp
@@ -27,6 +27,11 @@
 #include <unordered_set>
 #include <vector>
 
+namespace {
+const uint64_t kChunkSizeLimit =
+    16 * 1024 * 1024;  // Limit the data to prevent OOM
+}
+
 namespace gapii {
 
 template <typename T>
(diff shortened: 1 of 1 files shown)


gapii/cc/vulkan_mid_execution.cpp
```diff
@@ -384,47 +384,58 @@ void VulkanSpy::serializeGPUBuffers(StateSerializer *serializer) {
         continue;
       }
       auto &deviceMemory = mState.DeviceMemories[bind.mmemory];
+
+      const uint64_t CHUNK_LIMIT =
+          16 * 1024 * 1024;  // Limit the data to prevent OOM
```
hevrard: For such constants, it's recommended to declare them at the top of the file in empty namespace:

```
namespace gapii {
namespace {
  // Limit the size of chunk data to prevent OOM during capture
  const uint64_t kChunkSizeLimit = 16 * 1024 * 1024;
}  // namespace

...
```

See:
https://google.github.io/styleguide/cppguide.html#Unnamed_Namespaces_and_Static_Variables
https://google.github.io/styleguide/cppguide.html#Constant_Names

AWoloszyn: If this is in the same file, it does not need to be in a nested namespace. A simple anonymous namespace would be fine.

AWoloszyn: Or rather,  we don't have to nest anonymous namespaces inside of other things. Because they should **only** ever be in C++ files.

PR #3182, opened 2019-09-23
Add DescriptorPools to Frame Looping

purvisa-at-google-com:
Add DescriptorPools to Frame Looping

### DIFF
--- a/gapis/api/vulkan/frame_loop.go
+++ b/gapis/api/vulkan/frame_loop.go
@@ -92,25 +92,34 @@ type frameLoop struct {
 	loopStartState *api.GlobalState
 	loopEndState   *api.GlobalState
 
-	bufferCreated   map[VkBuffer]bool
+	bufferToDestroy map[VkBuffer]bool
+	bufferToCreate  map[VkBuffer]bool
 	bufferChanged   map[VkBuffer]bool
-	bufferDestroyed map[VkBuffer]bool
 	bufferToBackup  map[VkBuffer]VkBuffer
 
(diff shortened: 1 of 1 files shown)


gapis/api/vulkan/frame_loop.go
```diff
@@ -709,3 +736,21 @@ func (f *frameLoop) resetDescriptorSetLayouts(ctx context.Context, stateBuilder
...
+
+	// For every DescriptorSetLayout that was created during the loop...
+	for created := range f.descriptorPoolCreated {
+		// Write the command to delete the created object
+		descPool := GetState(f.loopEndState).descriptorPools.Get(created)
+		stateBuilder.write(stateBuilder.cb.VkDestroyDescriptorPool(descPool.Device(), descPool.VulkanHandle(), memory.Nullptr))
+	}
+
+	// For every DescriptorSetLayout that was destroyed during the loop...
+	for destroyed := range f.descriptorPoolDestroyed {
```
RenfengLiu: As described in https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#vkDestroyDescriptorPool, this will makes all the descriptor sets implicitly freed and become invalid. Do we need to mark the descriptor sets in those pools invalid?

purvisa-at-google-com: Oops. Sorry for the accidental resolve. Reopened.

Yeah, I was just recreating the pools for now. I was going to handle rebuilding the DescriptorSets inside a recreated pool as part of the DescriptorSet work in a follow-up commit.

PR #3183, opened 2019-09-23
Handle buffer create/destroy during frame loop

RenfengLiu:


gapis/api/vulkan/frame_loop.go
```diff
@@ -253,6 +257,24 @@ func (f *frameLoop) Transform(ctx context.Context, cmdId api.CmdID, cmd api.Cmd,
 				// Iterate through the loop contents, emitting instructions one by one.
 				for cmdIndex, cmd := range f.capturedLoopCmds {
 
+					switch cmd.(type) {
+					case *VkUnmapMemory:
+						vkCmd := cmd.(*VkUnmapMemory)
+						vkCmd.Extras().Observations().ApplyReads(out.State().Memory.ApplicationPool())
+						memID := vkCmd.Memory()
```
purvisa-at-google-com: Do we need this ApplyReads? What does it do?

gapis/api/vulkan/frame_loop.go
```diff
@@ -279,6 +301,8 @@ func (f *frameLoop) Transform(ctx context.Context, cmdId api.CmdID, cmd api.Cmd,
 					log.E(ctx, "FrameLoop: Failed to reset changed resources %v.", err)
 					return
 				}
+				// Flush out the reset commands
+				stateBuilder.scratchRes.Free(stateBuilder)
```
purvisa-at-google-com: I see lots of explicit flushing going on here. Can you explain why it is needed?

gapis/api/vulkan/frame_loop.go
```diff
@@ -599,8 +630,21 @@ func (f *frameLoop) resetResource(ctx context.Context, stateBuilder *stateBuilde
...
+			stateBuilder.write(stateBuilder.cb.VkUnmapMemory(bufObj.Device(), memID))
+		}
+		stateBuilder.write(stateBuilder.cb.VkDestroyBuffer(bufObj.Device(), buf, memory.Nullptr))
+		stateBuilder.write(stateBuilder.cb.VkFreeMemory(bufObj.Device(), memID, memory.Nullptr))
+	}
+
+	for buf := range f.bufferDestroyed {
+		log.D(ctx, "Recreate buffer %v which was destroyed during loop.", buf)
+		buffer := GetState(f.loopStartState).Buffers().Get(buf)
+		f.recreateDestroyedBuffer(ctx, stateBuilder, buffer)
```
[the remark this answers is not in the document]
RenfengLiu: Yes, the remap is handled in that method in line:708

gapis/api/vulkan/frame_loop.go
RenfengLiu: The `scratchRes.Free` will actully write out the mutation: https://github.com/google/gapid/blob/master/gapis/api/vulkan/scratch_resources.go#L55.  This `Free` function is really confusing.

gapis/api/vulkan/frame_loop.go
RenfengLiu: My understanding is that when we need to get the value of parameters from the capture, we need to `ApplyRead` or `ApplyWrite` depends on the command.

AWoloszyn: You only need to ApplyRead or ApplyWrite if you are reading memory. (MustRead()) if you are just reading the parameters, you do not have to.

gapis/api/vulkan/frame_loop.go
```diff
@@ -626,10 +671,67 @@ func (f *frameLoop) resetBuffers(ctx context.Context, stateBuilder *stateBuilder
...
+	)
+	createWithMemReq.Extras().Add(memReq)
+	stateBuilder.write(createWithMemReq)
+	stateBuilder.write(stateBuilder.cb.VkGetBufferMemoryRequirements(
+		buffer.Device(),
+		buffer.VulkanHandle(),
+		stateBuilder.MustAllocWriteData(memReq).Ptr(),
+	))
+
+	mem := buffer.Memory()
```
AWoloszyn: I think this might be wrong.
This is recreating the entire device memory object, regardless of it is actually deleted or not. I think that is incorrect.

AWoloszyn: We should have a separate pass that re-creates device mmory as needed, (that happens before buffers/images)

RenfengLiu: thanks Andrew for the review, I've create another CL the deal with the device memory. After that I'll update this part of the CL. Change the CL for WIP  for now.

RenfengLiu: Just added code to handle the device memory free/allocate separately. Tested with the sample which free/re-allocate vertex buffer and the dump images are as expected.

RenfengLiu:
Anyone has more comments on this CL?

gapis/api/vulkan/frame_loop.go
```diff
@@ -355,6 +365,28 @@ func (f *frameLoop) Transform(ctx context.Context, cmdId api.CmdID, cmd api.Cmd,
...
 
+					switch cmd.(type) {
+					case *VkUnmapMemory:
+						vkCmd := cmd.(*VkUnmapMemory)
+						memID := vkCmd.Memory()
+						mem := GetState(out.State()).DeviceMemories().Get(memID)
+						addr := mem.MappedLocation().Address()
+						if addr != 0 {
+							// Only remember the first mapped target.
+							if _, ok := f.mappedAddress[addr]; !ok {
```
purvisa-at-google-com: As discussed offline-- we need to take care of when something is mapped and unmapped inside the loop. This code only seems to work when the map is before the loop start and the unmap is inside the loop.

gapis/api/vulkan/frame_loop.go
```diff
@@ -1182,10 +1238,67 @@ func (f *frameLoop) resetResources(ctx context.Context, stateBuilder *stateBuild
...
+				if err != nil {
+					return err
+				}
+				b.Load(protocol.Type_AbsolutePointer, newTarget)
+				b.Store(originalTarget)
+
+				return nil
+			}))
+		}
+	}
```
purvisa-at-google-com: What about binding? When we reallocate device memory that was freed in the last iteration of the loop, we need to make sure that it is rebound to all of the buffers and images that were using it when the loop starts?

RenfengLiu: the bindings for affected images/imageviews/buffer/bufferviews will need to addressed. Those need to be done in another PR.

gapis/api/vulkan/frame_loop.go
```diff
@@ -333,6 +341,8 @@ func (f *frameLoop) Transform(ctx context.Context, cmdId api.CmdID, cmd api.Cmd,
 					log.E(ctx, "FrameLoop: Failed to backup changed resources: %v", err)
 					return
 				}
+				// Flush out the backup commands
+				stateBuilder.scratchRes.Free(stateBuilder)
```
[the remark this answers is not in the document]
RenfengLiu: yes, the flush will actually build the instructions to submit the command buffers do the backup, as defined here at https://github.com/google/gapid/blob/master/gapis/api/vulkan/scratch_resources.go#L53

gapis/api/vulkan/frame_loop.go
RenfengLiu: After further consideration, I'll split this CL into two. one deal with buffer create/destroy and the other for device memory. As your comments reminds me that the memory mapped also need to be unmapped,and the unmapped need to mapped.

PR #3185, opened 2019-09-24
Reset semaphore/fence/event during frame looping

RenfengLiu:
This is the rebase onto the latest code.

### DIFF
--- a/gapis/api/vulkan/frame_loop.go
+++ b/gapis/api/vulkan/frame_loop.go
@@ -113,6 +113,18 @@ type frameLoop struct {
 	descriptorSetDestroyed     map[VkDescriptorSet]bool
 	descriptorSetAutoDestroyed map[VkDescriptorSet]bool
 
+	semaphoreCreated   map[VkSemaphore]bool
+	semaphoreChanged   map[VkSemaphore]bool
+	semaphoreDestroyed map[VkSemaphore]bool
+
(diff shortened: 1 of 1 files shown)


gapis/api/vulkan/frame_loop.go
```diff
@@ -404,7 +428,57 @@ func (f *frameLoop) buildStartEndStates(ctx context.Context, startState *api.Glo
...
+			f.semaphoreCreated[sem] = true
+
+		case *VkDestroySemaphore:
+			vkCmd := cmd.(*VkDestroySemaphore)
+			sem := vkCmd.Semaphore()
+			log.D(ctx, "Semaphore %v is destroyed during loop.", sem)
+			if _, ok := f.semaphoreCreated[sem]; !ok {
+				f.semaphoreDestroyed[sem] = true
+			} else {
+				delete(f.semaphoreCreated, sem)
```
pmuetschard: should all these deletes also delete the mapping from f.<type>Changed?

RenfengLiu: Thanks Pascal Muetschard for helping review the code. 

Yes you're right,need to exclude the f.change*. However the f.changed* is only available at the end of the simulation loop. So updated the code at the reset part to filter out the create/destroy for changed resource in the resetting step.

PR #3186, opened 2019-09-24
Fix the search of descriptorSets in vkUpdateDescriptorSets for copy operation.

yalcinmelihyasin:
b/140988159

### DIFF
--- a/gapis/api/vulkan/api/descriptor.api
+++ b/gapis/api/vulkan/api/descriptor.api
@@ -643,13 +643,13 @@ cmd void vkUpdateDescriptorSets(
   for _ , _ , c in copies {
     srcSet := DescriptorSets[c.SrcSet]
     dstSet := DescriptorSets[c.DstSet]
-    srcBinding := switch (c.SrcBinding in srcSet.Bindings) {
+    srcBinding := switch srcSet.Bindings[c.SrcBinding] == null {
       case false:
         srcSet.Bindings[c.SrcBinding]
       case true:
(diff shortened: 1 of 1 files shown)


gapis/api/vulkan/api/descriptor.api
```diff
@@ -643,13 +643,13 @@ cmd void vkUpdateDescriptorSets(
   for _ , _ , c in copies {
     srcSet := DescriptorSets[c.SrcSet]
     dstSet := DescriptorSets[c.DstSet]
-    srcBinding := switch (c.SrcBinding in srcSet.Bindings) {
+    srcBinding := switch srcSet.Bindings[c.SrcBinding] == null {
       case false:
```
hevrard: For readability, I would prefer keeping `(c.SrcBinding in srcSet.Bindings)` and inverse the true/false of the `case` expressions.

You would also need to mirror this style by changing the condition on line 576

yalcinmelihyasin:
I am not sure about this fix. Please comment on it.

gapis/api/vulkan/api/descriptor.api
AWoloszyn: `c.SrcBinding in srcSet.Bindings` is NOT semantically equivalent to `srcSet.Bindings[c.SrcBinding] == null`. 
The second being correct in more cases.

hevrard: Oh ok, so we have to keep the `== null` test.

yalcinmelihyasin:
Forced push due to flaky(possibly) test.

PR #3187, opened 2019-09-25
Print out basic pipeline skeleton

hliatis:
Prints out a basic skeleton in GAPIT for dump_pipelines

### DIFF
--- a/cmd/gapit/dump_pipeline.go
+++ b/cmd/gapit/dump_pipeline.go
@@ -19,8 +19,6 @@ import (
 	"flag"
 	"fmt"
 	"os"
-	"strconv"
-	"strings"
 	"text/tabwriter"
 
 	"github.com/google/gapid/core/app"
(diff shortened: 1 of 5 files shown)


hliatis:
Bug: 141620836

cmd/gapit/dump_pipeline.go
```diff
@@ -79,9 +77,9 @@ func (verb *pipeVerb) getBoundPipelineResource(ctx context.Context, c client.Cli
 		return nil, err
 	}
 
-	targetType := api.Pipeline_GRAPHICS
+	targetType := "Graphics Pipeline"
```
pau-baiget: Why are we moving back to string literals?

hliatis: Andrew believes it will make it more extensible.

pmuetschard: I don't think this is going to need extending anytime soon. I can understand strings for the stages, but the pipeline type should stay an enum.

PR #3188, opened 2019-09-25
Hopefully final descriptor set looping

purvisa-at-google-com:
I think this is the final piece of the puzzle. Next-- to add tests and make sure it all works :)


gapis/api/vulkan/frame_loop.go
```diff
@@ -426,6 +437,35 @@ func (f *frameLoop) buildStartEndStates(ctx context.Context, startState *api.Glo
 			} else {
 				f.descriptorPoolDestroyed[descriptorPool] = true
 			}
```
RenfengLiu: Do we need to check if this is the first time this pool get destroyed? Though I think it's very unlikely for real application create and destroy the descriptor pool multiple times.

purvisa-at-google-com: I thought about it for a while and I'm not sure it matters how many times a pool is destroyed. At first I was worried that if we had multiple destroys we would only want to reconstruct the descriptor sets that were in the pool before the first destruction, but the code as it is written should already do this. It's more subtle than I would like, but the behaviour is basically encoded in line 847 where any descriptor pools that we recreate have their descriptor sets retroactively added to f.descriptorSetChanged. It does mean we need to call resetDescriptorPools before resetDescriptorSets, but that feels like a natural order for things to happen anyway.

RenfengLiu: thanks for the detailed explanation. I have no more questions.

PR #3194, opened 2019-09-26
Skip multi-sampled images for backup/reset during frame loop.

RenfengLiu:


gapis/api/vulkan/frame_loop.go
```diff
@@ -840,6 +840,12 @@ func (f *frameLoop) detectChangedImages(ctx context.Context) {
 			continue
 		}
 
+		// Skip the multi-sampled images.
+		if image.Info().Samples() != VkSampleCountFlagBits_VK_SAMPLE_COUNT_1_BIT {
+			log.D(ctx, "Multi-sampled image %v is not supported for backup/reset.", image)
```
pmuetschard: Should this be log.W?

PR #3195, opened 2019-09-26
Update to latest Perfetto and use their BUILD file.

pmuetschard:


hevrard:
Just a minor presubmit issue.

Also, perfetto fails to build on Windows, but that's an issue in their code?

lpy:
We need this patch in order to have the latest trace processor to get the extra data from render stages.

PR #3196, opened 2019-09-27
Empty tabs

hliatis:
This creates empty tabs for each stage of the bound pipeline(s). As of now, compute and graphics are combined. This is just preliminary. Expect an additional commit or PR that separates the two.

Bug: 141624419


gapic/src/main/com/google/gapid/views/PipelineView.java
```diff
@@ -0,0 +1,282 @@
+/*
+ * Copyright (C) 2017 Google Inc.
```
pau-baiget: Change copyright

PR #3198, opened 2019-09-27
Cleanup traces after they're pulled from the device

apbodnar:
Fix for b/140064402 
Perfetto fails to overwrite a trace when running as a different user than the creator of the trace.  This is the case for the `shell` user accessing a trace owned by `root` and vice versa. 

This is fixed mostly by deleting the trace immediately after pulling it from the device.

For the edge cases where the trace somehow still exists, we will attempt to delete the file before starting a new trace.  This now works when running as `root`, and will still fail when running as `shell` (b/141704436), but now we present the failure that Perfetto logged to the UI instead of `exited with status 1`

### DIFF
--- a/core/os/android/adb/perfetto.go
+++ b/core/os/android/adb/perfetto.go
@@ -16,9 +16,12 @@ package adb
 
 import (
 	"bufio"
+	"container/ring"
 	"context"
 	"encoding/base64"
+	"errors"
 	"io"
+	"regexp"
 	"strings"
 
 	perfetto_pb "perfetto/config"
@@ -31,7 +34,13 @@ import (
 
 // StartPerfettoTrace starts a perfetto trace on this device.
 func (b *binding) StartPerfettoTrace(ctx context.Context, config *perfetto_pb.TraceConfig, out string, stop task.Signal) error {
+	// Silently attempt to delete the old trace in case the user has changed
+	// If the user has changed from shell => root, this is fine. For root => shell,
+	// the root user will have still have to manually delete the old trace if it exists
+	b.RemoveFile(ctx, out)
+
 	reader, stdout := io.Pipe()
+	logRing := ring.New(10)
 	data, err := proto.Marshal(config)
 	if err != nil {
 		return err
@@ -42,6 +51,8 @@ func (b *binding) StartPerfettoTrace(ctx context.Context, config *perfetto_pb.Tr
 		buf := bufio.NewReader(reader)
 		for {
 			line, e := buf.ReadString('\n')
+			logRing.Value = line
+			logRing = logRing.Next()
 			switch e {
 			default:
 				log.E(ctx, "[perfetto] Read error %v", e)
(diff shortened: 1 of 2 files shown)


core/os/android/adb/perfetto.go
```diff
@@ -29,9 +33,25 @@ import (
...
+	}
+	return err
+}
+
 // StartPerfettoTrace starts a perfetto trace on this device.
 func (b *binding) StartPerfettoTrace(ctx context.Context, config *perfetto_pb.TraceConfig, out string, stop task.Signal) error {
+
+	// Silently attempt to delete the old trace in case the user has changed
+	// If the user has changed from shell => root, this is fine. For root => shell,
+	// the root user will have still have to manually delete the old trace if it exists
```
pau-baiget: Where does the user get the message "You need to delete this manually?"

apbodnar: Assuming perfetto failed because it couldn't overwrite the existing trace, the user will now see the message:
```Tracing Failed
   Cause: perfetto_cmd.cc:747      Failed to open /data/misc/perfetto-traces/gapis-trace. If you get permission denied in /data/misc/perfetto-traces, the file might have been created by another user, try deleting it first. (errno: 13, Permission denied)```

I chose to silently attempt to delete the file because it doesn't necessarily mean perfetto will fail if deletion fails, but it does fix an edge case failure where a root owned trace is still on the device.

File: gapis/perfetto/android/trace.go [hunk 2]
```diff
@@ -97,6 +97,10 @@ func (p *Process) Capture(ctx context.Context, start task.Signal, stop task.Sign
 		return 0, err
 	}
 
+	if err := p.device.DeleteFile(ctx, perfettoTraceFile); err != nil {
```
lpy: my understanding is this is not needed, gapis will end up calling `StartPerfettoTrace`

core/os/android/device.go
```diff
@@ -90,6 +90,8 @@ type Device interface {
 	DeleteSystemSetting(ctx context.Context, namespace, key string) error
 	// StartPerfettoTrace starts a perfetto trace.
 	StartPerfettoTrace(ctx context.Context, config *perfetto_pb.TraceConfig, out string, stop task.Signal) error
+	// DeleteFile deletes a file on the device.
+	DeleteFile(ctx context.Context, traceFile string) error
```
lpy: if I understand correctly, the `trace.go` change is not necessary. Hence no need to extend the interface

core/os/android/adb/perfetto.go
```diff
@@ -29,9 +33,25 @@ import (
...
+	b.DeleteFile(ctx, out)
```
lpy:
```
if err := deleteFile(b, ctx, out); err != nil {
  return err
}
```

gapis/perfetto/android/trace.go
apbodnar: I don't quite follow.  The purpose here is to delete the the using the same permissions as the creator of the trace to avoid any conflicts should the shell be un-rooted.

core/os/android/adb/perfetto.go
apbodnar: We don't want to return an error for the reasons mentioned in the description.

gapis/perfetto/android/trace.go
lpy: Discussed offline :D

lpy:
resolved some comments based on our discussion offline.

core/os/android/adb/perfetto.go
```diff
@@ -29,9 +32,24 @@ import (
 	"github.com/google/gapid/core/log"
 )
 
+// DeleteFile deletes a file on the device.
+func (b *binding) DeleteFile(ctx context.Context, traceFile string) error {
```
lpy: `filePath`

core/os/android/adb/perfetto.go
```diff
@@ -29,9 +32,24 @@ import (
...
+	_, err := b.Shell("rm", "-f", traceFile).Call(ctx)
+	if err != nil {
+		return log.Errf(ctx, err, "Failed to delete perfetto trace file %v. Try root.", traceFile)
```
lpy: `log.Errf(ctx, err, "Failed to delete file %v. Try root.", filePath)`, sorry, didn't notice this should be more generic. I feel like we should move this method to core/os/android/adb/command.go.

core/os/android/adb/perfetto.go
```diff
@@ -29,9 +32,24 @@ import (
...
+	}
+	return err
```
lpy: nil

PR #3199, opened 2019-09-27
Rename gpu renderstage producer

rschiu:


lpy:
Does the latest producer .so drop already have the name changed?

rschiu:
Yes, repackaging it into 9_26 has the name change.  I am repackaging the apk at the moment.

core/os/android/adb/commands.go
```diff
@@ -40,7 +40,7 @@ const (
 	ErrRootFailed      = fault.Const("Device failed to switch to root")
 
 	maxRootAttempts                            = 5
-	gpuRenderingStagesDataSourceDescriptorName = "gpu.renderingstages"
+	gpuRenderingStagesDataSourceDescriptorName = "gpu.renderstages"
```
lpy: I guess we should change the constant to `gpuRenderStagesDataSourceDescriptorName`

rschiu: Done

pmuetschard:
What is going on here? is this the final final final name? #3181

rschiu:
Yes, this is the final final final name.  It's always suppose to be this name, but the producer we have didn't use that name until now.

lpy:
The reason I changed it to `renderingstages` before was because our own prototype https://github.com/rschiu/perfetto-sdk/blob/master/android_client_api_example.cc#L104 also uses this name :D sorry for the confusion

PR #3203, opened 2019-09-30
Automatically run launch_producer for global GPU counter producer.

lpy:
Always attempt to run launch_producer to launch global GPU counter producer.

BUG: #3086


core/os/android/adb/device.go
```diff
@@ -110,6 +111,17 @@ func Devices(ctx context.Context) (DeviceList, error) {
 	return out, nil
 }
 
+func gapidApkPackageName(abi *device.ABI) string {
+	switch {
+	case abi.SameAs(device.AndroidARMv7a):
+		return "com.google.android.gapid.armeabiv7a"
```
rschiu: Did we package the 64-bit binary in the gapid-armeabi-v7a.apk?  It's good if we did, I just want to check.

What about on x86?  In that case, the launcher won't run.

core/os/android/adb/device.go
```diff
@@ -38,7 +38,8 @@ const (
 	// ErrInvalidStatus May be returned if the status string is not a known status.
 	ErrInvalidStatus = fault.Const("Invalid status string")
 	// Frequency at which to print scan errors.
-	printScanErrorsEveryNSeconds = 120
+	printScanErrorsEveryNSeconds            = 120
+	gapidPerfettoProducerLauncherBinaryName = "launch_producer"
```
pmuetschard: just `perfettoProducerLauncher`

core/os/android/adb/device.go
```diff
@@ -110,6 +111,17 @@ func Devices(ctx context.Context) (DeviceList, error) {
 	return out, nil
 }
 
+func gapidApkPackageName(abi *device.ABI) string {
```
pmuetschard: This is a dupe of `gapidapk.PackageName`

core/os/android/adb/perfetto.go
```diff
@@ -29,6 +29,95 @@ import (
 	"github.com/google/gapid/core/log"
 )
 
+const (
+	perfettoProducerLaunchersDirectory = "/data/local/tmp/perfetto_producer_launchers"
+)
+
+func (b *binding) preparePerfettoProducerLauncherFromApk(ctx context.Context, packageName string, launcherBinary string) (string, error) {
+	if b.Instance().GetConfiguration().GetOS().GetAPIVersion() < 29 {
+		return "", log.Errf(ctx, nil, "Querying perfetto capability requires Android API >= 29")
```
pmuetschard: The error message needs to be adjusted

core/os/android/adb/perfetto.go
```diff
@@ -29,6 +29,95 @@ import (
...
+	// unzip also creates the directory structure, clean it up.
+	b.Shell("mv", perfettoProducerLaunchersDirectory+"/assets/"+launcherBinary, perfettoProducerLaunchersDirectory).Call(ctx)
+	b.Shell("rm", "-rf", perfettoProducerLaunchersDirectory+"/assets").Call(ctx)
+
+	// Finally, make sure the binary is executable
+	b.Shell("chmod", "a+x", launcherPath).Call(ctx)
+	return launcherPath, nil
+}
+
+func (b *binding) LaunchPerfettoProducerFromApk(ctx context.Context, packageName string, launcherBinary string, started chan int) error {
```
pmuetschard: why are the package name and launcher path parameters? they are constants.

core/os/android/adb/perfetto.go
```diff
@@ -29,6 +29,95 @@ import (
...
+	if b.Instance().GetConfiguration().GetOS().GetAPIVersion() < 29 {
+		return "", log.Errf(ctx, nil, "Querying perfetto capability requires Android API >= 29")
+	}
+	launcherPath := perfettoProducerLaunchersDirectory + "/" + launcherBinary
+	if _, err := b.Shell("rm", "-f", launcherPath).Call(ctx); err != nil {
+		return "", log.Errf(ctx, err, "Can't clean up existing perfetto producer launcher %v", launcherBinary)
+	}
+
+	// Attempt to create the directory
+	b.Shell("mkdir", "-p", perfettoProducerLaunchersDirectory).Call(ctx)
```
pmuetschard: Use `b.TempFile` to give you a temporary file and extract the binary directly into that file. Then delete it when we are shutting down.

core/os/android/adb/perfetto.go
```diff
@@ -29,6 +29,95 @@ import (
...
+	res, err := b.Shell("pm", "path", packageName).Call(ctx)
+	if err != nil {
+		return "", log.Errf(ctx, err, "Failed to query path to apk %v", packageName)
+	}
+	packagePath := strings.Split(res, ":")[1]
+	if _, err := b.Shell("unzip", "-o", packagePath, "assets/"+launcherBinary, "-d", perfettoProducerLaunchersDirectory).Call(ctx); err != nil {
```
pmuetschard: use `unzip <apk> assets/launch_producer -p > <tmp file>` do not have to do the whole mv/rm thing.

core/os/android/adb/perfetto.go
```diff
@@ -29,6 +29,95 @@ import (
...
+	// Construct IO pipe, shell command outputs to stdout, GAPID reads from
+	// reader for logging purpose.
+	reader, stdout := io.Pipe()
+	fail := make(chan error, 1)
+	crash.Go(func() {
+		buf := bufio.NewReader(reader)
+		for {
+			line, e := buf.ReadString('\n')
+			// As long as there's output, consider the binary starting running.
+			started <- 1
```
pmuetschard: after the second line this will block, since you created the `started` chan with size 1 and only read from it once.

Use our task.Signal instead.

core/os/android/adb/device.go
```diff
@@ -172,6 +179,19 @@ func newDevice(ctx context.Context, serial string, status bind.Status) (*binding
 		}
 	}
 
+	started := make(chan int, 1)
+	go func() {
```
pmuetschard: `crash.Go`

core/os/android/adb/device.go
```diff
@@ -172,6 +179,19 @@ func newDevice(ctx context.Context, serial string, status bind.Status) (*binding
...
+		err := d.LaunchPerfettoProducerFromApk(ctx, gapidApkPackageName(d.To.Configuration.PreferredABI(nil)), gapidPerfettoProducerLauncherBinaryName, started)
+		if err != nil {
```
pmuetschard: should this retry?

core/os/android/adb/device.go
lpy: I can't use gapidapk.PackageName, there's cycle reference

pmuetschard: I suppose, then, move the `LaunchPerfettoProducer` function into the gapidapk package.

core/os/android/adb/device.go
lpy: No, we compile a 32-bit binary version of the `launch_producer` in that case which launches 32-bits producer if available. Not sure whether we pack the `launch_producer` for x86 but if it can't find the producer for that ABI it fails.

core/os/android/adb/device.go
lpy: That doesn't solve the problem, we still need to call `LaunchPerfettoProducer` when we construct the device info for a newly connected device, right?

core/os/android/adb/device.go
lpy: Let's make it fail for now

lpy:
PTAL, as suggested, move to gapidapk package

core/os/android/adb/device.go
rschiu: For x86, it's fine to just fail (assume we have proper error message of course)
For 32-bits, I'm questioning whether packaging the binary into gapid-armeabi-v7a.apk is the right call.  If we do, we should package the same binary.  The counter producer we run should not depend on the app we are profiling.  We should always run the 64-bit counter producer, because are not requiring the vendors to provide both 32-bits and 64-bits version.
It's fine to get something working first.  But please do file a new bug to follow up on this if it isn't resolve in this PR.

lpy: GAPID pushes the apk based on the primary ABI of the device. If the device is 32-bits and there's no 32-bits producer provided, then it will fail anyway.

gapidapk/deviceinfo.go
```diff
@@ -99,6 +103,18 @@ func fetchDeviceInfo(ctx context.Context, d adb.Device) error {
 		return nil
 	}
 
+	if d.Instance().GetConfiguration().GetOS().GetAPIVersion() >= 29 {
+		startSignal, startFunc := task.NewSignal()
+		startFunc = task.Once(startFunc)
+		crash.Go(func() {
+			launchPerfettoProducerFromApk(ctx, d, startFunc)
```
pmuetschard: We should at least log the error returned here.

gapidapk/deviceinfo.go
```diff
@@ -99,6 +103,18 @@ func fetchDeviceInfo(ctx context.Context, d adb.Device) error {
...
+
+			// Make sure we fire the start signal.
```
pmuetschard: Please update the comment to "Ensure start signal is fired on failure/immediate return." I was confused at first why we signal the start at the end of the goroutine, as I thought it was the only place it is fired.

PR #3204, opened 2019-10-01
Sampler desc set invalidation

purvisa-at-google-com:
When a Sampler is recreated it invalidates any Descriptor sets that were created before the loop start. Even though these DescriptorSets are not part of the loop, they must also be recreated.

This change makes that happen. It also renames a lot of variables to make it clearer how the recreation of one type of object can force the recreation of another.

### DIFF
--- a/gapis/api/vulkan/frame_loop.go
+++ b/gapis/api/vulkan/frame_loop.go
@@ -102,6 +102,9 @@ type frameLoop struct {
 	imageDestroyed map[VkImage]bool
 	imageToBackup  map[VkImage]VkImage
 
+	samplerCreated   map[VkSampler]bool
+	samplerDestroyed map[VkSampler]bool
+
 	descriptorSetLayoutCreated   map[VkDescriptorSetLayout]bool
 	descriptorSetLayoutDestroyed map[VkDescriptorSetLayout]bool
 
@@ -157,6 +160,9 @@ func newFrameLoop(ctx context.Context, graphicsCapture *capture.GraphicsCapture,
 		imageChanged:   make(map[VkImage]bool),
 		imageDestroyed: make(map[VkImage]bool),
 		imageToBackup:  make(map[VkImage]VkImage),
+
+		samplerCreated:   make(map[VkSampler]bool),
+		samplerDestroyed: make(map[VkSampler]bool),
 
 		descriptorSetLayoutCreated:   make(map[VkDescriptorSetLayout]bool),
 		descriptorSetLayoutDestroyed: make(map[VkDescriptorSetLayout]bool),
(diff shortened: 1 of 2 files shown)


gapis/api/vulkan/frame_loop.go
```diff
@@ -366,42 +389,105 @@ func (f *frameLoop) buildStartEndStates(ctx context.Context, startState *api.Glo
...
+		case *VkCreateSampler:
+			vkCmd := cmd.(*VkCreateSampler)
+			sampler := vkCmd.PSampler().MustRead(ctx, vkCmd, currentState, nil)
+			log.D(ctx, "Sampler %v created", sampler)
+			f.samplerToDestroy[sampler] = true
+
+		case *VkDestroySampler:
+			vkCmd := cmd.(*VkDestroySampler)
+			sampler := vkCmd.Sampler()
+			log.D(ctx, "Sampler %v created", sampler)
```
RenfengLiu: destroyed

gapis/api/vulkan/frame_loop.go
```diff
@@ -366,42 +389,105 @@ func (f *frameLoop) buildStartEndStates(ctx context.Context, startState *api.Glo
...
 			vkCmd := cmd.(*VkCreateDescriptorSetLayout)
 			descriptorSetLayout := vkCmd.PSetLayout().MustRead(ctx, vkCmd, currentState, nil)
 			log.D(ctx, "DescriptorSetLayout %v created", descriptorSetLayout)
-			f.descriptorSetLayoutCreated[descriptorSetLayout] = true
+			f.descriptorSetLayoutToDestroy[descriptorSetLayout] = true
 
 		case *VkDestroyDescriptorSetLayout:
 			vkCmd := cmd.(*VkDestroyDescriptorSetLayout)
 			descriptorSetLayout := vkCmd.DescriptorSetLayout()
 			log.D(ctx, "DescriptorSetLayout %v created", descriptorSetLayout)
```
RenfengLiu: same here

gapis/api/vulkan/frame_loop.go
```diff
@@ -366,42 +389,105 @@ func (f *frameLoop) buildStartEndStates(ctx context.Context, startState *api.Glo
...
+			}
+
+		// DescriptorSet(s)
+		case *VkAllocateDescriptorSets:
+			vkCmd := cmd.(*VkAllocateDescriptorSets)
+			allocInfo := vkCmd.PAllocateInfo().MustRead(ctx, vkCmd, currentState, nil)
+			descSetCount := allocInfo.DescriptorSetCount()
+			descriptorSets := vkCmd.PDescriptorSets().Slice(0, (uint64)(descSetCount), startState.MemoryLayout).MustRead(ctx, vkCmd, currentState, nil)
+			for index := range descriptorSets {
+				log.D(ctx, "DescriptorSet %v created", descriptorSets[index])
```
RenfengLiu: An alternative would be using `for index, descriptorSet := range...`

PR #3205, opened 2019-10-01
Fix Vulkan MEC image element size for stencil/depth

hevrard:
subGetDepthElementSize() should be used only for the DEPTH aspect, not
the STENCIL aspect.


hevrard:
This bug was introduced in https://github.com/google/gapid/commit/c107b1602d17587a697c5990600f7babe398733d

See the change around `get_element_size` definition. This function returns the size in bytes of the element, in the context of a image-to-buffer copy. The relevant spec is:
https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/VkBufferImageCopy.html

The Stencil case is clear:
> data copied to or from the stencil aspect of any depth/stencil format is tightly packed
> with one VK_FORMAT_S8_UINT value per texel.

This size-retrieving logic is implemented in `subGetElementAndTexelBlockSizeForAspect()` in `gapis/api/vulkan/api/utils.api`.

PR #3211, opened 2019-10-02
Upgrade to latest Perfetto.

lpy:
Upgrade to latest Perfetto so that we can use the trace processor for extraData
field in render stages.

Minor: Format bazel file.

BUG: N/A


lpy:
ptal

Issue #3178, opened 2019-09-19
[UI] Capture doesn't  start at the specified frame number


RenfengLiu:
When select `Frame` for the `Start at` option and input a number in the capture UI, the capture(Vulkan application) didn't start at the specified frame number. 

The application is built from the [latest source code.](https://github.com/google/gapid/commit/730849a13c7f0c93e08401e299a22b9e0a85d1a8) and is tested on gLinux

AWoloszyn:
Was this an off-by-1 error? or did it just not start at all or what?

RenfengLiu:
The capute did not started at all

RenfengLiu:
Update on this one. I thought the capture should automatically start after press the `ok` button. 
So there're two problems:
1. I need to click the start button, which will be hidden because the traced application window popped out. 
2. As the traced application is already running before I press the `Start` button, I think there's no way it could start at the frame specified in the input box. Maybe it calculated the frame number started at the time I press the `Start` button.

PR #3217, opened 2019-10-04
Fix "Start at a given frame" in trace dialog 

pau-baiget:
Two commits:
* The fix.
* Adding some semantics to the startType selector for code clarity.

Fixes https://github.com/google/gapid/issues/3178

### DIFF
--- a/gapic/src/main/com/google/gapid/views/TracerDialog.java
+++ b/gapic/src/main/com/google/gapid/views/TracerDialog.java
@@ -97,6 +97,7 @@ import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Logger;
+import java.util.Arrays;
 import java.util.HashMap;
 
 /**
@@ -258,16 +259,17 @@ public class TracerDialog {
       private static final int DEFAULT_START_FRAME = 100;
       private static final String PERFETTO_LABEL = "Profile Config: ";
 
-      private static final int TRACE_START_TYPE_BEGINNING = 0;
-      private static final int TRACE_START_TYPE_MANUAL = 1;
-      private static final int TRACE_START_TYPE_FRAME = 2;
+      private static enum StartType {
+        Beginning, Manual, Frame;
+      }
+
       private static final HashMap<Integer, String> startTypeTitles = new HashMap<Integer, String>();
 
       static
       {
-        startTypeTitles.put(TRACE_START_TYPE_BEGINNING, "Beginning");
-        startTypeTitles.put(TRACE_START_TYPE_MANUAL, "Manual");
-        startTypeTitles.put(TRACE_START_TYPE_FRAME, "Frame");
+        startTypeTitles.put(StartType.Beginning.ordinal(), "Beginning");
+        startTypeTitles.put(StartType.Manual.ordinal(), "Manual");
+        startTypeTitles.put(StartType.Frame.ordinal(), "Frame");
       };
 
       private final String date = TRACE_DATE_FORMAT.format(new Date());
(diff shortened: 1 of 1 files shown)


gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -257,6 +258,18 @@ protected void buttonPressed(int buttonId) {
...
+      private static final int TRACE_START_TYPE_FRAME = 2;
+      private static final HashMap<Integer, String> startTypeTitles = new HashMap<Integer, String>();
+
+      static
+      {
+        startTypeTitles.put(TRACE_START_TYPE_BEGINNING, "Beginning");
+        startTypeTitles.put(TRACE_START_TYPE_MANUAL, "Manual");
+        startTypeTitles.put(TRACE_START_TYPE_FRAME, "Frame");
+      };
+
```
pmuetschard: Use an enum:
```
startType.setItems(Arrays.stream(StartType.values()).map(Enum::name).toArray(String[]::new));
...
startType.select(<enum>.ordinal());
...
startType.getSelectionIndex() == <enum>.ordinal()
...
etc.
...
private static enum StartType {
  Beginning, Manual, Frame;
}
```

gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -787,7 +787,7 @@ public TraceRequest getTraceRequest(Settings settings) {
           options.addAllEnvironment(splitEnv(envVars.getText()));
         }
         if (config.getMidExecutionCaptureSupport() != Service.FeatureStatus.NotSupported) {
-          options.setDeferStart(settings.traceStartAt != 0);
+          options.setDeferStart(settings.traceStartAt < 0);
```
pmuetschard: Yikes. We probably should also fix this in the server... starting a trace at frame x is by definition a deferred start.

pau-baiget: Yes, it's a bit misleading. I think deferred start refers to "Start when the users clicks button".

gapic/src/main/com/google/gapid/views/TracerDialog.java
pau-baiget: Nifty! Will do that. Thanks!

gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -257,6 +259,19 @@ protected void buttonPressed(int buttonId) {
...
+
+      private static final HashMap<Integer, String> startTypeTitles = new HashMap<Integer, String>();
+
+      static
+      {
+        startTypeTitles.put(StartType.Beginning.ordinal(), "Beginning");
+        startTypeTitles.put(StartType.Manual.ordinal(), "Manual");
+        startTypeTitles.put(StartType.Frame.ordinal(), "Frame");
+      };
+
```
pmuetschard: Just use `startType.name()`. If, however, you don't want to tie yourself to the name of the enum, add a final field in the enum for it. Something like:
```
private static enum StartType {
  Beginning("Beginning"), ... ;

  public final String display;

  private StartType(...) { ... }
}
```

gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -294,6 +309,8 @@ protected void buttonPressed(int buttonId) {
       protected boolean userHasChangedOutputFile = false;
       protected boolean userHasChangedTarget = false;
 
+
+
```
pmuetschard: remove

gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -367,19 +384,19 @@ protected String createAndShowDialog(String current) {
             new GridData(GridData.FILL_HORIZONTAL));
         createLabel(durGroup, "Start at:");
         startType = Widgets.createDropDown(durGroup);
-        startType.setItems("Beginning", "Manual", "Frame");
+        startType.setItems(Arrays.stream(StartType.values()).map(Enum::name).toArray(String[]::new));
```
pmuetschard: This isn't using the map you created above, so if you choose to not use the name, then this needs to be fixed to whatever you choose.

gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -457,16 +474,16 @@ protected void configureDialog(DirectoryDialog dialog) {
 
         Listener mecListener = e -> {
           TraceTypeCapabilities config = getSelectedApi();
-          boolean beginning = startType.getSelectionIndex() == 0;
-          if (!beginning && config != null &&
+          boolean beginning = startType.getSelectionIndex() == StartType.Beginning.ordinal();
+          if (beginning && config != null &&
```
pmuetschard: oops, this inverts the test!

PR #3221, opened 2019-10-07
Update the Eclipse .classpath file for recent Perfetto changes.

pmuetschard:


pmuetschard:
> But just curious how can I identify which is the jar missing and which new jar should be used in replacement?

When I changed the Java proto target in the BUILD files, I should have remembered to update `.classpath`. It's not a very common thing, but when the Java parts of the BUILD files change, we have to make the `.classpath` match.

PR #3222, opened 2019-10-07
Initial time range selection UX.

pmuetschard:


rschiu:
LGTM functionality-wise.  Code looks reasonable, but probably not qualified to approve it.

Is it intentional that I need to press shift to highlight even though I already need to move the mouse to the timeline bar?  I am fine with it, just want to double check.

gapic/src/main/com/google/gapid/perfetto/canvas/Panel.java
```diff
@@ -148,6 +149,36 @@ public boolean click() {
         }
       };
     }
+
+    public default Panel.Hover withClick(BooleanSupplier onClick) {
+      return new Hover() {
+        @Override
+        public Area getRedraw() {
```
pau-baiget: [Possibly ignorant question]: I'm wondering, is the overridden function doing anything different?

gapic/src/main/com/google/gapid/perfetto/models/Selection.java
```diff
@@ -58,6 +59,14 @@ public Composite buildUi(Composite parent, State state) {
         return new MultiSelectionView(parent, selections, state);
       }
     }
+
+    @Override
+    public void mark(State state) {
+      // TODO: is this good enough?
```
pau-baiget: Looks good enough to me.

gapic/src/main/com/google/gapid/perfetto/canvas/Panel.java
pmuetschard: `withClick` creates a delegate that augments the `click()` function below. However, all other calls have to be propagated to the delegate.

PR #3223, opened 2019-10-07
Small improvements to the loading screen.

pmuetschard:
- use the system defined colors for hover
 - better spacing/padding


gapic/src/main/com/google/gapid/LoadingScreen.java
```diff
@@ -91,8 +90,7 @@ public LoadingScreen(Composite parent, Theme theme) {
     statusLabel = createLabel(container, "Starting up...");
     statusLabel.setLayoutData(new GridData(SWT.CENTER, SWT.TOP, true, false));
 
-    optionsContainer = createComposite(
-        container, withSpacing(filling(new RowLayout(SWT.VERTICAL), true, true), 8));
+    optionsContainer = createComposite(container, filling(new RowLayout(SWT.VERTICAL), true, true));
     createOptions();
```
stellama0208: Would suggest adding `optionsContainer.setLayoutData(new GridData(SWT.CENTER, SWT.TOP, true, false));` here in this PR, to avoid text jumping at resizing.

pmuetschard: Done.

PR #3227, opened 2019-10-09
Small fix to time formatting.

pmuetschard:


pmuetschard:
re formatting: that is something to consider. There are some other proposals floating around...

PR #3228, opened 2019-10-09
Updates to the slices UI

pmuetschard:
- use a newer Perfetto version
- update to use the new schema tables
- refactoring some of the shared slices code
- show extra data in the slice selection view
- show the size on slices that have a "width" and "height" extra data field

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/models/ArgSet.java
+++ b/gapic/src/main/com/google/gapid/perfetto/models/ArgSet.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright (C) 2019 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.gapid.perfetto.models;
+
+import static com.google.gapid.util.MoreFutures.transform;
+import static java.lang.String.format;
+
+import com.google.common.collect.ImmutableMap;
+import com.google.common.util.concurrent.ListenableFuture;
+
+import java.util.Map;
+
+/**
+ * Collection of "extra data" from the args table.
+ */
+public class ArgSet {
+  private static final String ARGS_QUERY =
+      "select key, int_value, real_value, string_value from args where arg_set_id = %d";
+
+  public static final ArgSet EMPTY = new ArgSet(ImmutableMap.of());
+
+  private final ImmutableMap<String, Object> values;
+
+  public ArgSet(ImmutableMap<String, Object> values) {
+    this.values = values;
+  }
+
+  public static ListenableFuture<ArgSet> get(QueryEngine qe, long id) {
+    return transform(qe.query(sql(id)), ArgSet::of);
+  }
+
+  private static String sql(Long id) {
+    return format(ARGS_QUERY, id);
+  }
+
+  public static ArgSet of(QueryEngine.Result res) {
+    if (res.getNumRows() == 0) {
+      return EMPTY;
+    }
+
+    ImmutableMap.Builder<String, Object> map = ImmutableMap.builder();
+    res.forEachRow(($, r) -> {
+      if (!r.isNull(1)) {
+        map.put(r.getString(0), r.getInt(1));
+      } else if (!r.isNull(2)) {
+        map.put(r.getString(0), r.getDouble(2));
+      } else {
+        map.put(r.getString(0), r.getString(3));
+      }
+    });
+    return new ArgSet(map.build());
+  }
+
+  public boolean isEmpty() {
+    return values.isEmpty();
+  }
+
+  public int size() {
+    return values.size();
+  }
+
+  public Iterable<String> keys() {
+    return values.keySet();
+  }
+
+  public Iterable<Map.Entry<String, Object>> entries() {
+    return values.entrySet();
+  }
+
+  public Object get(String key) {
+    return values.get(key);
+  }
+}
--- a/gapic/src/main/com/google/gapid/perfetto/models/GpuInfo.java
+++ b/gapic/src/main/com/google/gapid/perfetto/models/GpuInfo.java
@@ -15,57 +15,61 @@
  */
 package com.google.gapid.perfetto.models;
 
-import static com.google.common.collect.Iterators.transform;
-import static com.google.common.collect.Iterators.unmodifiableIterator;
 import static com.google.gapid.util.MoreFutures.transform;
 
+import com.google.common.collect.Iterables;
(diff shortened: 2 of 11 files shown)


lpy:
I took a pass on the code yesterday and also tried it with existing trace, it looks great to me :) I will do another code pass today after I finish my morning meeting.

gapic/src/main/com/google/gapid/perfetto/models/GpuInfo.java
```diff
@@ -15,59 +15,63 @@
...
+      this.trackId = trackId;
       this.maxDepth = maxDepth;
     }
 
+    public Queue(int id, QueryEngine.Row row) {
+      this(id, row.getLong(0), row.getInt(1));
+    }
+
     public String getDisplay() {
       return "GPU Queue " + id;
```
rschiu: The track name is part of the gpu_track table now.  Field name is "name"

gapic/src/main/com/google/gapid/perfetto/models/GpuInfo.java
```diff
@@ -15,59 +15,63 @@
...
-public class GpuInfo implements Iterable<Gpu> {
-  public static final GpuInfo NONE = new GpuInfo(Collections.emptySortedMap());
+public class GpuInfo {
+  public static final GpuInfo NONE = new GpuInfo(Collections.emptyList());
 
   private static final String MAX_DEPTH_QUERY =
-      "select ref, max(depth) + 1 from slices where ref_type = 'gpu' group by ref";
+      "select t.id, max(depth) + 1 " +
+      "from gpu_track t, gpu_slice s " +
+      "where t.id = s.track_id and t.scope = 'gpu_render_stage' " +
```
rschiu: Can we make this "where t.id = s.track_id " + "group by t.id " + "order by t.scope, t.id"?
We aren't doing anything special with the gpu_render_stage track.  If we don't limit the scope, we will be able to see other gpu tracks instead of having to change the UI every time we have a new type of track.

If we end up needing something special for gpu_render_stage, we can do refactor this so that we have a basic UI for all gpu tracks and a special case just for gpu_render_stage.

gapic/src/main/com/google/gapid/perfetto/models/GpuInfo.java
pmuetschard: Yes, but it is ugly ("queue 0"). I'm planning on surfacing the description later though.

gapic/src/main/com/google/gapid/perfetto/models/GpuInfo.java
pmuetschard: No, i specifically don't want that. There already is a bug open to group slices vs counters in the UI. Since we have to update GAPID to use a newer version of the trace processor anyways for these future tracks to even be in the data, making the UI changes at the same time is no big deal.

lpy: I agree with Pascal. That's why I didn't point out. The P0.9 is to get render stages show up.

rschiu: Maybe not in this PR, but if that's the argument, the code needs to be architect such that there is a list of scope it will display.  Adding a new GPU type should not require more than one or two lines of change.

gapic/src/main/com/google/gapid/perfetto/models/GpuInfo.java
rschiu: That's the decision from the vendor and we should respect that.  When we have ARM producer 
they will have something else that's more descriptive.

I can ask Steve to change the name.

gapic/src/main/com/google/gapid/perfetto/models/QueryEngine.java
```diff
@@ -95,6 +107,24 @@ public QueryEngine(Client client, Path.Capture capture, StatusBar status) {
     });
   }
 
+  public ListenableFuture<ArgSet> getArgs(long id) {
+    return argsCache.get(id);
+  }
+
+  public ArgSet getArgsFromCache(long id) {
+    return argsCache.getIfPresent(id);
+  }
```
stellama0208: Just curious, will this method going to be used in some scenarios in the future, when it is a soft cache?

gapic/src/main/com/google/gapid/perfetto/models/ThreadTrack.java
```diff
@@ -71,12 +74,12 @@
 
 
   private final ThreadInfo thread;
-  private final SliceTrack sliceTrack;
+  private final SliceFetcher sliceTrack;
```
stellama0208: The variable name here and below is kept as before, is this intended because it's just a very thin wrapper?

gapic/src/main/com/google/gapid/perfetto/views/GpuQueuePanel.java
```diff
@@ -127,6 +128,11 @@ public void renderTrack(RenderContext ctx, Repainter repainter, double w, double
     });
   }
 
+  private static String getTitle(String title, ArgSet args) {
```
pau-baiget: Bad function name. It's not getting the title, it's constructing it. Can you change it with createTitle or constructTitle?

gapic/src/main/com/google/gapid/perfetto/models/ArgSet.java
```diff
@@ -0,0 +1,86 @@
...
+
+  public boolean isEmpty() {
+    return values.isEmpty();
+  }
+
+  public int size() {
+    return values.size();
+  }
+
+  public Iterable<String> keys() {
```
pau-baiget: Just wondering why you keep the names for all the other functions (size, isEmpty, ..) but not for keySet and entrySet?

gapic/src/main/com/google/gapid/perfetto/models/GpuInfo.java
pmuetschard: once we have more than one, i'm happy to make that change. Until we have more than one, it's not worth it.

gapic/src/main/com/google/gapid/perfetto/models/QueryEngine.java
pmuetschard: removed, since it's not actually used.

gapic/src/main/com/google/gapid/perfetto/models/ThreadTrack.java
pmuetschard: i didn't think it was worth renaming.

gapic/src/main/com/google/gapid/perfetto/models/GpuInfo.java
rschiu: As I mentioned, we don't have to do it in this PR, but I want to note that we already have more than one.
On an R device, surface flinger register itself with data source name of "android.surfaceflinger.frame".  The trace processor only produces instant events at the moment.  So it still won't show with the change, but it would make it easier to test and develop.

gapic/src/main/com/google/gapid/perfetto/views/GpuQueuePanel.java
pmuetschard: done. 🙄

gapic/src/main/com/google/gapid/perfetto/models/ArgSet.java
pmuetschard: because they are not returning the sets, but `Iterables`.

lpy:
\o/

PR #3229, opened 2019-10-09
Small fix to the Perfetto error handling.

pmuetschard:
- Don't drop the original error on the floor
 - Simplify the null & string check


pmuetschard:
+CC @apbodnar

PR #3239, opened 2019-10-11
Add explicit memory read observation for commandbuffers in a submission

chrisforbes:
Allows the tracer to successfully capture a vkQueueSubmit call where the
VkSubmitInfo::commandBufferCount > 1.

Google Bug: b/142511853

### DIFF
--- a/gapis/api/vulkan/api/queue.api
+++ b/gapis/api/vulkan/api/queue.api
@@ -123,6 +123,7 @@ cmd VkResult vkQueueSubmit(
       }
     }
     read(info.pWaitDstStageMask[0:info.waitSemaphoreCount])
+    read(info.pCommandBuffers[0:info.commandBufferCount])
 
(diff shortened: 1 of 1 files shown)


gapis/api/vulkan/api/queue.api
```diff
@@ -123,6 +123,7 @@ cmd VkResult vkQueueSubmit(
       }
     }
     read(info.pWaitDstStageMask[0:info.waitSemaphoreCount])
+    read(info.pCommandBuffers[0:info.commandBufferCount])
```
AWoloszyn: This should not be necessary.
`command_buffers[j]` will do the read.
The case that might be a problem is if we run into
```vkErrorInvalidCommandBuffer(command_buffers[j])```

If there is an error in the command buffer we could see this. We should check this, because this change simply hides that other issue.

PR #3244, opened 2019-10-11
Recreate invalidated ImageViews caused by Image destruction

purvisa-at-google-com:


gapis/api/vulkan/frame_loop.go
```diff
@@ -815,7 +815,7 @@ func (f *frameLoop) detectChangedBuffers(ctx context.Context) {
 			poolID := data.Pool()
 
 			// Did we see this buffer get written to during the loop? If we did, then we need to capture the values at the start of the loop.
-			// TODO: This code does not handle the possibility of new DeviceMemory being bound to the object during the loop. Talk to @purvisa.
+			// TODO: This code does not handle the possibility of new DeviceMemory being bound to the object during the loop. TODO(purvisa).
```
hevrard: I meant adding your name in the TODO at the beginning of the comment :)

gapis/api/vulkan/frame_loop.go
```diff
@@ -869,7 +869,7 @@ func (f *frameLoop) detectChangedImages(ctx context.Context) {
 						poolID := data.Pool()
 
 						// Did we see this part of this image get written to during the loop? If we did, then we need to capture the values at the start of the loop.
-						// TODO: This code does not handle the possibility of new DeviceMemory being bound to the object during the loop. Talk to @purvisa.
+						// TODO: This code does not handle the possibility of new DeviceMemory being bound to the object during the loop. TODO(purvisa).
```
hevrard: idem

gapis/api/vulkan/frame_loop.go
pau-baiget: I'm not a fan of adding people's usernames (specially if they don't match the Github one). Whoever has questions around this code should just "git blame" and ask the corresponding person.

hevrard: This `TODO(username)` is Google style recommendation:  https://google.github.io/styleguide/cppguide.html#TODO_Comments

And it is already quite used in this codebase, try `git grep 'TODO('` to see many instances.

So I think it is worth sticking to this style for consistency.

pau-baiget: Woopsie! sorry :S

Issue #3234, opened 2019-10-10
Push constant handling is broken, leading to gapis crashes while scrolling through commands


chrisforbes:
GAPID Version: 1.7.0:developer
OS: linux amd64

Scrolling up and down through the command list with the arrow keys:

```
panic: context canceled [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x100927e]

goroutine 7472 [running]:
github.com/google/gapid/core/app/crash.Crash(0x566a060, 0xc02e670)
	core/app/crash/crash.go:89 +0xa7
github.com/google/gapid/core/app/crash.handler()
	core/app/crash/crash.go:56 +0x52
panic(0x566a060, 0xc02e670)
	GOROOT/src/runtime/panic.go:522 +0x1b5
github.com/google/gapid/gapis/database.(*memory).resolvePanicHandler(0xc000495140, 0x6c631e0, 0xc003df70e0)
	gapis/database/debug.go:96 +0x43e
panic(0x5505ba0, 0xc0001aa1b0)
	GOROOT/src/runtime/panic.go:522 +0x1b5
github.com/google/gapid/gapis/api/vulkan.panicOnError(...)
	bazel-out/k8-fastbuild/bin/gapis/api/vulkan/api.go:554
github.com/google/gapid/gapis/api/vulkan.U8ˢ.MustReadʷ(0x0, 0x0, 0x10, 0x10, 0xb, 0x6c631e0, 0xc003df7350, 0x0, 0x0, 0xc0044cdea0, ...)
	bazel-out/k8-fastbuild/bin/gapis/api/vulkan/api_slices.go:3888 +0x126
github.com/google/gapid/gapis/api/vulkan.U8ˢ.MustRead(...)
	bazel-out/k8-fastbuild/bin/gapis/api/vulkan/api_slices.go:3882
github.com/google/gapid/gapis/api/vulkan.rebuildVkCmdPushConstants(0x6c631e0, 0xc003df7350, 0x0, 0x7fc9100015f0, 0x556d927c3520, 0xc0044cdea0, 0xc0044cdea0, 0xc001b75c00, 0x1dc7a4, 0x54bdaa0, ...)
	gapis/api/vulkan/command_buffer_rebuilder.go:911 +0xe4
[...]

PR #3251, opened 2019-10-11
Fix push constant state tracking

chrisforbes:
Tracing of vkCmdPushConstants missed capturing the actual data; there
was a write observation created for the push constant state, but the
data in the command was not observed.

This led to two issues:
- LastPushConstants state would never show the actual push constants
- rebuildVkCmdPushConstants would sporadically explode

Bug: #3234


AWoloszyn:
I can't work out why 
```
rebuildVkCmdPushConstants would sporadically explode
```
Do you have some more context here? Was it that we were dropping a dependency in the dependency graph? or something else, since the rebuild command only reads the Args.

Issue #2557, opened 2019-01-23
Unhandled exception in the UI thread.


paulthomson:
GAPID Version: 1.2.1:developer
OS: linux amd64

I had searched for "texture" in the commands panel. I clicked on a few commands. Not really sure what happened. This is running a debug build so might be slower than normal. 

```
org.eclipse.swt.SWTException: Failed to execute runnable (java.lang.IllegalStateException: Querying children before loaded)
	at org.eclipse.swt.SWT.error(SWT.java:4595)
	at org.eclipse.swt.SWT.error(SWT.java:4510)
	at org.eclipse.swt.widgets.Synchronizer.runAsyncMessages(Synchronizer.java:188)
	at org.eclipse.swt.widgets.Display.runAsyncMessages(Display.java:5026)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:4586)
	at org.eclipse.jface.window.Window.runEventLoop(Window.java:821)
	at org.eclipse.jface.window.Window.open(Window.java:797)
	at com.google.gapid.Main$UI.show(Main.java:119)
	at com.google.gapid.Main.main(Main.java:68)
Caused by: java.lang.IllegalStateException: Querying children before loaded
	at com.google.common.base.Preconditions.checkState(Preconditions.java:507)
	at com.google.gapid.models.ApiState$Node.getOrCreateChildren(ApiState.java:212)
	at com.google.gapid.models.ApiState$Node.getChildren(ApiState.java:207)
	at com.google.gapid.views.StateView$StateTree$1.getChildNodes(StateView.java:349)
	at com.google.gapid.views.StateView$StateTree$1.getChildNodes(StateView.java:341)
	at com.google.gapid.widgets.LinkifiedTree$ContentProvider.getChildren(LinkifiedTree.java:250)
	at org.eclipse.jface.viewers.AbstractTreeViewer.getRawChildren(AbstractTreeViewer.java:1422)
	at org.eclipse.jface.viewers.TreeViewer.getRawChildren(TreeViewer.java:350)
	at org.eclipse.jface.viewers.StructuredViewer.getFilteredChildren(StructuredViewer.java:903)
	at org.eclipse.jface.viewers.AbstractTreeViewer.getSortedChildren(AbstractTreeViewer.java:621)
	at org.eclipse.jface.viewers.AbstractTreeViewer.createChildren(AbstractTreeViewer.java:815)
[...]

PR #3252, opened 2019-10-11
Skip updating expansion state if the state model is not available

chrisforbes:
It's possible to have a race here with multiple overlapping updates,
where the state model is not available. The crash here is not fatal, but
throws a dialog in the user's face which is fairly awful.


chrisforbes:
OK, will do.

chrisforbes:
Fixed. Verified that the state view is still well-behaved with this fix.

pmuetschard:
I think this fixes #2557.

chrisforbes:
Yes, that's the same stack.

PR #3260, opened 2019-10-14
Change the non-userdebug build error message to not mention "root".

pmuetschard:


core/os/android/adb/commands.go
```diff
@@ -31,7 +31,7 @@ import (
 const (
 	// ErrDeviceNotRooted is returned by Device.Root when the device is running a
 	// production build as is not 'rooted'.
-	ErrDeviceNotRooted = fault.Const("Device is not rooted")
+	ErrDeviceNotRooted = fault.Const("Device is not a userdebug build")
```
lpy: debuggable?

rschiu: "debuggable" can be confusing.  I understand what you mean given the context, but might have misinterpreted the message if I am not aware of the reason.  Maybe "userdebug/eng"?  Personally, I also don't mind leaving it just userdebug.

Are external users familiar with what a userdebug build is?  I imagine they might not know as they almost always just have user build.

pmuetschard: I think userdebug is fairly well known. Note that this error is only shown to the user if they put in the package manually and will always be shown along with the "cannot trace non-debuggable app" message.

In general, I don't like that these errors surface to the user as is, and we really should make API level errors that the UI can then display nicer than it is now. But that is out of the scope for this. Here I just want to ... root out ...:) the word root as it's confusing people into thinking that if they use an exploit to root a device it would suddenly work, which it won't.

PR #3275, opened 2019-10-16
Reset command buffers that are changed during loop

RenfengLiu:
1. Re-allocate if freed during loop;
2. Free if allocated during loop;
3. Re-record it if it was submitted and not allocated during the loop;

### DIFF
--- a/gapis/api/vulkan/frame_loop.go
+++ b/gapis/api/vulkan/frame_loop.go
@@ -159,11 +159,9 @@ type frameLoop struct {
 	renderPassToDestroy map[VkRenderPass]bool
 	renderPassToCreate  map[VkRenderPass]bool
 
-	commandBufferToFree      map[VkCommandBuffer]bool
-	commandBufferToAllocate  map[VkCommandBuffer]bool
-	commandBufferChanged     map[VkCommandBuffer]bool
-	commandBufferRecorded    map[VkCommandBuffer]bool // Command buffers recorded during loop.
-	commandBufferNotRecorded map[VkCommandBuffer]bool // Command buffers submitted but not recorded during loop.
+	commandBufferToFree     map[VkCommandBuffer]bool
+	commandBufferToAllocate map[VkCommandBuffer]bool
+	commandBufferToRecord   map[VkCommandBuffer]bool
 
 	loopCountPtr value.Pointer
 
@@ -267,11 +265,9 @@ func newFrameLoop(ctx context.Context, graphicsCapture *capture.GraphicsCapture,
 		renderPassToDestroy: make(map[VkRenderPass]bool),
 		renderPassToCreate:  make(map[VkRenderPass]bool),
 
-		commandBufferToFree:      make(map[VkCommandBuffer]bool),
-		commandBufferToAllocate:  make(map[VkCommandBuffer]bool),
-		commandBufferChanged:     make(map[VkCommandBuffer]bool),
-		commandBufferRecorded:    make(map[VkCommandBuffer]bool),
-		commandBufferNotRecorded: make(map[VkCommandBuffer]bool),
+		commandBufferToFree:     make(map[VkCommandBuffer]bool),
+		commandBufferToAllocate: make(map[VkCommandBuffer]bool),
+		commandBufferToRecord:   make(map[VkCommandBuffer]bool),
 
 		loopTerminated:      false,
 		lastObservedCommand: api.CmdNoID,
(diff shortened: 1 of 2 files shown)


gapis/api/vulkan/frame_loop.go
```diff
@@ -779,14 +791,75 @@ func (f *frameLoop) buildStartEndStates(ctx context.Context, startState *api.Glo
...
+			for _, cmdBuf := range cmdBufs {
+				// The command buffer deleted in this call was created during loop, no action needed.
+				log.D(ctx, "Command buffer %v freed", cmdBufs)
+				if _, ok := f.commandBufferToFree[cmdBuf]; ok {
+					delete(f.commandBufferToFree, cmdBuf)
+				} else { // The command buffer deleted in this call was not created during loop, need to back up it
+					f.commandBufferToAllocate[cmdBuf] = true
+				}
+			}
+		case *VkBeginCommandBuffer:
```
purvisa-at-google-com: Tiny "nit", but the other commands all have a single empty line between the cases.

gapis/api/vulkan/frame_loop.go
```diff
@@ -779,14 +791,75 @@ func (f *frameLoop) buildStartEndStates(ctx context.Context, startState *api.Glo
...
+			vkCmd := cmd.(*VkAllocateCommandBuffers)
+			cmdBuffers := vkCmd.PCommandBuffers().MustRead(ctx, vkCmd, currentState, nil)
+			f.commandBufferToFree[cmdBuffers] = true
+			log.D(ctx, "Command buffer %v allcoated.", cmdBuffers)
+		case *VkFreeCommandBuffers:
+			vkCmd := cmd.(*VkFreeCommandBuffers)
+			cmdBufCount := vkCmd.CommandBufferCount()
+			cmdBufs := vkCmd.PCommandBuffers().Slice(0, uint64(cmdBufCount), currentState.MemoryLayout).MustRead(ctx, cmd, currentState, nil)
+			for _, cmdBuf := range cmdBufs {
+				// The command buffer deleted in this call was created during loop, no action needed.
```
purvisa-at-google-com: comment inside if statement

gapis/api/vulkan/frame_loop.go
```diff
@@ -779,14 +791,75 @@ func (f *frameLoop) buildStartEndStates(ctx context.Context, startState *api.Glo
...
+			// If this command buffer is allocated during loop then it is reset by the re-allocation step.
+			if _, ok := f.commandBufferToFree[cmdBuf]; ok {
+				break
+			}
+			// If this command buffer has not been submitted before no action needed.
+			if _, ok := f.commandBufferNotRecorded[cmdBuf]; !ok {
+				break
+			}
+			// Only backup the fist time it changes
+			if _, ok := f.commandBufferChanged[cmdBuf]; !ok {
```
purvisa-at-google-com: Does this if statement actually do anything?

gapis/api/vulkan/frame_loop.go
```diff
@@ -779,14 +791,75 @@ func (f *frameLoop) buildStartEndStates(ctx context.Context, startState *api.Glo
...
```
purvisa-at-google-com: same: this if statement does nothing?

chrisforbes:
This appears not to break the world like the original :)

gapis/api/vulkan/frame_loop.go
```diff
@@ -779,14 +791,75 @@ func (f *frameLoop) buildStartEndStates(ctx context.Context, startState *api.Glo
...
+			cmdBuf := vkCmd.CommandBuffer()
+			f.commandBufferRecorded[cmdBuf] = true
+			log.D(ctx, "Command buffer %v began", cmdBuf)
+
+			// If this command buffer is allocated during loop then it is reset by the re-allocation step.
+			if _, ok := f.commandBufferToFree[cmdBuf]; ok {
+				break
+			}
+			// If this command buffer has not been submitted before no action needed.
+			if _, ok := f.commandBufferNotRecorded[cmdBuf]; !ok {
```
purvisa-at-google-com: What if it was recorded just before the loop began?

gapis/api/vulkan/frame_loop.go
```diff
@@ -779,14 +791,75 @@ func (f *frameLoop) buildStartEndStates(ctx context.Context, startState *api.Glo
...
+			}
+		case *VkQueueSubmit:
+			vkCmd := cmd.(*VkQueueSubmit)
+			submitCount := vkCmd.SubmitCount()
+			submitInfos := vkCmd.pSubmits.Slice(0, uint64(submitCount), currentState.MemoryLayout).MustRead(ctx, cmd, currentState, nil)
+			for _, si := range submitInfos {
+				cmdBuffers := si.PCommandBuffers().Slice(0, uint64(si.CommandBufferCount()), currentState.MemoryLayout).MustRead(ctx, cmd, currentState, nil)
+				for _, cmdBuf := range cmdBuffers {
+					if _, ok := f.commandBufferRecorded[cmdBuf]; !ok {
+						f.commandBufferNotRecorded[cmdBuf] = true
```
purvisa-at-google-com: I think you are assuming vk_command_buffer_usage_one_time_submit_bit. We should check this bit explicitly as it changes the behavior of this logic.

gapis/api/vulkan/state_rebuilder.go
```diff
@@ -2754,6 +2756,13 @@ func (sb *stateBuilder) createCommandBuffer(cb CommandBufferObjectʳ, level VkCo
 		VkResult_VK_SUCCESS,
 	))
 
+}
+
+func (sb *stateBuilder) recordCommandBuffer(cb CommandBufferObjectʳ, level VkCommandBufferLevel, srcState *api.GlobalState) {
+	if cb.Level() != level {
+		return
```
purvisa-at-google-com: Should this be an error?

gapis/api/vulkan/frame_loop.go
purvisa-at-google-com: What if the begin was just before the loop start? Under this scheme, do we not need to consider VkEndCommandBuffer and VkCmd* as editing commands too?

I was talking to Andrew and I think we need to reconstruct every CommandBuffer submitted between the start and end of a loop that was *not* created inside the loop body itself. This will obviously touch more command buffers than strictly necessary in some situations, but it's the safest simplest option for now. We can optimise it to be less conservative over time if it turns out to be a performance issue.

purvisa-at-google-com: That would also let us get rid of the complexity of the "commandBufferNotRecorded" map.

gapis/api/vulkan/frame_loop.go
RenfengLiu: so here we're checking this condition: 
If it was recorded before the start of the loop:
a), if it was submitted during the loop and then at this point start to recording again, then we need to mark is as changed.
b), if it was not submitted before this point, we don't need to to anything

gapis/api/vulkan/frame_loop.go
RenfengLiu: sure, we can do that and that is simpler to implement.

gapis/api/vulkan/state_rebuilder.go
RenfengLiu: this follows the same pattern as the createCommandbuffer

purvisa-at-google-com: That's fair. I am always in favour of consistency.

In this case I think the createCommandbuffer does the wrong thing. But for now I agree we should just copy what is already in the code.

In the future, I want to share a document with some coding guidelines, including defensively programming and crashing early when we get into an illegal state. I think one problem across the GAPID code base is we tend to silently ignore illegal states and continue program execution "hoping for the best". This makes it hard to diagnose the original point of failure when the program finally crashes or does the wrong thing.

gapis/api/vulkan/frame_loop.go
```diff
@@ -869,66 +865,38 @@ func (f *frameLoop) buildStartEndStates(ctx context.Context, startState *api.Glo
...
 		case *VkAllocateCommandBuffers:
 			vkCmd := cmd.(*VkAllocateCommandBuffers)
-			cmdBuffers := vkCmd.PCommandBuffers().MustRead(ctx, vkCmd, currentState, nil)
-			f.commandBufferToFree[cmdBuffers] = true
-			log.D(ctx, "Command buffer %v allcoated.", cmdBuffers)
+			cmdBufCount := vkCmd.PAllocateInfo().MustRead(ctx, vkCmd, currentState, nil).CommandBufferCount()
+			cmdBuffers := vkCmd.PCommandBuffers().Slice(0, uint64(cmdBufCount), currentState.MemoryLayout).MustRead(ctx, vkCmd, currentState, nil)
+			for _, cmdBuf := range cmdBuffers {
+				f.commandBufferToFree[cmdBuf] = true
+				log.D(ctx, "Command buffer %v allcoated.", cmdBuf)
```
purvisa-at-google-com: allcoated  -> allocated

gapis/api/vulkan/state_rebuilder.go
RenfengLiu: yes, having a single coding guidelines is good idea.

PR #3276, opened 2019-10-16
Fix crash during system trace if app is empty

rschiu:


gapis/perfetto/android/trace.go
```diff
@@ -83,16 +83,18 @@ func setupRenderStagesEnvironment(ctx context.Context, d adb.Device, packageName
...
+
+	var cleanup app.Cleanup
+	var err error
 	if a != nil {
 		ctx = log.V{
 			"package":  a.Package.Name,
 			"activity": a.Activity,
 		}.Bind(ctx)
+		// Before we start the activity, attempt to setup environment if gpu.renderstages is selected.
+		cleanup, err = setupRenderStagesEnvironment(ctx, d, a.Package.Name, opts.PerfettoConfig)
```
lpy: let's just print the error here so that we don't have to do `var err error` outside.

pmuetschard: Please make sure to not drop the error on the floor. I know it's the fault of the original change, but can you please fix it here?

PR #3277, opened 2019-10-17
add new WAIT instruction and trigger replay pausing/continuation

apbodnar:
An initial pass in preparation for replay profiling.  After a discussion with @pmuetschard we decided that adding a new `WAIT` instruction would be better than overloading the `NOTIFICATION` instruction.  Currently, executing the wait instruction is coupled with the logic to _eventually_ start a perfetto trace, but it should be extensible to allow waiting for other types of events from the server.

CC: @purvisa-at-google-com

### DIFF
--- a/gapir/api.go
+++ b/gapir/api.go
@@ -41,6 +41,10 @@ type (
 	Notification = replaysrv.Notification
 	// Severity represents the severity level of notification messages. It uses the same enum as gapis
 	Severity = severity.Severity
+	// FenceReadyRequest is sent when the device is waiting for the server perform a task
+	FenceReadyRequest = replaysrv.FenceReadyRequest
+	// FenceReady signals that the server finished a task and replay can continue
+	FenceReady = replaysrv.FenceReady
 )
 
 // ReplayResponseHandler handles all kinds of ReplayResponse messages received
@@ -58,6 +62,8 @@ type ReplayResponseHandler interface {
 	HandleNotification(context.Context, *Notification, Connection) error
 	// HandleFinished handles the replay complete
 	HandleFinished(context.Context, error, Connection) error
+	// HandleFenceReadyRequest handles the profiler ready message.
+	HandleFenceReadyRequest(context.Context, *FenceReadyRequest, Connection) error
 }
 
 // Connection represents a connection between GAPIS and GAPIR. It wraps the
@@ -79,6 +85,8 @@ type Connection interface {
 	SendResources(ctx context.Context, resources []byte) error
 	// SendPayload sends the given payload to the connected GAPIR device.
 	SendPayload(ctx context.Context, payload Payload) error
+	// SendFenceReady signals the device to continue a replay.
+	SendFenceReady(ctx context.Context, id uint32) error
 	// PrewarmReplay requests the GAPIR device to get itself into the given state
 	PrewarmReplay(ctx context.Context, payload string, cleanup string) error
 	// HandleReplayCommunication handles the communication with the GAPIR device on
--- a/gapir/cc/grpc_replay_service.cpp
+++ b/gapir/cc/grpc_replay_service.cpp
@@ -95,6 +95,26 @@ std::unique_ptr<ReplayService::Payload> GrpcReplayService::getPayload(
       std::unique_ptr<replay_service::Payload>(req->release_payload())));
 }
 
+std::unique_ptr<ReplayService::FenceReady> GrpcReplayService::getFenceReady(
+    const uint32_t& id) {
+  // Send a replay response with payload request
+  replay_service::ReplayResponse res;
+  auto frr = new replay_service::FenceReadyRequest();
+  frr->set_id(id);
+  res.set_allocated_fence_ready_request(frr);
+  mGrpcStream->Write(res);
+  std::unique_ptr<replay_service::ReplayRequest> req = getNonReplayRequest();
+  if (!req) {
+    return nullptr;
+  }
+  if (req->req_case() != replay_service::ReplayRequest::kFenceReady) {
+    return nullptr;
+  }
+  return std::unique_ptr<ReplayService::FenceReady>(
+      new ReplayService::FenceReady(std::unique_ptr<replay_service::FenceReady>(
+          req->release_fence_ready())));
+}
+
 std::unique_ptr<ReplayService::Resources> GrpcReplayService::getResources(
     const Resource* resources, size_t resCount) {
   if (!mGrpcStream) {
--- a/gapir/replay_service/service.proto
+++ b/gapir/replay_service/service.proto
@@ -52,12 +52,17 @@ message Replay {
   string dependent_id = 2;
 }
 
+message FenceReady {
+  uint32 id = 1;
+}
+
 message ReplayRequest {
   oneof req {
     Replay replay = 1;
     PrewarmRequest prewarm = 2;
     Payload payload = 3;
     Resources resources = 4;
+    FenceReady fence_ready = 5;
   }
 }
 
@@ -139,6 +144,10 @@ message Notification {
   }
 }
 
+message FenceReadyRequest {
+  uint32 id = 1;
+}
+
 message ReplayResponse {
   oneof res {
     Finished finished = 1;
@@ -147,6 +156,7 @@ message ReplayResponse {
     CrashDump crash_dump = 4;
     PostData post_data = 5;
     Notification notification = 6;
+    FenceReadyRequest fence_ready_request = 7;
   }
 }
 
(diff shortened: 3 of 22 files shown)


gapir/api.go
```diff
@@ -79,6 +85,8 @@ type Connection interface {
 	SendResources(ctx context.Context, resources []byte) error
 	// SendPayload sends the given payload to the connected GAPIR device.
 	SendPayload(ctx context.Context, payload Payload) error
+	// SendPayload sends the given payload to the connected GAPIR device.
```
pmuetschard: Wrong comment.

gapir/replay_service/service.proto
```diff
@@ -52,12 +52,17 @@ message Replay {
   string dependent_id = 2;
 }
 
+message ProfilerReady {
```
pmuetschard: This is the message sent when a WAIT is encountered, correct? Let's un-tie it from the fact that it's used for profiling and simply call it "FenceReady".

gapir/cc/grpc_replay_service.cpp
```diff
@@ -95,6 +95,27 @@ std::unique_ptr<ReplayService::Payload> GrpcReplayService::getPayload(
...
 
+std::unique_ptr<ReplayService::ProfilerReady>
+GrpcReplayService::getProfilerReady(const uint64_t& id) {
+  // Send a replay response with payload request
+  replay_service::ReplayResponse res;
+  auto prc = new replay_service::ProfilerReadyRequest();
+  prc->set_id(id);
+  res.set_allocated_profiler_ready_request(prc);
+  mGrpcStream->Write(res);
+  std::unique_ptr<replay_service::ReplayRequest> req = getNonReplayRequest();
```
pmuetschard: I see this is a pattern we've used before, but here this is getting a bit tricky in that, once a wait ready event is sent, we can only receive a wait ready response. For now, I think that's OK, but it's definitely something to keep in mind.

gapis/api/vulkan/replay.go
```diff
@@ -943,6 +943,7 @@ func (a API) Replay(
 			}
 			profile.AddResult(rr.Result)
 			optimize = false
+			// TODO(apbodnar) add the WaitForProfiler transform _somewhere_. Possibly here
```
pmuetschard: remove this comment, it's implied that we need to actually start using the WAIT operation.

gapis/replay/asm/instructions.go
```diff
@@ -390,3 +390,14 @@ func (a Notification) Encode(r value.PointerResolver, w binary.Writer) error {
 	}
 	return opcode.Notification{}.Encode(w)
 }
+
+type Wait struct {
+	ID uint64
+}
+
+func (a Wait) Encode(r value.PointerResolver, w binary.Writer) error {
+	if err := encodePush(protocol.Type_Uint64, a.ID, w); err != nil {
```
pmuetschard: Instead of push-popping the id, can we make it an immediate to the instruction? I don't think we'll need that many IDs.

gapis/replay/wait_for_profiler.go
```diff
@@ -0,0 +1,47 @@
...
+
+func (t *WaitForProfiler) AddResult(r Result) {
+}
+
+func (t *WaitForProfiler) Transform(ctx context.Context, id api.CmdID, cmd api.Cmd, out transform.Writer) {
+	out.MutateAndWrite(ctx, id, cmd)
+}
+
+func (t *WaitForProfiler) Flush(ctx context.Context, out transform.Writer) {
+	t.AddWaitInstruction(ctx, out)
```
pmuetschard: This will only add wait instruction at the end, which is not really useful. Can you extend this to insert the wait instruction just after a command with a given ID? This would match with some of the other transforms we have (such as the read framebuffer one) and already allow us to use it.

gapis/replay/wait_for_profiler.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func (t *WaitForProfiler) PreLoop(ctx context.Context, out transform.Writer)  {}
+func (t *WaitForProfiler) PostLoop(ctx context.Context, out transform.Writer) {}
+
+func (t *WaitForProfiler) AddWaitInstruction(ctx context.Context, out transform.Writer) {
+	out.MutateAndWrite(ctx, api.CmdNoID, Custom{T: 0, F: func(ctx context.Context, s *api.GlobalState, b *builder.Builder) error {
+		waitID := b.GetNotificationID()
```
pmuetschard: Instead of using the same namespace as the notification ids, simply make the wait ID a parameter to the transform.

gapis/replay/replay_connection.go
```diff
@@ -108,6 +108,19 @@ func (e *backgroundConnection) HandlePayloadRequest(ctx context.Context, payload
...
 
+// HandleProfilerReadyRequest implements gapir.ReplayResponseHandler interface.
+func (e *backgroundConnection) HandleProfilerReadyRequest(ctx context.Context, req *gapir.ProfilerReadyRequest, conn gapir.Connection) error {
+	ctx = status.Start(ctx, "Profile Request")
+	defer status.Finish(ctx)
+	waitId := req.GetId()
+	if waitId == 0 {
+		return log.Err(ctx, nil, "Invalid wait ID")
+	}
+	// TODO(apbodnar) Start a perfetto trace here
```
pmuetschard: This function should just forward it to the e.executor like the notifications above, that way the transform can handle this. It's OK to leave it as a TODO, but maybe mention that in this TODO.

gapis/replay/replay_connection.go
```diff
@@ -108,6 +108,19 @@ func (e *backgroundConnection) HandlePayloadRequest(ctx context.Context, payload
 	return log.Errf(ctx, err, "Payload type is unexpected: %T", boxed)
 }
 
+// HandleProfilerReadyRequest implements gapir.ReplayResponseHandler interface.
+func (e *backgroundConnection) HandleProfilerReadyRequest(ctx context.Context, req *gapir.ProfilerReadyRequest, conn gapir.Connection) error {
+	ctx = status.Start(ctx, "Profile Request")
+	defer status.Finish(ctx)
+	waitId := req.GetId()
+	if waitId == 0 {
```
pmuetschard: Why is 0 a bad id?

apbodnar: It's just an artifact of originally using notification IDs. I'll remove this check.

gapis/replay/asm/instructions.go
apbodnar: I'll pack it into the instruction, but the existing convention seems to be to unpack it in interpreter.cpp and push it onto the stack anyways.

gapir/cc/grpc_replay_service.cpp
chrisforbes: @apbodnar file a bug for fixing this properly?

apbodnar: I've created the bug:  b/143093342

gapis/replay/wait_for_fence.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/google/gapid/gapis/api"
+	"github.com/google/gapid/gapis/api/transform"
+	"github.com/google/gapid/gapis/replay/builder"
+)
+
+type WaitForFence struct{}
+
+func (t *WaitForFence) Transform(ctx context.Context, id api.CmdID, cmd api.Cmd, out transform.Writer) {
+	if cmd.CmdFlags(ctx, id, out.State()).IsStartOfFrame() || id == 0 {
```
AWoloszyn: Is this going to insert a wait event at the start of every frame?

Is this what we want?

apbodnar: After talking with @chrisforbes It seems like we should only insert it at the beginning of the replay.  I'll remove the `IsStartOfFrame` check unless we think we want to profile each frame individually.

chrisforbes: SGTM. We can always adjust this later..

gapis/replay/replay_connection.go
```diff
@@ -108,6 +108,15 @@ func (e *backgroundConnection) HandlePayloadRequest(ctx context.Context, payload
 	return log.Errf(ctx, err, "Payload type is unexpected: %T", boxed)
 }
 
+// HandleFenceReadyRequest implements gapir.ReplayResponseHandler interface.
+func (e *backgroundConnection) HandleFenceReadyRequest(ctx context.Context, req *gapir.FenceReadyRequest, conn gapir.Connection) error {
+	ctx = status.Start(ctx, "Profile Request")
```
pmuetschard: `"Profile Request"` needs to be updated.

PR #3278, opened 2019-10-17
Preserve error when framebuffer attachments can't be fetched

chrisforbes:
The next layer down produces a variety of meaningful error messages for
why a framebuffer attachment is not available. Unfortunately, we were
ignoring them, and instead failing to convert VK_FORMAT_UNDEFINED to our
internal format representation and complaining about that instead.

Google bug: b/142860636


gapis/api/vulkan/vulkan.go
```diff
@@ -106,6 +106,10 @@ func (API) GetFramebufferAttachmentInfo(
 	attachment api.FramebufferAttachment) (info api.FramebufferAttachmentInfo, err error) {
 
 	w, h, form, i, r, err := GetState(state).getFramebufferAttachmentInfo(attachment)
+	if err != nil {
+		return api.FramebufferAttachmentInfo{w, h, i, nil, r}, err
```
hevrard: In case of error, default to `api.FramebufferAttachmentInfo{}` ?

PR #3282, opened 2019-10-18
Clean capture termination of concurrent app with in-flight commands

hevrard:
When capturing multithreaded applications, thread A may emit an
command which ends the capture, such that should_trace(api) returns
false, while thread B emits a command which eventually leads to
sendResource(). In this case, it is valid to reach sendResource() with
should_trace(api) returning false.

It looks OK to keep on trying sending resource even if the
capture has ended, as getEncoder(api)->object(&resource) will
eventually call ChunkWriterImpl::write() which will not try to write
to the closed stream.

Google bug: b/141847511

### DIFF
--- a/gapii/cc/spy_base.cpp
+++ b/gapii/cc/spy_base.cpp
@@ -52,7 +52,7 @@ void SpyBase::init(CallObserver* observer) {
   mIsSuspended = false;
 }
 
-void SpyBase::lock(CallObserver* observer) { mSpinLock.Lock(); }
+void SpyBase::lock() { mSpinLock.Lock(); }
 
 void SpyBase::unlock() { mSpinLock.Unlock(); }
 
--- a/gapii/cc/call_observer.cpp
+++ b/gapii/cc/call_observer.cpp
@@ -132,6 +132,21 @@ void CallObserver::encode(const ::google::protobuf::Message* cmd) {
   encoder()->object(cmd);
 }
 
+void CallObserver::reenter() {
+  if (!mShouldTrace) {
+    // This observer was disabled from the start of the command, nothing to do.
+    return;
+  }
+  mShouldTrace = mSpy->should_trace(mApi);
+  if (!mShouldTrace) {
+    // This branch is taken when this observer was enabled for pre-fence
+    // observations, but a concurrent command terminated the trace while this
+    // command was passed on to the driver. Pop the encoder which was pushed at
+    // creation.
+    mEncoderStack.pop();
+  }
+}
+
 void CallObserver::exit() {
   if (!mShouldTrace) {
     return;
--- a/gapii/cc/call_observer.h
+++ b/gapii/cc/call_observer.h
@@ -171,6 +171,11 @@ class CallObserver : public context_t {
   template <typename T, typename = enable_if_encodable<T> >
   inline void encode(const T& obj);
 
+  // reenter updates whether the observer should keep on tracing or not. It is
+  // meant to be called when a threadsafe command is re-entered after the actual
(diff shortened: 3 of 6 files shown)


chrisforbes:
@pmuetschard if we choose to end tracing, should we consider calls that other threads are in the middle of to be part of the trace, or not? It seems either policy is likely to be fine, but should be consistent.

pmuetschard:
@chrisforbes, yes I was thinking about that, too and was wondering if the situation happens that we're trying to fix, does it mean we have partially serialized commands? i.e. commands where we have the "begin" message, but not the end? It looks to me like there is a tricky underlying problem and just removing the assert seems like we are ignoring it. In the end, having partially serialized commands at the end of the trace may not even be an issue. Maybe the solution here is just to change the `should_trace` to not return false at the end of the trace?

We still need to fix the underlying issue, especially for the multi-trace future as then the statement "ChunkWriterImpl::write() which will not try to write to the closed stream" will no longer apply.

hevrard:
Let me try to recap: there's a single Spy with a single mSpinLock. This Spy creates a new CallObserver for each API command to trace, and each CallObserver locks on mSpinLock, yet for commands marked as `@threadsafe` the CallObserver will unlock just before calling the driver, and re-lock right after the driver returned. This is to enable some other commands to be traced while the driver code executes thread-safe commands.

Now what happens if the capture is terminated (at the end of an end-of-frame command observation) while other commands are in-flight, not yet completely observed?
Currently, we (may) hit the assertion discussed here.
We can either:
1. stop the capture at that point and maybe have only half of the observations of the commands that are in-flight, or
2. make sure to wait for all in-flight commands to be terminated before actually closing the capture.

(2) seems cleaner to me. It would require the Spy to know whether there are any commands that are in-flight, and wait for them to finish, before actually terminating the capture. I feel like this could be achieved by:
- a mSpinLock-protected counter of in-flight commands
- make the Spy check that there is no in-flight commands before closing the trace
  - if there are still commands in flight, it waits on a condition variable, which is notified in `Spy::exit()` when the counter reaches zero
- another flag `shouldNotTraceNewCommands` to indicate that we want to terminate the capture, so no new commands should be traced---only pending observations of in-flight commands should be added to the trace.

@pmuetschard @chrisforbes this is a rough draft, but how does it sound?

AWoloszyn:
For #2 above, we have to make sure that we do NOT drop commands in this case. One of the main reasons that @threadsafe even exists is for synchronization primitives. If someone is holding an event across the end-frame, we will now have to keep all commands up to and including the event signal operation.

Of course, now some OTHER command may be waiting on such a synchronization primitive, so we have to wait on that, etc etc

hevrard:
Some decisions after team meeting:
- keep the pre-fence observations of in-flight commands
- do not wait for these commands to terminate (they may never terminate, e.g. with a Vulkan `event`...)
- mark these commands as "not terminated" in the server, such that they are not replayed, and highlighted as such in the UI

hevrard:
Only fix the proper capture termination for now. I suggest to add the "not terminated" flag to commands as a separate PR.

gapii/cc/spy_base.cpp
```diff
@@ -52,7 +52,7 @@ void SpyBase::init(CallObserver* observer) {
   mIsSuspended = false;
 }
 
-void SpyBase::lock(CallObserver* observer) { mSpinLock.Lock(); }
+void SpyBase::lock() { mSpinLock.Lock(); }
```
chrisforbes: Commit message should mention that you're doing this :)

hevrard: Done!

PR #3283, opened 2019-10-18
Rename some of the maps for synchronization for frame loop

RenfengLiu:


RenfengLiu:
> Good, but please re-order the declarations to match alpha ordering, i.e. `ToDestroy` comes after `ToCreate`
> 
> Just to check: do you sync with @purvisa-at-google-com to get homogeneous naming?

Thanks very much for the quick review. Currently all the destroy/free comes before the create/allocate, which would be the followed logic to destroy/free first and then create/allocate them. The toDestroy/toCreate name is suggested by @purvisa-at-google-com and I'd like to follow this rule to make the code looks the same style. 

@purvisa-at-google-com Do you have any suggestion on this?

hevrard:
I see, and the other declarations around also match the logical order (destroy first, create second). All good then!

PR #3286, opened 2019-10-19
Fix NPE in the trace dialog, when no device is connected.

pmuetschard:


lpy:
conflict with https://github.com/google/gapid/pull/3285

PR #3288, opened 2019-10-22
Add Input Assembly Information

hliatis:
This shows the basics of the Key-Value Pair List. It handles the info from VkPipelineInputAssemblyStateCreateInfo (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/VkPipelineInputAssemblyStateCreateInfo.html)


cmd/gapit/dump_pipeline.go
```diff
@@ -20,10 +20,12 @@ import (
 	"fmt"
 	"os"
 	"text/tabwriter"
+	//"strconv"
```
AWoloszyn: Avoid commented out code.

gapis/api/vulkan/resources.go
```diff
@@ -1049,6 +1071,21 @@ func InputAssembly() *api.Stage {
 
 		&api.DataGroup{
 			GroupName: "Input Assembly State",
+			Data: &api.DataGroup_KeyValues{
+				&api.KeyValuePairList{
+					KeyValues: []*api.KeyValuePair{
```
AWoloszyn: This is quite verbose, I would prefer to extract the repeat KeyValuePair{ Name: Value: } into a separate method, or these will start getting out of hand.

PR #3290, opened 2019-10-22
updated external package list for gofuse

johnplate:


cmd/gofuse/main.go
```diff
@@ -44,15 +44,16 @@ import (
 // Map of bazel external package names to the expected import names.
 var externals = map[string]string{
 	"com_github_golang_protobuf":       filepath.Join("github.com", "golang", "protobuf"),
-	"com_github_google_protobuf":       filepath.Join("github.com", "google", "protobuf"),
```
pmuetschard: Is this really not needed anymore?

johnplate: Pretty sure. Before I removed that line I checked that there is none of those imports left in the entire codebase. I also checked that the bazel cache did not contain that package after a fresh compilation, to make sure that no generated code would use it.

pmuetschard: Ha! I just noticed, `com_github_google_protobuf` doesn't even exist :) I thought this was the main protobuf one, but that one is `com_google_protobuf` and doesn't contain any go code.

PR #3291, opened 2019-10-22
Fix the infinite wait on semaphores

yalcinmelihyasin:
Fixes the infinite wait on semaphores used as wait semaphore without submitted for signaling.

### DIFF
--- a/gapis/api/vulkan/api/queue.api
+++ b/gapis/api/vulkan/api/queue.api
@@ -113,9 +113,16 @@ cmd VkResult vkQueueSubmit(
           wait_semaphores_all_valid.b = false
           vkErrorInvalidSemaphore(ws)
         } else {
-          LastBoundQueue.PendingCommands[len(LastBoundQueue.PendingCommands)]
-          = new!CommandReference(as!VkCommandBuffer(0), 0, cmd_vkNoCommand, 0,
-            Unsignal,    ws,    null, as!VkFence(0))
+          semaphore := Semaphores[ws]
+
+          if(semaphore.SubmitCount == 0) {
+            vkErrorSemaphoreNotSubmitted(ws)
+          } else {
+            LastBoundQueue.PendingCommands[len(LastBoundQueue.PendingCommands)]
+            = new!CommandReference(as!VkCommandBuffer(0), 0, cmd_vkNoCommand, 0,
+              Unsignal,    ws,    null, as!VkFence(0))
+            semaphore.SubmitCount = semaphore.SubmitCount - 1
+          }
         }
       }
     }
@@ -206,6 +213,8 @@ cmd VkResult vkQueueSubmit(
           signal_semaphores_all_valid.b = false
           vkErrorInvalidSemaphore(ss)
         } else {
+          semaphore := Semaphores[ss]
+          semaphore.SubmitCount = semaphore.SubmitCount + 1
           LastBoundQueue.PendingCommands[len(LastBoundQueue.PendingCommands)]
           = new!CommandReference(as!VkCommandBuffer(0), 0, cmd_vkNoCommand, 0,
             Signal,      ss,  null, as!VkFence(0))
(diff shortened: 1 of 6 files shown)


yalcinmelihyasin:
b/143145690

gapis/api/vulkan/api/queue.api
```diff
@@ -113,9 +113,15 @@ cmd VkResult vkQueueSubmit(
           wait_semaphores_all_valid.b = false
           vkErrorInvalidSemaphore(ws)
         } else {
-          LastBoundQueue.PendingCommands[len(LastBoundQueue.PendingCommands)]
-          = new!CommandReference(as!VkCommandBuffer(0), 0, cmd_vkNoCommand, 0,
-            Unsignal,    ws,    null, as!VkFence(0))
+          semaphore := Semaphores[ws]
+
+          if(semaphore.State != Submitted) {
+            vkErrorInvalidSemaphore(ws)
```
chrisforbes: You need a better error here

AWoloszyn: We should create another error, vkErrorSempahoreNotSubmitted().

gapis/api/vulkan/api/queue.api
```diff
@@ -206,6 +212,7 @@ cmd VkResult vkQueueSubmit(
           signal_semaphores_all_valid.b = false
           vkErrorInvalidSemaphore(ss)
         } else {
+          Semaphores[ss].State = Submitted
```
AWoloszyn: I think this is not perfectly correct, but it might be fine enough.
The issue here is that this will get submitted "before" the execute. I feel like we probably want to mark them as submitted after the execute.

yalcinmelihyasin: After the execution, should not they pass to the "signaled" state?

gapis/api/vulkan/api/synchronization.api
```diff
@@ -166,6 +166,7 @@ cmd VkResult vkWaitForFences(
   @unused VkDevice                  Device
   @unused VkSemaphore               VulkanHandle
   @unused VkQueue                   LastQueue
+  @unused u32                       Submitted
```
AWoloszyn: SubmitCount or something else might be a better name. Submitted by itself sounds like it should be a bool.

gapis/api/vulkan/api/queue.api
```diff
@@ -113,9 +113,16 @@ cmd VkResult vkQueueSubmit(
...
-            Unsignal,    ws,    null, as!VkFence(0))
+          semaphore := Semaphores[ws]
+
+          if(semaphore.Submitted == 0) {
+            vkErrorSemaphoreNotSubmitted(ws)
+          } else {
+            LastBoundQueue.PendingCommands[len(LastBoundQueue.PendingCommands)]
+            = new!CommandReference(as!VkCommandBuffer(0), 0, cmd_vkNoCommand, 0,
+              Unsignal,    ws,    null, as!VkFence(0))
+            semaphore.Submitted = semaphore.Submitted - 1
```
AWoloszyn: I am a bit worried we might be able to get in an underflow state, which would screw everything up.

yalcinmelihyasin: Thats true. I will change it to a signed variable.

AWoloszyn: Not just a signed variable, I think we actually want to clamp it

AWoloszyn: A negative submit count is invalid (this is just to handle invalid cases)
So I think the solution will be
```
semaphore.Submitted = switch semaphore.Submitted {
 case 0:
      0
  default:
      semaphore.Submitted - 1
}
```

yalcinmelihyasin: There we are already checking if it is 0 before. But we don't check in vkQueuePresentKHR though. I think if we add a check there too, it shouldn't be an issue.

AWoloszyn: Sure that works for me.

yalcinmelihyasin:
Test is flaky and not related

PR #3294, opened 2019-10-23
Add Vulkan validation layers to GAPID APK

hevrard:
We need to have access to the ANDROID_NDK_HOME environment variable,
hence the new repository_rule.

Google bug: b/141360600

### DIFF
--- a/gapidapk/android/apk/rules.bzl
+++ b/gapidapk/android/apk/rules.bzl
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-load("//tools/build:rules.bzl", "android_native_vulkan_validation_layers", "extract", "filehash")
+load("//tools/build:rules.bzl", "extract", "filehash")
 load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain")
 
 def _strip_impl(ctx):
@@ -100,10 +100,11 @@ def gapid_apk(name = "", abi = "", pkg = "", libs = {}, bins = {}):
         name = name + "_native",
         linkstatic = 1,
         srcs = select({
-            "//tools/build:android-" + name: [
-                name + "_libs",
-                "@android_native_vulkan_validation_layers//:vulkan_validation_layer_" + abi,
-            ],
+            "//tools/build:android-" + name: [name + "_libs"],
+            "//conditions:default": [],
+        }),
+        deps = select({
+            "//tools/build:android-" + name: ["@ndk_vk_validation_layer//:" + abi],
             "//conditions:default": [],
         }),
     )
--- a/gapidapk/android/apk/rules.bzl
+++ b/gapidapk/android/apk/rules.bzl
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-load("//tools/build:rules.bzl", "extract", "filehash")
+load("//tools/build:rules.bzl", "android_native_vulkan_validation_layers", "extract", "filehash")
 load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain")
 
 def _strip_impl(ctx):
@@ -100,9 +100,12 @@ def gapid_apk(name = "", abi = "", pkg = "", libs = {}, bins = {}):
         name = name + "_native",
         linkstatic = 1,
         srcs = select({
-            "//tools/build:android-" + name: [name + "_libs"],
+            "//tools/build:android-" + name: [
+                name + "_libs",
+                "@android_native_vulkan_validation_layers//:vulkan_validation_layer_" + abi,
+            ],
             "//conditions:default": [],
-        })
+        }),
     )
     native.android_binary(
         name = name,
--- a/tools/build/rules.bzl
+++ b/tools/build/rules.bzl
@@ -16,6 +16,7 @@ load("//tools/build/rules:android.bzl",
     _android_native_app_glue = "android_native_app_glue",
     _android_native = "android_native",
     _android_native_binary = "android_native_binary",
+    _android_native_vulkan_validation_layers = "android_native_vulkan_validation_layers",
 )
 load("//tools/build/rules:apic.bzl",
     _apic_compile = "apic_compile",
@@ -76,6 +77,7 @@ load("//tools/build/rules:zip.bzl",
 android_native_app_glue = _android_native_app_glue
 android_native = _android_native
 android_native_binary = _android_native_binary
+android_native_vulkan_validation_layers = _android_native_vulkan_validation_layers
 apic_compile = _apic_compile
 apic_template = _apic_template
 cc_copts = _cc_copts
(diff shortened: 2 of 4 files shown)


hevrard:
@pmuetschard this is my first meaningful contribution in Bazel, not sure whether I'm doing it right or not, feedback very welcome!

I'm still facing Bazel warnings like the following:

```
WARNING: /path/to/work/gapid/gapidapk/android/apk/BUILD.bazel:31:1:
in srcs attribute of cc_library rule //gapidapk/android/apk:armeabi-v7a_native:
please do not import '@android_native_vulkan_validation_layers//:armeabi-v7a/libVkLayer_threading.so'
directly. You should either move the file to this package or depend on an appropriate
rule there. Since this rule was created by the macro 'gapid_apk', the error might have
been caused by the macro implementation in /path/to/work/gapid/gapidapk/android/apk/rules.bzl:102:16
```

Do you have any idea what needs hot to properly get rid of these?

AWoloszyn:
Does the NDK not get pulled in as a part of the build?
Can we pull the NDK in as part of the build itself?

AWoloszyn:
Looks like the ndk already comes in like this:

```
        maybe_repository(
            native.android_ndk_repository,
            name = "androidndk",
            locals = locals,
            api_level = 23, # This is the minimum API
        )
```

IT would be good if we can just point to the existing repository, rather than needing the ANDROID_NDK_HOME as well?

tools/build/rules/android.bzl
```diff
@@ -105,3 +105,35 @@ android_native_app_glue = repository_rule(
         "ANDROID_NDK_HOME",
     ]
 )
+
+# Retrieve Vulkan validation layers from the Android NDK
+
+def _android_native_vulkan_validation_layers(ctx):
+    build = ""
+    for abi in ["armeabi-v7a", "arm64-v8a", "x86", "x86_64"]:
```
AWoloszyn: We don't ship x86_64 for Android.

tools/build/rules/android.bzl
```diff
@@ -105,3 +105,35 @@ android_native_app_glue = repository_rule(
...
+
+        for layer in ["core_validation", "object_tracker", "parameter_validation", "threading", "unique_objects"]:
```
AWoloszyn: If we update the NDK version (if we can), these have been renamed to be a single layer, might be worth updating GAPID for this.

tools/build/rules/android.bzl
```diff
@@ -105,3 +105,35 @@ android_native_app_glue = repository_rule(
...
+            layerpath = abi + "/libVkLayer_" + layer + ".so"
+            ctx.symlink(
```
AWoloszyn: Does this work on Windows?

tools/build/rules/android.bzl
hevrard: The unified layer ships in NDK r21, which is still in Beta as of today (https://developer.android.com/ndk/downloads)

hevrard:
The Android NDK Bazel rules are auto-generated by Bazel itself, and they offer a predefined set of rules. The BUILD file for the NDK can be found in `~/.cache/bazel/.../external/androidndk/BUILD.bazel`

I can't see an easy way to refer to the validation layers using the predefined NDK rules / targets. I opened https://github.com/bazelbuild/bazel/issues/10094 to ask for new filegroups to easily access the Validation layers.

tools/build/rules/android.bzl
hevrard: Yes, confirmed by the CI build, I manually checked that the validation layers are present in the generated APKs

tools/build/workspace.bzl
```diff
@@ -278,6 +278,12 @@ def gapid_dependencies(android = True, mingw = True, locals = {}):
             locals = locals,
         )
 
+        maybe_repository(
+            android_native_vulkan_validation_layers,
```
pmuetschard: I would just call the rule and the repository `ndk_vk_validation_layers`

tools/build/rules/android.bzl
```diff
@@ -105,3 +105,35 @@ android_native_app_glue = repository_rule(
...
+                ctx.path(ctx.os.environ["ANDROID_NDK_HOME"] +
+                         "/sources/third_party/vulkan/src/build-android/jniLibs/" + layerpath),
+                ctx.path(layerpath),
+            )
+
+        build += "\n".join([
+            "filegroup(",
+            "    name = \"vulkan_validation_layer_" + abi + "\",",
```
pmuetschard: Just use the abi as the rule name. The `vulkan_validation_layer` part is redundant with the workspace name.

tools/build/workspace.bzl
hevrard: done

tools/build/rules/android.bzl
hevrard: done

tools/build/rules.bzl
```diff
@@ -16,6 +16,7 @@ load("//tools/build/rules:android.bzl",
     _android_native_app_glue = "android_native_app_glue",
     _android_native = "android_native",
     _android_native_binary = "android_native_binary",
+    _android_native_vulkan_validation_layers = "android_native_vulkan_validation_layers",
```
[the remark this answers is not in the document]
hevrard: I was following the native app glue indeed! Fixed.

gapidapk/android/apk/rules.bzl
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-load("//tools/build:rules.bzl", "extract", "filehash")
+load("//tools/build:rules.bzl", "android_native_vulkan_validation_layers", "extract", "filehash")
```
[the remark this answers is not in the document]
hevrard: removed

gapidapk/android/apk/rules.bzl
```diff
@@ -102,7 +102,11 @@ def gapid_apk(name = "", abi = "", pkg = "", libs = {}, bins = {}):
         srcs = select({
             "//tools/build:android-" + name: [name + "_libs"],
             "//conditions:default": [],
-        })
+        }),
+        deps = select({
+            "//tools/build:android-" + name: ["@ndk_vk_validation_layer//:" + abi],
+            "//conditions:default": [],
+        }),
```
pmuetschard: does it not work if you just make it a dep below in the android_binary target?

hevrard: I tried this in native.android_binary:

```
        deps = [
            "//gapidapk/android/app/src/main:gapid",
            ":" + name + "_native",
        ] + select({
            "//tools/build:android-" + name: ["@ndk_vk_validation_layer//:" + abi],
        },
	no_match_error = "ERROR"
	),
```

This does hit the error, I'm not sure why we cannot rely on the same condition as in native.android_library?

Another option is:

```
        deps = [
            "//gapidapk/android/app/src/main:gapid",
            ":" + name + "_native",
            "@ndk_vk_validation_layer//:" + abi
        ],
```

But this leads to all APKs embedding validation layers for all ABIs.

PR #3296, opened 2019-10-23
Change Frame Looping logic to not rewind state on last iteration

purvisa-at-google-com:
Changes the loop in frame looping so that we do not rewind the state on the last iteration.

This required some changes to the bytecode interpreter, since forward branching did not work in the old code.

### DIFF
--- a/gapir/cc/interpreter.cpp
+++ b/gapir/cc/interpreter.cpp
@@ -105,14 +105,44 @@ void Interpreter::resetInstructions() {
   mInstructions = nullptr;
   mInstructionCount = 0;
   mCurrentInstruction = 0;
+  mJumpLabels.clear();
+}
+
+bool Interpreter::updateJumpTable(uint32_t jumpLabel) {
+  uint32_t instruction = 0;
+
+  if (mJumpLabels.rbegin() != mJumpLabels.rend()) {
+    instruction = mJumpLabels.rbegin()->second + 1;
+  }
+
+  for (; instruction < mInstructionCount; instruction++) {
+    auto opcode = mInstructions[instruction];
+
+    InstructionCode code =
+        static_cast<InstructionCode>(opcode >> OPCODE_BIT_SHIFT);
+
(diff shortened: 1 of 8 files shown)


gapir/cc/interpreter.cpp
```diff
@@ -107,12 +107,32 @@ void Interpreter::resetInstructions() {
   mCurrentInstruction = 0;
 }
 
+void Interpreter::buildJumpTable() {
```
AWoloszyn: I would prefer to see this done on-demand. (I.e. when we hit a jump label, add it to the jump table, if we hit a jump that is not in the jump table, then scan forward until we find the jump label).

That way for the (more common case) where we don't have a jump table, we don't have to parse every opcode first.

gapir/cc/interpreter.cpp
```diff
@@ -487,20 +509,45 @@ Interpreter::Result Interpreter::switchThread(uint32_t opcode) {
...
   auto should_jump = mStack.pop<int32_t>();
+
   if (!mStack.isEmpty()) {
     GAPID_WARNING("Error: stack is not empty before jumping to label %d",
                   jump_id);
     return ERROR;
   }
+
   if (should_jump != 0) {
+    if (mJumpLables.find(jump_id) == mJumpLables.end()) {
```
AWoloszyn: mJumpLabels

gapir/cc/interpreter.cpp
purvisa-at-google-com: Even if we have 100mb of opcodes, I'd expect this to take a few tens of miliseconds. Is the extra complexity worth it for a speed-up that is only theoretical?

purvisa-at-google-com: Done it anyway!

AWoloszyn: heh, thanks!
It practice it probably does not matter terribly much, but it just struck me as a future potentially irritating thing.

PR #3298, opened 2019-10-23
Updates to the timeline panel.

pmuetschard:
- Show the total trace time.
 - Fix rendering of the scale (and scale it to be 1:1).
 - Don't drop rendering of the ticks of the first segment.
 - Increase the top margin by 2px.


pmuetschard:
> The scale marker is visually the size of a minor division, but is labeled with the duration of major division.

Oops. Fixed.

PR #3302, opened 2019-10-24
Add Surfaces and Swapchains to Frame Looping

purvisa-at-google-com:


gapis/api/vulkan/frame_loop.go
```diff
@@ -1460,6 +1538,53 @@ func (f *frameLoop) resetBufferViews(ctx context.Context, stateBuilder *stateBui
...
+	// For every Surface that we need to create at the end of the loop...
+	for toCreate := range f.surfaceToCreate {
+		// Write the commands needed to recreate the destroyed object
+		surface := GetState(f.loopStartState).surfaces.Get(toCreate)
+		stateBuilder.createSurface(surface)
+	}
+
+	// The shadow state for Surfaces does not contain reference to the Swapchains they are used in. So we have to loop around finding the story.
+	for _, swapchainObject := range GetState(f.loopStartState).Swapchains().All() {
+		surface := swapchainObject.Surface()
```
RenfengLiu: Not sure if need to check if the surface is nilobject or not, have seen other objects that is nilobject and calling VulkanHandle() on it will cause crash.

PR #3303, opened 2019-10-24
Added a method of enabling Custom Perfetto Layers from GAPID.

AWoloszyn:
This means you can select which data sources you want,
and they will show up in the GAPID UI for replay, automatically
enabling any Vulkan layers as necessary.


gapis/trace/android/trace.go
```diff
@@ -476,6 +476,22 @@ func (t *androidTracer) SetupTrace(ctx context.Context, o *service.TraceOptions)
...
+			// For NativeBridge emulated devices opt for the native ABI of the emulator.
+			abi = d.NativeBridgeABI(ctx, abi)
+			layers := tracer.LayersFromOptions(ctx, o)
+
+			c, err := android.SetupLayers(ctx, d, pkg.Name, gapidapk.PackageName(abi), layers, true)
+			if err != nil {
+				return nil, cleanup.Invoke(ctx), log.Err(ctx, err, "Setting up the layer")
+			}
+			cleanup = cleanup.Then(c)
+		}
```
pmuetschard: This won't work, as we're overriding the layers again inside `perfetto/android/trace.go` (i.e. the line above you had to fix). This should be done in there instead of here.
Also, this does not ensure that `gapid.apk` is installed so the layers can be found.

gapis/trace/tracer/tracer.go
```diff
@@ -70,6 +71,20 @@ type Tracer interface {
...
+		return ret
+	}
+	for _, x := range o.PerfettoConfig.GetDataSources() {
+		if _, err := layout.LibraryFromLayerName(x.GetConfig().GetName()); err == nil {
+			ret = append(ret, x.GetConfig().GetName())
+		}
+	}
+	return ret
+}
+
```
pmuetschard: Yikes. I don't like this. It does expose that we have a bit of an issue with where knowledge exists about trace config vs env setup. For now, to keep with established pattern, let's have the UI create the config proto, but then add the list of layers to add as an option in the trace request. What do you think? We can then possibly find a better way to do this in the future, I'm even open to moving the entire Perfetto config construction into the server...

pmuetschard: Actually, after talking F2F, this is OK and we can keep it and key of the layers to enable based on the enabled data sources. This simplifies, for example, taking a trace via `gapit`, as it won't need to be updated every time a new layer is added.

cmd/gapit/BUILD.bazel
```diff
@@ -96,7 +96,7 @@ go_stripped_binary(
     name = "gapit",
     data = [
         "//cmd/gapis",
-        "//core/vulkan/vk_api_timing_layer/cc:libVkLayer_APITiming",
+        "//core/vulkan/vk_api_timing_layer/cc:libVkLayer_VulkanCPUTiming",
```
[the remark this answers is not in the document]
AWoloszyn: Done

core/app/layout/layout.go
```diff
@@ -100,14 +106,39 @@ func LibraryName(lib LibraryType, abi *device.ABI) string {
...
+	var dev device.OSKind
+	switch runtime.GOOS {
+	case "windows":
+		dev = device.Windows
+	case "linux":
+		dev = device.Linux
+	case "darwin":
+		dev = device.OSX
+	}
+	return dev
```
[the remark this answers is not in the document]
AWoloszyn: This was so we could get the host OS before we had actually resolved host.Instance().
That was to help break circular deps, but I no longer think it is  needed.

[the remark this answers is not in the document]
AWoloszyn: Nope, I lied, this does in fact remove the dependency on "device/host" from this package.

core/os/android/layers.go
```diff
@@ -43,9 +44,9 @@ func SupportsLayersViaSystemSettings(d Device) bool {
 // SetupLayer initializes d to use either a Vulkan or GLES layer from layerPkg
 // limited to the app with package appPkg using the system settings and returns
 // a cleanup to remove the layer settings.
-func SetupLayer(ctx context.Context, d Device, appPkg, layerPkg, layer string, vulkan bool) (app.Cleanup, error) {
+func SetupLayers(ctx context.Context, d Device, appPkg string, layerPkg string, layers []string, vulkan bool) (app.Cleanup, error) {
 	var cleanup app.Cleanup
-
+	log.E(ctx, "Setting up layers %+v", layers)
```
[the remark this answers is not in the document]
AWoloszyn: Removed

core/os/device/bind/simple.go
```diff
@@ -82,6 +84,50 @@ func (b *Simple) FileContents(ctx context.Context, path string) (string, error)
...
+	if err != nil {
+		return err
+	}
+	defer out.Close()
+
+	_, err = io.Copy(out, in)
+	if err != nil {
+		return err
+	}
+	return out.Close()
```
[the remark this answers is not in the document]
AWoloszyn: Done

core/os/device/device.proto
```diff
@@ -175,8 +175,13 @@ message Configuration {
   PerfettoCapability perfetto_capability = 5;
 }
 
+message VulkanProfilingLayers {
+  bool cpu_timing = 1;
+}
```
pmuetschard: I would just put this into the `GPUProfiling` message below...

core/os/device/deviceinfo/cc/android/query.cpp
```diff
@@ -40,14 +40,14 @@ typedef uint8_t GLubyte;
 
 namespace {
 
-device::DataTypeLayout* new_dt_layout(int size, int alignment) {
+device::DataTypeLayout *new_dt_layout(int size, int alignment) {
```
pmuetschard: I think we've been putting the stars next to the type, not the name.

gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
```diff
@@ -191,6 +191,21 @@ public static String getConfigSummary(Settings settings, Device.PerfettoCapabili
                   .addAllMeminfoCounters(Arrays.asList(MEM_COUNTERS));
     }
 
+    boolean large_buffer = false;
+    if (settings.perfettoVulkanCPUTiming) {
+      large_buffer = true;
+      config.addDataSourcesBuilder()
+        .getConfigBuilder()
+          .setName("CPUTiming");
```
pmuetschard: Should we use the full "VkLayer_CPUTiming" name, so that we don't risk name collisions?
Or maybe "vulkan.CPUTiming".

gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
```diff
@@ -26,6 +26,7 @@
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static java.util.stream.Collectors.toSet;
 
+import com.google.gapid.models.Devices.DeviceCaptureInfo;
```
pmuetschard: unused import?

gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
```diff
@@ -191,6 +191,21 @@ public static String getConfigSummary(Settings settings, Device.PerfettoCapabili
...
+          .setName("CPUTiming");
+    }
+
+    if (large_buffer) {
+      config.addBuffers(PerfettoConfig.TraceConfig.BufferConfig.newBuilder()
+              .setSizeKb(BUFFER_SIZE * 8));
+    } else {
+      config.addBuffers(PerfettoConfig.TraceConfig.BufferConfig.newBuilder()
+              .setSizeKb(BUFFER_SIZE));
+    }
```
pmuetschard:
```
config.addBuffers(PerfettoConfig.TraceConfig.BufferConfig.newBuilder()
    .setSizeKb((largeBuffer ? 8 : 1) * BUFFER_SIZE));
```

gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
```diff
@@ -191,6 +191,21 @@ public static String getConfigSummary(Settings settings, Device.PerfettoCapabili
                   .addAllMeminfoCounters(Arrays.asList(MEM_COUNTERS));
     }
 
+    boolean large_buffer = false;
```
pmuetschard: `boolean largeBuffer = false`

core/os/device/device.proto
AWoloszyn: I had wanted to keep this out, because the Vulkan profiling is complementary to the standard "GPU Profiling". It's going to be more than just GPU profiling as time goes on. (E.g. this layer actually does nothing GPU related)

core/os/device/deviceinfo/cc/android/query.cpp
AWoloszyn: Yep, I assumed clang-format/presubmit would have fixed that up for me. I will fix.

gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
AWoloszyn: We can do either, it was nice to have all of the `layer names == producer name`, but I'm not too strongly tied.

core/os/device/remotessh/device.go
```diff
@@ -316,6 +306,7 @@ func GetConnectedDevice(ctx context.Context, c Configuration) (Device, error) {
 	}
 
 	connection, err := ssh.Dial("tcp", fmt.Sprintf("%s:%d", c.Host, c.Port), sshConfig)
+
```
pmuetschard: this is a crazy nit, but I think if we're going to add an empty line, it should go right after the error handling if-block. That seems to be our pattern.

gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
```diff
@@ -50,6 +50,7 @@
 import org.eclipse.swt.widgets.TableColumn;
 import org.eclipse.swt.widgets.TableItem;
 
+import java.util.ArrayList;
```
pmuetschard: unused import, please remove.

gapis/perfetto/android/trace.go
```diff
@@ -59,8 +61,21 @@ func hasRenderStageEnabled(perfettoConfig *perfetto_pb.TraceConfig) bool {
 	return false
 }
 
-func setupRenderStagesEnvironment(ctx context.Context, d adb.Device, packageName string, perfettoConfig *perfetto_pb.TraceConfig) (app.Cleanup, error) {
+func setupVulkanLayers(ctx context.Context, d adb.Device, packageName string, abi *device.ABI, layers []string) (app.Cleanup, error) {
```
pmuetschard: I'm a bit confused by this function. Reading it first, I expected it to handle both cases - that is, be the one and only thing that will call `android.SetupLayers`.

However, I don't want to hold up the review due to this, as it does do the right thing.

gapis/perfetto/android/trace.go
```diff
@@ -95,10 +117,11 @@ func Start(ctx context.Context, d adb.Device, a *android.ActivityAction, opts *s
 		}.Bind(ctx)
 		// Before we start the activity, attempt to setup environment if gpu.renderstages is selected.
 		var err error
-		cleanup, err = setupRenderStagesEnvironment(ctx, d, a.Package.Name, opts.PerfettoConfig)
+		c, err := setupRenderStagesEnvironment(ctx, d, a.Package.Name, opts.PerfettoConfig, abi, layers)
 		if err != nil {
 			return nil, cleanup.Invoke(ctx), err
 		}
+		cleanup = cleanup.Then(c)
```
pmuetschard: Oops, the change to the cleanup actually made things worse. `cleanup` is nil throughout this block. Functions that return a cleanup and an error should only return a cleanup that is expected to be invoked in both success and error situations. Thus to fix your change you'd have to call `c.Invoke(ctx)`, but why create another local var?

core/os/device/deviceinfo/cc/query.cpp
```diff
@@ -32,7 +32,7 @@ inline bool isLittleEndian() {
 }
 
 template <typename T>
-device::DataTypeLayout* new_dt_layout() {
+device::DataTypeLayout *new_dt_layout() {
```
pmuetschard: dancing stars throughout this file.

gapis/perfetto/android/trace.go
AWoloszyn: Fixed

core/os/device/deviceinfo/cc/query.cpp
AWoloszyn: Fixed

gapis/perfetto/android/trace.go
```diff
@@ -76,26 +91,39 @@ func setupRenderStagesEnvironment(ctx context.Context, d adb.Device, packageName
...
+	}
 	if a != nil {
 		ctx = log.V{
 			"package":  a.Package.Name,
 			"activity": a.Activity,
 		}.Bind(ctx)
 		// Before we start the activity, attempt to setup environment if gpu.renderstages is selected.
 		var err error
-		cleanup, err = setupRenderStagesEnvironment(ctx, d, a.Package.Name, opts.PerfettoConfig)
+		cleanup, err := setupRenderStagesEnvironment(ctx, d, a.Package.Name, opts.PerfettoConfig, abi, layers)
```
pmuetschard: remove the `:`. With it, go creates the new `cleanup` var inside this scope and the outside one stays nil.

Issue #3295, opened 2019-10-23
Unhandled exception in the UI thread.


zhangyiwei:
GAPID Version: 1.7.0:developer
OS: mac os x x86_64

ToT build and fresh start. An error box popped up and showed below:

org.eclipse.swt.SWTException: Failed to execute runnable (java.lang.NullPointerException)
	at org.eclipse.swt.SWT.error(SWT.java:4595)
	at org.eclipse.swt.SWT.error(SWT.java:4510)
	at org.eclipse.swt.widgets.Synchronizer.runAsyncMessages(Synchronizer.java:188)
	at org.eclipse.swt.widgets.Display.runAsyncMessages(Display.java:4068)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3735)
	at org.eclipse.jface.window.Window.runEventLoop(Window.java:821)
	at org.eclipse.jface.window.Window.open(Window.java:797)
	at com.google.gapid.Main$UI.show(Main.java:120)
	at com.google.gapid.Main.main(Main.java:69)
Caused by: java.lang.NullPointerException
	at com.google.gapid.LoadingScreen.showOptions(LoadingScreen.java:115)
	at com.google.gapid.MainWindow.updateLoadingScreen(MainWindow.java:101)
	at com.google.gapid.Main$UI.uiStartup(Main.java:161)
	at com.google.gapid.Main$UI.lambda$scheduleIfStillOpen$5(Main.java:183)
	at com.google.gapid.widgets.Widgets.ifNotDisposed(Widgets.java:125)
	at com.google.gapid.widgets.Widgets.lambda$scheduleIfNotDisposed$1(Widgets.java:153)
	at com.google.gapid.widgets.Widgets.lambda$schedule$0(Widgets.java:145)
	at org.eclipse.swt.widgets.RunnableLock.run(RunnableLock.java:40)
	at org.eclipse.swt.widgets.Synchronizer.runAsyncMessages(Synchronizer.java:185)
	... 6 more

pmuetschard:
Fixed by #3304

PR #3304, opened 2019-10-24
Fix NPE in the recent-file popup.

stellama0208:
- The bug happens when launching GAPID for the first time, or when
 there's no recent files recorded in the $HOME/.gapic setting file.


chrisforbes:
Is this https://github.com/google/gapid/issues/3295 ?

pmuetschard:
> Is this #3295 ?

Indeed it is.

PR #3305, opened 2019-10-24
Handle image create/destroy during frame loop.

RenfengLiu:


gapis/api/vulkan/frame_loop.go
```diff
@@ -1470,6 +1470,17 @@ func (f *frameLoop) resetBufferViews(ctx context.Context, stateBuilder *stateBui
...
 func (f *frameLoop) resetImages(ctx context.Context, stateBuilder *stateBuilder) error {
 
+	for toDestroy := range f.imageToDestroy {
+		log.D(ctx, "Destroy image %v which was created during loop.", toDestroy)
+		image := GetState(f.loopEndState).Images().Get(toDestroy)
+		stateBuilder.write(stateBuilder.cb.VkDestroyImage(image.Device(), toDestroy, memory.Nullptr))
+
+		imageViewUsers := image.Views().All()
+		for imageView := range imageViewUsers {
+			f.imageViewToDestroy[imageView] = true
```
purvisa-at-google-com: Not needed here. You need to add all the *created* images' imageViews to both the imageViewToDestroy and imageViewToCreate maps below. I'll add a comment in the appropriate place.

gapis/api/vulkan/frame_loop.go
```diff
@@ -1479,6 +1490,13 @@ func (f *frameLoop) resetImages(ctx context.Context, stateBuilder *stateBuilder)
 	imgPrimer := newImagePrimer(stateBuilder)
 	defer imgPrimer.Free()
 
+	for toCreate := range f.imageToCreate {
+		log.D(ctx, "Recreate image %v which was destroyed during loop.", toCreate)
+		image := GetState(f.loopStartState).Images().Get(toCreate)
+		stateBuilder.createImage(image, f.loopStartState, toCreate)
+		// For image creation, the associated image views changes are handled in the restore step below.
+	}
+
```
purvisa-at-google-com: here. (see above)

purvisa-at-google-com: I saw you removed the code above, but you were supposed to move it down here! And also add the imageView to imageViewToCreate. Am I just reading the diff incorrectly?

purvisa-at-google-com: Sorry, my bad. I see it now!!!

gapis/api/vulkan/state_rebuilder.go
```diff
@@ -1498,20 +1501,24 @@ func (sb *stateBuilder) createImage(img ImageObjectʳ, imgPrimer *imagePrimer) {
 	} else {
 		// Otherwise, we have no sparse bindings, we are either non-sparse, or empty.
 		if !isDenseBound(img) {
-			return
+			return nil, fmt.Errorf("no sparse bindings found")
```
AWoloszyn: This is not an error IIRC. This means we are sparse bound, but have no bindings, this should be totally valid.

gapis/api/vulkan/state_rebuilder.go
```diff
@@ -1498,20 +1501,24 @@ func (sb *stateBuilder) createImage(img ImageObjectʳ, imgPrimer *imagePrimer) {
...
-		vkBindImageMemory(sb, img.Device(), img.VulkanHandle(),
+		planeMemInfo, _ := subGetImagePlaneMemoryInfo(sb.ctx, nil, api.CmdNoID, nil, srcState, GetState(srcState), 0, nil, nil, img, VkImageAspectFlagBits(0))
+		vkBindImageMemory(sb, img.Device(), vkImage,
 			planeMemInfo.BoundMemory().VulkanHandle(), planeMemInfo.BoundMemoryOffset())
 	}
 	// opaqueRanges should contain all the bound image subresources by now.
 	if len(opaqueRanges) == 0 {
 		// There is no valid data in this image at all
-		return
+		return nil, fmt.Errorf("no valid data in the image")
```
AWoloszyn: I think same as before, this might? not actually be an error condition. Please double-check the spec.

PR #3309, opened 2019-10-24
Hide the query tab, turn it into a menu command.

pmuetschard:


gapic/src/main/com/google/gapid/MainWindow.java
```diff
@@ -433,6 +433,7 @@ public MainView updateAndGet(Service.TraceType traceType) {
 
     ViewThumbnails("Show Filmstrip"),
     ViewDarkMode("Dark Mode", 'D'),
+    ViewQueryShell("Open &Query Shell", 'M'),
```
stellama0208: Just a quick mention that it's using the same accelerator key with the GotoMemory(...) menu item. If it is intended design I'm also fine with it, because memory is disabled in perfetto trace.

PR #3310, opened 2019-10-25
Fix the crash due to null image bindings in vkCopyDescriptorSet

yalcinmelihyasin:


yalcinmelihyasin:
b/143274544

gapis/api/vulkan/api/descriptor.api
```diff
@@ -661,26 +661,23 @@ cmd void vkUpdateDescriptorSets(
...
-        srcBinding.ImageBinding[c.SrcArrayIndex]
-        dstBinding.ImageBinding = imageBinding
 
         if srcBinding.ImageBinding[c.SrcArrayIndex] == null {
-          vkErrorInvalidSampler(null)
+           vkErrInvalidDescriptorArrayElement(as!u64(c.SrcSet), c.SrcBinding, c.SrcArrayIndex)
         } else {
+          imageBinding := dstBinding.ImageBinding
+          imageBinding[c.DstArrayIndex] =
+          srcBinding.ImageBinding[c.SrcArrayIndex]
```
pmuetschard: can you combine these two lines? i know they were separate before, but they really shouldn't be.... or at the very least indent the second one.

chrisforbes:
:+1: confirmed this fixes the issue and we can now trace that workload

gapis/api/vulkan/api/descriptor.api
yalcinmelihyasin: Surely.

AWoloszyn:
In chatting offline, we want to make sure this isn't just hiding a different issue related to state reconstruction.

yalcinmelihyasin:
I am merging this pr while still looking at the suspected underlying issue.

PR #3317, opened 2019-10-28
Look for dev-releases in new separate repo gapid-dev-releases

hevrard:
The dev-releases are now published on the new repo
https://github.com/google/gapid-dev-releases.

Also, refactor variable names from "prerelease" to "dev-release" for
the sake of consistent naming.


gapic/src/main/com/google/gapid/util/Messages.java
```diff
@@ -64,7 +64,7 @@
   public static final String CRASH_REPORTING_OPTION =
       "Help GAPID identify issues by sending crash reports to Google";
   public static final String UPDATE_CHECK_OPTION = "Automatically check for GAPID updates (please restart GAPID to force an update check)";
-  public static final String UPDATE_CHECK_PRERELEASE_OPTION = "Include developer preview releases";
+  public static final String UPDATE_CHECK_DEV_RELEASE_OPTION = "Include unstable preview dev-releases";
```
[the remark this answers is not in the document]
hevrard: Good point, I wanted to be extra-explicit, but your proposition sounds better and fits best the variable renaming.

gapis/server/server.go
```diff
@@ -113,29 +113,37 @@ func (s *server) GetServerInfo(ctx context.Context) (*service.ServerInfo, error)
...
+		githubRepo    = "gapid"
+		devGithubRepo = "gapid-dev-releases"
 	)
 	ctx = status.Start(ctx, "RPC CheckForUpdates")
 	defer status.Finish(ctx)
 	ctx = log.Enter(ctx, "CheckForUpdates")
-	client := github.NewClient(nil)
+
 	options := &github.ListOptions{}
+	client := github.NewClient(nil)
```
pmuetschard: why move the client?

gapis/server/server.go
```diff
@@ -113,29 +113,37 @@ func (s *server) GetServerInfo(ctx context.Context) (*service.ServerInfo, error)
...
 	releases, _, err := client.Repositories.ListReleases(ctx, githubOrg, githubRepo, options)
 	if err != nil {
 		return nil, log.Err(ctx, err, "Failed to list releases")
 	}
+
+	if includeDevReleases {
+		devClient := github.NewClient(nil)
```
pmuetschard: Is it not possible to use the same client?

gapis/server/server.go
hevrard: Fixed

gapis/server/server.go
hevrard: Yes indeed, fixed.

gapis/server/server.go
```diff
@@ -113,29 +113,36 @@ func (s *server) GetServerInfo(ctx context.Context) (*service.ServerInfo, error)
...
+
 	var mostRecent *service.Release
 	mostRecentVersion := app.Version
 	mostRecentDevVersion := app.Version.GetDevVersion()
 
 	for _, release := range releases {
-		if !includePrereleases && release.GetPrerelease() {
-			continue
-		}
-
```
pmuetschard: If you remove this filtering here, you're going to have to make the logic below smarter to filter dev releases when `includeDevReleases` is false.

hevrard: Sorry I forgot to say: the code now assumes all dev-releases will be on the new repo. So the `includeDevReleases` boolean now means: do list dev-releases from the new repo (whereas includePrerelease filtered on the GitHub notion of a pre-release).

Then the logic to select the most recent is oblivious to whether we want dev-releases or not.

I will take care of cleaning up the release page of the main repo before this change is merged in.

How does that sound?

pmuetschard: Yes, I understand that. However, we then need to do either
1) always filter pre-releases
2) add another boolean to indicate whether to include pre-releases (and set it to false in the UI)

With this change as is, pre-releases are treated as regular releases and would cause a notification. This is different behavior than before where we were either ignoring them or users had to explicitly opt-in to be notified of pre-releases.

PR #3324, opened 2019-10-29
Fix vkUpdateDescriptorSet copy operation crash 

yalcinmelihyasin:
Fix the crash due to copying non existent srcBinding during vkUpdateDescriptorSet.


yalcinmelihyasin:
This pr replaces previous workaround(#3310) with the actual fix.

PR #3325, opened 2019-10-29
Increase semaphore error warning level and fix the function signature

yalcinmelihyasin:


yalcinmelihyasin:
Due to it's a misuse of the api increased the warning level. Also fixed the function signature.

PR #3329, opened 2019-10-30
Fix VkCmdCopyImage when we are copying more than one aspect.

AWoloszyn:
This fixes some titles that copy depth+stencil images.


AWoloszyn:
See b/143601310

PR #3336, opened 2019-10-31
Add code to dump the image from virtual swapchain.

RenfengLiu:
Use environment variable to control  whether to dump the images from virtual swapchain to file. This is used for test the output of frameloop.

### DIFF
--- a/core/vulkan/vk_virtual_swapchain/cc/virtual_swapchain.cpp
+++ b/core/vulkan/vk_virtual_swapchain/cc/virtual_swapchain.cpp
@@ -17,12 +17,14 @@
 #include "virtual_swapchain.h"
 #include <cassert>
 #include <chrono>
+#include <cstring>
 #include <fstream>
 #include <functional>
 #include <iomanip>
 #include <iostream>
 #include <sstream>
 #include <string>
+#include <thread>
 
 namespace {
 
@@ -41,6 +43,15 @@ int32_t FindMemoryType(
 }
 
 void null_callback(void *, uint8_t *, size_t) {}
+
+struct ImageDumpInfo {
+  std::string image_dump_path;
+  uint32_t image_width;
+  uint32_t image_height;
+  VkFormat image_format;
+};
+
+constexpr char kImageDumpPathEnv[] = "IMAGE_DUMP_PATH";
 }  // namespace
 
 namespace swapchain {
@@ -63,6 +74,7 @@ VirtualSwapchain::VirtualSwapchain(
       device_(device),
       should_close_(false),
       callback_(null_callback),
+      callback_user_data_(nullptr),
       queue_(queue),
       functions_(functions),
       pending_image_timeout_in_milliseconds_(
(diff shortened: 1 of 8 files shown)


core/vulkan/vk_virtual_swapchain/cc/virtual_swapchain.cpp
```diff
@@ -267,6 +279,87 @@ VirtualSwapchain::VirtualSwapchain(
...
+              << std::endl;
+    static int frame_count = 1;
+    ImageDumpInfo *dump_info = new ImageDumpInfo{
+        std::string(image_dump_path),
+        width_,
+        height_,
+        _swapchain_info->imageFormat,
+    };
+
+    SetCallback(
```
AWoloszyn: What happens if the user wants to set a callback, but we have this env var set. that means we will end up overwriting it.

core/vulkan/vk_virtual_swapchain/cc/virtual_swapchain.cpp
```diff
@@ -292,6 +385,11 @@ void VirtualSwapchain::Destroy(const VkAllocationCallbacks *pAllocator) {
   }
 
   functions_->vkDestroyCommandPool(device_, command_pool_, pAllocator);
+
+  if (callback_user_data_ != nullptr) {
```
AWoloszyn: Same as above, if this comes from the user, we will be deleting a user object. That is very bad.

gapis/api/vulkan/frame_loop.go
```diff
@@ -478,9 +478,13 @@ func (f *frameLoop) Transform(ctx context.Context, cmdId api.CmdID, cmd api.Cmd,
 
 				// Add unconditional jump instruction to bring us back to the start of the loop if we made it past the conditional break.
 				stateBuilder.write(stateBuilder.cb.Custom(func(ctx context.Context, s *api.GlobalState, b *builder.Builder) error {
-					b.Push(value.S32(1))
```
purvisa-at-google-com: You plan on reverting the loop structure to before I fixed the last iteration rewind problem?

RenfengLiu:
sorry, I pushed my testing code to the run branch. I comments that out as it seems has problem for destroy the resources.

purvisa-at-google-com:
This might be related to a crash I have been seeing today. In theory, the old loop structure should be the one causing crashes, not the new one. Can you tell me about what you're seeing? Do you have a stack trace or an idea of which vkFunction that we're crashing in? What is the test app or trace?

RenfengLiu:
The sample I usde is from SaschaWillems/Vulkan, with slight modification at https://github.com/RenfengLiu/Vulkan/commit/e82531bc0bb6ebb20fc8991146e13cd8ef667155#diff-3268f9dd8f19a1993cdccc85d1775373R606, which will casue the gapir crash after looping. there's much log in gapir for the crash, so that's why i'm adding logs in the tmpl files .

purvisa-at-google-com:
I am working on a PR to add frame looping to replay export. You can then launch gapir directly on your host with a file representing the protos that GAPIS would have sent. If you launch gapir with verbose logging and GDB directly, you can learn a lot about where it broke. Will send you over repro steps soon!

RenfengLiu:
yes, that would be very helpful for debugging.

PR #3337, opened 2019-10-31
Add support for VK_KHR_device_group_creation

AWoloszyn:
This is preliminary support. We make all device groups contain a single device, we can expand on this later.


AWoloszyn:
Closing this was supersceded

PR #3338, opened 2019-10-31
Added support for VK_EXT_scalar_block_layout

AWoloszyn:
This also required updating our version of the vulkan api.


core/vulkan/cc/include/vulkan/vulkan_fuchsia.h
```diff
@@ -1,10 +1,6 @@
 #ifndef VULKAN_FUCHSIA_H_
 #define VULKAN_FUCHSIA_H_ 1
 
-#ifdef __cplusplus
-extern "C" {
-#endif
-
 /*
```
yalcinmelihyasin: Should not the license be at the top of the file or does it not matter?

AWoloszyn: This is pulled directly from the vulkan repo, I don't want to change it much.

AWoloszyn:
Looks like the failure is a flake.

PR #3341, opened 2019-11-01
Fix termination of observations of concurrent capture

hevrard:
This is a fixed version of 0b0ceed9900d9f6231de6b4162f0abecdcedf235,
which was reverted in 5effe3537faaa83492f1728098ff937914fd85a4.

This change adds the necessary checks for CallObserver of
"@threadsafe" commands to know if they have to keep on observing after
the actual driver call.

When capturing multithreaded applications, a thread may emit a command
which ends the capture, while other threads have threadsafe commands
in-flight. We add a CallObsever::reenter() function to check whether
observations should keep on happening when an observer re-enters after
a driver call.

A CallObserver caches whether it should trace or not in its
mShouldTrace field. This field must be set/updated when the
CallObserver enters/reenters. Any modification that impact whether a
call observer should trace or not should be guarded by the Spy
lock. Because of this:

 - the SpyBase::lock() argument of type CallObserver is removed as we
   do need to acquire this lock *before* the creation of the call
   observer, where mShouldTrace is set.

 - in Spy::onPostFrameBoundary(), set_suspended() is called *before*
   the spy lock is released in exit().

Google bug: b/141847511

### DIFF
--- a/gapii/cc/call_observer.cpp
+++ b/gapii/cc/call_observer.cpp
@@ -132,7 +132,7 @@ void CallObserver::encode(const ::google::protobuf::Message* cmd) {
   encoder()->object(cmd);
 }
 
-void CallObserver::reenter() {
+void CallObserver::resume() {
   if (!mShouldTrace) {
     // This observer was disabled from the start of the command, nothing to do.
     return;
--- a/gapii/cc/call_observer.cpp
+++ b/gapii/cc/call_observer.cpp
@@ -132,6 +132,21 @@ void CallObserver::encode(const ::google::protobuf::Message* cmd) {
   encoder()->object(cmd);
 }
 
+void CallObserver::reenter() {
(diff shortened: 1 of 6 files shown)


gapii/cc/call_observer.cpp
```diff
@@ -132,6 +132,21 @@ void CallObserver::encode(const ::google::protobuf::Message* cmd) {
   encoder()->object(cmd);
 }
 
+void CallObserver::reenter() {
```
purvisa-at-google-com: The name of this function made me think we were doing something with reentrancy in the observer. Sugguest changing to resumeObservation, finishObservation, or something similar.

hevrard: Good point, I renamed to `resume()` to stay homogeneous with other methods names of this class (enter(), exit()).

PR #3346, opened 2019-11-04
Add missing dependency to build instructions

hevrard:
Just tried to build on my reseted laptop, `zlib.h` was not
found. Adding the zlib development package fixes this.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -186,7 +186,7 @@ tools/bin/sdkmanager platform-tools
 
 ```
 sudo apt-get update
-sudo apt-get install mesa-common-dev libncurses5-dev libgl1-mesa-dev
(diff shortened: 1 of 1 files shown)


pmuetschard:
This does seem right. We shouldn't be depending on system libraries like that. We have our own version of zlib, so I'd rather we fix our code/build than to just depend on the system library.
We should revert this, and fix the compile error.

hevrard:
Agreed, sorry for that.

The compilation issue is:
```
In file included from external/com_google_protobuf/src/google/protobuf/io/gzip_stream.cc:38:
external/com_google_protobuf/src/google/protobuf/io/gzip_stream.h:49:10: fatal error: zlib.h: No such file or directory
 #include <zlib.h>
          ^~~~~~~~
compilation terminated.
```

pmuetschard:
yikes. It being in protobuf may make this hard to fix.

PR #3348, opened 2019-11-04
Make replaying an archive that does not exist error out

purvisa-at-google-com:
At the moment, if you replay an archive that does not exist, you get the message "Replay Complete".

This is a lie. The code here prints an error and exists.

### DIFF
--- a/cmd/gapir/cc/main.cpp
+++ b/cmd/gapir/cc/main.cpp
@@ -478,6 +478,8 @@ static int replayArchive(core::CrashHandler* crashHandler,
                          gapir::ReplayService* replayArchiveService) {
   std::shared_ptr<MemoryAllocator> allocator = createAllocator();
 
+  const std::string payloadName = "payload";
+
   // The directory consists an archive(resources.{index,data}) and payload.bin.
   MemoryManager memoryManager(allocator);
 
@@ -487,7 +489,11 @@ static int replayArchive(core::CrashHandler* crashHandler,
   std::unique_ptr<Context> context = Context::create(
       replayArchiveService, *crashHandler, resLoader.get(), &memoryManager);
 
-  if (context->initialize("payload")) {
+  if (replayArchiveService->getPayload(payloadName) == NULL) {
+    GAPID_ERROR("Replay payload could not be found.");
+  }
+
+  if (context->initialize(payloadName)) {
     GAPID_DEBUG("Replay context initialized successfully");
   } else {
     GAPID_ERROR("Replay context initialization failed");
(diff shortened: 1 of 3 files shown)


cmd/gapir/cc/main.cpp
```diff
@@ -487,6 +487,10 @@ static int replayArchive(core::CrashHandler* crashHandler,
   std::unique_ptr<Context> context = Context::create(
       replayArchiveService, *crashHandler, resLoader.get(), &memoryManager);
 
+  if (replayArchiveService->getPayload("payload") == NULL) {
```
hevrard: `"payload"` string should probably be stored as a constant as it is used in different places in this file.

AWoloszyn:
LGTM other than Hugues' comment.

cmd/gapir/cc/main.cpp
```diff
@@ -478,6 +478,8 @@ static int replayArchive(core::CrashHandler* crashHandler,
                          gapir::ReplayService* replayArchiveService) {
   std::shared_ptr<MemoryAllocator> allocator = createAllocator();
 
+  const std::string payload = "payload";
```
pau-baiget: I see you're trying to avoid avoid string literals throughout the code, however this name suggests this variable will actually contain a payload, not just the string "payload".

gapir/cc/archive_replay_service.cpp
```diff
@@ -25,8 +25,15 @@ namespace gapir {
 std::unique_ptr<ReplayService::Payload> ArchiveReplayService::getPayload(
     const std::string&) {
   std::fstream input(mFilePrefix, std::ios::in | std::ios::binary);
+  if (!input) {
```
pau-baiget: Why are you using NULL instead of nullptr? And in the "if" statement, why are you using "!" instead of "nullptr == "?

gapir/cc/replay_service.h
```diff
@@ -89,6 +89,8 @@ class ReplayService {
     Payload(std::unique_ptr<replay_service::Payload> protoPayload);
 
     ~Payload();
+
+    Payload() = delete;
```
pau-baiget: Is this part of the change?

purvisa-at-google-com: It's an oversight I noticed while looking at where the Payload comes from. I thought I'd correct it when I saw it.

gapir/cc/archive_replay_service.cpp
purvisa-at-google-com: "input" is an fstream, not a pointer. The ! operator is defined for std::fstreams to return if the stream is good or not,

cmd/gapir/cc/main.cpp
purvisa-at-google-com: Changed to payloadName

gapir/cc/archive_replay_service.cpp
purvisa-at-google-com: http://www.cplusplus.com/reference/ios/ios/operator_not/

PR #3350, opened 2019-11-04
Add download link on front page, mention dev-releases

hevrard:


about.md
```diff
@@ -8,6 +8,10 @@ GAPID is a developer tool for recording and inspecting calls made by an applicat
 
 It is open-source: [https://github.com/google/gapid](https://github.com/google/gapid)
 
+[Download a GAPID release](https://github.com/google/gapid/releases)
+
+Unstable developer releases are also available [here](https://github.com/google/gapid-dev-releases).
```
pau-baiget: minor nit: Do we need to explicitly state they are unstable?

hevrard: I think it's best to say unstable here.

We clearly announce dev-releases as unstable everywhere else, e.g. in https://github.com/google/gapid-dev-releases repo front page.

PR #3354, opened 2019-11-04
Wait the device to be idle when backup and reset during frame loop

RenfengLiu:


purvisa-at-google-com:
Pushed this in myself so I can get early access, hope you don't mind!

PR #3364, opened 2019-11-05
Add the application being traced to the atrace apps.

pmuetschard:


gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
```diff
@@ -167,6 +171,10 @@ public static String getConfigSummary(Settings settings, Device.PerfettoCapabili
       }
       if (settings.perfettoCpuSlices) {
         ftrace.addAllAtraceCategories(Arrays.asList(CPU_SLICES_ATRACE));
```
AWoloszyn: Can we only add the atrace stuff for Android devices?
We get a warning if you have them enabled on desktop, that would be nice to avoid.

pmuetschard: Yes that is a good point. I didn't know there was a warning and thought it would just be ignored. There is another issue, which will require the device info to communicate some atrace support information. Can I do it when fixing that?

pmuetschard:
Ping.

gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
ronaldfw: What's the plan here? We found that this change breaks ggp system profiling; see https://github.com/google/gapid/issues/3529.

PR #3367, opened 2019-11-05
Make vkbuffer can be copied

RenfengLiu:
Add VK_BUFFER_USAGE_TRANSFER_SRC_BIT to buffer usage when create
the buffer, this will make the buffer can be copied to be backed up.

### DIFF
--- a/gapis/api/vulkan/state_rebuilder.go
+++ b/gapis/api/vulkan/state_rebuilder.go
@@ -2880,7 +2880,7 @@ func (sb *stateBuilder) createSameBuffer(src BufferObjectʳ, buffer VkBuffer) er
 				pNext,                    // pNext
 				src.Info().CreateFlags(), // flags
 				src.Info().Size(),        // size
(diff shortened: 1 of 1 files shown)


gapis/api/vulkan/state_rebuilder.go
```diff
@@ -2880,7 +2880,7 @@ func (sb *stateBuilder) createSameBuffer(src BufferObjectʳ, buffer VkBuffer) er
 				pNext,                    // pNext
 				src.Info().CreateFlags(), // flags
 				src.Info().Size(),        // size
-				VkBufferUsageFlags(uint32(src.Info().Usage())|uint32(VkBufferUsageFlagBits_VK_BUFFER_USAGE_TRANSFER_DST_BIT)), // usage
+				VkBufferUsageFlags(uint32(src.Info().Usage())|uint32(VkBufferUsageFlagBits_VK_BUFFER_USAGE_TRANSFER_DST_BIT)|uint32(VkBufferUsageFlagBits_VK_BUFFER_USAGE_TRANSFER_SRC_BIT)), // usage
```
AWoloszyn: We cannot do this in the state_rebuilder. This will change the flags on **all** buffers. We are not allowed to change properties in the initial state block.

PR #3376, opened 2019-11-06
Print Jumps in Verbose logging

purvisa-at-google-com:


gapir/cc/interpreter.cpp
```diff
@@ -538,7 +538,12 @@ Interpreter::Result Interpreter::jumpNZ(uint32_t opcode) {
       GAPID_WARNING("Error: unknown jumpLabel %i", jump_id);
     }
 
-    mCurrentInstruction = mJumpLabels[jump_id];
+    GAPID_VERBOSE("JUMP TAKEN");
```
RenfengLiu: If we're going to add debug log here, maybe we can add more information to log jumped from which instruction to which instruction and is for which jump label

purvisa-at-google-com: It already prints that info in verbose mode

RenfengLiu: I see, the `DEBUG_OPCODE_26` printed out the the arguments.

PR #3378, opened 2019-11-06
Remove duplicated image layout transition barriers.

RenfengLiu:
Currently, if the image has both depth and stencil aspect, when
transition its layout, it will generate a duplicated image barrier.

### DIFF
--- a/gapis/api/vulkan/image_primer.go
+++ b/gapis/api/vulkan/image_primer.go
@@ -322,7 +322,7 @@ func ipImageLayoutTransitionBarriers(sb *stateBuilder, imgObj ImageObjectʳ, old
 			if newLayout == VkImageLayout_VK_IMAGE_LAYOUT_UNDEFINED || newLayout == VkImageLayout_VK_IMAGE_LAYOUT_PREINITIALIZED || oldLayout == newLayout {
 				return
 			}
-			barriers = append(barriers, ipImageSubresourceLayoutTransitionBarrier(
+			imageBarrier := ipImageSubresourceLayoutTransitionBarrier(
 				sb,
 				imgObj,
 				aspect,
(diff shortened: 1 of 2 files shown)


gapis/api/vulkan/image_primer.go
```diff
@@ -322,15 +322,21 @@ func ipImageLayoutTransitionBarriers(sb *stateBuilder, imgObj ImageObjectʳ, old
...
+			imageBarrier := ipImageSubresourceLayoutTransitionBarrier(
 				sb,
 				imgObj,
 				aspect,
 				layer,
 				level,
 				oldLayout,
 				newLayout,
-			))
+			)
```
chrisforbes: I don't think this is correct. walkImageSubresourceRange() is going to invoke this separately the depth and stencil aspects [stateBuilder::imageAspectFlagBits() doesn't know to keep them together].

You're going to end up producing a barrier with just the depth aspect, and a second barrier with just the stencil aspect. Neither is actually valid in unextended Vulkan 1.1 -- you must transition the two aspects together if the image has both depth and stencil.

RenfengLiu: Hi Chris,
Thanks very much for the review. 
The code is to remove duplicated image barrier. For image with depth and stencil and for the specific format, the `ipImageBarrierAspectFlags` will always return the aspect combined both depth and stencil. As you said, this is run on both aspects, so it returns a duplicated beerier.

https://github.com/google/gapid/blob/master/gapis/api/vulkan/image_primer.go#L369

chrisforbes: Oh, I see. This is quite tangled then :)

RenfengLiu:
Thanks Chris for helping review the code.

PR #3381, opened 2019-11-07
Add device flags to run GPU profiling validation on specific devices.

lpy:
BUG: b/138717619
TesT: bazel run gapit -- validate_gpu_profiling --serial SERIAL_NUMBER

### DIFF
--- a/cmd/gapit/flags.go
+++ b/cmd/gapit/flags.go
@@ -401,7 +401,7 @@ type (
 	}
 
 	ValidateGpuProfilingFlags struct {
+		DeviceFlags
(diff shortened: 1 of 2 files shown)


lpy:
We don't really need everything in `DeviceFlags`, besides the `Os` filed in `DeviceFlags` conflicts with the `OS` filed we have

lpy:
Comment addressed, ptal :)

AWoloszyn:
Thanks!

lpy:
I should have seen there's a `filterDevices` function that I can just use

PR #3384, opened 2019-11-07
New drawer widget to show the profiler selection.

pmuetschard:


gapic/src/main/com/google/gapid/widgets/DrawerComposite.java
```diff
@@ -0,0 +1,186 @@
...
+  private static final int BORDER_HEIGHT = 3;
+
+  private final int titleHeight;
+  private final Composite main;
+  private final Sash sash;
+  private final Label border;
+  private final Composite drawer;
+  private String title = "";
+  private boolean expanded = false;
+  private int drawerHeight = 150;
```
stellama0208: Just a UX feedback: Would it be good if we make the initial hint for height larger? Or having the Selection view scrollable vertically would be even better.
It's just that at this height some rows are hided "perfectly" based on the font size on my machine, if I am a user for the first time and I don't see some half-covered text, I may never know there's still some fields hiding underneath.

pmuetschard: You're right, this is not a good value and what I just used to get going... Clearly this isn't quite ready yet. I will make it default to a better value and then remember it between runs, since the user can adjust the height.

pmuetschard: Done.

PR #3404, opened 2019-11-14
Add api.Cmd.Terminated(), propagate it to UI

hevrard:
"terminated" indicated whether the api command has terminated, i.e. there was some post-fence observations for it.


hevrard:
This identifies whether a command has properly terminated or not.

Mutating a command that has not terminated results in applying pre-fence observation, and then aborting (`api.Abort()`) at the fence. `api.ForeachCmd` has an extra argument to filter out non-terminated commands, or not. The only places where we actually want to consider the non-terminated commands is when the UI CommandTree is generated.

In the CommandTree view, non-terminated commands are marked in the UI by stating that they did not terminate (alternative UI presentation is welcome).

I could not properly test this today, as I need a trace that leads to one or more non-terminated commands -- I don't have a reproduction case at hand, I will try next week.

AWoloszyn:
I specifically suggest you get(or create) a trace that has a VkQueueSubmit that fails half-way through.

AWoloszyn:
As long as this WAI with the test, I am good with this

hevrard:
This new method identifies whether a command has properly terminated or not.

Mutating a command that has not terminated results in applying pre-fence observation, and then aborting (api.Abort()) at the fence. An extra argument is added to api.ForeachCmd to filter out non-terminated commands, or not. The only place where we actually want to consider the non-terminated commands is when the UI CommandTree is generated.

In the CommandTree view, non-terminated commands are marked in the UI by stating that they did not terminate.

This was tested on a few apps, including one with heavy concurrency such that at the end of a capture some commands in other threads are non-terminated, and another where we forced a crash in a VkQueueSubmit by passing invalid data for arguments related to semaphores.

PR #3405, opened 2019-11-14
Add some link support for Vulkan objects

chrisforbes:
- VkDevice
- VkImage
- VkImageView
- VkCommandBuffer
- VkRenderPass


chrisforbes:
This now supports every Vulkan handle type except for VkPipeline (which is trouble due to the compute/graphics pipelines being split)

chrisforbes:
macos CI failure is unrelated.

PR #3408, opened 2019-11-15
Put filmstrip in a tab

chrisforbes:
In preparation for having a GPU slices view in the (default) top pane,
convert the filmstrip into a normal tab and get rid of the special fixed
splitter machinery.

TODO: alter the initial layout to place this sensibly; currently it's
appended to the tabs in the main pane, which is not useful.


chrisforbes:
WIP

pmuetschard:
FYI @chrisforbes, I've taken this change and done the "TODO". So don't spend any more time on it :).

chrisforbes:
OK great :)

PR #3409, opened 2019-11-15
Add a vulkan sample application into gapid.

AWoloszyn:
This will let us have an available application for testing.
This is currently only windows, TODO: expand this to the other platforms in question.


cmd/vulkan_sample/BUILD.bazel
```diff
@@ -0,0 +1,25 @@
+# Copyright (C) 2018 Google Inc.
```
hevrard: 2019

cmd/vulkan_sample/frag.h
```diff
@@ -0,0 +1,67 @@
+{0x07230203,0x00010000,0x000d0007,0x00000026,
```
hevrard: Missing copyright header

Also, I would add a small comment to say this is binary SPIR-V, generated from cmd/vulkan_sample/shader.frag (which I assume is the case)

cmd/vulkan_sample/shader.frag
```diff
@@ -0,0 +1,27 @@
+/* Copyright 2017 Google Inc.
```
hevrard: 2019

cmd/vulkan_sample/shader.vert
```diff
@@ -0,0 +1,43 @@
+/* Copyright 2017 Google Inc.
```
hevrard: 2019

cmd/vulkan_sample/vert.h
```diff
@@ -0,0 +1,111 @@
+{0x07230203,0x00010000,0x000d0007,0x0000003a,
```
hevrard: Add header comment: copyright and "SPIR-V generated from shader.vert"

cmd/vulkan_sample/main.cpp
```diff
@@ -0,0 +1,1771 @@
...
+#define REQUIRE_SUCCESS(fn) \
+  do { \
+    if (VK_SUCCESS != fn) { \
+      write_error(kOutHandle, "Vulkan Error: " #fn); \
+      return -1; \
+    } \
+  } while(0)
+
+#define LOAD_INSTANCE_FUNCTION(name, instance) \
+  PFN_##name name = reinterpret_cast<PFN_##name>(vkGetInstanceProcAddr(instance, #name));
```
hevrard: Style: put macro definitions outside of main() body ?

AWoloszyn:
Re: GLFW, I don't want to add another dependency into GAPID proper. It's already a mess of dependencies, and for a single .cpp file added, I would avoid more deps.

cmd/vulkan_sample/main.cpp
```diff
@@ -0,0 +1,1771 @@
...
+    break;
+  }
+  if (memory_index ==  properties.memoryTypeCount) {
+    return static_cast<uint32_t>(-1);
+  }
+  return memory_index;
+}
+
+
+int main(int argc, const char** argv) {
```
yalcinmelihyasin: I would prefer to split this up to functions but if it is ok for everyone else, I am good too.

lpy:
would prefer to have the main function split into functions like `createInstance`, `createSwapchain`, etc

Assuming Android one is coming in follow-up patch, I am fine with this patch

PR #3411, opened 2019-11-15
Remove the dead connection from the connections

yalcinmelihyasin:


yalcinmelihyasin:
b/139480352

gapis/replay/replay_connection.go
```diff
@@ -176,22 +176,37 @@ func (e *backgroundConnection) HandleResourceRequest(ctx context.Context, req *g
...
+// MakeBackgroundConnectionForTest creates a connection to the replay device that persists in the background for test.
+func MakeBackgroundConnectionForTest(ctx context.Context, device bind.Device, conn gapir.Connection, replayABI *device.ABI) (*backgroundConnection, error) {
+	// TODO: Refactor the test and remove this.
+	m := manager{connections: make(map[id.ID]*backgroundConnection)}
+	return m.makeBackgroundConnection(ctx, device, conn, replayABI)
+}
+
 // MakeBackgroundConnection creates a connection to the replay device that persists in the background.
-func MakeBackgroundConnection(ctx context.Context, device bind.Device, conn gapir.Connection, replayABI *device.ABI) (*backgroundConnection, error) {
+func (m *manager) makeBackgroundConnection(ctx context.Context, device bind.Device, conn gapir.Connection, replayABI *device.ABI) (*backgroundConnection, error) {
```
pmuetschard: also make it lower case in the comment

gapis/replay/replay_connection.go
```diff
@@ -176,22 +176,37 @@ func (e *backgroundConnection) HandleResourceRequest(ctx context.Context, req *g
...
 	bgc := &backgroundConnection{conn: conn, ABI: replayABI, OS: device.Instance().GetConfiguration().GetOS()}
+	m.connections[device.Instance().ID.ID()] = bgc
```
pmuetschard: why did you move this? I don't like that it's now so far away from where the mutex is locked.

yalcinmelihyasin: I thought that the logic of storing and removing connection will be split between two functions. Although I agree with the mutex comment.

pmuetschard: I understand that, but I think keeping the put with the mutex is more important than keeping the put with the delete.

PR #3412, opened 2019-11-15
add vkDeviceWaitIdle to WaitForPerfetto transform

apbodnar:
We should wait for GPU work to finish before starting or stopping a trace.  This makes the perfetto wait logic specifically geared towards Vulkan.  If we want to support GLES, I can pull stuff into a cross API interface instead.


apbodnar:
Nagging @AWoloszyn and @chrisforbes

PR #3415, opened 2019-11-18
Fix MacOS builds on Kokoro, after Kokoro update to Mojave

hevrard:
Kokoro experimentation, do not review.

### DIFF
--- a/kokoro/macos/package.sh
+++ b/kokoro/macos/package.sh
@@ -70,10 +70,8 @@ iconutil -c icns -o GAPID.app/Contents/Resources/GAPID.icns GAPID.iconset
 pip install --upgrade --user dmgbuild pyobjc-framework-Quartz
 cp "$SRC"/background*.png .
 cp "$SRC/dmg-settings.py" .
-# This path to dmgbuild (installed via pip --user) may change with Kokoro MacOS
-# updates. Value of 'site.USER_BASE' can give a hint of where to look, but
-# beware that default Python version may be different from the pip one.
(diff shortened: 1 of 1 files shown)


kokoro/macos/package.sh
```diff
@@ -70,7 +70,10 @@ iconutil -c icns -o GAPID.app/Contents/Resources/GAPID.icns GAPID.iconset
 pip install --upgrade --user dmgbuild pyobjc-framework-Quartz
 cp "$SRC"/background*.png .
 cp "$SRC/dmg-settings.py" .
-~/Library/Python/2.7/bin/dmgbuild -s dmg-settings.py GAPID gapid-$VERSION-macos.dmg
+# This path to dmgbuild (installed via pip --user) may change with Kokoro MacOS
+# updates. Value of 'site.USER_BASE' can give a hint of where to look, but
+# beware that default Python version may be different from the pip one.
+/Users/kbuilder/Library/Python/3.7/bin/dmgbuild -s dmg-settings.py GAPID gapid-$VERSION-macos.dmg
```
pmuetschard: use `~/Library...` so that the script at least has a chance of working outside of the buildbots.

hevrard: Done

hevrard:
> Does not look like @pmuetschard s comment has actually been addressed

I added commit 4951c1f which makes that change, can you see it? Commit message is "Apply review comments"

AWoloszyn:
Aha, it was not there when I was adding the comment.

PR #3416, opened 2019-11-18
Fix layer config for foreign desktop ABIs.

AWoloszyn:


core/app/layout/layout.go
```diff
@@ -329,8 +332,11 @@ func (l *runfilesLayout) Library(ctx context.Context, lib LibraryType, abi *devi
 	return file.Path{}, ErrCannotFindPackageFiles
 }
 
-func (l *runfilesLayout) Json(ctx context.Context, lib LibraryType) (file.Path, error) {
-	return l.find(libTypeToJsonPath[lib])
+func (l *runfilesLayout) Json(ctx context.Context, lib LibraryType, abi *device.ABI) (file.Path, error) {
+	if hostOS(ctx) == abi.OS {
```
pmuetschard: `abi == nil || ...` here, too?

core/app/layout/layout.go
```diff
@@ -462,8 +468,11 @@ func (l *ZipLayout) Library(ctx context.Context, lib LibraryType, abi *device.AB
 }
 
 // Json returns the path to the Vulkan layer JSON definition for the given library.
-func (l *ZipLayout) Json(ctx context.Context, lib LibraryType) (*zip.File, error) {
-	return l.file("lib/" + libTypeToJson[lib])
+func (l *ZipLayout) Json(ctx context.Context, lib LibraryType, abi *device.ABI) (*zip.File, error) {
+	if abi == nil || l.os == abi.OS {
+		return l.file("lib/" + libTypeToJson[lib])
+	}
+	return l.file(osToDir(abi.OS) + "lib/" + libTypeToJson[lib])
```
RenfengLiu: should we use path.Join here?

PR #3419, opened 2019-11-18
Change CPU slices' color at multi-selection.

stellama0208:
- Before this commit, we only change CPU slices' color at
 single-selection, support changing color at multi-selection on CPU
 slices as well, so that the UX behavior is more consistent.
 - Use HashMultimap to record state, so that both upid and utid lookup
 is fast.


stellama0208:
The above force push is only to resolve merge conflict.

PR #3420, opened 2019-11-19
Use pid file to avoid launching multiple producers.

lpy:
Previously we couldn't run multiple instances of gapid for system profiler
because of launch_producer launches multiple profiling producers. This patch
makes launch_producer writes its pid file and have the tool always kill the
existing one and run a new one.

Minor: Removed the unused path of profiling producer, we no longer want them to
exist in /assets folder.

BUG: b/143532393
Test: launch multiple instances of gapid

### DIFF
--- a/cmd/launch_producer/main.cpp
+++ b/cmd/launch_producer/main.cpp
@@ -16,9 +16,12 @@
 
 #include <android/log.h>
 #include <dlfcn.h>
+#include <fcntl.h>
 #include <stdio.h>
 #include <string.h>
 #include <sys/system_properties.h>
+#include <unistd.h>
+
 #include <string>
 
 #define _LOG(lvl, name, msg, ...)                          \
@@ -60,9 +63,9 @@ typedef void (*FN_PTR)(void);
 
 const char* kDriverProperty = "ro.gfx.driver.1";
 const char* kProducerPaths[] = {
-    "/assets/libgpudataproducer.so",
     "/lib/" ABI "/libgpudataproducer.so",
 };
+const char* kPidFileName = "/data/local/tmp/gapid_launch_producer.pid";
 
 std::string getDriverPackageOverride() {
   std::string driver;
@@ -159,8 +162,48 @@ FN_PTR loadLibrary(const std::string& path, const char* lib) {
   return startFunc;
 }
 
+// If a previous producer has died without cleaning up its pidfile,
+// here we kill a PID that may be related to another process.
+// This is a risk we take, it would be rare for a previous PID to be reused,
+// and in the worst case we kill a non-critical application as core services
+// are not killable that easily.
+void killExistingProcess() {
+  int fd = open(kPidFileName, O_RDONLY);
+  if (fd == -1) {
+    return;
+  }
+  char pidString[10];
+  if (read(fd, pidString, 10) > 0) {
+    int pid = -1;
+    sscanf(pidString, "%d", &pid);
+    if (pid > 0) {
+      kill(pid, SIGINT);
+    }
+  }
+  close(fd);
+}
+
+bool writeToPidFile() {
+  killExistingProcess();
+  int fd = open(kPidFileName, O_CREAT | O_WRONLY | O_TRUNC,
+                S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH);
+  if (fd == -1) {
+    return false;
+  }
+  pid_t pid = getpid();
+  char pidString[10];
+  sprintf(pidString, "%d", pid);
+  write(fd, pidString, strlen(pidString));
+  close(fd);
+  return true;
+}
+
 // Program to load the GPU Perfetto producer .so and call start().
 int main(int argc, char** argv) {
+  if (!writeToPidFile()) {
+    LOG_ERR("Could not open %s", kPidFileName);
+    std::abort();
+  }
   std::string driver = getDriver();
   std::string path = getApkPath(driver);
 
--- a/gapidapk/deviceinfo.go
+++ b/gapidapk/deviceinfo.go
@@ -25,7 +25,6 @@ import (
 	"time"
 
 	"github.com/golang/protobuf/proto"
-	"github.com/google/gapid/core/app"
 	"github.com/google/gapid/core/app/crash"
 	"github.com/google/gapid/core/event/task"
 	"github.com/google/gapid/core/log"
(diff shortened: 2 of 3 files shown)


lpy:
ptal

cmd/launch_producer/main.cpp
```diff
@@ -60,9 +62,9 @@ typedef void (*FN_PTR)(void);
 
 const char* kDriverProperty = "ro.gfx.driver.1";
 const char* kProducerPaths[] = {
-    "/assets/libgpudataproducer.so",
     "/lib/" ABI "/libgpudataproducer.so",
 };
+const char* kPidFileName = "/data/local/tmp/launch_producer.pid";
```
hevrard: As anyone can read/write to `/data/local/tmp`, I would prefer to prefix this name with gapid:
`/data/local/tmp/gapid_launch_producer`

gapidapk/deviceinfo.go
```diff
@@ -35,12 +36,13 @@ import (
...
-	startServiceAttempts     = 3
-	portListeningAttempts    = 5
-	perfettoProducerLauncher = "launch_producer"
+	sendDevInfoAction               = "com.google.android.gapid.action.SEND_DEV_INFO"
+	sendDevInfoService              = "com.google.android.gapid.DeviceInfoService"
+	sendDevInfoPort                 = "gapid-devinfo"
+	startServiceAttempts            = 3
+	portListeningAttempts           = 5
+	perfettoProducerLauncher        = "launch_producer"
+	perfettoProducerLauncherPidFile = "/data/local/tmp/launch_producer.pid"
```
hevrard: idem, `/data/local/tmp/gapid_launch_producer.pid`

gapidapk/deviceinfo.go
```diff
@@ -172,8 +174,16 @@ func preparePerfettoProducerLauncherFromApk(ctx context.Context, d adb.Device) (
...
 	}
 
-	// Finally, make sure the binary is executable
+	// Make sure the binary is executable
 	d.Shell("chmod", "a+x", launcherPath).Call(ctx)
+
+	// Finally, if there's an existing instance running, kill it.
+	if res, err := d.Shell("cat", perfettoProducerLauncherPidFile).Call(ctx); err == nil {
+		if _, err := strconv.Atoi(res); err == nil {
+			d.Shell("kill -2", res).Call(ctx)
```
hevrard: GAPID never deletes that file. What if the producer is stopped, then another process gets the same PID: we kill this other process?
Would there be a way, maybe using `ps`, to check that this PID is indeed a perfetto producer launcher?

lpy: hmmmmm, let me think about it. I think maybe we can check whether there are producers registered from querying perfetto. Will explore it

lpy:
There's a problem with `gapit` as well, will address it :(

Hence mark as WIP

gapidapk/deviceinfo.go
```diff
@@ -172,8 +174,18 @@ func preparePerfettoProducerLauncherFromApk(ctx context.Context, d adb.Device) (
...
+
+	// Finally, if there's an existing instance running, kill it.
+	if pid, err := d.Shell("cat", perfettoProducerLauncherPidFile).Call(ctx); err == nil {
+		if _, err := strconv.Atoi(pid); err == nil {
+			if res, err := d.Shell("perfetto", "--query").Call(ctx); err == nil && (strings.Contains(res, "gpu.counters") || strings.Contains(res, "gpu.renderstages")) {
+				d.Shell("kill -2", pid).Call(ctx)
+			}
+		}
+	}
+
```
pmuetschard: Instead of this, we should read the file in the producer and just kill the pid in the file, if found.

lpy: I add check to only kill process when producers already exist. It's easier for us to do it here I think

lpy:
ptal

gapidapk/deviceinfo.go
pmuetschard: i don't think we should do that check, I don't see any upside to the check.

lpy:
ptal

lpy:
Actually, this patch causes an issue that if latter gapid instance is closed, then existing gapid instances will lose the producer

gapidapk/deviceinfo.go
```diff
@@ -115,13 +117,16 @@ func fetchDeviceInfo(ctx context.Context, d adb.Device) error {
 		startSignal, startFunc := task.NewSignal()
 		startFunc = task.Once(startFunc)
 		crash.Go(func() {
-			err := launchPerfettoProducerFromApk(ctx, d, startFunc)
-			if err != nil {
-				log.E(ctx, "[launchPerfettoProducerFromApk] error: %v", err)
+			for true {
```
hevrard: I don't understand why you have a loop here?

gapidapk/deviceinfo.go
```diff
@@ -156,6 +161,14 @@ func fetchDeviceInfo(ctx context.Context, d adb.Device) error {
 }
 
 func preparePerfettoProducerLauncherFromApk(ctx context.Context, d adb.Device) (string, app.Cleanup, error) {
+	if res, err := d.Shell("perfetto", "--query").Call(ctx); err == nil && (strings.Contains(res, "gpu.counters") || strings.Contains(res, "gpu.renderstages")) {
+		if pid, err := d.Shell("cat", perfettoProducerLauncherPidFile).Call(ctx); err == nil {
+			if _, err := strconv.Atoi(pid); err == nil {
+				d.Shell("kill -2", pid).Call(ctx)
+			}
+		}
+		return "", nil, log.Errf(ctx, err, "A producer is already running.")
```
hevrard: Here, a producer was running, but we just killed it. Why is it not OK to continue with the rest of the function?

lpy: The previous approach causes an issue that if latter gapid instance is closed, then existing gapid instances will lose the producer because we always kill the previous existing producer.

So instead, I changed it to:

1. Check whether there are producers running;
2. If yes, abort existing producers and return, meaning not starting new one;
3. The previous gapid instance will get a nil return in this case, so have a for loop to restart producers.

hevrard: Does it mean that the `for true` loop is expected to loop until it actually hits an `err != nil`, at which point we are sure there is a producer running?

lpy:
briefly synced with @pmuetschard offline about this. Feedback is to make producers run in background so that we don't lose producers when the instance that launches producers dies.

Still figuring out how to do a `trap` in order to delete the file properly or an alternative way to delete the file

lpy:
Based on offline discussion, I use a name for the producer and run it in background.

cmd/launch_producer/main.cpp
```diff
@@ -159,8 +162,43 @@ FN_PTR loadLibrary(const std::string& path, const char* lib) {
...
+    sscanf(pidString, "%d", &pid);
+    if (pid > 0) {
+      kill(pid, SIGINT);
+    }
+  }
+  close(fd);
+}
+
+bool writeToPidFile() {
+  killExistingProcess();
```
hevrard: Add a comment along these lines:
```
If a previous producer has died without cleaning up its pidfile, here we kill a PID that may be related to
another process. This is a risk we take, it would be rare for a previous PID to be reused, and in the
worst case we kill a non-critical application (core services are not killable that easily).
```

PR #3421, opened 2019-11-19
Implement selection events in counter panel.

stellama0208:


stellama0208:
The above force push is only to resolve merge conflict.

PR #3424, opened 2019-11-19
Give gapidapk internet permissions needed for Adreno render stages producer.

apbodnar:


pmuetschard:
Please update your commit message to match Google style (see go/commit, for example).

PR #3428, opened 2019-11-19
Make vulkan_sample work on Android

hevrard:
Added as a new activity in the gapid APK

### DIFF
--- a/cmd/vulkan_sample/vulkan_sample.exports
+++ b/cmd/vulkan_sample/vulkan_sample.exports
@@ -1,0 +1,2 @@
+ANativeActivity_onCreate
+android_main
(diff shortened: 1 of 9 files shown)


cmd/vulkan_sample/vulkan_sample.exports
```diff
@@ -0,0 +1,2 @@
+ANativeActivity_onCreate
```
lpy: Looks like we don't have this entry point

hevrard: This is a special one defined by android_native_app_glue, which is provided by the NDK (see https://android.googlesource.com/platform/ndk.git/+/refs/heads/master/sources/android/native_app_glue/android_native_app_glue.c#424 )

hevrard:
Replay was broken because the code starting GAPIR assumed only a single activity in the APK. I gave the replayer activity a proper name, which is used to make sure we launch it (and not the VkCube activity) for doing replays.

PR #3429, opened 2019-11-20
Removed out-dated counters.

lpy:
BUG: b/138717619


lpy:
We no longer want to validate this counter, ptal

PR #3430, opened 2019-11-20
Keep Android Vulkan sample single threaded

hevrard:


cmd/vulkan_sample/BUILD.bazel
```diff
@@ -27,7 +27,7 @@ cc_library(
     linkopts = select({
         "//tools/build:linux": [
             "-ldl",
-            "-lpthread",
+            "-lpthread", # Needed to run on Swiftshader (Nov. 2019)
```
AWoloszyn: Thanks for the comment, I definitely should have added it.

PR #3431, opened 2019-11-20
Make android.Validate() use our own vulkan_sample

hevrard:


hevrard:
@lpy I think we'll need to update the expected values of counters accordingly.

PR #3432, opened 2019-11-20
Update Kokoro script to Bazel 1.2.0 and Android NDK r20b

johnplate:
Note: Bazel versions 0.26.1 until 1.1.0 are incompatible with mingw-gcc on Windows.


kokoro/windows/build.bat
```diff
@@ -24,10 +24,11 @@ set JAVA_HOME=c:\Program Files\Java\jdk1.8.0_144
 
 REM Install the Android SDK components and NDK.
 set ANDROID_HOME=%LOCALAPPDATA%\Android\Sdk
-echo y | %ANDROID_HOME%\tools\bin\sdkmanager build-tools;26.0.1 platforms;android-26
-wget -q https://dl.google.com/android/repository/android-ndk-r18b-windows-x86_64.zip
-unzip -q android-ndk-r18b-windows-x86_64.zip
-set ANDROID_NDK_HOME=%CD%\android-ndk-r18b
+copy /Y "%SRC%\kokoro\windows\android-sdk-license" "%ANDROID_HOME%\licenses\"
+call %ANDROID_HOME%\tools\bin\sdkmanager.bat platforms;android-26 build-tools;29.0.2
```
pmuetschard: This seems like it's not flexible - why not use the `echo y` anymore?

johnplate: echo y doesn't work with sdkmanager.bat, and I guess it never did. It didn't cause problems before because build-tools 26.0.0 were already installed.

johnplate: An alternative would be to use yes.exe from mingw if we change the installation order, but it was also triggering a warning that might cause problems in the future.

kokoro/windows/build.bat
```diff
@@ -24,10 +24,11 @@ set JAVA_HOME=c:\Program Files\Java\jdk1.8.0_144
 
 REM Install the Android SDK components and NDK.
 set ANDROID_HOME=%LOCALAPPDATA%\Android\Sdk
-echo y | %ANDROID_HOME%\tools\bin\sdkmanager build-tools;26.0.1 platforms;android-26
-wget -q https://dl.google.com/android/repository/android-ndk-r18b-windows-x86_64.zip
-unzip -q android-ndk-r18b-windows-x86_64.zip
-set ANDROID_NDK_HOME=%CD%\android-ndk-r18b
+copy /Y "%SRC%\kokoro\windows\android-sdk-license" "%ANDROID_HOME%\licenses\"
```
hevrard: Add a comment saying "This may break in the future"

johnplate: comment added

sallybelly: This may break in the future

sallybelly: > comment added

sallybelly:
Enough she's on to us. She knows ..

PR #3434, opened 2019-11-21
Clear color grouping after selecting a non-cpu selection.

stellama0208:
- Clear color grouping after selecting a non-cpu selection.
 - Allow de-selection by clicking empty area on panel group's title bar.

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/views/State.java
+++ b/gapic/src/main/com/google/gapid/perfetto/views/State.java
@@ -226,7 +226,6 @@ public class State {
   public boolean resetSelections() {
     boolean hasDeselection = selection != null || selectedThreads.size() > 0;
     setSelection((Selection.MultiSelection)null);
-    selectedThreads = HashMultimap.create();
     return hasDeselection;
   }
 
@@ -255,6 +254,10 @@ public class State {
   public void setSelection(Selection.MultiSelection selection) {
     lastSelectionUpdateId.incrementAndGet();
     this.selection = selection;
+    // If selection is cleared or set to a non-cpu one, don't do color grouping for cpu slices.
+    if (selection == null || selection.getSelection(Selection.Kind.Cpu) == Selection.EMPTY_SELECTION) {
+      clearSelectedThreads();
+    }
     listeners.fire().onSelectionChanged(selection);
   }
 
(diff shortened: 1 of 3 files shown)


gapic/src/main/com/google/gapid/perfetto/views/State.java
```diff
@@ -255,6 +254,10 @@ public void setSelection(ListenableFuture<Selection.MultiSelection> futureSel) {
   public void setSelection(Selection.MultiSelection selection) {
     lastSelectionUpdateId.incrementAndGet();
     this.selection = selection;
+    // If selection is cleared or set to a non-cpu one, don't do color grouping for cpu slices.
+    if (selection == null || selection.getSelection(Selection.Kind.Cpu) == Selection.EMPTY_SELECTION) {
```
pmuetschard: Instead of comparing against the empty selection, we should add a `isEmpty()` method.

gapic/src/main/com/google/gapid/perfetto/views/TrackContainer.java
```diff
@@ -326,7 +326,9 @@ public Dragger onDragStart(double x, double y, int mods) {
 
     @Override
     public Hover onMouseMove(Fonts.TextMeasurer m, double x, double y) {
-      if (y < TITLE_HEIGHT && (expanded || x < LABEL_WIDTH)) {
+      if (y < TITLE_HEIGHT && expanded && x >= LABEL_WIDTH) {
+        return Hover.NONE;
+      } else if (y < TITLE_HEIGHT && (expanded || x < LABEL_WIDTH)) {
```
pmuetschard: Oops, no that doesn't quite work. If `expanded` is `true`, the title may be longer than `LABEL_WIDTH`.

gapic/src/main/com/google/gapid/perfetto/views/State.java
stellama0208: Done in the new commit.

gapic/src/main/com/google/gapid/perfetto/views/TrackContainer.java
stellama0208: Remove this part in the new commit, will handle the deselection at empty title area after #3523 is checked in.

stellama0208:
Hi Pascal, a change is made, please take another look, thanks! Will make this PR only focus on solving color change problem for now. After #3523 get checked in, will look at how to handle title empty area de-selection again.

PR #3435, opened 2019-11-21
Chore: Remove ANSI color codes from perfetto logs

apbodnar:
Improves log readability by removing incorrectly rendered ANSI color codes from perfetto stdout/stderr

### DIFF
--- a/core/os/android/adb/perfetto.go
+++ b/core/os/android/adb/perfetto.go
@@ -33,7 +33,7 @@ import (
 )
 
 var (
-	ansiRegex = regexp.MustCompile("\u001b\\[\\d+m")
+	AnsiRegex = regexp.MustCompile("\u001b\\[\\d+m")
 )
 
 // StartPerfettoTrace starts a perfetto trace on this device.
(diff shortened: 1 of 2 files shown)


core/os/android/adb/perfetto.go
```diff
@@ -33,7 +33,7 @@ import (
 )
 
 var (
-	ansiRegex = regexp.MustCompile("\u001b\\[\\d+m")
+	AnsiRegex = regexp.MustCompile("\u001b\\[\\d+m")
```
AWoloszyn: Have we considered filing a bug with perfetto to remove this via a command-line option (or better yet they should be removing them automatically if they detect stdout is not going to shell)

apbodnar: I'll file a bug. My assumption was that we'd still have to work around it for current android version(s).

apbodnar: Internal bug: b/145150184

pau-baiget:
If no other reviewers find issues with the current patch, please merge it :)

PR #3439, opened 2019-11-22
Update the list of GPU counters for validation.

lpy:
Based on the new Vulkan sample cube app, update the list of GPU counters for
device validation.

BUG: b/138717619


lpy:
ptal

PR #3440, opened 2019-11-22
Do not continue API logic after an error detection

hevrard:
Once an error has been detected, we should not try to keep
executing the rest of the subroutine.

### DIFF
--- a/gapis/api/vulkan/extensions/khr_descriptor_update_template.api
+++ b/gapis/api/vulkan/extensions/khr_descriptor_update_template.api
@@ -102,7 +102,6 @@ sub dense_map!(u32, VkDescriptorUpdateTemplateEntry) rewriteDescriptorUpdateEntr
   return ret_val.Map
 }
 
-
 @internal
 class DescriptorUpdateTemplateObject {
     @unused VkDevice                           Device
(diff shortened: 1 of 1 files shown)


gapis/api/vulkan/extensions/khr_descriptor_update_template.api
```diff
@@ -124,7 +123,10 @@ sub ref!DescriptorUpdateTemplateObject createDescriptorUpdateTemplate(
     if !(device in Devices) { vkErrorInvalidDevice(device) }
     if pCreateInfo == null { vkErrorNullPointer("pDescriptorUpdateTemplate") }
     info := pCreateInfo[0]
-    if !(info.descriptorSetLayout in DescriptorSetLayouts) { vkErrorInvalidDescriptorSetLayout(info.descriptorSetLayout) }
+    if !(info.descriptorSetLayout in DescriptorSetLayouts) {
+        vkErrorInvalidDescriptorSetLayout(info.descriptorSetLayout)
+        abort
```
hevrard: I am not sure how to handle the error here, as a return value is expected? Should we return nil, and if so how do we express this?

hevrard: @AWoloszyn a friendly ping :)

AWoloszyn: You can make a nil object, but it is a bit awkward.

```
class NilDescriptorUpdateTemplateObject {
    ref!DescriptorUpdateTemplateObject v;
}

....
....
obj := NilDescriptorUpdateTemplateObject().v
if (info.descriptorSetLayout in DescriptorSetLayouts) {
  obj = .....
}
return obj
```

hevrard: Thanks, I ended up opting for:
```
...
obj := new!DescriptorUpdateTemplateObject()
if (check error) {
  // report error
} else {
  obj.Device = device
  // etc...
}
return obj
```

gapis/api/vulkan/extensions/khr_descriptor_update_template.api
```diff
@@ -122,102 +121,110 @@ sub ref!DescriptorUpdateTemplateObject createDescriptorUpdateTemplate(
...
-        FromKHR: fromKHR,
-    )
-
-    obj.Entries = rewriteDescriptorUpdateEntries(
-        info.descriptorUpdateEntryCount,
-        info.pDescriptorUpdateEntries,
-        info.descriptorSetLayout)
-
+    obj := new!DescriptorUpdateTemplateObject()
+    if !(device in Devices) {
```
AWoloszyn: I would switch the device section back, so it matches every other command.

hevrard: Done

PR #3445, opened 2019-11-22
Add Perfetto resolvable to trigger replay profiling

apbodnar:
My understanding is that we'll ultimately have a different UI, but I've been using @pmuetschard 's patches as a foundation for testing the in-progress replay profiling work.  @chrisforbes suggested merging the UI patches to avoid having maintain a pile of local dev patches.

My changes integrate pmuetschard's UI logic with the Perfetto synchronization and can be used as a simple demo.


apbodnar:
@pmuetschard Is removing the UI logic and keeping the resolvable stuff acceptable?  Having to manually resolve merge conflicts every time I rebase is quite tedious.  @chrisforbes ?

AWoloszyn:
Separate from Pascal's comment. Do we want this to be a resolvable?
This means the results will be cached, maybe expected but maybe not.

chrisforbes:
I think we *dont* want this getting cached.

apbodnar:
Closing since we'd rather not cache the result. I've moved the necessary parts of this to https://github.com/google/gapid/pull/3490

PR #3447, opened 2019-11-22
Add Vulkan Handle Mappings to QueryTimestamp as an ad-hoc solution.

yalcinmelihyasin:


yalcinmelihyasin:
It is mostly there but I am leaving for a vacation for a week. So please feel free to update the return type with a proper object with mapping instead of string array such as VulkanHandleMapping{typename, traceValue, replayValue}

gapis/api/vulkan/query_timestamps.go
```diff
@@ -72,16 +73,19 @@ type queryTimestamps struct {
...
-		queryPools:   make(map[VkQueue]*queryPoolInfo),
-		willLoop:     willLoop,
-		handler:      handler,
-		results:      make(map[uint64]queryResults),
+		cmds:          Cmds,
+		commandPools:  make(map[commandPoolKey]VkCommandPool),
+		queryPools:    make(map[VkQueue]*queryPoolInfo),
+		willLoop:      willLoop,
+		handler:       handler,
+		results:       make(map[uint64]queryResults),
```
AWoloszyn: Bit confused between "results" and "ResultsToSend"

gapis/replay/mapping_exporter.go
```diff
@@ -0,0 +1,150 @@
...
+		} else {
+			size = uint64(typ.Size())
+		}
+		if size != 1 && size != 2 && size != 4 && size != 8 {
+			// Ignore objects that are not handles
+			continue
+		}
+		// Count the number of actual Posts we expect
+		total += 1
+		func(k interface{}, typ reflect.Type, size uint64) {
```
AWoloszyn: I know this is just moving code around, but we may have to be more intelligent here later, since this is a crazy number of posts. (posts will have a performance impact)

gapis/api/vulkan/replay.go
```diff
@@ -919,6 +920,10 @@ func (a API) Replay(
 		case timestampsRequest:
 			var numInitialCommands int
 			var err error
+			if exporter == nil {
```
AWoloszyn: The exporter should probably be optional here, depending on if the request asked for handle mappings.

gapis/service/service.proto
```diff
@@ -1299,6 +1299,13 @@ message GetTimestampsRequest {
 // is specified in a TimestampsItem message.
 message Timestamps {
```
AWoloszyn: The request should probably specify whether or not it wants handle mappings at all.
Also if we have a non-0 loop count, I think your handle mappings wont do the right thing, maybe a TBD, but if that is the case then we DEFINITELY should not make it default to on.

gapis/replay/mapping_exporter.go
```diff
@@ -0,0 +1,150 @@
...
+			b.Post(v, size, func(r binary.Reader, err error) {
+				defer func() {
+					total -= 1
```
apbodnar: Are there any concurrent access concerns here?  I'm assuming gapis is single threaded, but could the scheduling of grpc callbacks/goroutines cause a race here?

yalcinmelihyasin: This is called after Vulkan calls. They are sequential(as far as I know @AWoloszyn )

gapis/replay/mapping_exporter.go
yalcinmelihyasin: I agree

yalcinmelihyasin:
Created a new pr(#3600) instead of this.

PR #3449, opened 2019-11-23
Fix semaphore submit count: increment on vkAcquireNextImageKHR

hevrard:
vkAcquireNextImageKHR signals this semaphore, so its submit count must
be incremented accordingly.


hevrard:
Already fixed by #3400

PR #3454, opened 2019-11-25
Used the arena for allocations in the memory tracker.

### LINKED ISSUE #3426: Use the arena allocator in the memory tracker.
We currently use a std::map in the memory_tracker. This can be problematic when memory happens to be mapped to locations that are too close to the data we use in the memory tracker. 
If we swapped this to use our arena allocator, we would be fine, since the arena allocator allocates 2MB chunks, and this TYPICALLY will be page aligned in practice, meaning it wont share a page with other stuff.

AWoloszyn:
This will help trace applications that hit edge-cases
with memory locations. Swiftshader for example.

Fixes #3426

### DIFF
--- a/core/memory/arena/cc/stl_compatible_allocator.h
+++ b/core/memory/arena/cc/stl_compatible_allocator.h
@@ -1,0 +1,113 @@
+/* Copyright 2019 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


core/memory/arena/cc/stl_compatible_allocator.h
```diff
@@ -0,0 +1,113 @@
+/* Copyright 2017 Google Inc.
```
RenfengLiu: 2019

core/memory/arena/cc/stl_compatible_allocator.h
```diff
@@ -0,0 +1,113 @@
...
+
+  template <typename U>
+  StlCompatibleAllocator(const StlCompatibleAllocator<U>& other) {
+    arena_ = other.arena_;
+  }
+
+  // Copy constructs an object of type T at the location given by p.
+  void construct(pointer p, const_reference val) { new (p) T(val); }
+
+  // Constructs an object of Type U a the location given by P passing
```
RenfengLiu: at

PR #3460, opened 2019-11-27
Add a flag to enable pprof profiling.

AWoloszyn:
This will let you attach with go tool pprof to port :6060
in order to collect data.

### DIFF
--- a/core/app/profile.go
+++ b/core/app/profile.go
@@ -17,6 +17,8 @@ package app
 import (
 	"bufio"
 	"context"
+	"net/http"
+	_ "net/http/pprof"
 	"os"
 	"runtime"
 	"runtime/pprof"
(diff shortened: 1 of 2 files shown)


core/app/profile.go
```diff
@@ -70,6 +72,12 @@ func applyProfiler(ctx context.Context, flags *ProfileFlags) func() {
 			log.I(ctx, "Trace profile written")
 		})
 	}
+	if flags.Pprof {
+		log.I(ctx, "Enabling pprof on :6060")
+		go func() {
+			http.ListenAndServe("localhost:6060", nil)
```
RenfengLiu: is it possible that the cmd will be run with multiple instance

AWoloszyn: You technically COULD, but its unlikely.
EVen if you do it with gapit, you would have to 

gapit -profile-pprof -gapis-args -profile-pprof

Which is probably .... less useful.

RenfengLiu: As we're binding to port 6060 here, so that assume that it can be only with one instance, which is not like other commands such as  `gapit report`, you can run multiple of `gapit report` at the same time.

AWoloszyn: This isn't a verb though, this is a flag that can run to ANY/all applications. It would like:
gapit -profile-pprof report ......
You just cannot run 2 applications with the same arg at the same time. 
We could make gapit -profile-pprof 6060 report, but that is a lot more typing, since I don't expect users to use the hidden flags that way.

(remember these flags are hidden)

PR #3461, opened 2019-11-27
Allow apic to specify an independent template search path.

AWoloszyn:


cmd/apic/template.go
```diff
@@ -42,11 +43,12 @@ func init() {
...
-	Tracer     string        `help:"The template function trace expression"`
-	Gopath     string        `help:"the go path to use when looking up packages"`
-	GlobalList []string      `help:"A global value setting for the template"`
-	Search     file.PathList `help:"The set of paths to search for includes"`
+	Dir            string        `help:"The output directory"`
+	Tracer         string        `help:"The template function trace expression"`
+	Gopath         string        `help:"the go path to use when looking up packages"`
+	GlobalList     []string      `help:"A global value setting for the template"`
+	Search         file.PathList `help:"The set of paths to search for includes"`
+	TemplateSearch file.PathList `help:"Search path apic includes"`
```
hevrard: I found this help string a bit hard to parse, maybe mirror the previous one:
`The set of paths to search for templates`

cmd/apic/template.go
```diff
@@ -42,11 +43,12 @@ func init() {
...
-	Dir        string        `help:"The output directory"`
-	Tracer     string        `help:"The template function trace expression"`
-	Gopath     string        `help:"the go path to use when looking up packages"`
-	GlobalList []string      `help:"A global value setting for the template"`
-	Search     file.PathList `help:"The set of paths to search for includes"`
+	Dir            string        `help:"The output directory"`
+	Tracer         string        `help:"The template function trace expression"`
+	Gopath         string        `help:"the go path to use when looking up packages"`
+	GlobalList     []string      `help:"A global value setting for the template"`
+	Search         file.PathList `help:"The set of paths to search for includes"`
```
hevrard: rename variable `Search` to `IncludeSearch`, now that we also have a `TemplateSearch`

cmd/apic/template.go
```diff
@@ -74,7 +76,17 @@ func (v *templateVerb) Run(ctx context.Context, flags flag.FlagSet) error {
 	options := template.Options{
 		Dir:     v.Dir,
 		APIFile: api.Name(),
-		Loader:  ioutil.ReadFile,
+		Loader: func(filename string) ([]byte, error) {
```
hevrard: Style: I would be keen to reduce our use of lambdas, would it be OK to declare this function outside? (esp. here it doesn't seem to hoist any local vars.)

AWoloszyn: It hoists v, but I can make it first-class function. That being said, for short functions, I tend to prefer lambdas if they are used only once. Otherwise you have to spend more time hunting through code.

cmd/apic/template.go
AWoloszyn: I don't want to change IncludeSearch, as it breaks the public interface of this program. We have down-stream folks using this.

hevrard: Makes sense. Maybe document the fact that this interface is public, and should not be changed lightheartedly?

cmd/apic/template.go
hevrard: That is a matter of taste, but I tend to not like lambdas because they appear unnamed in stack traces :)
I let you make the decision.

PR #3462, opened 2019-11-28
Shorten Enums and Bitfields in UI

hliatis:


gapis/api/service.proto
```diff
@@ -289,6 +289,7 @@ message EnumValue {
 message BitfieldValue {
   repeated uint64 set_bits = 1;
   repeated string set_bitnames = 2;
+  bool is_color = 3;
```
pmuetschard: Instead of `is_color`, we should call this something more generic. Maybe `short`?

hliatis: What other instances would this exception cover?

pmuetschard: I don't have a specific case in mind, but I could imagine an XYZW, enum, for example.

pmuetschard:
Please change this PR to provide both the long and short version in the API. The long version is what we have now, so instead of changing it to the short version, just keep both in the proto. You can either change the UI to use the new short field or have the server provide the long version in a different field and use the current field for the short version.

The tooltip and copy paste stuff can be done later.

PR #3464, opened 2019-11-28
Reduce desktop window size

hevrard:
Having a small window size can help when running on top of swiftshader on CI bots.

### DIFF
--- a/cmd/vulkan_sample/main.cpp
+++ b/cmd/vulkan_sample/main.cpp
@@ -17,6 +17,7 @@
 #include <math.h>
 #include <chrono>
 #include <cstdint>
+#include <cstdlib>
 #include <cstring>
 #include <iostream>
 #include <string>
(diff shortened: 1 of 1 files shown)


cmd/vulkan_sample/main.cpp
```diff
@@ -377,12 +377,34 @@ uint32_t inline GetMemoryIndex(
...
+  for (int i = 1; i < argc; i++) {
+    std::string arg(argv[i]);
+    if (arg.compare(0, 3, "-h=") == 0) {
+      height = std::stoi(&(arg[3]));
+    } else if (arg.compare(0, 3, "-w=") == 0) {
+      width = std::stoi(&(arg[3]));
+    } else {
+      std::cout << "Unrecognized argument: " << arg << '\n';
+      usage();
+      std::exit(EXIT_FAILURE);
```
AWoloszyn: To be consistent with the rest of the application, return -1

hevrard: Done

PR #3473, opened 2019-11-28
Update Perfetto version in GAPID

zakerinasab:
This change updates the Perfetto version in GAPID to the latest
version.


pmuetschard:
Since this was committed without any other changes to our code, could you please comment here why we needed to update Perfetto? In the future, having the commit message say why we're updating is helpful.

PR #3478, opened 2019-12-02
Update vulkan memory tracking layer to allow selecting categories

zakerinasab:
This change updates the GAPID profiler UI such that device and/or
driver memry tracking can be selected for Vulkan memory tracking.

### DIFF
--- a/core/vulkan/perfetto_producer/perfetto_threadlocal_emitter.h
+++ b/core/vulkan/perfetto_producer/perfetto_threadlocal_emitter.h
@@ -51,6 +51,10 @@ class ThreadlocalEmitter : ThreadlocalEmitterBase {
       const typename perfetto::DataSourceBase::SetupArgs&) override;
   void StopTracing() override { enabled_ = false; }
   bool Enabled() { return enabled_; }
+  bool CategoryEnabled(const char* category) {
+    return enabled_categories_.empty() ||
+           enabled_categories_.contains(category);
+  }
   void StartEvent(const char* catagory, const char* name);
   void EndEvent(const char* category);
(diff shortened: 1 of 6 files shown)


gapis/trace/android/trace.go
```diff
@@ -634,7 +634,7 @@ func (t *androidTracer) SetupTrace(ctx context.Context, o *service.TraceOptions)
 			}
 			// For NativeBridge emulated devices opt for the native ABI of the emulator.
 			packageABI = d.NativeBridgeABI(ctx, abi)
-			layers = tracer.LayersFromOptions(ctx, o)
+			layers, _ = tracer.LayersFromOptions(ctx, o)
```
pmuetschard: You cannot ignore the env vars here.

core/vulkan/perfetto_producer/perfetto_threadlocal_emitter.h
```diff
@@ -51,6 +51,9 @@ class ThreadlocalEmitter : ThreadlocalEmitterBase {
       const typename perfetto::DataSourceBase::SetupArgs&) override;
   void StopTracing() override { enabled_ = false; }
   bool Enabled() { return enabled_; }
+  bool CategoryEnabled(const char* category) {
+    return enabled_categories_.contains(category);
```
AWoloszyn: My other PR also makes the assumption if enabled_categories_ is empty, then all categories are enabled.

return enabled_categories.empty() || ...

zakerinasab: Done.

PR #3484, opened 2019-12-03
Patch up Frame Looping transform so it works with EOS

purvisa-at-google-com:
The FrameLooping transform invalidated the api state for all transforms
following it. This is a shortcoming of the transform system which was
not designed to be aware of control flow. This change patches up the
Frame Looping transform so we can avoid the problem for now. I am going
to follow this commit with a design doc for a revised transform system
that will not have this problem (and require this patch).

### DIFF
--- a/gapis/api/vulkan/frame_loop.go
+++ b/gapis/api/vulkan/frame_loop.go
@@ -318,6 +318,12 @@ func newFrameLoop(ctx context.Context, graphicsCapture *capture.GraphicsCapture,
 
 func (f *frameLoop) Transform(ctx context.Context, cmdId api.CmdID, cmd api.Cmd, out transform.Writer) {
 
+	// If we're looping only once we can just passthrough commands
+	if f.loopCount == 1 {
+		out.MutateAndWrite(ctx, cmdId, cmd)
+		return
+	}
+
 	ctx = log.Enter(ctx, "FrameLoop Transform")
 	log.D(ctx, "FrameLoop: looping from %v to %v. Current CmdID/CmD = %v/%v", f.loopStartIdx, f.loopEndIdx, cmdId, cmd)
 	log.D(ctx, "f.loopTerminated = %v, f.lastObservedCommand = %v", f.loopTerminated, f.lastObservedCommand)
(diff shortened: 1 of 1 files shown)


gapis/api/vulkan/frame_loop.go
```diff
@@ -371,6 +377,22 @@ func (f *frameLoop) Transform(ctx context.Context, cmdId api.CmdID, cmd api.Cmd,
...
+				f.capturedLoopCmds = make([]api.Cmd, 0)
+				f.capturedLoopCmdIds = make([]api.CmdID, 0)
+				return
+			}
+
+			// Some things we're going to need for the next work...
+			apiState := GetState(out.State())
+			stateBuilder := apiState.newStateBuilder(ctx, newTransformerOutput(out))
+			defer stateBuilder.ta.Dispose()
+
```
hevrard: I think a small comment giving the big picture would be nice here, something like:

```
Frame looping is achieved with a payload structure like this:
1. <pre-loop>
2. loop body
3. state rewind
4. loop body
5. jump to 3
6. <post-loop>
This is to enable a smooth transition to <post-loop> when the conditional jump is not taken.
```

gapis/api/vulkan/frame_loop.go
```diff
@@ -318,6 +318,12 @@ func newFrameLoop(ctx context.Context, graphicsCapture *capture.GraphicsCapture,
 
 func (f *frameLoop) Transform(ctx context.Context, cmdId api.CmdID, cmd api.Cmd, out transform.Writer) {
 
+	// If we're looping only once we can just passthrough commands
+	if f.loopCount == 1 {
```
RenfengLiu: This check has been done in gapis/api/vulkan/replay.go:927. And if want to make sure, maybe we want to check `if f.loopCount <= 1 `

purvisa-at-google-com: There is code handling f.loopCount == 0 elsewhere in this file :)

As for having two checks, I think this is fine. The transform should behave intelligently with any representable value passed in as it's dangerous and irresponsible to do otherwise. But there is also no reason to create the transform in the chain if it is of no benefit at the point of instantiation.

RenfengLiu:
Have you success loop the trace from `bloom` that being captured from beginning? For me even with this patch the loop will crash

purvisa-at-google-com:
Works for me!

PR #3485, opened 2019-12-03
Update the changed descriptor set which are caused by the resetting of images and buffers

RenfengLiu:


gapis/api/vulkan/frame_loop.go
```diff
@@ -2485,6 +2525,8 @@ func (f *frameLoop) resetDescriptorSets(ctx context.Context, stateBuilder *state
 		stateBuilder.allocateDescriptorSets(descPoolObj, descSetHandles, descSetLayoutHandles)
 	}
 
+	// Update the map of changed descriptorset due to the buffer/image recreateion etc.
```
AWoloszyn: recreation*

PR #3486, opened 2019-12-03
Release resource only after they are not used(after vkQueueWaitIdle)

RenfengLiu:


gapis/api/vulkan/frame_loop.go
```diff
@@ -1492,7 +1492,16 @@ func (f *frameLoop) backupChangedImages(ctx context.Context, stateBuilder *state
 
 func (f *frameLoop) resetResources(ctx context.Context, stateBuilder *stateBuilder) error {
 
+	// TODO: remove those waitdeviceidle after we're sure it is safe to do so.
 	f.waitDeviceIdle(stateBuilder)
```
AWoloszyn: General comment about this, we will need to be more clever once we start to optimize this. (in the future) but vkDeviceWaitIdle is a big hammer, that will cause all work on the GPU to be flushed, creating a large bubble.

RenfengLiu: created an issue to tracking this: https://github.com/google/gapid/issues/3496

PR #3490, opened 2019-12-04
Add replay profile command to GAPIT

apbodnar:
This renames the current `profile` command to `coarse_profile`.  `profile` now takes a perfetto trace of a replay using a provided config.  It will use a simple render stages only config if no config is provided.

### DIFF
--- a/cmd/gapit/coarse_profile.go
+++ b/cmd/gapit/coarse_profile.go
@@ -1,0 +1,112 @@
+// Copyright (C) 2019 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"context"
+	"encoding/csv"
+	"flag"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/google/gapid/core/app"
+	"github.com/google/gapid/core/log"
+	"github.com/google/gapid/gapis/service"
+	"github.com/google/gapid/gapis/service/path"
+)
+
+type coarseProfileVerb struct{ GetTimestampsFlags }
+
+func init() {
+	verb := &coarseProfileVerb{GetTimestampsFlags{LoopCount: 1}}
+	app.AddVerb(&app.Verb{
+		Name:      "coarse_profile",
+		ShortHelp: "Profile a replay to get the time of executing the commands.",
+		Action:    verb,
+	})
+}
+
+func (verb *coarseProfileVerb) Run(ctx context.Context, flags flag.FlagSet) error {
+	if flags.NArg() != 1 {
+		app.Usage(ctx, "Exactly one gfx trace file expected, got %d", flags.NArg())
+		return nil
+	}
+	capture, err := filepath.Abs(flags.Arg(0))
+	if err != nil {
+		log.Errf(ctx, err, "Could not find capture file: %v", flags.Arg(0))
+	}
+
+	client, err := getGapis(ctx, verb.Gapis, verb.Gapir)
+	if err != nil {
+		return log.Err(ctx, err, "Failed to connect to the GAPIS server")
+	}
+	defer client.Close()
+
+	capturePath, err := client.LoadCapture(ctx, capture)
+	if err != nil {
+		return log.Err(ctx, err, "Failed to load the capture file")
+	}
+
+	device, err := getDevice(ctx, client, capturePath, verb.Gapir)
+	if err != nil {
+		return err
+	}
+
+	var out io.Writer = os.Stdout
+	if verb.Out != "" {
+		f, err := os.OpenFile(verb.Out, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)
+		if err != nil {
+			return log.Err(ctx, err, "Failed to open report output file")
+		}
+		defer f.Close()
+		out = f
+	}
+
+	reportWriter := csv.NewWriter(out)
+	defer reportWriter.Flush()
+
+	header := []string{"BeginCmd", "EndCmd", "Time(ns)"}
+	if err = reportWriter.Write(header); err != nil {
+		log.Err(ctx, err, "Failed to write header")
+	}
+
+	cmdToString := func(cmd *path.Command) string {
+		return strings.Trim(strings.Join(strings.Fields(fmt.Sprint(cmd.Indices)), "."), "[]")
+	}
+
+	req := &service.GetTimestampsRequest{
+		Capture:   capturePath,
+		Device:    device,
+		LoopCount: int32(verb.LoopCount),
+	}
+
+	client.GetTimestamps(ctx, req, func(r *service.GetTimestampsResponse) error {
+		if ts := r.GetTimestamps(); ts != nil {
+			for _, t := range ts.Timestamps {
+				begin := cmdToString(t.Begin)
+				end := cmdToString(t.End)
+				record := []string{begin, end, fmt.Sprint(t.TimeInNanoseconds)}
+				if err := reportWriter.Write(record); err != nil {
+					log.Err(ctx, err, "Failed to write record")
+				}
+			}
+		}
+		return nil
+	})
+	return nil
+}
--- a/cmd/gapit/profile.go
+++ b/cmd/gapit/profile.go
@@ -16,27 +16,24 @@ package main
 
 import (
 	"context"
-	"encoding/csv"
+	"encoding/json"
 	"flag"
 	"fmt"
-	"io"
 	"os"
 	"path/filepath"
-	"strings"
 
 	"github.com/google/gapid/core/app"
 	"github.com/google/gapid/core/log"
 	"github.com/google/gapid/gapis/service"
-	"github.com/google/gapid/gapis/service/path"
 )
 
-type profileVerb struct{ GetTimestampsFlags }
+type profileVerb struct{ GpuProfileFlags }
 
 func init() {
-	verb := &profileVerb{GetTimestampsFlags{LoopCount: 1}}
+	verb := &profileVerb{GpuProfileFlags{}}
 	app.AddVerb(&app.Verb{
 		Name:      "profile",
-		ShortHelp: "Profile a replay to get the time of executing the commands.",
+		ShortHelp: "Profile a replay to get time slices for gpu render stages.",
 		Action:    verb,
 	})
 }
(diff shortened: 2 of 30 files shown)


cmd/gapit/profile.go
```diff
@@ -16,31 +16,53 @@ package main
...
+	"github.com/golang/protobuf/proto"
 	"github.com/google/gapid/core/app"
 	"github.com/google/gapid/core/log"
 	"github.com/google/gapid/gapis/service"
-	"github.com/google/gapid/gapis/service/path"
 )
 
-type profileVerb struct{ GetTimestampsFlags }
+const (
+	bufferSizeKb                            = uint32(131072)
```
RenfengLiu: can we have some comments about those numbers?

apbodnar: Sure, I shamelessly took them from gapis/trace/android/trace.go  It's possible that we should use different values in any case.

gapis/replay/gpu_profile.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+// GpuProfile replays the trace and writes a Perfetto trace of the replay
+func GpuProfile(ctx context.Context, capturePath *path.Capture, device *path.Device, opts *service.TraceOptions) (*service.GpuProfileResponse, error) {
+	c, err := capture.ResolveGraphicsFromPath(ctx, capturePath)
+	if err != nil {
+		return nil, err
+	}
+
+	if device != nil {
```
RenfengLiu: if we're checking `device != nil ` then maybe we also need to deal with the `else` condition.

gapir/client/session.go
```diff
@@ -327,6 +328,8 @@ func (s *session) newADB(ctx context.Context, d adb.Device, abi *device.ABI, lau
 	}
 
 	log.I(ctx, "Launching GAPIR...")
+	// Configure GAPIR to be traceable for replay profiling
+	perfetto_android.SetupRenderStagesEnvironment(ctx, d, apk.Name, true, abi, []string{})
```
pmuetschard: Don't ignore the error or the cleanup here.

gapis/service/service.proto
```diff
@@ -1287,6 +1290,16 @@ message TraceTypeCapabilities {
...
+message GpuProfileRequest {
+  path.Capture capture = 1;
+  path.Device device = 2;
+  TraceOptions trace_options = 3;
+}
+
+message GpuProfileResponse {
+  string trace_save_path = 1;
+}
+
```
pmuetschard: This is not the API we've agreed upon and looks to me like a step in the wrong direction. If you need to save the trace to disk to debug things, add a `config.go` option that will dump it somewhere, similar to the "dump replay" stuff.

cmd/gapit/profile.go
```diff
@@ -16,27 +16,33 @@ package main
...
 	"flag"
-	"fmt"
-	"io"
-	"os"
 	"path/filepath"
-	"strings"
 
+	"github.com/google/gapid/core/os/file"
+	"github.com/google/gapid/gapis/config"
+	perfetto_pb "perfetto/config"
```
pmuetschard: This one should probably be in it's own group after all the gapid imports.

gapis/perfetto/android/trace.go
```diff
@@ -71,8 +71,8 @@ func setupVulkanLayers(ctx context.Context, d adb.Device, packageName string, ab
 	return cleanup, nil
 }
 
-func setupRenderStagesEnvironment(ctx context.Context, d adb.Device, packageName string, perfettoConfig *perfetto_pb.TraceConfig, abi *device.ABI, layers []string) (app.Cleanup, error) {
-	if !hasRenderStageEnabled(perfettoConfig) {
+func SetupRenderStagesEnvironment(ctx context.Context, d adb.Device, packageName string, hasRenderStages bool, abi *device.ABI, layers []string) (app.Cleanup, error) {
```
pmuetschard: Since you are making this function public, please rename it. It's a bit silly that it says "do X" and takes a boolean that controls whether "do do X"... Maybe `SetupProfileLayers`

Also, since it's now public, it needs a doc comment.

gapis/service/service.proto
```diff
@@ -1287,6 +1290,15 @@ message TraceTypeCapabilities {
   bool requires_application = 6;
 }
 
+message GpuProfileRequest {
+  path.Capture capture = 1;
+  path.Device device = 2;
+  TraceOptions trace_options = 3;
```
pmuetschard: Don't pass the trace options here. That is way too generic. I think the only thing we need here are the capture and the device.

Another reason why passing the perfetto config from the client doesn't work here: you'd have to setup all the layers (not just the GPU slices layer) when starting gapir.

I think all you have to do is move the code that creates the trace options from gapit into gapis.

AWoloszyn: We shouldn't have to do much for the layers, we could put "all" of the layers in the right place when starting GAPIR always, then we can turn them on/off at instance creation time I think.

pmuetschard: The system setting overrides the `VkInstanceCreateInfo`. I don't think you can "turn-off" any layers the system requests.

apbodnar:
@AWoloszyn @pmuetschard I addressed your feedback.  Instead of returning a local path to the trace, it now returns a (currently) empty `ProfilingData` proto with the option of dumping the trace via config.go

gapis/replay/gpu_profile.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/google/gapid/gapis/capture"
+	"github.com/google/gapid/gapis/service"
+	"github.com/google/gapid/gapis/service/path"
+
+	perfetto_pb "perfetto/config"
+)
+
+// Eyeball some generous trace config parameters
+const (
+	bufferSizeKb                            = uint32(262144)
```
purvisa-at-google-com: 256 * 1024 is clearer to read. The compiler will constant fold to 262144 for you.

gapis/service/service.proto
AWoloszyn: But we can **not** add these layers to the system setting, and just add them to the VkInstanceCreateInfo (in a transform).

gapir/client/session.go
```diff
@@ -327,6 +328,17 @@ func (s *session) newADB(ctx context.Context, d adb.Device, abi *device.ABI, lau
 	}
 
 	log.I(ctx, "Launching GAPIR...")
+	// Configure GAPIR to be traceable for replay profiling
+	cleanup, err := perfetto_android.SetupProfileLayers(ctx, d, apk.Name, true, abi, []string{})
```
AWoloszyn: This makes me nervous. This means no matter WHAT we are doing the profiling layers will be turned on in GAPIR. That seems like something we do not want. I do not trust that the profiling layers will be .. stable enough .. to handle all of the odd stuff we do in GAPID replay that ISN'T just "replay this frame".

cmd/gapit/coarse_profile.go
```diff
@@ -0,0 +1,112 @@
+// Copyright (C) 2018 Google Inc.
```
hevrard: copyright 2019

gapis/trace/android/adreno/profiling_data.go
```diff
@@ -0,0 +1,12 @@
+package adreno
```
hevrard: Missing copyright header

gapir/client/session.go
hevrard: Would the solution be to add the profiling layers using a transform like the FindIssue one which adds the validation layers?
See https://github.com/google/gapid/blob/1e1473feeb407daa5ddfa7491b0dc1032b70a8d9/gapis/api/vulkan/find_issues.go#L85

gapis/trace/android/trace.go
```diff
@@ -118,6 +118,25 @@ func (t *androidTracer) GetDevice() bind.Device {
...
 }
 
+func (t *androidTracer) ProcessProfilingData(ctx context.Context, buffer *bytes.Buffer) (*service.ProfilingData, error) {
+	// Load Perfetto trace and create trace processor.
+	rawData := make([]byte, buffer.Len())
+	_, err := buffer.Read(rawData)
+	if err != nil {
+		return nil, log.Err(ctx, err, "Failed to read trace buffer")
+	}
+	processor, err := perfetto.NewProcessor(ctx, rawData)
```
pmuetschard: This leaks the processor. It needs to be closed.

gapis/server/server.go
```diff
@@ -914,6 +914,13 @@ func (s *server) GetTimestamps(ctx context.Context, req *service.GetTimestampsRe
 	return replay.GetTimestamps(ctx, req.Capture, req.Device, req.LoopCount, h)
 }
 
+func (s *server) GpuProfile(ctx context.Context, req *service.GpuProfileRequest) (*service.GpuProfileResponse, error) {
```
pmuetschard: This should return `(service.ProfilingData, error)` instead.

pmuetschard: See https://github.com/pmuetschard/gapid/commit/e8909ff5093f12ff5d25912c55aa6883d0bad6e2

gapir/client/session.go
apbodnar: I've added an additional commit that does the layer loading explicitly. It still requires some config, but at least gapir will not load the stages layer when not in a replay.

gapis/api/vulkan/replay.go
```diff
@@ -781,13 +783,15 @@ type profileRequest struct {
...
 	handler      *replay.SignalHandler
+	buffer       *bytes.Buffer
 }
 
 func (a API) GetInitialPayload(ctx context.Context,
 	capture *path.Capture,
 	device *device.Instance,
 	out transform.Writer) error {
 	transforms := transform.Transforms{}
+	transforms.Add(&profilingLayers{})
```
pmuetschard: This transform should only be added if we're profiling **and** if the layer is available.

apbodnar: Oops, actually didn't mean to leave it in.  For now, replay profiling will not use split replays

gapis/api/vulkan/profiling_layers.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+	}()
+	mustAlloc := func(ctx context.Context, v ...interface{}) api.AllocResult {
+		res := s.AllocDataOrPanic(ctx, v...)
+		allocated = append(allocated, res)
+		return res
+	}
+
+	switch cmd := cmd.(type) {
+	case *VkCreateInstance:
```
pmuetschard: This will cause the replay to fail if the layer is not available. I would prefer if we either don't do anything if the layer is not available. At least add a TODO for now.

gapis/trace/android/trace.go
```diff
@@ -118,6 +118,26 @@ func (t *androidTracer) GetDevice() bind.Device {
...
+	defer processor.Close()
+	if err != nil {
+		return nil, log.Errf(ctx, err, "Failed to create trace processor")
+	}
+	gpu := t.b.Instance().GetConfiguration().GetHardware().GetGPU()
+	gpuName := gpu.GetName()
+	if strings.Contains(gpuName, "Adreno") {
+		return adreno.ProcessProfilingData(ctx, processor)
+	}
```
pmuetschard: I'm not certain this really needs to be GPU specific, but I'm OK to keep it in for now.

gapis/replay/gpu_profile.go
```diff
@@ -0,0 +1,88 @@
...
+
+// Eyeball some generous trace config parameters
+const (
+	bufferSizeKb                            = uint32(256 * 1024)
+	durationMs                              = 30000
+	gpuRenderStagesDataSourceDescriptorName = "gpu.renderstages"
+)
+
+// GpuProfile replays the trace and writes a Perfetto trace of the replay
+func GpuProfile(ctx context.Context, capturePath *path.Capture, device *path.Device) (*service.GpuProfileResponse, error) {
```
pmuetschard: This should return `(*service.ProfilingData, error)`. You are currently wrapping things, then unwrapping them in `server.go` just to wrap them again in `grpc.go` :) I know this is because of the change I've asked for before.

PR #3491, opened 2019-12-04
Add GPU render stages validation for Adreno.

lpy:
This patch adds the simple shape of gpu render stages validation for Adreno.
The gpu render stages validation on Adreno will succeed if:

1) there's at least one 'Surface' and/or one 'Render' slice, and
2) 'Surface' slices are top-level slices, and
3) 'Render' slices are nested slices.

Minor: Remove out-dated GPU counter.

BUG: b/138717619

### DIFF
--- a/gapis/trace/android/adreno/validate.go
+++ b/gapis/trace/android/adreno/validate.go
@@ -16,9 +16,19 @@ package adreno
 
 import (
 	"context"
+	"fmt"
 
+	"github.com/google/gapid/core/log"
 	"github.com/google/gapid/gapis/perfetto"
 	"github.com/google/gapid/gapis/trace/android/validate"
+)
+
+const (
+	renderStageSlicesQuery = "" +
+		"select name, depth, parent_stack_id " +
+		"from gpu_slice " +
+		"where track_id = %v " +
+		"order by slice_id"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


gapis/trace/android/trace.go
```diff
@@ -171,12 +171,13 @@ func (t *androidTracer) Validate(ctx context.Context) error {
 		doneSignal, doneFunc := task.NewSignal()
 
 		crash.Go(func() {
+			// TODO(b/142824856): This is a workaround to a problem that render stages data
```
hevrard: If this is not fixed soon, we might want to document it to the users?

lpy: The users don't need to know this I think.

gapis/trace/android/adreno/validate.go
```diff
@@ -34,17 +44,78 @@ var (
...
+		}
+		if !hasRenderSlice {
+			return log.Errf(ctx, err, "Render stage verification failed: No Render slice found")
+		}
+
+		// Skip slices until we hit the first 'Surface' slice.
+		skipNum := 0
+		for names[0] != "Surface" {
+			skipNum++
+		}
```
pmuetschard: won't this loop for ever if `names[0]` is something different?
You should combine this with the loop above.

gapis/trace/android/validate/validate.go
```diff
@@ -155,3 +158,15 @@ func ValidateGpuCounters(ctx context.Context, processor *perfetto.Processor, cou
 	}
 	return nil
 }
+
+func GetRenderStageTrackId(ctx context.Context, processor *perfetto.Processor) ([]int64, error) {
```
pmuetschard: `GetRenderStageTrackIds` and document it, since it's public.

lpy:
Thanks, ptal!

gapis/trace/android/adreno/validate.go
```diff
@@ -34,17 +44,77 @@ var (
...
+		for i, name := range names {
+			if name == "Surface" {
+				hasSurfaceSlice = true
+			}
+			if name == "Render" {
+				hasRenderSlice = true
+			}
+			if skipNum == -1 && name == "Surface" {
+				skipNum = i
+			}
```
pmuetschard: Just put this into the `if name == "Surface"` above?

gapis/trace/android/adreno/validate.go
```diff
@@ -34,17 +44,77 @@ var (
...
+			return log.Errf(ctx, err, "Render stage verification failed: No Surface slice found")
+		}
+		if !hasRenderSlice {
+			return log.Errf(ctx, err, "Render stage verification failed: No Render slice found")
+		}
+		names = names[skipNum:]
+		depths := columns[1].GetLongValues()[skipNum:]
+		parentStackId := columns[2].GetLongValues()[skipNum:]
+
+		for i := 0; i < len(names); i++ {
```
pmuetschard: maybe just `for i := skipNum; ...` then you don't need to slice things, but either is fine.

lpy:
Thanks for the feedback, I addressed the latest comments :)

PR #3499, opened 2019-12-05
Implement vulkan enum translation in memory struct view.

stellama0208:
- Visualize the enum fields in meaningful strings rather than
 abstract index numbers. The translation is based on enum mapping
 defined in gapis/api/<api>/api/enums.api file.


stellama0208:
The above force pushes were to solve presubmit failure, code ready for review, thanks!

stellama0208:
So far this PR only handles the single enum's translation, mark it as WIP for now.
Will update it with further implementation for bitfield enum's translation.

stellama0208:
This PR will only focus on the single vulkan enum translation, will do the bitfield enum translation in a separate PR in the future. 
PR ready for review again, thanks!

PR #3501, opened 2019-12-05
Remove stale file

hevrard:
gapir/cc/server_connection_test.cpp should have been removed when
GAPIR-GAPIS communication moved to gRPC, with commit
a32d98794e079dbf69be3f88d3d10b9613671124

b/145725163


hevrard:
@AWoloszyn this is a test for a module that was removed in May 2018, I think we can get rid of it?

AWoloszyn:
Ahh sorry about that, yes, this was the old gapir protocol. This shoudl be remove.d

PR #3502, opened 2019-12-05
GAPID profiler - add Vulkan memory usage counters to the UI

zakerinasab:
This change adds Vulkan memory usage counters to the profiler UI and also turns on driver memory usage tracking for now, until we merge https://github.com/google/gapid/pull/3478.

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/models/Tracks.java
+++ b/gapic/src/main/com/google/gapid/perfetto/models/Tracks.java
@@ -34,6 +34,7 @@ import com.google.gapid.perfetto.views.GpuQueuePanel;
 import com.google.gapid.perfetto.views.ProcessSummaryPanel;
 import com.google.gapid.perfetto.views.ThreadPanel;
 import com.google.gapid.perfetto.views.TitlePanel;
+import com.google.gapid.perfetto.views.VulkanCounterPanel;
 
 import java.util.Collections;
 import java.util.List;
@@ -188,6 +189,21 @@ public class Tracks {
         final int idleCount = threads.size() - firstIdle;
         data.tracks.addLabelGroup(summary.getId(), summary.getId() + "_idle", "Idle Threads",
             group(state -> new TitlePanel(idleCount + " Idle Threads (< 0.1%)"), false));
+      }
+
+      // For each process, add Vulkan memory usage counters if any exist
+      List<CounterInfo> counters = data.getCounters().values().stream()
+      .filter(c -> (c.count > 0) && (c.name.startsWith("vulkan")))
+      .collect(toList());
+
+      if (!counters.isEmpty()) {
+        data.tracks.addLabelGroup(summary.getId(), "vulkan_counters", "Vulkan Counters",
+            group(state -> new TitlePanel("Vulkan Memory Usage Counters"), true));
+        for (CounterInfo counter : counters) {
+          CounterTrack track = new CounterTrack(counter);
+          data.tracks.addTrack("vulkan_counters", track.getId(), counter.name,
+              single(state -> new VulkanCounterPanel(state, track), true));
+        }
       }
     });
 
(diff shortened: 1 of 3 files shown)


gapic/src/main/com/google/gapid/perfetto/models/CounterInfo.java
```diff
@@ -23,10 +23,11 @@
...
   private static final String LIST_SQL =
-      "select counter_id, name, ref, ref_type, description, count(value)," +
-      " min(value), max(value), avg(value) " +
-      "from counter_definitions left join counter_values using (counter_id) " +
-      "group by counter_id";
+      "select cd.counter_id, cd.name, cd.ref, cd.ref_type, cd.description, "
+      + "count(cv.value), min(cv.value), max(cv.value), avg(cv.value), p.pid "
+      + "from counter_definitions cd left join counter_values cv "
+      + "on (cd.counter_id = cv.counter_id) left join process p "
+      + "on (cd.ref = p.upid) group by cd.counter_id";
```
pmuetschard: You are not actually using the pid, so undo this change.

gapic/src/main/com/google/gapid/perfetto/models/CounterInfo.java
```diff
@@ -38,13 +39,29 @@
   public final double max;
   public final double avg;
 
+  private String GenerateCounterName(String name) {
```
pmuetschard: This is very specific to your counters. Let's not do it this way, but handle it like we are handling the system memory and battery counters. If, for now, you just want to show them as counters, simply expand the `CounterPanel` to take a name override and in `Tracks` look for counters with the specific names.

For one thing, it's impossible to tell from this what the actual labels will be.

gapic/src/main/com/google/gapid/perfetto/models/CounterInfo.java
```diff
@@ -38,13 +39,29 @@
...
+
   public CounterInfo(long id, String name, long ref, String refType, String description, long count,
       double min, double max, double avg) {
     this.id = id;
-    this.name = name;
+    this.name = GenerateCounterName(name);
     this.ref = ref;
     this.refType = refType;
-    this.description = description;
+    this.description = (description != null && description.length() > 0) ? description : name;
```
pmuetschard: The description is never null.
Use `!.empty()` instead of `.length() != 0`.

However, repeating the name for the description is not helpful, so please revert. The description is shown in a hover along with the name. So this, in essence, would make the hover card for these counters look like:
```
Foo
Foo
```
... which is not better than the current:
```
Foo
```

gapic/src/main/com/google/gapid/perfetto/models/CounterInfo.java
zakerinasab: Done.

gapic/src/main/com/google/gapid/perfetto/models/CounterInfo.java
zakerinasab: Done.

gapic/src/main/com/google/gapid/perfetto/models/CounterInfo.java
zakerinasab: Done.

gapic/src/main/com/google/gapid/perfetto/models/Tracks.java
```diff
@@ -189,6 +190,25 @@ private Tracks() {
...
       }
+
+      // For each process, add Vulkan memory usage counters if any exist
+      List<CounterInfo> counters =
+          data.getCounters()
+              .values()
+              .stream()
+              .filter(c
+                  -> (c.count > 0) && (c.name.startsWith("vulkan")))
+              .collect(toList());
```
pmuetschard: Please format this as:
```
List<CounterInfo> counters = data.getCounters().values().stream()
    .filter(c -> (c.count > 0) && (c.name.startsWith("vulkan")))
    .collect(toList());
```

zakerinasab: Done.

PR #3503, opened 2019-12-05
Change the crash folder to a writable location

yalcinmelihyasin:
The crash reports were trying to write to a read-only location. This pr changes it to a writable location.

### DIFF
--- a/core/cc/android/crash_handler.cpp
+++ b/core/cc/android/crash_handler.cpp
@@ -32,10 +32,10 @@ static bool handleCrash(const google_breakpad::MinidumpDescriptor& descriptor,
 
 namespace core {
 
-CrashHandler::CrashHandler()
+CrashHandler::CrashHandler(const std::string& crashDir)
     : mNextHandlerID(0),
       mExceptionHandler(new google_breakpad::ExceptionHandler(
-          google_breakpad::MinidumpDescriptor("./crashes"), NULL, ::handleCrash,
+          google_breakpad::MinidumpDescriptor(crashDir), NULL, ::handleCrash,
           reinterpret_cast<void*>(this), true, -1)) {
   registerHandler(defaultHandler);
 }
--- a/core/cc/crash_handler.h
+++ b/core/cc/crash_handler.h
@@ -22,6 +22,8 @@
 #include <string>
 #include <unordered_map>
 
+#include "core/cc/target.h"
+
 namespace google_breakpad {
 class ExceptionHandler;
 }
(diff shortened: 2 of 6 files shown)


core/cc/android/crash_handler.cpp
```diff
@@ -34,9 +38,24 @@ namespace core {
 
 CrashHandler::CrashHandler()
     : mNextHandlerID(0),
-      mExceptionHandler(new google_breakpad::ExceptionHandler(
-          google_breakpad::MinidumpDescriptor("./crashes"), NULL, ::handleCrash,
-          reinterpret_cast<void*>(this), true, -1)) {
+      mExceptionHandler(nullptr) {
+
+    std::string appInfo;
```
AWoloszyn: Is there not a standard temp directory that we could use?
std::mkdtmp?

pmuetschard: Since this is Android, just use `/data/local/tmp`

yalcinmelihyasin: Android no more gives permission to write under that folder as far as I can see.

yalcinmelihyasin: if accumulating the crash files a problem, we can delete the crash log after we upload it.

yalcinmelihyasin: Changed it to under cache folder.

core/cc/crash_handler.h
```diff
@@ -36,6 +36,7 @@ class CrashHandler {
   typedef std::function<void()> Unregister;
 
   CrashHandler();
```
AWoloszyn: If only Android is implementing this, we should #ifdef it.

core/cc/crash_handler.h
```diff
@@ -35,7 +35,12 @@ class CrashHandler {
       Handler;
   typedef std::function<void()> Unregister;
 
+#if TARGET_OS == GAPID_OS_ANDROID && TARGET_OS != GAPID_OS_LINUX
```
AWoloszyn: This statement is weird 
`TARGET_OS == GAPID_OS_ANDROID implies TARGET_OS != GAPID_OS_LINUX`
Since TARGET_OS == GAPID_OS_ANDROID.

yalcinmelihyasin: b/145808085

pmuetschard: That cannot be right. Your `#if` is `a == 1 && a != 2`. If `a == 1` it can never also `a == 2`.

yalcinmelihyasin: I don't know what is the reason, I have not inspect it but it does not compile on Linux(at least on my machine). I can take a look at it after vacation.

hevrard: @yalcinmelihyasin the compilation error was due to a missing `#include "core/cc/target.h"`, this header is the one that properly defines TARGET_OS. I just pushed the fix.

yalcinmelihyasin:
Please merge this.

hevrard:
Merged, thanks @yalcinmelihyasin for your work on this!

PR #3511, opened 2019-12-09
Setup prerelease driver when available.

lpy:
Always use prerelease driver when it's available.

BUG: b/145000150
Test: check adb shell dumpsys gpu

### DIFF
--- a/gapis/perfetto/android/trace.go
+++ b/gapis/perfetto/android/trace.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 	"sync/atomic"
 
 	perfetto_pb "protos/perfetto/config"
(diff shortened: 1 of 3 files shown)


gapis/perfetto/android/trace.go
```diff
@@ -105,6 +105,23 @@ func setupRenderStagesEnvironment(ctx context.Context, d adb.Device, packageName
...
+	if err != nil {
+		return nil, err
+	}
+	if driverPackageName == "" {
+		return nil, nil
+	}
+	// TODO(b/145893290) Check whether application has developer driver enabled once b/145893290 is fixed.
+	if err := d.SetSystemSetting(ctx, "global", "game_driver_prerelease_opt_in_apps", p.Name); err != nil {
+		return nil, log.Errf(ctx, err, "Failed to set up prerelease driver for app: %v.", p.Name)
+	}
```
pmuetschard: Shouldn't this be an "add to list"?

lpy: I think we should just overwrite the list like we do for layers?

hevrard: In the layer case, we probably want not to have any other externally-imposed layers other than our own Spy layer, and I guess that's why we overwrite the list.

Here, there may already be some apps listed in game_driver_prerelease_opt_in_apps. Maybe we should save it here and restore them in the cleanup lambda returned by setupPrereleaseDriver()?

Or just add the package to the list, in which case we should also check if it is already there or not?

AWoloszyn: For desktop, IIRC we don't overwrite the list, we append to it. This lets us insert additional layers (for example the validation layers) to make sure WE are doing the right thing

lpy:
Thanks for the feedback, changed to restore the original list after tracing ends. PTAL :)

PR #3513, opened 2019-12-09
Set write_info_file with big period with Vulkan CPU Timing

dpallotti:
Sets write_into_file and one-hour file_write_period_ms in the Perfetto
config when collecting Vulkan CPU Timing. This was suggested in
b/145341050#comment21 and makes traced write the trace file directly,
instead of sending the content of the big trace back to the perfetto
command over several IPC messages, making the finalization of the
trace at least one third faster.

Bug: b/145341050

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
+++ b/gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
@@ -229,17 +229,22 @@ public class TraceConfigDialog extends DialogBase {
     if (settings.perfettoVulkanCPUTiming) {
       largeBuffer = true;
       config.addDataSourcesBuilder()
-        .getConfigBuilder()
-          .setName("VulkanCPUTiming");
+          .getConfigBuilder()
+              .setName("VulkanCPUTiming");
+    }
+
+    config.addBuffers(PerfettoConfig.TraceConfig.BufferConfig.newBuilder()
+        .setSizeKb((largeBuffer ? 8 : 1) * BUFFER_SIZE));
+
+    if (largeBuffer) {
       config.setWriteIntoFile(true);
       config.setFileWritePeriodMs((int)HOURS.toMillis(1));
     }
 
-    config.addBuffers(PerfettoConfig.TraceConfig.BufferConfig.newBuilder()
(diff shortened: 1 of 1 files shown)


gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
```diff
@@ -230,6 +231,8 @@ public static String getConfigSummary(Settings settings, Device.PerfettoCapabili
       config.addDataSourcesBuilder()
         .getConfigBuilder()
           .setName("VulkanCPUTiming");
+      config.setWriteIntoFile(true);
+      config.setFileWritePeriodMs((int)HOURS.toMillis(1));
```
rschiu: Wouldn't the buffer gets filled frequently with such a large period?

AWoloszyn: So with largeBuffer = true, not likely.
Also we should do this any time largebuffer == true, and so should probably move this down into the block there.

rschiu: Correct me if I am misunderstanding something, but I have a few questions about this.
1. If we don't expect to fill the buffer, does setting this actually solve any problem?
2. largeBuffer = true sets the buffer size to 1GB.  Is that really what we want?  1GB seems quite a lot on a mobile device.
3. Looking at the traces I have, a 5 second trace takes about 15-30MB.  This means it would take about 3-6 mins to fill 1GB.  1 hour seem way too long.  Am I misunderstanding what writeIntoFile actually do?

rschiu: Sorry, I just read through the bug.  I understand what this is doing now.

dpallotti: So for the record, even if you read the comment:
- one hour is just a large period to prevent traced from writing to file while tracing, while still making it write to file directly thanks to write_into_file;
- with non-mobile games, traces can get to 50 MB/s with Vulkan CPU Timing.

dpallotti: It seems that in some cases these settings can trigger the memory watchdog for traced. I'm investigating.

Issue #3512, opened 2019-12-09
CPU tracks don't show correct CPU ids from Perfetto trace


ronaldfw:
This is most likely observable when taking traces on a mostly idle system.

GAPID takes the count of non-idle CPUs in a Perfetto trace and uses that to create the CPU tracks. I have an example trace where CPU with id 4 is idle and the CPUs in the UI show up as 1, 2, 3, 4, 5, 6, 7 (total CPUs is 8). Opening the same trace in the trace viewer on perfetto.dev will show the correct CPU ids 0, 1, 2, 3, 5, 6, 7. 

I suggest to use the actual CPU ids as recorded in the Perfetto traces in the UI to make sure this mapping is correct and for consistency with the Perfetto web UI. 

I have the code changes for this already done but please let me know if there is any disagreement with the plan here.

PR #3514, opened 2019-12-09
Fix CPU ids in CPU tracks

ronaldfw:
Fix issue https://github.com/google/gapid/issues/3512, which contains more information about this change.


pmuetschard:
This is subsumed by #3512

PR #3521, opened 2019-12-09
Handle completely idle CPU core/threads correctly.

### LINKED ISSUE #3512: CPU tracks don't show correct CPU ids from Perfetto trace

pmuetschard:
- Adds a new CpuInfo class to track data about the CPU cores/threads.
 - When determining the available CPU IDs, use both the sched and idle counter tables. Thus, if a completely idle CPU is present, it will show up as long as the idle counter is present.
 - Don't assume CPU ids are contiguous, i.e. handle absent CPU ids.
 - Use the new cpu_counter_track table to get rid of ref/ref_type.

Fixes #3512

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/views/CpuPanel.java
+++ b/gapic/src/main/com/google/gapid/perfetto/views/CpuPanel.java
@@ -67,7 +67,7 @@ public class CpuPanel extends TrackPanel<CpuPanel> implements Selectable {
 
   @Override
   public String getTitle() {
-    return "CPU " + (track.getCpu() + 1);
+    return "CPU " + (track.getCpu().id + 1);
   }
 
   @Override
@@ -97,7 +97,7 @@ public class CpuPanel extends TrackPanel<CpuPanel> implements Selectable {
   private void renderSummary(RenderContext ctx, CpuTrack.Data data, double w, double h) {
     long tStart = data.request.range.start;
     int start = Math.max(0, (int)((state.getVisibleTime().start - tStart) / data.bucketSize));
-    ctx.setBackgroundColor(StyleConstants.Palette.getColor(track.getCpu()));
+    ctx.setBackgroundColor(StyleConstants.Palette.getColor(track.getCpu().id));
     ctx.path(path -> {
       path.moveTo(0, h);
       double y = h, x = 0;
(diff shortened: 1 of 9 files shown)


gapic/src/main/com/google/gapid/perfetto/views/CpuPanel.java
```diff
@@ -67,7 +67,7 @@ public CpuPanel copy() {
 
   @Override
   public String getTitle() {
-    return "CPU " + (track.getCpu() + 1);
```
ronaldfw: Is there a specific reason why you want CPU ids to show up as their id + 1? I would prefer to show CPUs as their actual ids (typically starting at 0).

pmuetschard: The IDs are rather arbitrary and it makes more sense to start counting at 1 when displaying.

ronaldfw: Are they really arbitrary? I think they are the IDs in the ftrace data that is collected by traced_probes. Also, if you want consistency with perfetto.dev, the actual IDs should be used.

ronaldfw: At least on Linux, the CPU ids used directly map to the ids in the ftrace filesystem; see for example https://android.googlesource.com/platform/external/perfetto/+/refs/heads/master/src/traced/probes/ftrace/ftrace_procfs.cc

pmuetschard: Done.

PR #3524, opened 2019-12-10
Modify timing layer to emit debug marker event

rschiu:
The PR needs to be update once https://android-review.googlesource.com/c/platform/external/perfetto/+/1183295 lands.  But I would appreciate it if someone can take a look to make sure it's mostly fine.  In particular, I am not sure if this is the intended approach to add a new emit function in the in Emitter.

### DIFF
--- a/core/app/layout/layout.go
+++ b/core/app/layout/layout.go
@@ -91,6 +91,7 @@ var dataSourceNameToLayerName = map[string]string{
 	"VirtualSwapchain":    "VirtualSwapchain",
 	"VulkanCPUTiming":     "CPUTiming",
 	"VulkanMemoryTracker": "MemoryTracker",
+	"VulkanAPI":           "CPUTiming",
 }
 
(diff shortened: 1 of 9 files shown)


tools/build/workspace.bzl
```diff
@@ -158,6 +158,11 @@ def gapid_dependencies(android = True, mingw = True, locals = {}):
         build_file = "@gapid//tools/build/third_party:lss.BUILD",
     )
 
+    native.local_repository(
+        name = "perfetto",
+        path = "/usr/local/google/home/chiur/dev/perfetto"
```
AWoloszyn: Well this doesn't seem right,

rschiu: Right, I am waiting for some perfetto changes.  But I want to put this up for review because this is the first time I am modifying the layer and not sure if this is how things are expected to be done.

rschiu: Fixed, PTAL

PR #3527, opened 2019-12-10
Update the CPU timing layer to be able to allow selecting categories.

AWoloszyn:


gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
```diff
@@ -226,10 +226,47 @@ public static String getConfigSummary(Settings settings, Device.PerfettoCapabili
 
     boolean largeBuffer = false;
     if (settings.perfettoVulkanCPUTiming) {
+      String s = "";
```
pmuetschard: If you want to build up a string, use a `StringBuilder`. However, the code will be more readable if you build up a `List<String>` and then use `Joiner.on(":").join(...)`.

ronaldfw:
Happy to see this change, looks really useful!

core/vulkan/perfetto_producer/perfetto_data_source.inc
```diff
@@ -91,6 +92,16 @@ void PerfettoProducer<T>::OnStop(const typename perfetto::DataSourceBase::StopAr
     Get().OnStop(a);
 }
 
+template<typename T>
+void PerfettoProducerData<T>::OnSetup(const typename perfetto::DataSourceBase::SetupArgs& a) {
+    emitter_lock_.Lock();
+    started_ = true;
+    for (auto e : emitters_) {
```
purvisa-at-google-com: Guess it doesn't matter so much because of the map's type, but auto&& is safer here to avoid copying data

AWoloszyn: We probably dont want auto&&, since we really wouldn't want move semantics, auto& however is probably ok.

PR #3533, opened 2019-12-11
Revert "Implemented new GAPII-GAPIS protocol (#3493)"

AWoloszyn:
This reverts commit 8fc5aef2de328d800788899d1f8f7bcee7408ecf.

We can re-land once we have a fix for:
b/146054078


AWoloszyn:
It is 2-fold. One the message is never sent, 2 we simply don't stop tracing.

PR #3534, opened 2019-12-11
Use SHA256 instead of SHA1 for all the maven rules.

pmuetschard:
This takes care of all the bazel warnings associated with SHA1.


AWoloszyn:
That one LOOKS like it might be spurious, I will try and kick off another build

pmuetschard:
Yeah the MacOS failure was clearly unrelated and kicking the build again made it pass.

PR #3535, opened 2019-12-11
Update Eclipse project files for new bazel layout.

pmuetschard:


.project
```diff
@@ -23,7 +23,7 @@
 	</linkedResources>
 	<filteredResources>
 		<filter>
-			<id>1555091862383</id>
+			<id>1576009800852</id>
```
stellama0208: Just curious, where does this number come from?

pmuetschard: IDK... looks like a timestamp. These files are managed by eclipse. I.e. I made the change in the UI and it updated the files.

PR #3538, opened 2019-12-12
Make validate_gpu_profiling return non-zero on validation failure

hevrard:
This enables to use `gapit validate_gpu_profiling` return code to detect failing validation in automated testing.

### DIFF
--- a/cmd/gapit/validate_gpu_profiling.go
+++ b/cmd/gapit/validate_gpu_profiling.go
@@ -47,14 +47,20 @@ func (verb *validateGpuProfilingVerb) Run(ctx context.Context, flags flag.FlagSe
 		return log.Err(ctx, err, "Failed to get device list.")
 	}
 	stdout := os.Stdout
+	someDeviceFailed := false
 	for i, p := range devices {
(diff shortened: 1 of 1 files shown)


cmd/gapit/validate_gpu_profiling.go
```diff
@@ -47,14 +47,20 @@ func (verb *validateGpuProfilingVerb) Run(ctx context.Context, flags flag.FlagSe
...
 		if err != nil {
 			fmt.Fprintf(stdout, "%v\n", log.Err(ctx, err, "Failed to validate device"))
+			someDeviceFailed = true
 			continue
 		}
 		fmt.Fprintf(stdout, "Device is validated.\n")
 	}
+	if someDeviceFailed {
+		err := fmt.Errorf("Some device failed validation")
+		return err
```
johnplate: If there is nothing to format and no need for a variable, wouldn't it be simpler to use the following?
`return errors.New("Some device failed validation")`

hevrard: Good point, fixed.

PR #3546, opened 2019-12-13
Refactor copyright header generation and checking

hevrard:
This change replace the text/core/copyright module and the
cmd/copyright command with simpler, much smaller code.

The text/core/copyright module served two purpose: checking copyright
header presence, and producing copyright headers for generated code.

The copyright checking was incorrectly updating the copyright date to
the current year, whereas the copyright date must be the date of the
file creation. This feature was used in cmd/copyright, which we did
not use in presubmit. This has been replaced by 23 lines of shell in
the presubmit check script.

Copyright header generation was done through Go templating, with a
surprising amount of features/templates that were actually never
used (maybe a leftover from GAPID history). Its actual use boils down
to producing rather static copyright headers, which do not need such a
complex module to get generated. This has been replaced by writing
such copyright headers directly as simple templates in the API files.

Also add a missing copyright header.

Also speedup presubmit check by processing files in parallel.


hevrard:
> presubmit failed

Apparently when running on Kokoro, shell script dislikes single quote escapes, I'm trying a fix.

hevrard:
@pau-baiget  it now passes presubmit!

kokoro/presubmit/presubmit.sh
```diff
@@ -50,17 +50,36 @@ function check() {
...
+    # Filter out third party files
+    # Grep seems to match "(standard input)", filter this out in the for loop output
+    find . -type f -name "*.${suffix}" | grep -v "./tools/build/third_party" | xargs grep -L "Copyright .* Google"
+  done | grep -v "(standard input)" > ${tmpfile}
+  if test -s ${tmpfile}; then
+    # tempfile is NOT empty
+    echo "Copyright issue in these files:"
+    cat ${tmpfile}
+    rm ${tmpfile}
+    return 1
```
pmuetschard: It's disappointing that this presubmit doesn't just fix the files in question, like all the others do.

PR #3549, opened 2019-12-13
Non mec looping

purvisa-at-google-com:
NoCmdID inside a loop was causing the FrameLoop transform to prematurely end the loop. This prevents that.

### DIFF
--- a/gapis/api/vulkan/frame_loop.go
+++ b/gapis/api/vulkan/frame_loop.go
@@ -364,7 +364,7 @@ func (f *frameLoop) Transform(ctx context.Context, cmdId api.CmdID, cmd api.Cmd,
 	} else if f.loopTerminated == false { // We're not before or at the start of the loop: thus, are we inside the loop or just at the end of it?
 
 		// This is the end of the loop. We have a lot of deferred things to do.
-		if api.CmdID.Real(cmdId) >= f.loopEndIdx {
+		if api.CmdID.Real(cmdId) >= f.loopEndIdx && cmdId != api.CmdNoID {
 
 			if lastObservedCommand == api.CmdNoID {
 				log.F(ctx, true, "FrameLoop: Somehow, the FrameLoop ended before it began. Did an earlier transform delete the whole loop? Were your loop indexes realistic?")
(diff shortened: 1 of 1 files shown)


gapis/api/vulkan/frame_loop.go
```diff
@@ -371,6 +377,22 @@ func (f *frameLoop) Transform(ctx context.Context, cmdId api.CmdID, cmd api.Cmd,
 			f.loopTerminated = true
 			log.D(ctx, "FrameLoop: end loop at frame %v cmdId %v, cmd is %v.", f.frameNum, cmdId, cmd)
 
+			// This command is the last in the loop so lets add it to the captured commands so we don't need to special case it.
+			f.capturedLoopCmds = append(f.capturedLoopCmds, cmd)
+			f.capturedLoopCmdIds = append(f.capturedLoopCmdIds, cmdId)
+
+			// If we are looping zero times we can just drop the commands inside the loop.
+			if f.loopCount == 0 {
```
RenfengLiu: so `loopcount == 0` means skip all commands and  `loopcount == 1`  means no loop and just replay normally, right?

purvisa-at-google-com: Correct. Same as a c program.

gapis/api/vulkan/frame_loop.go
```diff
@@ -380,108 +402,51 @@ func (f *frameLoop) Transform(ctx context.Context, cmdId api.CmdID, cmd api.Cmd,
...
-				apiState := GetState(out.State())
-
-				stateBuilder := apiState.newStateBuilder(ctx, newTransformerOutput(out))
-				defer stateBuilder.ta.Dispose()
-
-				// Write out the command that ended the loop. That one is outside of the captured commands since it's the one that called this code.
-				out.MutateAndWrite(ctx, cmdId, cmd)
-
-				// Notify the other transforms that we have just emitted the end of the loop.
-				out.NotifyPostLoop(ctx)
```
RenfengLiu: In the new code, the `out.NotifyPostLoop(ctx)` is gone. This is need to notify other transform to clean up at the end of loop

PR #3550, opened 2019-12-13
Fixed gapii protocol

johnplate:
New version of the new GAPII protocol which fixes the bug b/146054078 that compute trace could not be stopped.

### DIFF
--- a/gapii/cc/spy.cpp
+++ b/gapii/cc/spy.cpp
@@ -250,7 +250,10 @@ Spy::Spy()
                   GAPID_DEBUG("Received end trace message");
                   if (!is_suspended()) {
                     GAPID_DEBUG("Ending capture");
-                    mCaptureFrames = 1;
+                    // If app uses frame boundaries, end capture at next one
+                    // otherwise at next traced graphics API call
+                    const bool usesFrameBounds = mFrameNumber > 0u;
+                    mCaptureFrames = usesFrameBounds ? 1 : -1;
                   }
                   break;
                 default:
@@ -265,6 +268,11 @@ Spy::Spy()
   }
   set_suspended(mSuspendCaptureFrames != 0);
   set_observing(mObserveFrameFrequency != 0 || mObserveDrawFrequency != 0);
+}
+
+Spy::~Spy() {
+  mCaptureFrames = -1;
+  checkEndTrace();
 }
 
 void Spy::resolveImports() { GlesSpy::mImports.resolve(); }
@@ -478,6 +486,22 @@ void Spy::gvr_frame_submit(CallObserver* observer, gvr_frame** frame,
   GvrSpy::gvr_frame_submit(observer, frame, list, head_space_from_start_space);
 }
 
+bool Spy::checkEndTrace() {
+  if (!is_suspended() && mCaptureFrames < 0) {
+    GAPID_DEBUG("Ended capture");
+    mEncoder->flush();
+    // Error messages can be transferred any time during the trace, e.g.:
+    // auto err = protocol::createError("end of the world");
+    // mConnection->write(err.data(), err.size());
+    auto msg = protocol::createHeader(protocol::MessageType::kEndTrace);
+    mConnection->write(msg.data(), msg.size());
+    mConnection->close();
+    set_suspended(true);
+    return true;
+  }
+  return false;
+}
+
 void Spy::onPostDrawCall(CallObserver* observer, uint8_t api) {
   if (is_suspended()) {
     return;
@@ -569,15 +593,8 @@ void Spy::onPostFrameBoundary(bool isStartOfFrame) {
   } else {
     if (mCaptureFrames > 0) {
       if (--mCaptureFrames == 0) {
-        GAPID_DEBUG("Ended capture");
-        mEncoder->flush();
-        // Error messages can be transferred any time during the trace, e.g.:
-        // auto err = protocol::createError("end of the world");
-        // mConnection->write(err.data(), err.size());
-        auto msg = protocol::createHeader(protocol::MessageType::kEndTrace);
-        mConnection->write(msg.data(), msg.size());
-        mConnection->close();
-        set_suspended(true);
+        mCaptureFrames = -1;
+        checkEndTrace();
       }
     }
   }
--- a/gapii/cc/spy.cpp
+++ b/gapii/cc/spy.cpp
@@ -18,6 +18,7 @@
 
 #include "connection_header.h"
 #include "connection_stream.h"
+#include "protocol.h"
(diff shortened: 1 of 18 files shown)


gapii/cc/spy.cpp
```diff
@@ -460,6 +486,22 @@ void Spy::gvr_frame_submit(CallObserver* observer, gvr_frame** frame,
   GvrSpy::gvr_frame_submit(observer, frame, list, head_space_from_start_space);
 }
 
+bool Spy::checkEndTrace() {
+  if (!is_suspended() && mCaptureFrames < 0) {
```
hevrard: Sould this be `mCaptureFrames <= 0` ?

gapii/cc/spy.cpp
```diff
@@ -536,22 +578,24 @@ void Spy::onPostFrameBoundary(bool isStartOfFrame) {
...
-      // We must change suspended state BEFORE releasing the Spy lock with
-      // exit(), because the suspended state affects concurrent CallObservers.
-      set_suspended(false);
-      exit();
-      saveInitialState();
-      enter("RecreateState", 2);
+  } else {
+    if (mCaptureFrames > 0) {
+      if (--mCaptureFrames == 0) {
+        mCaptureFrames = -1;
```
hevrard: And here `if (mCaptureFrames <= 0)` instead of the two lines

gapii/cc/spy.cpp
johnplate: No, mCaptureFrames == 0 means continue until manual stop, -1 means end of trace.
I have added a comment to the variable declaration.

gapii/cc/spy.cpp
johnplate: No, the second line includes a pre-decrement, that's why it should be separate.

gapii/cc/spy.cpp
```diff
@@ -495,6 +496,8 @@ bool Spy::checkEndTrace() {
     // mConnection->write(err.data(), err.size());
     auto msg = protocol::createHeader(protocol::MessageType::kEndTrace);
     mConnection->write(msg.data(), msg.size());
+    // allow some time for the message to arrive
+    std::this_thread::sleep_for(std::chrono::milliseconds(200));
```
hevrard: Can't we have a call to some kind of `flush()` instead?

gapii/client/capture.go
```diff
@@ -212,7 +212,10 @@ func (p *Process) Capture(ctx context.Context, start task.Signal, stop task.Sign
 	}
 	go func() {
 		if stop.Wait(ctx) {
-			if err := writeEndTrace(conn); err != nil {
+			if err := writeEndTrace(conn); err == nil {
+				time.Sleep(2 * time.Second)
+				writeErr <- errors.New("traced app is unresponsive")
```
hevrard: Does this mean whenever we do a manual stop, we have a 2seconds delay and this message?
We may do a manual stop for a different reason than an unresponsive app, and in that case it would be best to quickly stop.

gapii/cc/spy.cpp
johnplate: There is no "flush" functionality for sockets. It would be possible to enable the linger option of the socket, but this would require some time to implement and test in core\cc\socket_connection.cpp

gapii/client/capture.go
johnplate: No, if the app manages to respond with the end trace message within 2 seconds, the main loop and main function will be ended without error and the message will not appear.

gapii/cc/spy.cpp
hevrard: Thanks for clarifying this.

gapii/cc/spy.cpp
johnplate: Just tried the linger option for the socket, but it didn't fix the problem, so we have to keep the sleep or change the protocol to send an acknowledge or similar.

johnplate:
Performed more tests: The detection of normal app termination (triggered by the Spy desctructor) works with tracing apps on Linux, but not on Windows, as apparently the socket is forcibly closed before the Spy destructor (or atexit) is called. This is most probably caused by the Windows limitation of dlls.
I don't have any app on Android which actually terminates normally, so I am not able to test it there. Even our Replay apks don't terminate, Android just puts them to sleep in the background when they are done.

gapii/client/capture.go
AWoloszyn: "Traced Application is unresponsive."
Try to use full sentences in error messages, they show up better in logs.

gapii/cc/spy.cpp
```diff
@@ -460,6 +487,24 @@ void Spy::gvr_frame_submit(CallObserver* observer, gvr_frame** frame,
   GvrSpy::gvr_frame_submit(observer, frame, list, head_space_from_start_space);
 }
 
+bool Spy::checkEndTrace() {
```
AWoloszyn: I am not sure if "check" is the right word for this, this does not really check anything it is really more "EndTraceIfNeeded" or something like that. I would expect this to return true, if the trace was supposed to end (or had ended maybe), and true otherwise, but not both.

gapii/cc/spy_base.h
```diff
@@ -112,6 +112,9 @@ class SpyBase {
   void set_record_timestamps(bool record) { mRecordTimestamps = record; }
   bool should_record_timestamps() const { return mRecordTimestamps; }
 
+  // Checks if the current trace should be ended
+  virtual bool checkEndTrace() { return false; }
```
AWoloszyn: Since the **only** person who overrides this is (eventually) spy, this should probably be pure virtual.

gapii/cc/spy.h
```diff
@@ -73,6 +71,8 @@ class Spy : public GlesSpy, public GvrSpy, public VulkanSpy {
                         const gvr_buffer_viewport_list* list,
                         gvr_mat4_abi head_space_from_start_space);
 
+  bool checkEndTrace() override;
```
AWoloszyn: We never check the return value of this, so we should not be returning anything, or we should be checking it in all of the call-sites.

gapii/cc/spy.cpp
johnplate: Good point, I was unsure about that function name. Renamed.

gapii/cc/spy_base.h
johnplate: Leaving it for now, because pure virtual Spy function triggers compilation error in GlesSpy constructor.

gapii/cc/spy.h
johnplate: Removed return value

PR #3552, opened 2019-12-13
Add 'Report a bug' option in the menu.

pau-baiget:
Also, use the standard bug template by default in the menu and in the error dialog. This way the title is empty and the user is forced to give a meaningful one before submitting it).

### DIFF
--- a/gapic/src/main/com/google/gapid/MainWindow.java
+++ b/gapic/src/main/com/google/gapid/MainWindow.java
@@ -45,6 +45,7 @@ import com.google.gapid.util.Messages;
 import com.google.gapid.util.OS;
 import com.google.gapid.util.StatusWatcher;
 import com.google.gapid.util.UpdateWatcher;
+import com.google.gapid.util.URLs;
 import com.google.gapid.views.StatusBar;
 import com.google.gapid.widgets.CopyPaste;
 import com.google.gapid.widgets.LoadablePanel;
@@ -374,6 +375,8 @@ public class MainWindow extends ApplicationWindow {
     manager.add(MenuItems.HelpShowLogs.create(() -> showLogDir(models.analytics)));
     manager.add(MenuItems.HelpLicenses.create(
         () -> showLicensesDialog(getShell(), models.analytics, widgets.theme)));
+    manager.add(MenuItems.HelpFileBug.create(
+        () -> Program.launch(URLs.FILE_BUG_URL)));
     return manager;
   }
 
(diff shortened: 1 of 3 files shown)


gapic/src/main/com/google/gapid/MainWindow.java
```diff
@@ -438,7 +441,8 @@ public MainView updateAndGet(Service.TraceType traceType) {
     HelpOnlineHelp("&Online Help\tF1", SWT.F1),
     HelpAbout("&About"),
     HelpShowLogs("Open &Log Directory"),
-    HelpLicenses("&Licenses");
+    HelpLicenses("&Licenses"),
+    HelpFileBug("File a &Bug on Github");
```
hevrard: Knowing that it is on GitHub is not relevant here I think, and this is the kind of things that we would typically forget to update if we move outside of GitHub one day, so I would avoid to mention it:
`HelpFileBug("File a &Bug");`

gapic/src/main/com/google/gapid/views/ErrorDialog.java
```diff
@@ -137,18 +136,13 @@ private void createDetails(Composite container) {
         withLayoutData(createTextbox(inner, DETAILS_STYLE, detailString),
             new GridData(SWT.FILL, SWT.FILL, true, true));
         withLayoutData(createLink(inner, "<a>File a bug</a> on GitHub", e -> {
```
hevrard: Same remark, remove "on GitHub"

gapic/src/main/com/google/gapid/views/ErrorDialog.java
```diff
@@ -137,18 +136,13 @@ private void createDetails(Composite container) {
...
-          Program.launch(getFileBugUrl());
+          Program.launch(URLs.FILE_BUG_URL);
         }), new GridData(SWT.RIGHT, SWT.BOTTOM, false, false));
         withLayoutData(createLink(inner, "<a>Show logs</a> directory", e -> {
           AboutDialog.showLogDir(analytics);
         }), new GridData(SWT.RIGHT, SWT.BOTTOM, false, false));
       }
 
-      private String getFileBugUrl() {
-        Escaper esc = UrlEscapers.urlFormParameterEscaper();
```
stellama0208: nit: remove `import com.google.common.escape.Escaper; import com.google.common.net.UrlEscapers;` together to avoid JAVA warning.

PR #3553, opened 2019-12-13
Put filmstrip in a tab

pmuetschard:
Also fixes an issue with tab merging.


chrisforbes:
Thanks for finishing this off.

PR #3555, opened 2019-12-13
Release resource allocated in queryTimestamps transform

RenfengLiu:


gapis/api/vulkan/query_timestamps.go
```diff
@@ -544,4 +545,50 @@ func (t *queryTimestamps) PostLoop(ctx context.Context, out transform.Writer) {
...
+			info.device,
+			info.queryPool,
+			memory.Nullptr,
+		)
+		out.MutateAndWrite(ctx, api.CmdNoID, cmd)
+	}
+	t.queryPools = make(map[VkQueue]*queryPoolInfo)
+
+	// Free commandbuffer and commandpool allocated in this transform.
+	for commandPoolkey, commandPool := range t.commandPools {
```
AWoloszyn: You would have to check the spec, but IIRC you can just destroy the command-pool, you dont have to destroy each individual command buffer. But this is fine too.

RenfengLiu: Updated to just released the commandpool

AWoloszyn:
LGTM

PR #3568, opened 2019-12-17
Add null check for vulkanCPUTiming checkbox.

lpy:
The vulkanCPUTiming checkbox may not be created, hence do an early return if it
doesn't exist.

BUG: N/A
Test: build gapid, manually select System Profile and click on Configure button.


lpy:
Conflict with https://github.com/google/gapid/pull/3566

PR #3574, opened 2019-12-18
Refactor and expose PNG-writing functionality

afd:
This change refactors the PNG-writing functionality in the virtual
swapchain layer so that the PNG data is passed to a function (provided
as a function pointer), rather than directly written to a file.  When
invoked directly from the layer, a function that does indeed write the
image to a file is used.  This more general PNG-writing facility is
exposed to client code as an extern C function.

### DIFF
--- a/core/vulkan/vk_virtual_swapchain/cc/virtual_swapchain_android.exports
+++ b/core/vulkan/vk_virtual_swapchain/cc/virtual_swapchain_android.exports
@@ -4,3 +4,4 @@ vkEnumerateInstanceLayerProperties
 vkEnumerateInstanceExtensionProperties
 vkEnumerateDeviceLayerProperties
 vkEnumerateDeviceExtensionProperties
+virtual_swapchain_write_png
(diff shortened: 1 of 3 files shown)


core/vulkan/vk_virtual_swapchain/cc/virtual_swapchain_android.exports
```diff
@@ -4,3 +4,4 @@ vkEnumerateInstanceLayerProperties
 vkEnumerateInstanceExtensionProperties
 vkEnumerateDeviceLayerProperties
 vkEnumerateDeviceExtensionProperties
+virtual_swapchain_write_png
```
AWoloszyn: I don't think we should be passing this out directly. This bypasses the vulkan layer system. 

We should do this the same way as we set the callback. This callback is done by passing it into the vulkan layer.

afd:
Abandoning this as @johnplate has provided #3612 which factors out the common code this PR was attempting to make reusable.

PR #3583, opened 2019-12-19
Update device validation.

lpy:
BUG: b/138717619


hevrard:
@lpy the test failure is due to a OOM, I think it's OK to merge as-is.

PR #3591, opened 2019-12-20
Notify user when typically static data was set dynamically

hliatis:


AWoloszyn:
I will let @pmuetschard review this since its mostly in the UI code

PR #3592, opened 2019-12-20
Open trace automatically if it is a Perfetto trace

dpallotti:
As per b/143711475, when a Perfetto trace is successfully taken, in almost every
case the user would want to click "Open Trace". This opens the trace
automatically, removing the extra step.

Bug: b/143711475

### DIFF
--- a/gapic/src/main/com/google/gapid/views/TracerDialog.java
+++ b/gapic/src/main/com/google/gapid/views/TracerDialog.java
@@ -151,12 +151,16 @@ public class TracerDialog {
     TraceInputDialog input =
         new TraceInputDialog(shell, models, widgets, models.devices::loadDevices);
     if (loadDevicesAndShowDialog(input, models) == Window.OK) {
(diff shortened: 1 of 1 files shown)


dpallotti:
Not sure if this should be handled by the TraceProgressDialog like I did or by showTracingDialog directly.

dpallotti:
Also, if we think that there are relevant cases where users might not want the Perfetto trace to be opened automatically, we could add an option to open it automatically or not.

AWoloszyn:
See b/142317175

gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -961,6 +970,9 @@ public boolean successful() {
     public void onProgress(StatusResponse progress) {
       status = progress;
       update();
+      if (progress.getStatus() == Service.TraceStatus.Done) {
```
dpallotti: I obviously forgot to check this.openOnDone here

PR #3595, opened 2019-12-20
Merge Pipeline View branch into master

hliatis:


gapic/src/main/com/google/gapid/models/Settings.java
```diff
@@ -67,6 +67,8 @@
   public int[] texturesSplitterWeights = new int[] { 20, 80 };
   public String traceDeviceSerial = "";
   public String traceDeviceName = "";
+  public int[] pipelineSplitterWeights = new int[] { 20, 80 };
+  public String traceDevice = "";
```
pmuetschard: Remove `traceDevice`. This is a bad merge.

Actually remove both, you are not using `pipelineSplitterWeights` anymore.

PR #3597, opened 2019-12-20
Update Perfetto for a fix to the trace processor.

pmuetschard:
This is to get https://android-review.googlesource.com/c/platform/external/perfetto/+/1196781/
which fixes a data loss issue.


pmuetschard:
The test was broken. I've fixed it some time ago.

PR #3598, opened 2019-12-23
Android related fixes to the Vulkan sample.

pmuetschard:
- Don't ever use mailbox mode, always use the guaranteed FIFO mode.
 - On Android, always use 3 swapchain images.

### DIFF
--- a/cmd/vulkan_sample/main.cpp
+++ b/cmd/vulkan_sample/main.cpp
@@ -513,7 +513,6 @@ int main(int argc, const char** argv) {
 #endif
   VkSurfaceCapabilitiesKHR surface_capabilities = {};
   VkSurfaceFormatKHR surface_format = {};
-  VkPresentModeKHR present_mode = {};
 
   VkPhysicalDevice physical_device = {};
   uint32_t queue_family_index = static_cast<uint32_t>(-1);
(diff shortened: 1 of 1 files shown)


AWoloszyn:
For Android we should still check minimagecount for the swapchain. We can default to 3 but if min is 4 we should respect that.

AWoloszyn:
Also, out of curiosity why not mailbox?

pmuetschard:
> For Android we should still check minimagecount for the swapchain. We can default to 3 but if min is 4 we should respect that.

Done.

> Also, out of curiosity why not mailbox?

Even with the display running at 90Hz, we were rendering the cube 3-4 times per frame. This is a waste.

chrisforbes:
I have 2 reasons to prefer fifo here:
- mailbox is not guaranteed to be available
- when we get to validating the SF events producer, fifo is far less
chaotic (frames are consumed in order, and none get dropped)

AWoloszyn:
> I have 2 reasons to prefer fifo here: - mailbox is not guaranteed to be available - when we get to validating the SF events producer, fifo is far less chaotic (frames are consumed in order, and none get dropped)

- We don't explicitly choose Mailbox, we choose the first provided mode.
- FIFO is less chaotic, but using it in the sample may prevent us from finding some bugs. It takes a non-trivial amount of work for us to handle out-of-order swapchain images etc, and having tests is useful.

That being said, I totally understand for this type of sample switching off, but we do lose some normal coverage, but that code-path is already fairly well tested anyway.

I don't mind having this.

chrisforbes:
For test coverage I completely agree with you -- this sample is being used for device validation though, so ought to be as predictable as possible

PR #3600, opened 2019-12-24
Add Vulkan Handle Mappings for replay profiler.

yalcinmelihyasin:
replaces(#3447 )


gapis/replay/mapping_exporter.go
```diff
@@ -0,0 +1,158 @@
...
+		// Count the number of actual Posts we expect
+		total += 1
+		func(k interface{}, typ reflect.Type, size uint64) {
+			b.Post(v, size, func(r binary.Reader, err error) {
+				defer func() {
+					total -= 1
+
+					if total == 0 {
+						done()
+					}
```
apbodnar: I know this block isn't really your code, but I still think there's a very unlikely data race here.  Whenever we get a new PostData message, we spin up a new goroutine (https://github.com/google/gapid/blob/6aeb23ad58849ab89e564bf2c9ac5e44b69f5a60/gapis/replay/builder/builder.go#L810) which then calls this decrement and compare logic.  AFAIK we don't get to chose how goroutines get scheduled once they're in the pool without our own synchronization.  In the worst case, through some hiccup in grpc or go which leads to multiple goroutines in the pool, the running goroutine could be swapped out between the decrement and comparison.  Wrapping this in a `mutex.Lock()` and `defer mutex.Unlock()` should avoid any risk of a data race.

yalcinmelihyasin: I have moved it from the old mapping_printer. That will be rewritten with a different logic.

gapis/replay/mapping_exporter.go
```diff
@@ -0,0 +1,158 @@
...
+	"github.com/google/gapid/core/log"
+	"github.com/google/gapid/gapis/api"
+	"github.com/google/gapid/gapis/api/transform"
+	"github.com/google/gapid/gapis/memory"
+	"github.com/google/gapid/gapis/replay/builder"
+)
+
+type VulkanHandleMappingItem struct {
+	HandleType  string
+	TraceValue  string
```
chrisforbes: Shouldn't the values be uint64?

PR #3603, opened 2019-12-26
Fix classpath source locations.

pmuetschard:


.classpath
```diff
@@ -5,25 +5,25 @@
...
-	<classpathentry kind="lib" path="bazel-bin/core/stream/libstream_proto-speed.jar" sourcepath="bazel-genfiles/core/stream/stream_proto-speed-src.jar"/>
-	<classpathentry kind="lib" path="bazel-bin/external/com_google_protobuf/libprotobuf_java.jar"/>
+	<classpathentry kind="lib" path="bazel-bin/core/data/pod/libpod_proto-speed.jar" sourcepath="bazel-bin/core/data/pod/pod_proto-speed-src.jar"/>
+	<classpathentry kind="lib" path="bazel-bin/core/image/libimage_proto-speed.jar" sourcepath="bazel-bin/core/image/image_proto-speed-src.jar"/>
+	<classpathentry kind="lib" path="bazel-bin/core/log/log_pb/liblog_pb_proto-speed.jar" sourcepath="bazel-bin/core/log/log_pb/log_pb_proto-speed-src.jar"/>
+	<classpathentry kind="lib" path="bazel-bin/core/os/device/libdevice_proto-speed.jar" sourcepath="bazel-bin/core/os/device/device_proto-speed-src.jar"/>
+	<classpathentry kind="lib" path="bazel-bin/core/stream/libstream_proto-speed.jar" sourcepath="bazel-bin/core/stream/stream_proto-speed-src.jar"/>
+	<classpathentry kind="lib" path="bazel-bin/external/com_google_protobuf/libprotobuf_java.jar" sourcepath="bazel-external/com_google_protobuf/java"/>
```
hevrard: Should the sourcepath be a `.jar`, like all others around seem to be?

pmuetschard: No, in this case it's a source directory, since we build the protobuf runtime library from source.

pmuetschard:
bazel-genfiles is no longer created. If you do a `bazel clean` you will see it disappear and not come back with a new `bazel build`.

PR #3604, opened 2019-12-26
Highlight Surfaceflinger V-SYNC.

pmuetschard:


gapic/src/main/com/google/gapid/perfetto/models/VSync.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    @Override
+    public Data getData(Track.DataRequest req, Track.OnUiThread<Data> onUiThread) {
+      return new Data(req, new long[0], false);
+    }
+  };
+
+  public default boolean hasData() {
+    return true;
```
stellama0208: Is it good to assume true by default? 
Would it be better to have default being false, and for class FromSurfaceFlingerAppCounter judging the boolean value by saved CounterInfo?

pmuetschard: The idea here is that the only instance that should return false for `hasData` is the `VSync.EMPTY` instance. If the counter doesn't exist, we should not return a `FromSurfaceFlingerAppCounter` instance. This is in preparation of having a better data source for future versions of Android (from surfaceflinger events in R, for example).

stellama0208: Thanks for explaining, then it's good to go now. We can take care of this method in the future if necessary, considering how the new data source operates.

PR #3607, opened 2019-12-30
Show name for pseudonym type fields within Struct

silence-do-good:
If a member of a struct is typedef'd multiple times, the name is lost due to the way Pseudonym children are created. This fixes the creation of children by passing the name all the way down to the innermost pseudonym.
http://b/142882406

### DIFF
--- a/gapic/src/main/com/google/gapid/models/Memory.java
+++ b/gapic/src/main/com/google/gapid/models/Memory.java
@@ -605,7 +605,6 @@ public class Memory extends DeviceDependentModel<Memory.Data, Memory.Source, Voi
         case STRUCT:
           TypeInfo.StructType struct = type.getStruct();
           List<TypeInfo.StructField> childrenTypes = struct.getFieldsList();
-
           List<MemoryBox.Value> childrenValues = value.getStruct().getFieldsList();
           for (int i = 0; i < childrenValues.size(); i++) {
             StructNode childNode = new StructNode(api,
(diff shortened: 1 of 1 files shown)


gapic/src/main/com/google/gapid/models/Memory.java
```diff
@@ -618,7 +629,7 @@ private boolean mayHaveChildren() {
         case PSEUDONYM:
           TypeInfo.PseudonymType pseudonym = type.getPseudonym();
           childType = typesModel.getType(type(pseudonym.getUnderlying(), api));
-          children.add(new StructNode(api, childType, value, rootAddress, typesModel));
+          children.add(new StructNode(api, childType, value, rootAddress, typesModel, structName));
```
stellama0208: Nit: Would prefer to use the previous way to initialize node (use the original constructor and setStructName() method).
Reason for suggestion: Use the constructor to initialize those final variables, but use has/get/set method for those very optional variables.

gapic/src/main/com/google/gapid/models/Memory.java
```diff
@@ -594,12 +605,12 @@ private boolean mayHaveChildren() {
         case STRUCT:
           TypeInfo.StructType struct = type.getStruct();
           List<TypeInfo.StructField> childrenTypes = struct.getFieldsList();
+
```
stellama0208: Nit: can remove the line.

stellama0208:
> Looks OK. You still prefer to keep the inner node when collapsing, rather than the outer?
> 
> Would like Stella to take a look at this before you land it.

Will add a small fix later after this get merged in, to deal with Pseudonym's inner/outer disposal.

gapic/src/main/com/google/gapid/models/Memory.java
silence-do-good: Done.

gapic/src/main/com/google/gapid/models/Memory.java
silence-do-good: Done.

PR #3608, opened 2020-01-02
Fix bug with sharing mode in buffer/image creation.

AWoloszyn:


AWoloszyn:
Windows was a spurious failure. I will submit. (This didn't touch any OS specific code at all).

PR #3609, opened 2020-01-03
Pass Vulkan Handle Mappings to profiling_data to match with profiler values

yalcinmelihyasin:
Pass Vulkan Handle Mappings to profiling_data to match with profiler values


yalcinmelihyasin:
We have a bug in profiling data for now. We will add the actual matching when we fix that.

PR #3612, opened 2020-01-07
exposed PNG writing

johnplate:


core/vulkan/tools/image.h
```diff
@@ -0,0 +1,32 @@
...
+
+#ifndef VULKAN_TOOLS_IMAGE_H_
+#define VULKAN_TOOLS_IMAGE_H_
+
+#include <vulkan/vulkan.h>
+#include <ostream>
+
+namespace vk_tools {
+
+// If |image_format| is a suitable format, the image in |image_data| is
```
AWoloszyn: This needs to have a better interface.
Specifically, the fact that if the data is NOT in a suitable format, there is simply no way to know that nothing has happened.

Should at the very least return a boolean.

core/vulkan/tools/image.h
```diff
@@ -0,0 +1,32 @@
...
+// converted to a PNG, with the data for the PNG being sent to |stream|.
+extern void WritePng(std::ostream* stream, uint8_t* image_data, size_t size,
```
AWoloszyn: Extern is not needed here

core/vulkan/tools/image.cpp
```diff
@@ -0,0 +1,67 @@
...
+void WriteDataToStream(void* context, void* data, int size) {
+  std::ostream* stream = static_cast<std::ostream*>(context);
+  stream->write(static_cast<char*>(data), size);
+}
+
+}  // namespace
+
+namespace vk_tools {
+
+void WritePng(std::ostream* stream, uint8_t* image_data, size_t size,
```
AWoloszyn: image_data is being modified here. This is likely unexpected. (I know it is not const, but we should call it out explicitly).

core/vulkan/tools/image.cpp
```diff
@@ -0,0 +1,67 @@
...
+              uint32_t width, uint32_t height, VkFormat image_format) {
+  switch (image_format) {
+    case VK_FORMAT_B8G8R8A8_UNORM:
+    case VK_FORMAT_B8G8R8A8_UINT: {
+      // Convert the image data from BGRA to RGBA
+      auto data = image_data;
```
AWoloszyn: As this is now a library function, we should at least check that the width*height*data_size is what we actually expect, and return if the data is wrong. Rather than writing to arbitrary memory.

core/vulkan/tools/image.h
johnplate: Thanks, right, fixed.

core/vulkan/tools/image.h
johnplate: Removed

core/vulkan/tools/image.cpp
johnplate: Yep, clarified in the comment

core/vulkan/tools/image.cpp
johnplate: added check

PR #3614, opened 2020-01-07
GAPID profiler - Fix VMT issues

zakerinasab:
VMT was reported to cause games crash during or after tracing. This PR fixes some related issues.

BUG=146925630

### DIFF
--- a/core/vulkan/vk_memory_tracker_layer/cc/memory_tracker_layer_impl.cpp
+++ b/core/vulkan/vk_memory_tracker_layer/cc/memory_tracker_layer_impl.cpp
@@ -33,8 +33,11 @@ using namespace std::chrono;
 
 namespace memory_tracker {
 
+using scoped_read_lock = layer_helpers::threading::scoped_read_lock;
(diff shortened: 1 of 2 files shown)


core/vulkan/vk_memory_tracker_layer/cc/memory_tracker_layer_impl.cpp
```diff
@@ -1152,18 +1232,17 @@ void MemoryTracker::EmitAndClearAllStoredEvents() {
     }
     event_container_set->clear();
   }
-  rwl_devices.wunlock();
+  rwl_devices.runlock();
```
AWoloszyn: These manual lock/unlock events make me somewhat uneasy.

You can submit, but maybe consider creating a small RAII struct to clean this up.

```
class scoped_read_locker {
    public:
      scoped_read_locker(rwlock* _lock): lock(_lock) { _lock.rlock(); }
     ~scoped_read_locker() { _lock.runlock(); }
     rwlock* lock
}

That way we can't accidently keep them around un-necessarily.

PR #3616, opened 2020-01-07
Add GPU counters to ProfilingData message

apbodnar:


gapis/service/service.proto
```diff
@@ -1329,7 +1329,27 @@ message ProfilingData {
...
+
+    message Track {
+      uint32 id = 1;
+      string name = 2;
+      string description = 3;
+      string unit = 4;
+      bool default = 5;
+    }
+
+    repeated Counter counters = 1;
```
pmuetschard: Storing the counter data as arrays will be more efficient. So
- rename Track -> Counter
- Add a `repeated uint64 ts` field to Counter
- Add a `repeated double value` field to Counter

gapis/service/service.proto
```diff
@@ -1329,7 +1329,22 @@ message ProfilingData {
     repeated Group groups = 3;
   }
 
+  message GpuCounters {
```
pmuetschard: I would get rid of this message and simply have `repeated Counter counters = 2` in the `ProfilingData` message, but I'm OK if you feel strongly about the wrapper.

PR #3618, opened 2020-01-07
Add a 250ms sleep after starting a Perfetto trace

apbodnar:
Hack to unblock @pmuetschard . In some cases, the producer isn't quite ready.

### DIFF
--- a/core/event/task/task.go
+++ b/core/event/task/task.go
@@ -37,6 +37,17 @@ func Once(task Task) Task {
 	}
 }
 
+// Delay wraps a task in a coroutine to asynchronously execute after a specified duration
+func Delay(task Task, duration time.Duration) Task {
+	return func(ctx context.Context) error {
+		crash.Go(func() {
+			time.Sleep(duration)
+			task(ctx)
+		})
+		return nil
+	}
+}
+
 func Noop() Task {
 	return func(context.Context) error {
 		return nil
--- a/core/os/android/adb/perfetto.go
+++ b/core/os/android/adb/perfetto.go
@@ -23,6 +23,7 @@ import (
 	"io"
 	"regexp"
 	"strings"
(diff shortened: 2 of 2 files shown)


core/event/task/task.go
```diff
@@ -37,6 +37,13 @@ func Once(task Task) Task {
 	}
 }
 
+func Delay(task Task, duration time.Duration) Task {
+	return func(ctx context.Context) error {
+		time.Sleep(duration)
+		return task(ctx)
```
pmuetschard: This needs to be on another goroutine, as it would otherwise block. I think it makes sense to do it here.

core/event/task/task.go
```diff
@@ -37,6 +37,17 @@ func Once(task Task) Task {
 	}
 }
 
+// Wraps a task in a coroutine to asynchronously execute after a specified duration
```
pmuetschard: The godoc stuff requires a very specific format for documentation comments. Comments for functions need to start with the function name. So this needs to be `// Delay ...`.
Please see https://blog.golang.org/godoc-documenting-go-code for details.

core/os/android/adb/perfetto.go
```diff
@@ -45,7 +46,8 @@ func (b *binding) StartPerfettoTrace(ctx context.Context, config *perfetto_pb.Tr
 
 	reader, stdout := io.Pipe()
 	logRing := ring.New(10)
-	readyOnce := task.Once(ready)
+	// TODO(apbodnar) Find a way to reliably know when Perfetto/producers are ready
```
pau-baiget: Is there a task for this? It would be good to track this for the next iteration.

apbodnar: yep, added it to the TODO

PR #3621, opened 2020-01-08
Change postback to notification

yalcinmelihyasin:
b/147341925

Change the handle mapping due to frame looping.


AWoloszyn:
> This needs a proper commit message explaining why you're doing this and what you hope to gain. I assume this also gets you some way toward fixing the performance?

This is because POST is incompatible with with frame looping.

yalcinmelihyasin:
> This needs a proper commit message explaining why you're doing this and what you hope to gain. I assume this also gets you some way toward fixing the performance?

This does not improve performance. That will be in the next pr.

PR #3622, opened 2020-01-08
Do not expose functions that the underlying driver does not support.

AWoloszyn:
b/145952976


gapis/api/vulkan/templates/vulkan_layer.tmpl
```diff
@@ -23,6 +23,7 @@
     {{Global "Vulkan.LayerNamespace" "<The namespace of your vulkan functions>"}}
     {{Global "Vulkan.OverrideFunctions" List Of Functions That Should Be Overriden}}
     {{Global "Vulkan.OtherUsedFunctions" List Of Functions That your layer will call down into}}
+    {{Global "Vulkan.ImplementedFunctions" List Of Functions That your layer will call down into}}
```
pmuetschard: This needs a slightly different description than the one above it. It should say that these are functions that should be intercepted even if the lower levels don't provide them, correct?

AWoloszyn: Correct, done.

PR #3625, opened 2020-01-09
 Record debug markers to be emitted during OnStart

rschiu:


rschiu:
@AWoloszyn friendly ping.  Could you please take a look?  Thanks.

core/vulkan/vk_api_timing_layer/cc/vk_api_emitter.inc
```diff
@@ -0,0 +1,74 @@
...
+}
+
+template <typename T>
+VkApiEmitter<T>::~VkApiEmitter() {
+  core::PerfettoProducer<T>::Get().UnregisterEmitter(this);
+}
+
+template <typename T>
+void VkApiEmitter<T>::StartTracing() {
+  for (auto marker : debug_markers_) {
```
AWoloszyn: Does this list expand forever?
What happens if the same handle gets more than one debug marker (with different names)
  - This can happen if you create an object, delete an object, create a NEW object that gets the same handle.

We will accumulate for ever, probably not what we want.

rschiu: done, PTAL

PR #3632, opened 2020-01-10
Frame Lifecycle - Add highlight for selected events

silence-do-good:
The frame lifecycle events did not highlight when selected and this can get confusing. This patch adds a few fixes to it and a minor refactor to the color scheme as well.

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
+++ b/gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
@@ -35,7 +35,9 @@ import com.google.gapid.perfetto.TimeSpan;
 import com.google.gapid.perfetto.views.FrameEventsMultiSelectionView;
 import com.google.gapid.perfetto.views.FrameEventsSelectionView;
 import com.google.gapid.perfetto.views.State;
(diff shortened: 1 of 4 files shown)


gapic/src/main/com/google/gapid/perfetto/models/FrameEventsTrack.java
```diff
@@ -75,6 +77,19 @@ public static FrameEventsTrack forBuffer(QueryEngine qe, GpuInfo.Buffer buffer)
...
+
+  public static RGBA getBorderColor(String title) {
+    return colorForSlice(title, StyleConstants.isLight() ? -5 : 5);
+  }
+
+  private static RGBA colorForSlice(String title, int shadeIdx) {
+    return StyleConstants.Palette.getColor(
+        (title.hashCode() ^ title.length()) & 0x7fffffff, shadeIdx);
+  }
+
```
pmuetschard: This should be in the view class, not the model.

silence-do-good: I took this mainly from the SliceTrack class which is inside models, which is why I put it in models too.

PR #3637, opened 2020-01-10
System profiler - minor fix in VMT

zakerinasab:
This change adds a check for device memory tracking option to the memory tracker.


zakerinasab:
> LGTM, but can you update the commit message to say "What" you are fixing.

Done.

PR #3639, opened 2020-01-10
Update Perfetto to 975bbd74e6b1d7da8dbdaaeec9191fb28471ed2c

rschiu:
Include changes to display debug marker name
slice_id is rename id due to gpu_slice table refactor


rschiu:
@pmuetschard, could you take a look?  Thanks.

PR #3641, opened 2020-01-13
Expose GPU frequency trace point.

lpy:
Always enable GPU frequency trace point when GPU checkbox is selected.

BUG: b/143969077

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
+++ b/gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
@@ -101,6 +101,9 @@ public class TraceConfigDialog extends DialogBase {
       "ion", "memory", "memreclaim", "network", "nnapi", "pdx", "pm", "power", "res", "rro", "rs",
       "sched", "sm", "ss", "sync", "vibrator", "video", "view", "webview", "wm",
   };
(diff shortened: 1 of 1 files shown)


lpy:
ptal

gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
```diff
@@ -180,6 +183,12 @@ public static String getConfigSummary(Settings settings, Device.PerfettoCapabili
     PerfettoConfig.TraceConfig.Builder config = PerfettoConfig.TraceConfig.newBuilder();
     SettingsProto.PerfettoOrBuilder p = settings.perfetto();
 
+    PerfettoConfig.FtraceConfig.Builder ftrace = config.addDataSourcesBuilder()
+          .getConfigBuilder()
+              .setName("linux.ftrace")
+              .getFtraceConfigBuilder()
+                  .setBufferSizeKb(FTRACE_BUFFER_SIZE);
+
```
pmuetschard: This causes the ftrace datasource to be added even if we aren't collecting cpu or gpu data. This should be wrapped in an `if (p.getCpuOrBuilder().getEnabled() || (p.getGpuOrBuilder().getEnabled())`

lpy:
Thanks, ptal!

PR #3642, opened 2020-01-13
Remove the atrace categories that require userdebug.

pmuetschard:


pmuetschard:
This is a temporary fix to at least get tracing working on non-debug devices. We will need to go through the list and be more deliberate in what to include at a later time.

PR #3643, opened 2020-01-13
VSync Highlight fixes and enhancements

pmuetschard:
- show the highlight in the timeline and pinned tracks
- show/hide the highlights via the 'v' keyboard shortcut
- remember the enable state between runs


pmuetschard:
fixed the NPE. PTAL

PR #3645, opened 2020-01-14
Document Golang setup with gofuse

hevrard:


DEVDOC.md
```diff
@@ -1,5 +1,43 @@
...
+## Setup Golang development
+
+This project contains Golang code, but it does not have the file hierarchy of
+regular Golang projects (this is due to the use of Bazel as a build system).
+The `cmd/gofuse` utility enables to re-create the file hierarchy expected by Go
+tools:
+
+```sh
+# We assume GAPID has been checked out under ~/gapid
+cd ${HOME}/gapid
```
johnplate: I propose "cd <path-to-gapid-source>", which is independent of the OS, doesn't need the comment and is consistent with BUILDING.md

DEVDOC.md
```diff
@@ -1,5 +1,43 @@
...
+# Make sure to build to have all compile-time generated files
+bazel build //pkg
```
johnplate: "bazel build pkg", again to be consistent with BUILDING.md. Users who are new to bazel get confused if "//" is required

DEVDOC.md
```diff
@@ -1,5 +1,43 @@
...
+# Prepare a gapid-gofuse directory **outside of the gapid checkout directory**
+mkdir ${HOME}/gapid-gofuse
```
johnplate: "mkdir <path-outside-gapid-source>/gapid-gofuse", consistent with above

DEVDOC.md
johnplate: omg, github cuts my stuff.
I wrote '''cd <path-to-gapid-source>'''

johnplate: arg, again
```cd <path-to-gapid-source>```

File: DEVDOC.md (hunk 3 above)
Comment (johnplate) [reply DEVDOC.md]: ```mkdir <path-outside-gapid-source>/gapid-gofuse```

File: DEVDOC.md [hunk 4]
```diff
@@ -1,5 +1,43 @@
...
+# Run gofuse with the previous directory as a target, and the bazel-specific OS identifier
+OS=k8 # Pick up the right option: k8 (linux) / darwin (macos) / x64_windows (windows)
+bazel run //cmd/gofuse -- -bazelout ${OS}-fastbuild --dir ${HOME}/gapid-gofuse
```
johnplate: I think the option is -dir with a single dash.
I would use here only
```bazel run //cmd//gofuse -- -dir <path-outside-gapid-source>/gapid-gofuse```
and let gofuse use the default for bazelout. If we describe the -bazelout option, we should also mention that the configuration can be fastbuild, dbg or opt, and that in doubt users can just check the folders under bazel-out/

File: DEVDOC.md [hunk 5]
```diff
@@ -1,5 +1,43 @@
...
+
+## In your shell init script, e.g. ~/.bashrc, set up your GOPATH:
+export GOPATH=${HOME}/gapid-gofuse
```
johnplate: This would overwrite any existing GOPATH. We should just write to append the GOPATH

DEVDOC.md
hevrard: done

DEVDOC.md
hevrard: done

DEVDOC.md
hevrard: Done.

The gapid utility commands accept both one or two dash for options, but yes let's stick with what their help message displays.

DEVDOC.md
hevrard: done

DEVDOC.md
hevrard: done

DEVDOC.md
johnplate: The new version would end up with a leading colon if GOPATH is not yet defined. Additionally, under Windows the GOPATH is separated by semicolons.
So better just write in words "append GOPATH with ...". The user can just search how to do that on his system.

hevrard: updated

PR #3646, opened 2020-01-14
Check command Mutate errors when calling MutateCmds

hevrard:
cmd.Mutate() may return an error. MutateCmds() is wrapping calls to
cmd.Mutate(), but did not do proper error checking. This change adds
error checking there, and in all callers of MutateCmds().


hevrard:
> LGTM, but do we have a plan to check all of the rest of the places that we are running ForEachCmd?

I plan to do it gradually, this was a first "self-contained" PR focusing on MutateCmds().

Issue #2411, opened 2018-11-28
Allow batching of multiple subcommands in Vulkan.


AWoloszyn:
This will represent a significant performance improvement in some cases.
This will involve splitting render-passes around the commands in question, i.e. creating extra render-passes so that we can copy out the framebuffer in-between.

This should significantly improve replay time for viewing multiple draw-calls.

AWoloszyn:
Fixed

PR #3648, opened 2020-01-14
Add vulkan command-buffer-splitter.

AWoloszyn:
This splits the vulkan command-buffer at a given set of subcommands.
It drops the command-buffer out of a renderpass, since
the operations allowed in a renderpass are limited. It then inserts
a CommandBufferInsertionCommand into the current command-buffer.

It then re-enters a renderpass (as necessary) and continues on its way.
Furthermore it inlines all secondary command-buffers that are found
in the command-buffer that must be split. This also allows the same
functionality.

### DIFF
--- a/gapis/api/vulkan/command_splitter.go
+++ b/gapis/api/vulkan/command_splitter.go
@@ -1,0 +1,757 @@
+// Copyright (C) 2020 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package vulkan
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/google/gapid/core/memory/arena"
+	"github.com/google/gapid/gapis/api"
+	"github.com/google/gapid/gapis/api/transform"
+	"github.com/google/gapid/gapis/memory"
+	"github.com/google/gapid/gapis/replay/builder"
+)
+
+// CommandBufferInsertionCommand is a temporary command
+// that is expected to be replaced by a down-stream transform.
+type CommandBufferInsertionCommand struct {
+	cmdBuffer VkCommandBuffer
+	idx       api.SubCmdIdx
+	callee    api.Cmd
+}
+
+// Interface check
+var _ api.Cmd = &CommandBufferInsertionCommand{}
+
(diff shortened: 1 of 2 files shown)


gapis/api/vulkan/command_splitter.go
```diff
@@ -0,0 +1,756 @@
+// Copyright (C) 2017 Google Inc.
```
pau-baiget: Copyright year is not right.

gapis/api/vulkan/command_splitter.go
```diff
@@ -0,0 +1,756 @@
...
+	"fmt"
+
+	"github.com/google/gapid/core/memory/arena"
+	"github.com/google/gapid/gapis/api"
+	"github.com/google/gapid/gapis/api/transform"
+	"github.com/google/gapid/gapis/memory"
+	"github.com/google/gapid/gapis/replay/builder"
+)
+
+// commandSplitter is a transform that allows vulkan command-buffers
```
pau-baiget: Is this comment in the right place? It seems to describe the struct below, but it doesn't.

gapis/api/vulkan/command_splitter.go
```diff
@@ -0,0 +1,757 @@
...
+	idx       api.SubCmdIdx
+	callee    api.Cmd
+}
+
+// Interface check
+var _ api.Cmd = &CommandBufferInsertionCommand{}
+
+func (*CommandBufferInsertionCommand) Mutate(ctx context.Context, cmd api.CmdID, g *api.GlobalState, b *builder.Builder, w api.StateWatcher) error {
+	if b != nil {
+		return fmt.Errorf("This command should have been replaced replaced before it got to the builder")
```
RenfengLiu: duplicated

gapis/api/vulkan/command_splitter.go
```diff
@@ -0,0 +1,757 @@
...
+	cb := CommandBuilder{Thread: queueSubmit.Thread(), Arena: a}
+
+	newSubmit := MakeVkSubmitInfo(a)
+	newSubmit.SetSType(submit.SType())
+	newSubmit.SetPNext(submit.PNext())
+	newSubmit.SetWaitSemaphoreCount(submit.WaitSemaphoreCount())
+	newSubmit.SetPWaitSemaphores(submit.PWaitSemaphores())
+	newSubmit.SetPWaitDstStageMask(submit.PWaitDstStageMask())
+	newSubmit.SetCommandBufferCount(submit.CommandBufferCount())
+	// pCommandBuffers
```
RenfengLiu: Is this comments needed?

yalcinmelihyasin:
Do we have a bug id ?

AWoloszyn:
We do not, I have updated the commit message to reflect.

yalcinmelihyasin:
Do we have a test this functionality?

AWoloszyn:
This is unused in its current form. The follow-up CL that starts using this for read_framebuffer will be tested by anything that pulls framebuffers back.

AWoloszyn:
I actually found the bug
https://github.com/google/gapid/issues/2411

yalcinmelihyasin:
LGTM but I would suggest others to look at.

PR #3649, opened 2020-01-14
Send Vulkan Handles as a batch

yalcinmelihyasin:
Instead of notifying every handle individually, copy them to a buffer at the end of the replay and return the buffer as whole

### DIFF
--- a/gapis/replay/mapping_exporter.go
+++ b/gapis/replay/mapping_exporter.go
@@ -21,7 +21,6 @@ import (
 	"os"
 	"reflect"
 	"sort"
-	"sync"
 
 	"github.com/google/gapid/core/data/endian"
 	"github.com/google/gapid/core/log"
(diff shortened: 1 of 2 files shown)


lpy:
Please attach a bug id in your description :)
```
Bug: b/BUG_ID
```
or
```
Bug: BUG_ID
```

gapis/replay/mapping_exporter.go
```diff
@@ -30,74 +29,91 @@ import (
...
+		switch handle.size {
+		case 1:
+			replayValue = uint64(r.Uint8())
+		case 2:
+			replayValue = uint64(r.Uint16())
+		case 4:
+			replayValue = uint64(r.Uint32())
+		case 8:
+			replayValue = r.Uint64()
+		}
```
hevrard: Add a `default:` and treat it as an error (not sure what is best: early return? panic?)

gapis/replay/mapping_exporter.go
```diff
@@ -30,74 +29,91 @@ import (
...
+		notificationID: 0,
 	}
 }
 
 func (m *MappingExporter) processNotification(ctx context.Context, s *api.GlobalState, n gapir.Notification) {
-	notificationData := n.GetData()
-	notificationID := n.GetId()
-	mappingData := notificationData.GetData()
-
-	m.mutex.Lock()
```
hevrard: You're removing a mutex here, and concurrency is never trivial: can you briefly document why in the commit message?

gapis/replay/mapping_exporter.go
```diff
@@ -106,23 +122,34 @@ func (m *MappingExporter) ExtractRemappings(ctx context.Context, s *api.GlobalSt
...
+	}
+
+	m.notificationID = b.GetNotificationID()
+	b.Notification(m.notificationID, handleBuffer, bufferSize)
+	err := b.RegisterNotificationReader(m.notificationID, func(n gapir.Notification) {
+		m.processNotification(ctx, s, n)
+	})
+
+	if err != nil {
+		log.W(ctx, "Vulkan Mapping Notification could not registered: ", err)
```
hevrard: typo: "Vulkan Mapping Notification could not **be** registered: "

yalcinmelihyasin:
b/147341925

gapis/replay/mapping_exporter.go
hevrard: Given that this is something that should never happen, I think it's better to have an early crash through a fatal log, i.e. `log.F(ctx, true, ...)`. Otherwise, we just get an error in the log but the program keeps on running with replayValue set to its default (zero), and this can lead to other issues/crash much further down, which may be harder to link to an invalid size here. With an early crash, we know directly where things started to go wrong.

hevrard: Great, now last nit: upon error, it would be good to print the `handle.size` alongside the name. I approve the PR and let you add this last thing before merging. Thanks!

hevrard: Brilliant, thanks!

PR #3651, opened 2020-01-14
Add presubmit check to enforce bug id.

lpy:
Every commit message should associate with at least one bug id if possible.
Attaching bug ids give us the ability to back track history reasons and
understand the context behind a patch.

### DIFF
--- a/kokoro/presubmit/presubmit.sh
+++ b/kokoro/presubmit/presubmit.sh
@@ -101,6 +101,27 @@ function run_gazelle() {
   $BAZEL run gazelle
 }
 
+function run_commit_message() {
+  echo
+  currentbranch=`git symbolic-ref HEAD | sed -E 's/refs\/heads\/(.*)/\1/'`
+  upstream=`git config branch.${currentbranch}.merge`
+  remote=`git config branch.${currentbranch}.remote`
+  if [[ ${upstream} != *"refs/heads/"* ]]; then
+    upstream="refs/heads/${upstream}"
+  fi
+  remoteupstream=`echo ${upstream} | sed -E "s/heads/remotes\/${remote}/"`
+  allcommits=`git log ${remoteupstream}.. --format=%H`
+  commits=($(echo $allcommits))
+  IFS=
+  commit=`git log --format=%B ${commits[${#commits[@]} - 1]}^!`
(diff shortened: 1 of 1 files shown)


lpy:
PTAL.

Intentionally not attaching a bug field to see whether presubmit will fail :)

lpy:
hmmmm doesn't work on kokoro :/

lpy:
> This will just cause commits to have something like "Bug: None" in them, which is just noisy.

Those tend to be small fixes? Otherwise they should have bugs filed and associated with them.

> Also, presubmits run on a PR, which can contain multiple commits. Why should only the head commit of the PR contain a bug line?

I intentionally avoid checking every commit but the first one. If I understand correctly the first commit message will always show up in the PR description when creating the PR. Followup commits may just address comments and hence it makes sense to not enforce on those.

pmuetschard:
re multi commits: no that is not the way it works. Many of my PRs start out with multiple commits.

hevrard:
This is a process discussion. Enforcing bug id in commit is good for tracking history, although it can be cumbersome for small fixes without any related bug, yet these small fixes can be committed with "bug: none".

I think it would make sense to look for the "bug:" string in the body of all commits of the PR, for PRs that come as a series of commits. Note that this does NOT guarantee that the developer keeps the "bug: 1234" string when she squashes/merges, yet we can expect the dev to keep this text by default.

Is checking for the "bug:" string in all commits of the PR an acceptable middle-ground?

pau-baiget:
Ah, I brought this discussion a few months back (not very successfully), so obviously I like the idea.

I think it's good to have the Bug: ID line, but enforced at "Squash and Merge" time (so it contains all the commits related to that bug). We just need to agree on what word do we use when no bug is associated. "None" is good for me.

@lpy can you please look into into whether it's possible to have this behavior?

AWoloszyn:
I will also note: WHICH bug-id goes here?
We have multiple bug-tracking locations. How do we plan on differentiating?

hevrard:
We could do: b/1234 and #1234 ?

ben-clayton:
While I don't necessarily agree with requiring a bug line, SwiftShader uses this:
https://swiftshader.googlesource.com/SwiftShader/+/refs/heads/master/tests/presubmit.sh#41

PR #3653, opened 2020-01-14
Various counter UI enhancements

pmuetschard:
- show process (atrace) counters
- fix rendering of counters without data at the beginning/end of a trace
- remove the max value label
- render counters that are always 0 as a line, rather than a bar
- use the units to format the values
- don't make NaN and +/- infinity mess up the scaling
- percentage counters use a 0-100 range (or bigger)
- non-idle threads are expanded by default

Bug: http://b/143885962
Bug: http://b/142338319
Bug: http://b/141309596
Bug: http://b/140034544

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/Unit.java
+++ b/gapic/src/main/com/google/gapid/perfetto/Unit.java
@@ -1,0 +1,441 @@
+/*
+ * Copyright (C) 2020 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


lpy:
I have verified below UI behaviour on latest user build:

* show process (atrace) counters
* remove the max value label
* render counters that are always 0 as a line, rather than a bar
* use the units to format the values
* percentage counters use a 0-100 range (or bigger)
* non-idle threads are expanded by default

image
image

Not sure how to verify below:

* fix rendering of counters without data at the beginning/end of a trace
* don't make NaN and +/- infinity mess up the scaling

gapic/src/main/com/google/gapid/perfetto/Unit.java
```diff
@@ -0,0 +1,426 @@
...
+      private final int offset;
+
+      public Base2(String[] names, int offset) {
+        this.names = names;
+        this.offset = offset;
+      }
+
+      @Override
+      public String format(long value) {
+        if (value < 1536) {
```
stellama0208: nit: Can give a small comment explaining the magic number :)

PR #3655, opened 2020-01-15
Crash early upon cmd.Mutate() error in all MutateAndWrite()

hevrard:
If cmd.Mutate() leads to an error, then we cannot trust the state to
be valid anymore. MutateAndWrite() wraps calls to cmd.Mutate(), this
change makes sure to crash early (here, using log.F()) upon any
cmd.Mutate() error.

We can think of propagating the error above MutateAndWrite(). Yet,
MutateAndWrite() is part of the TransformWriter interface, and there
is no obvious way to enforce developers working on Transforms to
properly handle possible errors returned by MutateAndWrite() in the
logic of their transform.

As a middle ground, we choose to have MutateAndWrite() crash early
upon any cmd.Mutate(). Again, we cannot strictly enforce that, it is
just documented in the interface comment.

Bug: b/145003736

### DIFF
--- a/gapis/api/transform/transformer.go
+++ b/gapis/api/transform/transformer.go
@@ -55,7 +55,8 @@ type Writer interface {
 	// State returns the state object associated with this writer.
 	State() *api.GlobalState
 	// MutateAndWrite mutates the state object associated with this writer,
-	// and it passes the command to further consumers.
+	// and it passes the command to further consumers. It must panic if the
+	// mutation fails with an error.
 	MutateAndWrite(ctx context.Context, id api.CmdID, cmd api.Cmd)
 	//Notify next transformer it's ready to start loop the trace.
 	NotifyPreLoop(ctx context.Context)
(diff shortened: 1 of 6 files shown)


gapis/api/transform/transformer.go
```diff
@@ -55,7 +55,8 @@ type Writer interface {
 	// State returns the state object associated with this writer.
 	State() *api.GlobalState
 	// MutateAndWrite mutates the state object associated with this writer,
-	// and it passes the command to further consumers.
+	// and it passes the command to further consumers. It must not survive
```
purvisa-at-google-com: "must not survive" is strange english. Maybe something like "Will panic if mutate fails with an error".

purvisa-at-google-com: Actually are we sure log.F does panic?

hevrard: Thanks for the English wording suggestion, applied.

log.F(ctx, **true**) will panic here: https://github.com/google/gapid/blob/ca1cb2f8866c17ef258050ad34332748cab0ebb2/core/app/log.go#L46
Because gapis is started as an app, here:
https://github.com/google/gapid/blob/ca1cb2f8866c17ef258050ad34332748cab0ebb2/cmd/gapis/main.go#L71

hevrard:
After offline discussion, it is not desirable to crash/panic.
This PR is superseded by #3666 for now, the places where no error propagation can be easily done will be covered later.

PR #3656, opened 2020-01-15
Marshal ProfilingData message to text instead of json in gapit

apbodnar:
Fix for b/147736754  Infinities emitted by the QC producer break converting the proto to json when calling `gapit profile <some.gfxtrace>`  Writing text works fine though.

### DIFF
--- a/cmd/gapit/flags.go
+++ b/cmd/gapit/flags.go
@@ -384,6 +384,7 @@ type (
 	GpuProfileFlags struct {
 		Gapis GapisFlags
 		Gapir GapirFlags
+		Json  bool `help:"Return replay profiling data as JSON instead of text"`
 	}
 
 	CreateGraphVisualizationFlags struct {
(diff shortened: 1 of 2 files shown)


AWoloszyn:
Could we not just fix up the data?
Also why are infinities being returned at all? that sounds like a bug. The answer for any profiling thing should probably NOT be inf.

apbodnar:
It _is_ a bug (b/146891559) I think having special sanitization logic is excessive for this feature, and I rather not trust all present and future producers to not behave poorly.

AWoloszyn:
That is fair, it is just that text output is significantly less useful than json (as a rule). It's a shame to lose the json.

yalcinmelihyasin:
Is there a reason that we don't always return json? or Proto?

apbodnar:
Supporting both is close to trivial. JSON is preferable, but proto is more robust if a producer emitting bad values, so it's the default.

PR #3662, opened 2020-01-16
System profiler - Add interactive access to trace to GAPIT

zakerinasab:
This change adds a new verb 'perfetto' to GAPIT which allows
interaction with a trace through shell. This will be followed
by another change that adds running metrics to GAPIT.

Doc: go/gapit-perfetto-stats
Tests: go/gapit-perfetto-stats-tests

### DIFF
--- a/cmd/gapit/flags.go
+++ b/cmd/gapit/flags.go
@@ -43,6 +43,17 @@ const (
 	ExportTimestamps
 )
 
+const (
+	ModeMetrics PerfettoMode = iota
+	ModeInteractive
+)
+
+const (
+	OutputDefault PerfettoOutputFormat = iota
+	OutputText
+	OutputJson
+)
+
 type VideoType uint8
 
 var videoTypeNames = map[VideoType]string{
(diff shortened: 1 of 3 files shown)


cmd/gapit/flags.go
```diff
@@ -409,4 +409,13 @@ type (
 		DeviceFlags
 		Gapis GapisFlags
 	}
+
+	PerfettoFlags struct {
+		Mode				string	`help:"Run mode: {metrics|interactive}. Default: metrics."`
```
AWoloszyn: Instead of a string you can use an enum here.

See core/log/severity.go for how this is done for (--log-level "Debug|Verbose|....)

cmd/gapit/flags.go
```diff
@@ -409,4 +409,13 @@ type (
...
+		Trace				string	`help:"Trace file."`
```
AWoloszyn: For most other gapit commands, the trace is always the last argument (and is not in the named list)

cmd/gapit/flags.go
zakerinasab: Done.

cmd/gapit/flags.go
zakerinasab: Done.

Issue #3657, opened 2020-01-15
[Frame loop] detects all images and buffers as changed if there's memorybarrier in the trace.


RenfengLiu:
The reason is https://github.com/google/gapid/blob/master/gapis/api/vulkan/api/synchronization.api#L624  will mark all buffers and images as dirty.  This will cause OOM when looping trace for big game.

PR #3663, opened 2020-01-16
Fix an issue that frame loop detects all images and buffers as changed

RenfengLiu:
This is to fix issue #3657

### DIFF
--- a/gapis/api/vulkan/frame_loop.go
+++ b/gapis/api/vulkan/frame_loop.go
@@ -486,10 +486,6 @@ func (f *frameLoop) writeLoopContents(ctx context.Context, cmd api.Cmd, out tran
 
 	// Iterate through the loop contents, emitting instructions one by one.
 	for cmdIndex, cmd := range f.capturedLoopCmds {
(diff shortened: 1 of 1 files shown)


gapis/api/vulkan/frame_loop.go
```diff
@@ -527,6 +538,32 @@ func (f *frameLoop) buildStartEndStates(ctx context.Context, startState *api.Glo
...
+	st := GetState(currentState)
+	st.PreSubcommand = func(i interface{}) {
+		cr, ok := i.(CommandReferenceʳ)
+		if ok {
+			args := GetCommandArgs(ctx, cr, st)
+			switch ar := args.(type) {
+			case VkCmdBeginRenderPassArgsʳ:
+				rp := st.RenderPasses().Get(ar.RenderPass())
+				f.watcher.ignore = true
+				for i := uint32(0); i < uint32(rp.AttachmentDescriptions().Len()); i++ {
```
AWoloszyn: I think there are more cases here that I missed.

IF the renderpass has LOAD_OP_CLEAR, then it really is a write (it will overwrite the renderpass)
possibly also if the renderpass has LOAD_OP_DONT_CARE (since that may overwrite the image as well)

PR #3665, opened 2020-01-16
[NOT FOR SUBMIT] Always assume render stage producer exists.

lpy:
This patch is not for review nor submit.

Bug: b/147723153


lpy:
To unblock partners while we check in the real solution for render stage producer sensing, I am going to use this PR to get builds from kokoro and share with them.

PR #3667, opened 2020-01-17
Add defensive check on inFlightRPCs counter

hevrard:
This is to chase a flaky issue of connection being lost between a
client and a server, without apparent reason.

Bug: b/144158856


yalcinmelihyasin:
LGTM but please take another eye.

PR #3668, opened 2020-01-17
Remove unaccessible counter.

lpy:
Some counters are not available on production build, hence remove them.

Bug: b/138717619


lpy:
ptal

lpy:
rerun kokoro before submit

lpy:
Alright, looks like it's a bug on Adreno platform :)

PR #3674, opened 2020-01-21
Add translation for command buffer handles in GPU slices

chrisforbes:
This is preparation for determining proper "link" values to emit.

Bug: b/147982708


chrisforbes:
Note: this is not going to work properly on P4 until we also force use of the prerelease driver in this scenario.

Issue #3068, opened 2019-08-15
If the replay service crashes, we should restart gapir.


AWoloszyn:
Ideally we don't crash gapir at all, but that is a longer term problem.

pau-baiget:
Tracked internally.

yalcinmelihyasin:
Fixed #3411

yalcinmelihyasin:
The fix is actually fixing if only communication is broken not when crashed.

yalcinmelihyasin:
Suggested but rejected : #3613

yalcinmelihyasin:
b/139480352

yalcinmelihyasin:
PR: #3675

yalcinmelihyasin:
Fixed with #3675

PR #3675, opened 2020-01-21
Refactoring gapis-gapir connection

yalcinmelihyasin:
This is a general refactoring regarding the gapis-gapir connection which simplifies the architecture and fixes the restarting gapir issue

See b/139480352 for details


yalcinmelihyasin:
issue #3068

gapir/api.go
```diff
@@ -51,24 +51,26 @@ type (
...
 	// HandleFenceReadyRequest handles the profiler ready message.
-	HandleFenceReadyRequest(context.Context, *FenceReadyRequest, Connection) error
+	HandleFenceReadyRequest(context.Context, *FenceReadyRequest) error
 }
 
 // Connection represents a connection between GAPIS and GAPIR. It wraps the
 // internal gRPC connections and holds authentication token. A new Connection
 // should be created only by client.Client.
+
+// TODO: We should update this comment.
```
hevrard: Given the current lack of design, the original comment does not make much sense, and brings more confusion than anything else. I would be keen to just remove it, rather than commenting on top of it.

gapir/client/background_connection.go
```diff
@@ -0,0 +1,179 @@
+// Copyright (C) 2017 Google Inc.
```
hevrard: Copyright 2019

gapir/client/background_connection.go
```diff
@@ -0,0 +1,179 @@
...
+// ReplayExecutor handles just the bits related to a
+// specific replay.
+type ReplayExecutor interface {
+	// HandlePostData handles the given post data message.
+	HandlePostData(context.Context, *gapir.PostData) error
+	// HandleNotification handles the given notification message.
+	HandleNotification(context.Context, *gapir.Notification) error
+	// HandleFinished is notified when the given replay is finished.
+	HandleFinished(context.Context, error) error
+
```
hevrard: Missing description comment

gapir/client/background_connection.go
```diff
@@ -0,0 +1,179 @@
...
+func (bgc *backgroundConnection) BeginReplay(ctx context.Context, payload string, dependent string) error {
+	return bgc.conn.BeginReplay(ctx, payload, dependent)
+}
+
+func (bgc *backgroundConnection) PrewarmReplay(ctx context.Context, payload string, cleanup string) error {
+	return bgc.conn.PrewarmReplay(ctx, payload, cleanup)
+}
+
+func (bgc *backgroundConnection) SetReplayExecutor(ctx context.Context, x ReplayExecutor) (func(), error) {
```
hevrard: rename variable `x` by `executor`, or something more explicit

gapir/client/background_connection.go
```diff
@@ -0,0 +1,179 @@
...
+
+func (bgc *backgroundConnection) HandleFinished(ctx context.Context, err error) error {
+	if bgc.executor == nil {
+		return log.Err(ctx, nil, "No active replay connection for this returned data")
+	}
+	return bgc.executor.HandleFinished(ctx, err)
+}
+
+// HandlePostData handles the given post data message.
+func (bgc *backgroundConnection) HandlePostData(ctx context.Context, pd *gapir.PostData) error {
```
hevrard: rename `pd` to `postData`

gapir/client/background_connection.go
```diff
@@ -0,0 +1,179 @@
...
+
+	pid, err := id.Parse(payloadID)
+	if err != nil {
+		return log.Errf(ctx, err, "Parsing payload ID")
+	}
+	boxed, err := database.Resolve(ctx, pid)
+	if err != nil {
+		return log.Errf(ctx, err, "Getting replay payload")
+	}
+	if pl, ok := boxed.(*gapir.Payload); ok {
```
hevrard: `pl` -> `payload`

gapir/client/client.go
```diff
@@ -32,84 +37,195 @@ type tyLaunchArgsKey string
...
-	mutex    sync.Mutex
-	sessions map[deviceArch]*session
+	mutex       sync.Mutex
+	clientInfos map[ConnectionKey]clientInfo
 }
 
+const (
+	connectTimeout    = time.Second * 10
+	heartbeatInterval = time.Millisecond * 500
+)
```
hevrard: move these const declaration to the top, merge it with the declaration of `LaunchArgsKey`

gapir/client/client.go
```diff
@@ -32,84 +37,195 @@ type tyLaunchArgsKey string
...
+
+type deviceArch struct {
+	device bind.Device
+	arch   device.Architecture
+}
+
+// ConnectionKey is used by manager to obtain a connection
+type ConnectionKey deviceArch
+
 // Client is interface used to connect to GAPIR instances on devices.
```
hevrard: Update comment (especially, this is not an interface, so the current comment is misleading):
`// Client handles connections to GAPIR instances on devices. A single Client can handle multiple connections.`

gapir/client/client.go
```diff
@@ -32,84 +37,195 @@ type tyLaunchArgsKey string
...
+	deviceArch := deviceArch{device: device, arch: abi.GetArchitecture()}
+	key := ConnectionKey(deviceArch)
+
+	if clientInfo, ok := client.clientInfos[key]; ok {
+		if clientInfo.abi.SameAs(abi) {
+			return &key, nil
+		}
+
+		// TODO: Should we reconnect instead of disconnect?
+		client.closeConnection(ctx, key)
```
hevrard: The `key` depends on the ABI, so the same device with two different ABI should lead to two different keys.
If this is correct, then your test here is probably not useful, as the same key cannot lead to different ABIs. In other words, `clientInfo.abi.SameAs(abi)` is always true. If so, maybe we should check that the connection already registered in `client.clientInfos[key].connection` is still valid or not. If it is, just return `&key`, if not, then re-create it.

gapir/client/client.go
```diff
@@ -32,84 +37,195 @@ type tyLaunchArgsKey string
...
+
+	log.I(ctx, "Heartbeat connection setup done")
+
+	bgConnection, err := client.makeBackgroundConnection(ctx, device, connection)
+	if err != nil {
+		return nil, log.Err(ctx, err, "Background connection error")
 	}
 
-	return s.connect(ctx)
+	client.clientInfos[key] = clientInfo{deviceConnectionInfo: *newDeviceConnectionInfo, connection: connection, device: device, arch: abi.Architecture, abi: abi, bgConnection: bgConnection}
```
hevrard: I would prefer this indentation:
```
	client.clientInfos[key] = clientInfo{
		deviceConnectionInfo: *newDeviceConnectionInfo,
		connection:           connection,
		device:               device,
		arch:                 abi.Architecture,
		abi:                  abi,
		bgConnection:         bgConnection}
```

gapir/client/client.go
```diff
@@ -32,84 +37,195 @@ type tyLaunchArgsKey string
...
-	d bind.Device
-	a device.Architecture
+	return client
 }
 
 // Connect opens a connection to the replay device.
-func (c *Client) Connect(ctx context.Context, d bind.Device, abi *device.ABI) (gapir.Connection, error) {
+func (client *Client) Connect(ctx context.Context, device bind.Device, abi *device.ABI) (*ConnectionKey, error) {
+	client.mutex.Lock()
+	defer client.mutex.Unlock()
```
hevrard: Concurrency is never easy: can you add a small comment at the top of the file, saying why these internal functions needs to be thread-safe?

gapir/client/client.go
```diff
@@ -32,84 +37,195 @@ type tyLaunchArgsKey string
...
+		}
 	}
-	return r
+}
+
+func (client *Client) BeginReplay(ctx context.Context, conn *ConnectionKey, payload string, dependent string) error {
+	return client.clientInfos[*conn].bgConnection.BeginReplay(ctx, payload, dependent)
+}
+
+func (client *Client) SetReplayExecutor(ctx context.Context, conn *ConnectionKey, x ReplayExecutor) (func(), error) {
```
hevrard: `x` -> `executor`

gapir/client/background_connection.go
AWoloszyn: Copyright 2020

gapir/client/background_connection.go
yalcinmelihyasin: @apbodnar can you please give context here so that I can update the comment

gapir/client/client.go
yalcinmelihyasin: You are right it's a leftover from the old code. For the second part, if the connection is invalid, it's being recreated. So, if a connection exists then it's valid.

gapir/api.go
yalcinmelihyasin: I am actually waiting suggestion for it. We should update before the merge.

gapir/client/client.go
```diff
@@ -28,88 +33,201 @@ import (
...
+// ConnectionKey is used by manager to obtain a connection
+type ConnectionKey deviceArch
 
-// Client is interface used to connect to GAPIR instances on devices.
+// Client handles connections to GAPIR instances on devices.
+// A single Client can handle multiple connections.
 type Client struct {
-	mutex    sync.Mutex
-	sessions map[deviceArch]*session
+	// Mutex is needed due to reconnect may happen in another thread
```
hevrard: You cannot say "due to ... may happen", but rather: `// Mutex is needed because reconnect() may happen in another thread`

gapir/client/client.go
hevrard: still `x`...?

gapir/client/device_connection.go
```diff
@@ -384,72 +332,16 @@ func (s *session) newADB(ctx context.Context, d adb.Device, abi *device.ABI, lau
...
-			if err != nil {
-				log.E(ctx, "Error sending keep-alive ping. Error: %v", err)
-				s.close(ctx)
-				return
-			}
-		}
-	}
+	return &deviceConnectionInfo{port: port, authToken: "", cleanupFunc: func() {
+		cleanup.Invoke(ctx)
+		d.RemoveForward(ctx, localPort)
```
hevrard: Please declare cleanupFunc outside of the struct field initializer, as you did earlier in your refactoring:
```
cleanupFunc := func() { ... }
...
return &deviceConnectionInfo{..., cleanupFunc: cleanupFunc}
```

hevrard: Also, is there a good reason why we don't use authToken for ADB connections?

gapis/replay/batch.go
```diff
@@ -269,7 +269,7 @@ func (m *manager) execute(
 	}
 	ctx = log.V{"replay target ABI": replayABI}.Bind(ctx)
 
-	connection, err := m.connect(ctx, d, replayABI)
+	conn, err := m.connect(ctx, d, replayABI)
```
hevrard: I would be keen to keep the long, explicit `connection` name.

gapis/replay/executor.go
```diff
@@ -22,6 +22,7 @@ import (
 	"github.com/google/gapid/core/log"
 	"github.com/google/gapid/core/os/device"
 	"github.com/google/gapid/gapir"
+	gapirClient "github.com/google/gapid/gapir/client"
```
hevrard: There is no other "client" package here, so no need to have the `gapirClient` ?

gapis/replay/executor.go
```diff
@@ -50,7 +51,8 @@ func Execute(
 	handlePost builder.PostDataHandler,
 	handleNotification builder.NotificationHandler,
 	fenceReadyCallback builder.FenceReadyRequestCallback,
-	connection *backgroundConnection,
+	m Manager,
```
hevrard: `m` is too short, let's use `manager`

gapis/replay/executor.go
```diff
@@ -68,38 +70,38 @@ func Execute(
 		memoryLayout:       memoryLayout,
 		OS:                 os,
 		finished:           make(chan error),
-	}.execute(ctx, connection)
+	}.execute(ctx, m.(*manager), conn)
 }
 
-func (e executor) execute(ctx context.Context, connection *backgroundConnection) error {
+func (e executor) execute(ctx context.Context, m *manager, conn *gapirClient.ConnectionKey) error {
```
hevrard: `m` -> `manager`

test/integration/replay/postback_test.go
```diff
@@ -83,7 +78,9 @@ func doReplay(t *testing.T, f func(*builder.Builder)) error {
 		return err
 	}
 
-	err = replay.Execute(ctx, "", payload, decoder, notification, fenceReady, bgc, abi.MemoryLayout, os)
+	m := replay.NewManagerForTest(client)
```
hevrard: `m` -> `manager`

gapir/client/device_connection.go
yalcinmelihyasin: I don't see a reason for the first comment.

yalcinmelihyasin: For the second question, I do not know. I have not touched the internals of the implementations. I cannot give any information but it was never set to any value.

gapis/replay/batch.go
yalcinmelihyasin: there is a struct with the same name. I think in this situation shortening is better to not get confused.

gapir/client/device_connection.go
hevrard: OK for authToken, let's keep it to `""`.

For cleanupFunc, I would prefer to have it defined outside because:
- this is how it is done in the other similar `newXXX` functions
- having the lambda defined directly here reduces code readability, the `return` line becomes quite complex

gapis/replay/batch.go
hevrard: Fair enough.

gapir/client/device_connection.go
AWoloszyn: We should be generating an auth-token for Android as well. I am not sure why we are not, but we absolutely should be.
```
authToken := auth.GenToken()
``` 
in newADB should do it.

gapir/client/client.go
AWoloszyn: gofmt should be able to handle that?

PR #3676, opened 2020-01-21
Add error reporting to Transforms

hevrard:
Add systematic error reporting to Transforms. In practice, this means adding
error reporting the following:
- gapis/api/transform.Transforms.Tranform()
- gapis/api/transform.Writer:MutateAndWrite()
- gapis/api/transform.Writer:Flush()
- gapis/api/transform.Writer:Transform()

This PR is split in separate commits for each of those.

Bug: b/145003736

### DIFF
--- a/test/integration/gles/replay/shaders.go
+++ b/test/integration/gles/replay/shaders.go
@@ -29,7 +29,7 @@ uniform float angle;
 void main() {
 	float c = cos(angle);
 	float s = sin(angle);
-	mat3 rotation = mat3(vec3(c, -s, 0.0), vec3(s, c, 0.0), vec3(0.0, 0.0, 1.0));
+		mat3 rotation = mat3(vec3(c, -s, 0.0), vec3(s, c, 0.0), vec3(0.0, 0.0, 1.0));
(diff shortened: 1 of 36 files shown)


test/integration/gles/replay/shaders.go
```diff
@@ -29,7 +29,7 @@ uniform float angle;
 void main() {
 	float c = cos(angle);
 	float s = sin(angle);
-		mat3 rotation = mat3(vec3(c, -s, 0.0), vec3(s, c, 0.0), vec3(0.0, 0.0, 1.0));
+	mat3 rotation = mat3(vec3(c, -s, 0.0), vec3(s, c, 0.0), vec3(0.0, 0.0, 1.0));
```
AWoloszyn: Probably best left to a separate CL.

hevrard: ok, reverted.

hevrard:
> Have we done any performance measurements on this?

I just ran experiments on three non-trivial games, I cannot see any performance impact, results are in the same range with and without error reporting in place.

PR #3682, opened 2020-01-21
Always vulkan sample app run in portrait mode on Android.

lpy:
Currently on Android the vulkan sample app doesn't handle screen rotation, and
we don't need it to react to screen rotation. Hence always run it in portrait
mode.

This patch also enables the apk to use developer driver.

Bug: b/148104724


lpy:
That's exactly why I uploaded this patch :D

PR #3685, opened 2020-01-23
Adjust "Goto Command" box to support specifying a .-delimited subcommand

chrisforbes:
We're about to have the replay profiler slices that link to specific
subcommands. A reasonable step along the way is to be able to specify
subcommands using this dialog.

Converts the spinner widget to a textbox.

Navigating to a subcommand doesn't actually work yet -- we need to build
out a bit more backend plumbing for that, but this is a first step.

Bug: b/143735500


chrisforbes:
Not ready to merge:
- Failure to parse the command index is currently nasty as we just allow an exception to escape.
- In backend CommandTreeNodeForCommand needs to be enhanced to support subcommands

chrisforbes:
Moved to AGI.

Issue #3067, opened 2019-08-15
Fix gapir crash reporting.


AWoloszyn:
Our crash reporting is not working correctly.

natduca:
@pmuetschard will add some technical details and concrete steps to fix this shortly

pmuetschard:
- Get the minidump export to work properly again
- Make sure they get uploaded to the crash server
- Make uploading the symbols part of the release
- Investigate if we can get crash reporting when tracing working? should we?

yalcinmelihyasin:
#3503 ensures that the crash report uploaded correctly.

yalcinmelihyasin:
#3686 for building symbol files

chrisforbes:
What remains to be done here?

yalcinmelihyasin:
Actually uploading them

yalcinmelihyasin:
Fixed! b/139480214

yalcinmelihyasin:
Merged into this repo by #3787

PR #3686, opened 2020-01-23
Build symbol files of gapir for Android

yalcinmelihyasin:
Build symbol files of gapir for Android so that we can get a clear debug information in breakpad

b/139480214

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -91,6 +91,20 @@ filegroup(
     }),
 )
 
+filegroup(
+    name = "symbols",
+    srcs = [
+        "//cmd/gapir/cc:gapir.sym",
+    ] + select({
+        "//tools/build:linux": [
+            "//gapidapk/android/apk:arm64-v8a_gapir.sym",
+            "//gapidapk/android/apk:armeabi-v7a_gapir.sym",
+            "//gapidapk/android/apk:x86_gapir.sym",
+        ],
+        "//conditions:default": [],
+    }),
+)
+
 # Rules to build the expected installed structure for running
 filegroup(
     name = "gfxapi-ls",
--- a/kokoro/linux/build.sh
+++ b/kokoro/linux/build.sh
@@ -61,7 +61,7 @@ for api in gles vulkan gvr; do
 done
 
 # Build the package and symbol file.
-build //:pkg //cmd/gapir/cc:gapir.sym
+build //:pkg //:symbols
 
 # Build the Vulkan sample
 build //cmd/vulkan_sample:vulkan_sample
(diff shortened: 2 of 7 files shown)


yalcinmelihyasin:
issue: #3067

kokoro/linux/build.sh
```diff
@@ -81,6 +81,9 @@ do
 done
 echo $(date): Smoketests completed.
 
+# Build the symbol files for Android
+build -c dbg //gapidapk/android/apk:armeabi-v7a_gapir.sym //gapidapk/android/apk:arm64-v8a_gapir.sym //gapidapk/android/apk:x86_gapir.sym
```
hevrard: Does it has to be `-c dbg`? This probably means you will have a lot of things to recompile in debug mode.

yalcinmelihyasin: When I try to build without -c, the function name was still available but the line information disappeared. Is this something we accept?

AWoloszyn: -c dbg will also not produce the same binary at all.

pmuetschard: On the build systems, we build with `-config symbols`, which adds `-g`, which should be enough to get the complete debug info for symbols, is it not?

BUILD.bazel
```diff
@@ -91,6 +91,15 @@ filegroup(
...
 
+filegroup(
+    name = "android_symbols",
+    srcs = [
+        "//gapidapk/android/apk:arm64-v8a_gapir.sym",
+        "//gapidapk/android/apk:armeabi-v7a_gapir.sym",
+        "//gapidapk/android/apk:x86_gapir.sym",
+    ],
+)
+
```
pmuetschard: Change to:
```
filegroup(
  name = "symbols",
  srcs = [
    "//cmd/gapir/cc:gapir.sym",
  ] + select({
   "//tools/build:linux": [
       "//gapidapk/android/apk:arm64-v8a_gapir.sym",
       "//gapidapk/android/apk:armeabi-v7a_gapir.sym",
       "//gapidapk/android/apk:x86_gapir.sym",
    ],
    "//conditions:default": [],
  }),
)
```

Then change the build scripts on all platforms to build the symbols target, so that we can control what gets built from the build files, rather than mucking around with the scripts.

tools/build/rules/cc.bzl
```diff
@@ -93,7 +93,7 @@ strip = rule(
 # Symbol rule implementation, which invokes the _dump_syms binary to generate
 # a symbol dump file that can be uploaded to the crash server to symbolize
 # stack traces of uploaded crash dumps.
-def _symbols_impl(ctx):
+def symbols_impl(ctx):
```
pmuetschard: The implementation function should stay private.

PR #3692, opened 2020-01-24
A few small optimizations.

AWoloszyn:
Removing an unneeded std::mutex for every call to spy::get()
Replace a bunch of new/delete pairs with stack-scoped objects.

BUG: No Bug

### DIFF
--- a/gapii/cc/spy.h
+++ b/gapii/cc/spy.h
@@ -27,6 +27,7 @@
 #include <unordered_map>
 
 namespace gapii {
+struct spy_creator;
 class ConnectionStream;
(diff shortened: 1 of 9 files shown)


gapii/cc/spy.h
```diff
@@ -140,6 +141,8 @@ class Spy : public GlesSpy, public GvrSpy, public VulkanSpy {
 
   std::unordered_map<ContextID, GLenum_Error> mFakeGlError;
   std::unique_ptr<core::AsyncJob> mMessageReceiverJob;
+
+friend struct spy_creator;
```
johnplate: Needs indentation to make pre-submit happy

PR #3693, opened 2020-01-24
Detect render stage capability from prerelease driver

rschiu:
The gpu.renderstages producer can be implemented in prerelease driver
UMD or the VkRenderStagesProducer layer.  To properly detect render
stage capacbility, we need to properly set up the environement for
device info service.

Bug: b/142480442

### DIFF
--- a/core/os/android/adb/device.go
+++ b/core/os/android/adb/device.go
@@ -22,9 +22,11 @@ import (
 	"sync"
 	"time"
 
+	"github.com/google/gapid/core/app"
 	"github.com/google/gapid/core/event/task"
 	"github.com/google/gapid/core/fault"
 	"github.com/google/gapid/core/log"
+	"github.com/google/gapid/core/os/android"
 	"github.com/google/gapid/core/os/device"
 	"github.com/google/gapid/core/os/device/bind"
 	"github.com/google/gapid/core/os/shell"
@@ -39,6 +41,8 @@ const (
 	ErrInvalidStatus = fault.Const("Invalid status string")
 	// Frequency at which to print scan errors.
 	printScanErrorsEveryNSeconds = 120
+	// Global settings for opting to use prerelease driver.
+	prereleaseDriverSettingVariable = "game_driver_prerelease_opt_in_apps"
 )
 
 var (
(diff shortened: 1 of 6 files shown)


core/os/android/adb/device.go
```diff
@@ -183,6 +205,16 @@ func newDevice(ctx context.Context, serial string, status bind.Status) (*binding
 		d.To.Configuration.PerfettoCapability = perfettoCapability
 	}
 
+	// If the VkRenderStagesProducer layer exist, we assume the render stages producer is
+	// implemented in the layer.
+	for _, l := range d.To.Configuration.Drivers.Vulkan.Layers {
+		if l.Name == "VkRenderStagesProducer" {
+			d.To.Configuration.PerfettoCapability.GpuProfiling.HasRenderStageProducerLayer = true
+			d.To.Configuration.PerfettoCapability.GpuProfiling.HasRenderStage = true
```
pmuetschard: `PerfettoCapability` and `GpuProfiling` are pointers that could be nil here.

core/os/device/device.proto
```diff
@@ -191,6 +191,7 @@ message GPUProfiling {
   // The GPU performance counters.
   GpuCounterDescriptor gpu_counter_descriptor = 2;
   bool hasFrameLifecycle = 3;
+  bool has_render_stage_producer_layer = 4;
```
lpy: why do we need to distinguish the layer? All we need to know is whether the device has gpu.renderstages producer

rschiu: We need this because if we set gpu_debug_layers to VkRenderStagesProducer and the layer doesn't exist, the app will crash at vkCreateInstance

gapis/perfetto/android/trace.go
```diff
@@ -90,7 +88,7 @@ func setupProfileLayers(ctx context.Context, d adb.Device, driver adb.Driver, pa
 	}
 
 	// Setup render stage layer. Render stage layer should be at the bottom (end of list).
-	if hasRenderStages {
+	if hasRenderStages && d.Instance().GetConfiguration().GetPerfettoCapability().GetGpuProfiling().GetHasRenderStageProducerLayer() {
```
lpy: Maybe just `d.Instance().GetConfiguration().GetPerfettoCapability().GetGpuProfiling().GetHasRenderStageProducerLayer()`, it implies `hasRenderStages`.

rschiu: `hasRenderStages` is the TraceConfig selection.  If the user is not trying to get render stages, then we don't need to enable the layer.

core/os/android/adb/device.go
rschiu: Done.  PTAL.  Thanks.

core/os/android/adb/device.go
```diff
@@ -183,6 +205,27 @@ func newDevice(ctx context.Context, serial string, status bind.Status) (*binding
...
+		if capability == nil {
+			capability = &device.PerfettoCapability{
+				GpuProfiling: &device.GPUProfiling{},
+			}
+		}
+		gpu := capability.GpuProfiling
+		if gpu == nil {
+			gpu = &device.GPUProfiling{}
+			capability.GpuProfiling = gpu
+		}
```
pmuetschard: This should be in the if below, you are doing it for every layer.

rschiu: Done

PR #3694, opened 2020-01-25
Show command buffer vulkan handle for subcommand node.

stellama0208:
- Related ticket: http://b/143734139.

### DIFF
--- a/gapis/resolve/command_tree.go
+++ b/gapis/resolve/command_tree.go
@@ -22,6 +22,7 @@ import (
 	"github.com/google/gapid/core/log"
 	"github.com/google/gapid/core/math/interval"
 	"github.com/google/gapid/gapis/api"
+	"github.com/google/gapid/gapis/api/sync"
 	"github.com/google/gapid/gapis/capture"
 	"github.com/google/gapid/gapis/database"
 	"github.com/google/gapid/gapis/extensions"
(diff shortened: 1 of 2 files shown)


gapis/resolve/command_tree.go
```diff
@@ -512,6 +514,32 @@ func addFrameGroups(ctx context.Context, events *service.Events, p *path.Command
...
+func setSubCmdNodesName(
+	ctx context.Context,
+	r *api.SubCmdRoot,
+	refMap map[api.CmdID][]sync.SubcommandReference,
+	commands []api.Cmd) {
+
+	id := api.CmdID(r.Id[0])
+	refs := refMap[id]
+	r.TraverseSubCmdRoot(func(item *api.SubCmdRoot) {
+		switch len(item.Id) {
```
chrisforbes: This seems a fairly hacky way of distinguishing the different types of nodes. Can we test for the actual conditions we want, or is that data gone by this point?

AWoloszyn: This is not the right place to do this.
We should try and isolate the Vulkan-specific knowledge to the Vulkan package.

AWoloszyn: You can see how we handle debug-marker regions as an example.

stellama0208: Thanks for suggestions! Will learn how debug-marker regions do this for a while and may raise some questions during the process. Before finding a better solution will mark this PR as WIP for now.

chrisforbes:
Superceded; Stella has landed a proper solution in AGI instead.

PR #3697, opened 2020-01-27
Update Perfetto & UI to show vulkan debug markers in the properties pane

silence-do-good:
The debug markers are now direct fields in the gpu_slice table and this
requires some UI changes to be shown properly.

Bug: 140590067


gapic/src/main/com/google/gapid/perfetto/models/SliceTrack.java
```diff
@@ -67,8 +67,48 @@ protected Slice buildSlice(Row row, ArgSet args) {
...
+
+      @Override
+      public ListenableFuture<Slice> getSlice(long id) {
+        return transformAsync(expectOneRow(qe.query(gpuSliceSql(id))), r ->
+        transform(qe.getArgs(r.getLong(8)), args -> buildSlice(r, args)));
+      }
       @Override
       protected Slice buildSlice(Row row, ArgSet args) {
+        if (isParentSlice(args)) {
+          args.addEntry("VkFrameBuffer", row.getString(10) + "(0x" + row.getLong(9) + ")");
```
silence-do-good: Just noticed that these need to be converted to hex first.. Updating that now...

gapic/src/main/com/google/gapid/perfetto/models/SliceTrack.java
```diff
@@ -423,8 +463,8 @@ public Node build(Map<Long, List<Builder>> byParent) {
         "where ts < %d and ts + dur >= %d and depth >= %d and depth <= %d";
 
     private final QueryEngine qe;
-    private final String table;
-    private final long trackId;
+    protected final String table;
+    protected final long trackId;
```
lpy: Let's reorder these to put `protected` and `private` together

gapic/src/main/com/google/gapid/perfetto/models/ArgSet.java
```diff
@@ -30,11 +30,11 @@
   private static final String ARGS_QUERY =
       "select key, int_value, real_value, string_value from args where arg_set_id = %d";
 
-  public static final ArgSet EMPTY = new ArgSet(ImmutableMap.of());
+  public static final ArgSet EMPTY = new ArgSet(new LinkedHashMap());
```
rschiu: Adding the names to the args table effectively nullify the point of putting the debug names into the table.  If GAPID were to add custom UI logic to change how the properties are displayed, the UI code will be the same as before.

silence-do-good: I added it to the args table because we wanted to show the markers in the properties pane. If not, then this can be avoided. Adding some random field to the properties pane without them being in the argset will be a mess of a code.

gapic/src/main/com/google/gapid/perfetto/models/SliceTrack.java
```diff
@@ -67,8 +67,48 @@ protected Slice buildSlice(Row row, ArgSet args) {
...
+      }
+
+      @Override
+      public ListenableFuture<Slice> getSlice(long id) {
+        return transformAsync(expectOneRow(qe.query(gpuSliceSql(id))), r ->
+        transform(qe.getArgs(r.getLong(8)), args -> buildSlice(r, args)));
+      }
       @Override
       protected Slice buildSlice(Row row, ArgSet args) {
+        if (isParentSlice(args)) {
```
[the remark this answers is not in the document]
pmuetschard: - never show handles in the main slice UI (i.e. only show debug names if present)
- show the handles and debug names if present, for *all* slice types in the *selection view*
- only show debug names in the parent slices, do this in the rendering part. We can do this, at least for now with a TODO, by checking the depth. Only show debug names for slices with depth=0 in the main UI.

gapic/src/main/com/google/gapid/perfetto/models/SliceTrack.java
```diff
@@ -67,8 +67,48 @@ protected Slice buildSlice(Row row, ArgSet args) {
 
   public static SliceTrack forGpuQueue(QueryEngine qe, GpuInfo.Queue queue) {
     return new WithQueryEngine(qe, "gpu_slice", queue.trackId) {
+      private final String GPU_COLUMNS = "render_target, render_target_name, render_pass, render_pass_name, command_buffer, command_buffer_name";
```
pmuetschard: We can do this abstraction, I think, in a cleaner way, not requiring `ArgSet` to be mutable. After addressing some of the other comments, I can work with you to do it.

silence-do-good: I've actually cleaned it up myself and removed the changes from ArgSet. Please take a look now.

gapic/src/main/com/google/gapid/perfetto/models/SliceTrack.java
silence-do-good: I think its better to have all the queries together for a neater look. But I can reorder them if needed.

gapic/src/main/com/google/gapid/perfetto/models/SliceTrack.java
```diff
@@ -67,14 +67,76 @@ protected Slice buildSlice(Row row, ArgSet args) {
...
+                new String[rows], new String[rows], new ArgSet[rows]);
+            res.forEachRow((i, row) -> {
+              long start = row.getLong(1);
+              data.ids[i] = row.getLong(0);
+              data.starts[i] = start;
+              data.ends[i] = start + row.getLong(2);
+              data.categories[i] = row.getString(3);
+              data.titles[i] = row.getString(4);
+              data.depths[i] = row.getInt(5);
+              if (data.depths[i] == 0) {
```
rschiu: Add comment saying something like  "Append debug marker of target frame buffer if available."

gapic/src/main/com/google/gapid/perfetto/models/SliceTrack.java
```diff
@@ -242,6 +308,43 @@ public ThreadInfo getThread() {
         return thread;
       }
     }
+
+    public static class GpuSlice extends Slice {
+      private final RenderStageDebugMarkers debugMarkers;
```
rschiu: This should not be named RenderStageDebugMarkers.  The vulkan handles are important whether we have the debug markers are available or not.
Something like GpuInfo would be a better name, and it compliments ThreadInfo.

gapic/src/main/com/google/gapid/perfetto/views/SliceSelectionView.java
```diff
@@ -77,6 +80,29 @@ public SliceSelectionView(Composite parent, State state, SliceTrack.Slice slice)
       createLabel(main, slice.name);
     }
 
+    RenderStageDebugMarkers debugMarkers = slice.getDebugMarkers();
+    if (debugMarkers != null) {
+      LinkedHashMap<String, String> props = new LinkedHashMap<String, String>();
+      if (debugMarkers.frameBufferHandle != 0 || !debugMarkers.frameBufferName.isEmpty()) {
```
rschiu: The handles should be shown whether or not the debug marker exist.

gapic/src/main/com/google/gapid/perfetto/views/SliceSelectionView.java
```diff
@@ -77,6 +80,29 @@ public SliceSelectionView(Composite parent, State state, SliceTrack.Slice slice)
...
+      }
+      if (debugMarkers.renderPassHandle != 0 || !debugMarkers.renderPassName.isEmpty()) {
+        props.put("VkRenderPass:", debugMarkers.renderPassName + "<" + String.format("0x%08X", debugMarkers.renderPassHandle) + ">");
+      }
+      if (debugMarkers.commandBufferHandle != 0 || !debugMarkers.commandBufferName.isEmpty()) {
+        props.put("VkCommandBuffer:", debugMarkers.commandBufferName + "<" + String.format("0x%08X", debugMarkers.commandBufferHandle) + ">");
+      }
+
+      if (!props.isEmpty()) {
+        withLayoutData(createBoldLabel(main, "Debug Markers"),
```
rschiu: Should be named something like "GPU Info" instead of "Debug Markers"

gapic/src/main/com/google/gapid/perfetto/models/SliceTrack.java
silence-do-good: GpuInfo is already used as a class for obtaining the queues. Check GpuInfo.java under perfetto/models. Maybe something like a RenderStageInfo would do.

gapic/src/main/com/google/gapid/perfetto/views/SliceSelectionView.java
silence-do-good: Done

gapic/src/main/com/google/gapid/perfetto/views/SliceSelectionView.java
silence-do-good: Done

gapic/src/main/com/google/gapid/perfetto/models/SliceTrack.java
silence-do-good: Done.

gapic/src/main/com/google/gapid/perfetto/views/SliceSelectionView.java
```diff
@@ -77,6 +80,29 @@ public SliceSelectionView(Composite parent, State state, SliceTrack.Slice slice)
       createLabel(main, slice.name);
     }
 
+    RenderStageInfo renderStageInfo = slice.getRenderStageInfo();
+    if (renderStageInfo != null) {
+      LinkedHashMap<String, String> props = new LinkedHashMap<String, String>();
+      if (renderStageInfo.frameBufferHandle != 0) {
+        props.put("VkFrameBuffer:", renderStageInfo.frameBufferName + "<" + String.format("0x%08X", renderStageInfo.frameBufferHandle) + ">");
```
rschiu: Isn't frameBufferName going to be "NULL" if there is no debug name?

gapic/src/main/com/google/gapid/perfetto/views/SliceSelectionView.java
```diff
@@ -77,6 +80,29 @@ public SliceSelectionView(Composite parent, State state, SliceTrack.Slice slice)
       createLabel(main, slice.name);
     }
 
+    RenderStageInfo renderStageInfo = slice.getRenderStageInfo();
+    if (renderStageInfo != null) {
+      LinkedHashMap<String, String> props = new LinkedHashMap<String, String>();
```
rschiu: A bit nitpicky, but LinkedHashMap seems unnecessary.
You could have done check to see if any handles exist first to call `withLayoutData` and then call `createLabel` directly with each handle.

gapic/src/main/com/google/gapid/perfetto/views/SliceSelectionView.java
silence-do-good: It will be an empty string.

gapic/src/main/com/google/gapid/perfetto/views/SliceSelectionView.java
silence-do-good: It'll be easier to move this if we want, if its stored as a map. Even if we decide to move to the properties pane, we need to know the number of elements that is going to be shown before creating the properties grid. Also, performance wise as well, there wont be any difference.

gapic/src/main/com/google/gapid/perfetto/models/SliceTrack.java
pmuetschard: Please add a TODO here to remove the copy-pasted code.

gapic/src/main/com/google/gapid/perfetto/models/SliceTrack.java
```diff
@@ -242,6 +309,43 @@ public ThreadInfo getThread() {
...
+  }
+
+  public static class RenderStageInfo {
+    public final long frameBufferHandle;
+    public final String frameBufferName;
+    public final long renderPassHandle;
+    public final String renderPassName;
+    public final long commandBufferHandle;
+    public final String commandBufferName;
+    public RenderStageInfo(long frameBufferHandle, String frameBufferName, long renderPassHandle,
```
pmuetschard: add a newline before the constructor.

gapic/src/main/com/google/gapid/perfetto/models/SliceTrack.java
pmuetschard: Yes, let's not impose arbitrary ordering.

gapic/src/main/com/google/gapid/perfetto/views/SliceSelectionView.java
pmuetschard: Please use the factory functions in `Maps`.

gapic/src/main/com/google/gapid/perfetto/views/SliceSelectionView.java
pmuetschard: Put a space between the name and the `<`. Also, if the name is empty, don't show the brackets at all.

gapic/src/main/com/google/gapid/perfetto/models/SliceTrack.java
```diff
@@ -242,6 +309,43 @@ public ThreadInfo getThread() {
...
+      }
+
+      @Override
+      public RenderStageInfo getRenderStageInfo() {
+        return renderStageInfo;
+      }
+    }
+  }
+
+  public static class RenderStageInfo {
```
pmuetschard: I know you've been going the rounds about this name, so ignore this if you are tired of renaming this:

I don't like the name "Render Stages". It's not standard. We should call it `VulkanInfo` or maybe `ApiInfo`. Definitely change the label in the UI to not mention "Render Stages".

silence-do-good: Changed it to Vulkan Info in the UI for now.. Leaving the class name as is and can be changed later.

tools/build/workspace.bzl
```diff
@@ -163,7 +163,7 @@ def gapid_dependencies(android = True, mingw = True, locals = {}):
         name = "perfetto",
         locals = locals,
         remote = "https://android.googlesource.com/platform/external/perfetto",
-        commit = "975bbd74e6b1d7da8dbdaaeec9191fb28471ed2c",
+        commit = "15c8c3e5af6e843dfcaf036333b7ccdfbb1b725f",
```
pmuetschard: Please also update the shallow_since, so it doesn't warn on builds.

silence-do-good: I'll update it separately in a different patch as I'm not sure how to do it. I'm merging this now since stellama is blocked on this and needs the patch urgently.

PR #3698, opened 2020-01-27
Force prerelease driver for tracing and replay

apbodnar:
This forces use of the prerelease driver in all tracing and turns missing profiling capabilities into an error that blocks tracing.

### DIFF
--- a/gapic/src/main/com/google/gapid/views/TracerDialog.java
+++ b/gapic/src/main/com/google/gapid/views/TracerDialog.java
@@ -232,7 +232,7 @@ public class TracerDialog {
       createButton(parent, IDialogConstants.CANCEL_ID, IDialogConstants.CANCEL_LABEL, false);
 
       Listener modifyListener = e -> {
-        ok.setEnabled(traceInput.isReady());
+        ok.setEnabled(traceInput.isReady() && traceInput.getDeviceHasProfilingCapability());
       };
       traceInput.addModifyListener(modifyListener);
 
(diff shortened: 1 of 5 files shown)


gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -232,7 +232,7 @@ protected void createButtonsForButtonBar(Composite parent) {
       createButton(parent, IDialogConstants.CANCEL_ID, IDialogConstants.CANCEL_LABEL, false);
 
       Listener modifyListener = e -> {
-        ok.setEnabled(traceInput.isReady());
+        ok.setEnabled(traceInput.isReady() && traceInput.isDeviceValidated());
```
pmuetschard: Make `isReady` return false, rather than expose the `isDeviceValidated` method

gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -520,6 +511,18 @@ protected void configureDialog(DirectoryDialog dialog) {
         colorFilledInput(widgets.theme);
       }
 
+      private void cacheDeviceIsValidated() {
```
pmuetschard: This is a miss-leading name, please change it. We are planning on caching device validation results between runs, so lets reserve any reference to validation caching to that.

gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -520,6 +511,18 @@ protected void configureDialog(DirectoryDialog dialog) {
...
+        //Skip if the device is not Android device, or trace type is not Perfetto.
```
pmuetschard: Add a space after `//`.

lpy:
I have asked Adam to wait for Adi's validation patch and rebase on top of it once it's landed.

gapic/src/main/com/google/gapid/views/TracerDialog.java
apbodnar: Putting it in `isReady` would cause us to display the `requiredFieldMessage` which would confuse users IMO

silence-do-good:
The validation UI patch is now ready for review #3718

chrisforbes:
Landed in AGI instead.

PR #3700, opened 2020-01-28
Make stage name label bigger (Pipeline View)

hliatis:
Also refactors the layout code a bit as well as fix a minor bug with tables.

### DIFF
--- a/gapic/src/main/com/google/gapid/views/PipelineView.java
+++ b/gapic/src/main/com/google/gapid/views/PipelineView.java
@@ -71,6 +71,7 @@ import org.eclipse.swt.graphics.Point;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Group;
+import org.eclipse.swt.widgets.Label;
 import org.eclipse.swt.widgets.TabFolder;
 import org.eclipse.swt.widgets.TabItem;
 
@@ -197,7 +198,15 @@ implements Tab, Capture.Listener, CommandStream.Listener {
         for (API.Stage stage : stages) {
           TabItem item = createStandardTabItem(folder, stage.getDebugName());
 
-          Group stageGroup = createGroup(folder, stage.getStageName());
+          Composite stageGroup = createComposite(folder, new GridLayout());
+
+          FillLayout nameLayout = new FillLayout(SWT.VERTICAL);
+          nameLayout.marginHeight = 5;
+          Composite nameComposite = withLayoutData( createComposite(stageGroup, nameLayout),
+              new GridData(SWT.BEGINNING, SWT.BEGINNING, false, false));
+
+          Label stageName = createLabel(nameComposite, stage.getStageName() + " (" + stage.getDebugName() + ")");
+          stageName.setFont(theme.bigBoldFont());
 
           item.setControl(stageGroup);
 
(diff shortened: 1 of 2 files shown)


gapic/src/main/com/google/gapid/views/PipelineView.java
```diff
@@ -197,22 +200,35 @@ private void createPipelineTabs(TabFolder folder, List<API.Pipeline> pipelines)
...
+
+          FillLayout nameLayout = new FillLayout(SWT.VERTICAL);
+          nameLayout.marginHeight = 5;
+          Composite nameComposite = withLayoutData( createComposite(stageGroup, nameLayout),
+              new GridData(SWT.BEGINNING, SWT.BEGINNING, false, false));
+
+          Label stageName = createBoldLabel(nameComposite, stage.getStageName() + " (" + stage.getDebugName() + ")");
+          FontData[] fd = stageName.getFont().getFontData();
+          fd[0].setHeight(14);
+          stageName.setFont( new Font(getDisplay(), fd[0]));
```
pmuetschard: Fonts are resources that need to be life-time managed. (i.e. at some point you need to call `.dispose()` on them. Add it to the theme, so this is managed for you.

hliatis: Should I be use bigBoldFont() from Theme.java?

johnplate:
The font size and padding of the stage name label are good now, thanks. The group titles need also adjustments, sorry if this was not clear enough in the design doc.
For example currently "Vertex Attributes" in IA is the same font as the column titles in its table, and it looks squeezed between the tables above and below. So the group title font needs to be a bit bigger than the content titles, but still smaller than the stage name. And it needs more vertical padding, with the padding above the group title about double the size than below.

gapic/src/main/com/google/gapid/views/PipelineView.java
```diff
@@ -197,22 +200,36 @@ private void createPipelineTabs(TabFolder folder, List<API.Pipeline> pipelines)
...
 
-          Group stageGroup = createGroup(folder, stage.getStageName());
+          Composite stageGroup = createComposite(folder, new GridLayout());
+
+          FillLayout nameLayout = new FillLayout(SWT.VERTICAL);
+          nameLayout.marginHeight = 5;
+          Composite nameComposite = withLayoutData( createComposite(stageGroup, nameLayout),
+              new GridData(SWT.BEGINNING, SWT.BEGINNING, false, false));
+
+          Label stageName = createBoldLabel(nameComposite, stage.getStageName() + " (" + stage.getDebugName() + ")");
```
pmuetschard: use `createLabel`, since `createBoldLabel` sets a font, which you then override again.

gapic/src/main/com/google/gapid/widgets/Theme.java
```diff
@@ -293,7 +294,7 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
   @Target(ElementType.METHOD)
   @Retention(RetentionPolicy.RUNTIME)
   public static @interface Text {
-    public static final int Mono = 1, Big = 2, TabTitle = 3;
+    public static final int Mono = 1, Big = 2, TabTitle = 3, Subhead = 4;
```
pmuetschard: `SubTitle` maybe? at least `SubHead`.

pmuetschard:
Regarding font size/layout changes, please make sure to test it on Windows and make sure it looks OK there, don't rely on what it looks like on Linux.

johnplate:
Thanks Pascal, I have a Windows build at home and always check it there as well.

gapic/src/main/com/google/gapid/views/PipelineView.java
```diff
@@ -68,9 +68,12 @@
 import org.eclipse.swt.layout.GridLayout;
 import org.eclipse.swt.layout.RowLayout;
 import org.eclipse.swt.graphics.Point;
+import org.eclipse.swt.graphics.Font;
+import org.eclipse.swt.graphics.FontData;
```
pmuetschard: Unused imports?

PR #3701, opened 2020-01-28
Remove save option in menu for opened system trace.

stellama0208:
Related ticket: http://b/142335577.

### DIFF
--- a/gapic/src/main/com/google/gapid/MainWindow.java
+++ b/gapic/src/main/com/google/gapid/MainWindow.java
@@ -53,6 +53,7 @@ import com.google.gapid.widgets.Theme;
 import com.google.gapid.widgets.Widgets;
 
 import org.eclipse.jface.action.Action;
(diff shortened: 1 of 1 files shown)


gapic/src/main/com/google/gapid/MainWindow.java
```diff
@@ -483,6 +488,7 @@ public void run() {
     }
 
     private Action configure(Action action) {
+      action.setId(label);
```
pmuetschard: Just for our education, how unique do these IDs have to be? Maybe we should use the enum ordinal, since then we can be more certain that it'll be globally unique.

stellama0208: A quick browse into SWT's source code: I think it's fine to go this way. 
This ID is mainly used for finding an item in the menu manager, during the find process there's a basic iteration through all the items, and the find condition is String equality (ignoring case). 
Since all our menu items are named in different strings, I think it's safe here.

PR #3703, opened 2020-01-28
Avoid duplicate Vulkan layer

rschiu:
Bug: b/147832494

### DIFF
--- a/gapis/trace/tracer/tracer.go
+++ b/gapis/trace/tracer/tracer.go
@@ -84,10 +84,15 @@ func LayersFromOptions(ctx context.Context, o *service.TraceOptions) []string {
 	if o.PerfettoConfig == nil {
 		return ret
 	}
+
+	added := map[string]struct{}{}
 	for _, x := range o.PerfettoConfig.GetDataSources() {
 		if layer, err := layout.LayerFromDataSource(x.GetConfig().GetName()); err == nil {
 			if _, err := layout.LibraryFromLayerName(layer); err == nil {
-				ret = append(ret, layer)
+				if _, ok := added[layer]; !ok {
+					added[layer] = struct{}{}
+					ret = append(ret, layer)
+				}
 			}
 		}
 	}


lpy:
Should we do it in https://github.com/google/gapid/blob/master/core/os/android/layers.go#L47?

gapis/trace/tracer/tracer.go
```diff
@@ -87,7 +87,16 @@ func LayersFromOptions(ctx context.Context, o *service.TraceOptions) []string {
 	for _, x := range o.PerfettoConfig.GetDataSources() {
 		if layer, err := layout.LayerFromDataSource(x.GetConfig().GetName()); err == nil {
 			if _, err := layout.LibraryFromLayerName(layer); err == nil {
-				ret = append(ret, layer)
+				skip := false
```
AWoloszyn: I think this is correct, but can I recommend
```
Outer:
  for _, x := range o.PerfettoConfig.GetDataSources() {
    if layer, err := layout.LayerFromDataSource(x.GetConfig().GetName()); err == nil {
	 if _, err := layout.LibraryFromLayerName(layer); err == nil {
            for _, l := range ret {
	        if l == layer {
		   continue Outer
		}
             }
             ret = append(ret, layer)
          }
    }
  }
```

AWoloszyn:
> Should we do it in https://github.com/google/gapid/blob/master/core/os/android/layers.go#L47?

I think this is the correct location, there is no situation in which we would want the same perfetto layer to get added twice.

gapis/trace/tracer/tracer.go
rschiu: Done.  I did not know that was an option!

gapis/trace/tracer/tracer.go
```diff
@@ -84,9 +84,15 @@ func LayersFromOptions(ctx context.Context, o *service.TraceOptions) []string {
...
 	}
+Outer:
 	for _, x := range o.PerfettoConfig.GetDataSources() {
 		if layer, err := layout.LayerFromDataSource(x.GetConfig().GetName()); err == nil {
 			if _, err := layout.LibraryFromLayerName(layer); err == nil {
+				for _, l := range ret {
+					if l == layer {
+						continue Outer
+					}
+				}
```
pmuetschard: Maube put them in a map and then return the map's keys? I guess N^2 here is ok, but...

rschiu: I figure with the typical number of layers we have, the map might make it worse.  But I don't mind changing to map either.

rschiu: Done

rschiu:
@pmuetschard PTAL.  Thanks.

gapis/trace/tracer/tracer.go
```diff
@@ -84,9 +84,15 @@ func LayersFromOptions(ctx context.Context, o *service.TraceOptions) []string {
...
 	}
+
+	added_layer := make(map[string]bool)
 	for _, x := range o.PerfettoConfig.GetDataSources() {
 		if layer, err := layout.LayerFromDataSource(x.GetConfig().GetName()); err == nil {
 			if _, err := layout.LibraryFromLayerName(layer); err == nil {
+				if added_layer[layer] {
+					continue
+				}
+				added_layer[layer] = true
```
pmuetschard: This is OK, but the more, sigh, idiomatic go way to do this is:
```
added := map[string]struct{}{}
for _, x := ... {
  ...
  if _, ok := added[layer]; !ok {
    ret = append(ret, layer)
    added[layer] = struct{}{}
  }
}
```
(also, use camelCase not snake_case)

rschiu: Thanks! Done.

PR #3704, opened 2020-01-28
Don't show the GAPID app in the launcher.

pmuetschard:
Let's not pollute the user's app launcher. Our users don't really have a need to launch the spinning cube from the app launcher.


pmuetschard:
> I assume we can still select the cube app to trace through trace dialog

yes, any action can be selected (we treat MAIN specially as the default)

PR #3705, opened 2020-01-29
Render vk_queue_submit event

rschiu:


lpy:
Let's also update perfetto hash in this patch

silence-do-good:
I've updated perfetto and added some labels in this PR. https://github.com/google/gapid/pull/3697
Is it okay to merge this on top my PR with just updating the perfetto commit ID alone?

lpy:
Adi, you need to update perfetto to the latest commit that includes Raymond's patch, probably this one: https://android.googlesource.com/platform/external/perfetto/+/15c8c3e5af6e843dfcaf036333b7ccdfbb1b725f

gapic/src/main/com/google/gapid/perfetto/models/GpuInfo.java
```diff
@@ -76,6 +76,7 @@ public int bufferCount() {
       res.forEachRow(($, r) -> {
         switch (r.getString(2)) {
           case "gpu_render_stage":
+          case "vulkan_events":
```
pmuetschard: We will have custom rendering for this track, so please don't re-use the queue class, which is used only for GPU queue tracks.

gapic/src/main/com/google/gapid/perfetto/models/GpuInfo.java
```diff
@@ -93,20 +94,22 @@ public int bufferCount() {
   public static class Queue {
     public final int id;
     public final long trackId;
+    public final String name;
     public final int maxDepth;
 
-    public Queue(int id, long trackId, int maxDepth) {
+    public Queue(int id, long trackId, String name, int maxDepth) {
```
pmuetschard: This changes more than what your description says. This changes the name of all the queue tracks.

gapic/src/main/com/google/gapid/perfetto/models/GpuInfo.java
stellama0208: Hi Pascal, and then is it OK to custom this new track as a SliceTrack.forVulkanEvents(...)?  Is this a right place to put it?

pmuetschard: Yes, if the slice track model makes sense, we can use it. But I'm thinking that we may need a different model, since we are not really dealing with slices and need to deal with the submission id linking thing.

stellama0208: I see, I will go along the way of writing a separate track for vulkan_events, in case in the future there comes more interaction between vulkan_events and other panels. 
There might be some duplicates for this track in the very first draft, but it will make this panel more flexible and benefit the future.

gapic/src/main/com/google/gapid/perfetto/models/GpuInfo.java
rschiu: Updated description.

gapic/src/main/com/google/gapid/perfetto/models/GpuInfo.java
rschiu: I have updated the patch with change in perfetto.  This is just so that you can apply the patch and it should work as is.  

Stella, approve this if you like me to check this in.  But it sounds like you may want to do a different approach.  So feel free to just close this PR if it's not going to help you.

stellama0208: Hi Raymond, would do mind only keep the change for perfetto version hash and perfetto settings flag, and have them checked in? I will take care of the other UI bits, thanks!

rschiu: Let's wait for Adi's PR then (https://github.com/google/gapid/pull/3697).  He has updated the hash to be the same as this PR.  I don't want to delay his PR any longer.

silence-do-good: #3697 has now merged

PR #3706, opened 2020-01-29
Clean-up output of gapit devices

### LINKED ISSUE #3659: Redirect logging to stdlog (stderr)
It would be very useful to be able to do

    gapit devices --os android | jq '.devices[].configuration.perfetto_capability'

This would require two changes:

  * Logging can be redirected to `stderr`
  * Devices are dumped as part of the JSON structure, rather than under `-- Device <id>: <hash> --` headers

Effectively, by adding a new `-log-to-stderr` flag (or `-log-to-stdlog`, or whatever else)

    $ gapit -log-to-stderr devices --os android 2>/dev/null
    {
      devices: [
        { ... },
        { ... }
      ]
    }

The other option is to change logging to `stdlog` (i.e. `stderr`) by default

hevrard:
Output a single JSON list containing all info for all devices.
Remove any other output not going through core.log functionnality.

This enables to get a clean JSON output when suppressing gapid
logging, e.g.:
```
gapit -log-level Fatal devices
```

Bug: b/147805965
Fixes #3659

### DIFF
--- a/cmd/gapit/devices.go
+++ b/cmd/gapit/devices.go
@@ -55,12 +55,11 @@ func (verb *devicesVerb) Run(ctx context.Context, flags flag.FlagSet) error {
 		return log.Err(ctx, err, "Failed to get device list")
 	}
 
-	stdout := os.Stdout
 	deviceObjs := []deviceObj{}
 	for _, p := range devices {
 		o, err := client.Get(ctx, p.Path(), nil)
 		if err != nil {
-			fmt.Fprintf(stdout, "%v\n", log.Err(ctx, err, "Couldn't resolve device"))
+			log.Err(ctx, err, "Couldn't resolve device")
 			continue
 		}
 		d := o.(*device.Instance)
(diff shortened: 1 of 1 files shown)


cmd/gapit/devices.go
```diff
@@ -49,25 +55,29 @@ func (verb *devicesVerb) Run(ctx context.Context, flags flag.FlagSet) error {
...
 
-	stdout := os.Stdout
-	for i, p := range devices {
-		fmt.Fprintf(stdout, "-- Device %v: %v --\n", i, p.ID.ID())
+	deviceObjs := []deviceObj{}
+	for _, p := range devices {
 		o, err := client.Get(ctx, p.Path(), nil)
 		if err != nil {
-			fmt.Fprintf(stdout, "%v\n", log.Err(ctx, err, "Couldn't resolve device"))
+			log.Err(ctx, err, "Couldn't resolve device")
```
pmuetschard: This does not do what you think it does. `log.Err` creates a new error, which you are now ignoring.

mattyclarkson:
@hevrard thanks for this

cmd/gapit/devices.go
hevrard: Thanks for this, fix is here: https://github.com/google/gapid/pull/3737

PR #3707, opened 2020-01-29
Fix compilation errors for GCC 9

johnplate:
Fixes b/148513938


yalcinmelihyasin:
Let's wait until all the builds finishes, if they all succeed then we can merge.

PR #3708, opened 2020-01-29
Fix a problem that frame loop doesn't terminate

RenfengLiu:


gapis/api/vulkan/replay.go
```diff
@@ -944,6 +944,17 @@ func (a API) Replay(
 		return numInitialCmdWithoutOpt, nil
 	}
 
+	frameLoopEndCmdID := func(cmds []api.Cmd) api.CmdID {
+		lastCmdID := api.CmdID(0)
```
purvisa-at-google-com: is 0 the right default value here? DO we want len(cmds) -1 ?

RenfengLiu: 0 is correct here, as we want to update it to the last terminated cmd. If non is terminated,  and we set to len(cmds) -1, then it is incorrect.

purvisa-at-google-com: A terminated command is one that completed? Or one that was aborted half way through?

RenfengLiu: A terminated command is one that completed. See gapis/api/cmd.go:69

purvisa-at-google-com: OK, logic makes sense to me now! "Terminated" is an unfortunately ambiguous name though :(

PR #3711, opened 2020-01-30
Make sure Perfetto daemons are running and persistent.

lpy:
Perfetto daemons are not required to run on the device, and the daemons must be
running in order to allow the data producers register themselves, hence before
we do any interaction between the tool and Perfetto, we must make sure Perfetto
daemons are running and persistent.

Note that there's a delay between setting the system property and the daemons
finish starting, hence we need to set the system property as early as possible.

Also update the test faked device to Android 10.

Bug: b/148566762

### DIFF
--- a/core/os/android/adb/adb_data_test.go
+++ b/core/os/android/adb/adb_data_test.go
@@ -194,11 +194,12 @@ untagSocket(48) failed with errno -22
 `),
 
 		// Common responses to all devices
-		stub.Regex(`adb -s .* shell getprop ro\.build\.product`, stub.Respond("hammerhead")),
-		stub.Regex(`adb -s .* shell getprop ro\.build\.version\.release`, stub.Respond("6.0.1")),
-		stub.Regex(`adb -s .* shell getprop ro\.build\.description`, stub.Respond("hammerhead-user 6.0.1 MMB29Q 2480792 release-keys")),
-		stub.Regex(`adb -s .* shell getprop ro\.product\.cpu\.abi`, stub.Respond("armeabi-v7a")),
-		stub.Regex(`adb -s .* shell getprop ro\.build\.version\.sdk`, stub.Respond("26")),
+		stub.Regex(`adb -s .* shell getprop ro\.build\.product`, stub.Respond("flame")),
+		stub.Regex(`adb -s .* shell getprop ro\.build\.version\.release`, stub.Respond("10")),
+		stub.Regex(`adb -s .* shell getprop ro\.build\.description`, stub.Respond("flame-user 10 QQ1A.191003.005 5926727 release-keys")),
+		stub.Regex(`adb -s .* shell getprop ro\.product\.cpu\.abi`, stub.Respond("arm64-v8a")),
+		stub.Regex(`adb -s .* shell getprop ro\.build\.version\.sdk`, stub.Respond("29")),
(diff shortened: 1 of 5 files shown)


lpy:
ptal!

I have tested on Note 10 by doing `adb shell setprop persist.traced.enable 0`, and then starting GAPID, and then doing `adb shell getprop persist.traced.enable` to verify the result.

core/os/android/adb/adb_data_test.go
```diff
@@ -195,10 +195,11 @@ untagSocket(48) failed with errno -22
 
 		// Common responses to all devices
 		stub.Regex(`adb -s .* shell getprop ro\.build\.product`, stub.Respond("hammerhead")),
-		stub.Regex(`adb -s .* shell getprop ro\.build\.version\.release`, stub.Respond("6.0.1")),
-		stub.Regex(`adb -s .* shell getprop ro\.build\.description`, stub.Respond("hammerhead-user 6.0.1 MMB29Q 2480792 release-keys")),
-		stub.Regex(`adb -s .* shell getprop ro\.product\.cpu\.abi`, stub.Respond("armeabi-v7a")),
-		stub.Regex(`adb -s .* shell getprop ro\.build\.version\.sdk`, stub.Respond("26")),
+		stub.Regex(`adb -s .* shell getprop ro\.build\.version\.release`, stub.Respond("10")),
+		stub.Regex(`adb -s .* shell getprop ro\.build\.description`, stub.Respond("hammerhead-user 10 QQ1A.191003.005 5926727 release-keys")),
```
hevrard: hammerhead is Nexus 5, we could update to `flame` here (Pixel4)

core/os/android/adb/device_test.go
```diff
@@ -34,17 +34,17 @@ func TestParseDevices(t_ *testing.T) {
 		Configuration: &device.Configuration{
 			OS: &device.OS{
 				Kind:         device.Android,
-				Name:         "Marshmallow",
-				Build:        "hammerhead-user 6.0.1 MMB29Q 2480792 release-keys",
-				MajorVersion: 6,
+				Name:         "Android 10",
+				Build:        "hammerhead-user 10 QQ1A.191003.005 5926727 release-keys",
```
hevrard: hammerhead -> flame

lpy:
done :)

hevrard:
Thanks. Would there be a way to monitor when the daemons are actually running, to avoid a race condition between the daemons start and the data producer registration?

core/os/android/adb/commands.go
```diff
@@ -304,6 +304,27 @@ func (b *binding) ConnectPerfetto(ctx context.Context) (*perfetto.Client, error)
...
 
+// EnsurePerfettoPersistent ensures that Perfetto daemons, traced and
+// traced_probes, are running. Note that there is a delay between setting the
+// system property and daemons finish starting, hence this function needs to be
+// called as early as possible.
+func (b *binding) EnsurePerfettoPersistent(ctx context.Context) error {
+	if !b.SupportsPerfetto(ctx) {
+		return fmt.Errorf("Perfetto is not supported on this device")
+	}
+	res, err := b.SystemProperty(ctx, "persist.traced.enable")
```
rschiu: Couldn't we just set it without reading it?

PR #3712, opened 2020-01-30
Link Vulkan submission events with GPU queue slices.

stellama0208:
- Create a new panel for showing vulkan events.
 - Show submission ID for vulkan event slices and GPU queue slices.
 - When selecting a submission slice in Vulkan Event Panel, all the
 slices with the same submission ID in GPU Queue Panel will get a
 bounding box for easy noticing, other slices will gray out.

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/models/GpuInfo.java
+++ b/gapic/src/main/com/google/gapid/perfetto/models/GpuInfo.java
@@ -29,7 +29,8 @@ import java.util.List;
  * Data about a GPU in the trace.
  */
 public class GpuInfo {
-  public static final GpuInfo NONE = new GpuInfo(Collections.emptyList(), Collections.emptyList());
+  public static final GpuInfo NONE = new GpuInfo(Collections.emptyList(), Collections.emptyList(),
+      Collections.emptyList());
 
   private static final String MAX_DEPTH_QUERY =
       "select t.id, t.name, t.scope, max(depth) + 1 " +
@@ -38,10 +39,12 @@ public class GpuInfo {
       "order by t.id";
 
   private final List<Queue> queues;
+  private final List<VkEvent> vkEvents;
   private final List<Buffer> buffers;
 
-  private GpuInfo(List<Queue> queues, List<Buffer> buffers) {
+  private GpuInfo(List<Queue> queues, List<VkEvent> vkEvents, List<Buffer> buffers) {
     this.queues = queues;
+    this.vkEvents = vkEvents;
     this.buffers = buffers;
   }
 
@@ -55,6 +58,14 @@ public class GpuInfo {
 
   public Iterable<Queue> queues() {
     return Iterables.unmodifiableIterable(queues);
+  }
+
+  public int vkEventCount() {
+    return vkEvents.size();
+  }
+
+  public Iterable<VkEvent> vkEvents() {
+    return Iterables.unmodifiableIterable(vkEvents);
   }
 
   public int bufferCount() {
@@ -72,11 +83,15 @@ public class GpuInfo {
   private static ListenableFuture<GpuInfo> info(QueryEngine qe) {
     return transform(qe.queries(MAX_DEPTH_QUERY), res -> {
       List<Queue> queues = Lists.newArrayList();
+      List<VkEvent> vkEvents = Lists.newArrayList();
       List<Buffer> buffers = Lists.newArrayList();
       res.forEachRow(($, r) -> {
         switch (r.getString(2)) {
           case "gpu_render_stage":
             queues.add(new Queue(queues.size(), r));
+            break;
+          case "vulkan_events":
+            vkEvents.add(new VkEvent(r));
             break;
           case "graphics_frame_event":
             buffers.add(new Buffer(r));
@@ -86,7 +101,7 @@ public class GpuInfo {
 
       // Sort buffers by name, the query is sorted by track id for the queues.
       buffers.sort((b1, b2) -> b1.name.compareTo(b2.name));
-      return new GpuInfo(queues, buffers);
+      return new GpuInfo(queues, vkEvents, buffers);
     });
   }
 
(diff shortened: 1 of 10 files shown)


gapic/src/main/com/google/gapid/perfetto/models/GpuInfo.java
```diff
@@ -110,6 +125,26 @@ public String getDisplay() {
     }
   }
 
+  public static class VkEvent {
```
chrisforbes: This needs a different name. VkEvent is an object in the Vulkan API and so this is going to cause confusion.

lpy: Let's go with VulkanApiEvent or VkApiEvent. I am fine with full name or abbreviation.

gapic/src/main/com/google/gapid/perfetto/views/SliceSelectionView.java
```diff
@@ -104,6 +104,10 @@ public SliceSelectionView(Composite parent, State state, SliceTrack.Slice slice)
         }
       }
 
+      if (renderStageInfo.submissionId != 0) {
+        propsBuilder.put("VkSubmissionId:", Long.toString(renderStageInfo.submissionId));
```
rschiu: Should be just "Submission ID".  There is no VkSubmissionId in the vulkan API.

gapic/src/main/com/google/gapid/perfetto/views/VulkanEventPanel.java
```diff
@@ -0,0 +1,203 @@
...
+
+      // Draw bounding rectangles after all the slices are rendered, so that the border is on the top.
+      for (int index : visibleSelected) {
+        ctx.setForegroundColor(VulkanEventTrack.getBorderColor(data.names[index]));
+        double rectStart = state.timeToPx(data.starts[index]);
+        double rectWidth = Math.max(1, state.timeToPx(data.ends[index]) - rectStart);
+        double depth = data.depths[index];
+        ctx.drawRect(rectStart, SLICE_Y + depth * SLICE_HEIGHT, rectWidth, SLICE_HEIGHT, BOUNDING_BOX_LINE_WIDTH);
+
+        double mid = rectStart + rectWidth / 2;
```
rschiu: Just my opinion, so you might want to ask others about this.  But it would be more accurate to start the arrow at the beginning of the slice.  Otherwise, the render stages can start before the arrow.  Although I am not sure if that will look good.

gapic/src/main/com/google/gapid/perfetto/models/Tracks.java
```diff
@@ -125,6 +125,16 @@ private Tracks() {
       }
     }
 
+    if (data.getGpu().vkEventCount() > 0) {
+      data.tracks.addLabelGroup(
```
rschiu: The group may not be necessary.  Or maybe even put this into the GPU group so that the vulkan track is closer to the gpu track.  But maybe ignore this for now and try it out in another patch.

gapic/src/main/com/google/gapid/perfetto/models/GpuInfo.java
stellama0208: Done in a new commit. VkEvent -> VkApiEvent.

gapic/src/main/com/google/gapid/perfetto/views/SliceSelectionView.java
stellama0208: Done in a new commit.

gapic/src/main/com/google/gapid/perfetto/views/VulkanEventPanel.java
stellama0208: Thanks for suggestion! But personally I think it looks alright to grow arrow from the middle. Will keep it and mark as resolved for now, we can go back to change it in the future if more people raise the concern.

gapic/src/main/com/google/gapid/perfetto/models/Tracks.java
stellama0208: Done in a new commit. It is now handled in a same fashion as how GPU Queue Panel, Frame Buffer Event Panel are handled in #3719 : Only add a new group if there's more than one track.

stellama0208:
The above force push is only to do rebase and resolve merge conflicts with the top of tree.

stellama0208:
The comments above are addressed, PTAL!

gapic/src/main/com/google/gapid/perfetto/views/GpuQueuePanel.java
```diff
@@ -108,6 +116,12 @@ public void renderTrack(RenderContext ctx, Repainter repainter, double w, double
 
         StyleConstants.Gradient color = gradient(data.titles[i].hashCode());
         color.applyBase(ctx);
+        // Link slices in Vulkan Event Panel and Gpu Queue Panel through submission id.
+        if (!selectedSIds.isEmpty() && i < sIds.length && !selectedSIds.contains(sIds[i])) {
+          ctx.setBackgroundColor(color.disabled); // Grey out unlinked ones.
+        } else if (i < sIds.length && selectedSIds.contains(sIds[i])) {
+          linkedIdxs.add(i);
+        }
```
pmuetschard: This could cause slices to be selected and grayed out. I think we should update this logic to be:
```
if (selected.contains(new Slice.Key(...)) || selectedSIds.contains(sIds[i])) {
  visibleSelected.add(new Highlight(...));
} else if (!selectedSIds.isEmpty()) {
  ctx.setBackgroundColor(color.disabled);
}
```

Note how that also gets rid of the `linkedIdxs` list. The `selected.contains(...)` part comes from below, but now would have to be done before drawing.

stellama0208: Thanks for suggestion, it's cleaner after this change! Done in a new commit, Please take another look!

PR #3715, opened 2020-01-30
Simplified queued command tracking.

AWoloszyn:
This removes an edge-case that is not quite valid in the spec
but that we handled anyway. This significantly simplifies
the queued command tracking code.

It removes the repsonsibility of the API itself to handle
debug markers (that was a server detail leaking into the
API).

It also moves all of the tree-building logic into one
place.


chrisforbes:
Trace/replay performance with viking village seems much worse with this change.

gapis/api/vulkan/api/queue.api
```diff
@@ -137,140 +143,57 @@ cmd VkResult vkQueueSubmit(
...
           vkErrorInvalidSemaphore(ws)
         } else {
-          semaphore := Semaphores[ws]
+          subm.WaitSemaphores[len(subm.WaitSemaphores)] = ws
+        }
+      }
+    }
 
-          if(semaphore.SubmitCount == 0) {
-            vkErrorSemaphoreNotSubmitted(ws)
```
yalcinmelihyasin: We had a case that sometimes we get waits on semaphores without they are being submitted. This error was to prevent the deadlock and report it. I see that this is being removed. Are we sure we are not reintroducing the bug?

AWoloszyn:
> > Trace/replay performance with viking village seems much worse with this change.
> 
> Not able to reproduce this in additional testing.

That's good to hear, since this does n

gapis/api/vulkan/api/queue.api
yalcinmelihyasin: Although I see that we don't have pending commands anymore.

AWoloszyn: With this change, no commands ever are put in the pending state. So we now simply don't care what state a semaphore is in (from a simulation perspective).

PR #3718, opened 2020-01-30
Device validation UI

silence-do-good:
Adding the validation to the trace UI makes checking the GPU profiling
capabilities redundant and hence the check is removed.

Bug: 141712515


silence-do-good:
The caching isn't done in this PR as yet. Will decide if we want to do it in the same PR or not after having a discussion.

gapic/src/main/com/google/gapid/util/URLs.java
```diff
@@ -18,4 +18,6 @@
 public interface URLs {
   public static final String FILE_BUG_URL =
       "https://github.com/google/gapid/issues/new?template=standard-bug-report-for-gapid.md";
+  public static final String DEVICE_COMPATIBILITY =
```
rschiu: DEVICE_COMPATIBILITY_URL

gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -582,6 +592,30 @@ private void updateOnDeviceChange(Settings settings, DeviceCaptureInfo dev) {
...
+          validationStatusText.setText("Device is being Validated");
+          models.devices.validateDevice(dev, () -> {
+            setValidationStatus(dev.validationStatus);
+          });
+        }
+      }
+
+      private void setValidationStatus(boolean status) {
+        validationStatusLoader.updateStatus(status);
+        validationStatusText.setText("Validation " + (status ? "Passed" : "Failed" + VALIDATION_FAILED_LANDING_PAGE));
```
lpy: consider adding a white space before `VALIDATION_FAILED_LANDING_PAGE`

gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -582,6 +592,30 @@ private void updateOnDeviceChange(Settings settings, DeviceCaptureInfo dev) {
         updatePerfettoConfigLabel(settings);
       }
 
+      private void runValidationCheck(Models models, DeviceCaptureInfo dev, TraceTypeCapabilities config) {
+        if (dev != null && isPerfetto(config)) {
+          validationStatusLoader.startLoading();
+          validationStatusText.setText("Device is being Validated");
```
lpy: s/Validated/validated/

gapic/src/main/com/google/gapid/util/URLs.java
lpy: Could you please file a bug about adding device compatibility doc and assign it to me? And then add a TODO with the bug id here.

gapic/src/main/com/google/gapid/views/TracerDialog.java
rschiu: s/Validated/validated/

gapic/src/main/com/google/gapid/views/TracerDialog.java
rschiu: Add ". " after "Passed"/"Failed"

gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -582,6 +592,30 @@ private void updateOnDeviceChange(Settings settings, DeviceCaptureInfo dev) {
...
+      }
+
+      private void setValidationStatus(boolean status) {
+        validationStatusLoader.updateStatus(status);
+        validationStatusText.setText("Validation " + (status ? "Passed" : "Failed" + VALIDATION_FAILED_LANDING_PAGE));
+        validationStatusLoader.stopLoading();
+        validationStatus.set(status);
+      }
+
+      private void resetValidationStatus() {
```
rschiu: Is there a way to rerun validation?

gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -237,6 +242,14 @@ protected void createButtonsForButtonBar(Composite parent) {
       traceInput.addModifyListener(modifyListener);
 
       modifyListener.handleEvent(null); // Set initial state of widgets.
+
+      traceInput.validationStatus.addListener(new PropertyChangeListener() {
+
+        @Override
+        public void propertyChange(PropertyChangeEvent event) {
+          ok.setEnabled(traceInput.isReady());
```
rschiu: There should be an command override.  Please do **not** submit this until override is implemented.  We are working with the vendors to ensure integration works.  Adding validation is going to slow down the progress.

silence-do-good: Yeah. I'm just starting with the override. Will update the PR with it

gapic/src/main/com/google/gapid/views/TracerDialog.java
silence-do-good: Refreshing the device list will rerun it.

gapic/src/main/com/google/gapid/views/TracerDialog.java
silence-do-good: Regarding the Pixel failure, try to run `bazel run gapit validate_gpu_profiling` and confirm if it passes there. If it fails there, then the device failed validation indeed. Currently, there is some textures/fragment that is a bit flaky and Peiyong is working on them.

silence-do-good: Added the commandline override. The help menu will show the command as well.

gapic/src/main/com/google/gapid/Main.java
```diff
@@ -196,6 +196,7 @@ private Analytics getAnalytics() {
     Flags.help,
     Flags.fullHelp,
     Flags.version,
+    Flags.skipValidation,
```
lpy: skipDeviceValidation

silence-do-good: Done.

gapic/src/main/com/google/gapid/models/Devices.java
```diff
@@ -24,6 +24,7 @@
 import com.google.gapid.proto.device.Device;
 import com.google.gapid.proto.device.Device.Instance;
 import com.google.gapid.proto.service.Service;
+import com.google.gapid.proto.service.Service.ValidateDeviceResponse;
```
pmuetschard: Please don't import this class, instead refer to it via `Service.ValidateDeviceResponse`. This makes it easier to know what proto file the message came from.

gapic/src/main/com/google/gapid/models/Devices.java
```diff
@@ -142,6 +144,48 @@ public void selectReplayDevice(Device.Instance dev) {
...
+      @Override
+      protected ResultOrError<ValidateDeviceResponse, ValidateDeviceResponse> onRpcThread(Rpc.Result<ValidateDeviceResponse> response)
+          throws RpcException, ExecutionException {
+        try {
+          return success(response.get());
+        } catch (RpcException | ExecutionException e) {
+          throttleLogRpcError(LOG, "LoadData error", e);
+          return error(null);
+        }
+      }
```
pmuetschard: Please move this above the `onUiThread` methods, because it is run before them. Make it easier to read.

gapic/src/main/com/google/gapid/util/Flags.java
```diff
@@ -40,6 +40,9 @@
   public static final Flag<Boolean> fullHelp =
       value("fullhelp", false, "Print help information incuding hidden flags.", true);
   public static final Flag<Boolean> version = value("version", false, "Print GAPID version.");
+  public static final Flag<Boolean> skipDeviceValidation = value("skip_device_validation", false,
```
pmuetschard: This does not belong here. Please move it into the `Devices` class.
Also, this should be a hidden flag.

gapic/src/main/com/google/gapid/models/Devices.java
```diff
@@ -278,6 +322,7 @@ public default void onCaptureDevicesLoaded() { /* empty */ }
     public final Device.Instance device;
     public final Service.DeviceTraceConfiguration config;
     public final TraceTargets targets;
+    public boolean validationStatus = false;
```
pmuetschard: This makes this class mutatable, which is not good, especially public mutatable like this.

I propose, instead, let's create a `DeviceValidation` class, which currently just contains a `boolean passed`. Then, add a method to the device model that does the validation. Something like:
`public ListenableFuture<DeviceValidation> getDeviceValidation(Path.Device device)`.
If the the skip flag is set, always return `DeviceValidation.PASSED`. We can then also add the cache handling in this method.

Also, if you use a `SingleInFlight` `RpcController` than it will make sure only one validation request is in flight.

gapic/src/main/com/google/gapid/models/Devices.java
```diff
@@ -142,6 +144,48 @@ public void selectReplayDevice(Device.Instance dev) {
...
+
+      @Override
+      protected void onUiThreadSuccess(Service.ValidateDeviceResponse response) {
+        updateValidationStatus(device, response);
+        callback.run();
+      }
+
+      @Override
+      protected void onUiThreadError(Service.ValidateDeviceResponse response) {
+        LOG.log(Level.WARNING, response.toString());
```
pmuetschard: nit: we typically statically import the levels

gapic/src/main/com/google/gapid/util/Flags.java
silence-do-good: Moved to Devices class and changed it to hidden.

gapic/src/main/com/google/gapid/models/Devices.java
silence-do-good: Created a new DeviceValidationInfo class and restructured the way validation requests are handled. The rpcController was SingleInFlight already and it is the same in the updated class as well. Could you please take a look at the new class now?

gapic/src/main/com/google/gapid/models/Devices.java
silence-do-good: Done.

gapic/src/main/com/google/gapid/models/Devices.java
silence-do-good: Done.

gapic/src/main/com/google/gapid/models/Devices.java
silence-do-good: Done.

gapic/src/main/com/google/gapid/models/Devices.java
```diff
@@ -142,6 +153,14 @@ public void selectReplayDevice(Device.Instance dev) {
...
   }
 
+  public static void validateDevice(DeviceCaptureInfo device, Function<Boolean, Void> callback) {
+    DeviceValidationInfo.validateDevice(device, callback);
+  }
+
+  public static boolean getValidationStatus(DeviceCaptureInfo device) {
+    return DeviceValidationInfo.getValidationStatus(device);
+  }
+
```
pmuetschard: please no static singletons, please follow the pattern from my earlier review: have the model return a `ListenableFuture` of the result. Have the view code do the `UICallback` stuff. There are two patterns we follow in the models: event based (primarily for loading things) and by returning a `ListenableFuture`.

gapic/src/main/com/google/gapid/models/Devices.java
```diff
@@ -295,4 +315,87 @@ public boolean isStadia() {
...
+      instance = new DeviceValidationInfo(client, shell);
+    }
+
+    public static void validateDevice(DeviceCaptureInfo device, Function<Boolean, Void> callback) {
+      if (instance != null) {
+        instance.validate(device, callback);
+      }
+    }
+
+    public static void reset() {
```
pmuetschard: why? I don't think there is any reason to reset the validation results. In fact with caching we want it to persist between runs.

gapic/src/main/com/google/gapid/models/Devices.java
silence-do-good: I was following the event based pattern in this. Changed it to now to have the views do all the UICallback stuff and removed the static singletons as well.

gapic/src/main/com/google/gapid/models/Devices.java
silence-do-good: Removed reset now. I had it as a way of not re-running validation, if it failed, unless the user presses the refresh button. Now, if the validation fails, just changing the device to something else and back will trigger it.

core/os/device/device.proto
```diff
@@ -269,3 +271,18 @@ message VulkanPhysicalDevice {
...
+// successfully validated.
+message DeviceValidationCache {
+  repeated DeviceValidationCacheEntry entries = 1;
+}
+
+// DeviceValidationCacheEntry describes the properties of the device that can
+// change validation results if updated.
+message DeviceValidationCacheEntry {
+  OS os = 1;
+  repeated VulkanPhysicalDevice physical_devices = 2;
```
lpy: let's store the driver_version instead. And make sure to put a comment saying this is per `vkPhysicalDevice`

silence-do-good: Done.

core/os/device/device.proto
```diff
@@ -245,6 +245,8 @@ message VulkanDriver {
   repeated string icd_and_implicit_layer_extensions = 2;
   // Physical devices that have Vulkan support.
   repeated VulkanPhysicalDevice physical_devices = 3;
+  // Driver package version
+  string version_code = 4;
```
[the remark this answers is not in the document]
silence-do-good: Done.

gapic/src/main/com/google/gapid/widgets/LoadingIndicator.java
```diff
@@ -186,4 +196,30 @@ public Point computeSize(int wHint, int hHint, boolean changed) {
       return new Point(SMALL_SIZE, SMALL_SIZE);
     }
   }
+
+  public class WithImage extends Widget {
```
pmuetschard: I think it would be a bit easier to understand and maintain if instead of extending the `Widget` class, we enhance the `Widget` version to optionally show an image on completion. Simply move the success and failure images there, in the current constructor initialize them with null, add a new constructor with the images, and in its paint method show them if they are non-null.

gapic/src/main/com/google/gapid/models/Devices.java
```diff
@@ -57,11 +71,19 @@
   private List<Device.Instance> replayDevices;
   private Device.Instance selectedReplayDevice;
   private List<DeviceCaptureInfo> devices;
+  private DeviceValidationInfo deviceValidationInfo;
+  private List<DeviceCaptureInfo> validatedDevices;
```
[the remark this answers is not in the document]
silence-do-good: This is the local list of validated devices. The problem with using the list (created from the proto) is that if you call the contains function, it defaults to == check instead of calling equals() check. == checks if the two protobuf messages are the same object while equals() checks if two protobuf messages have the same contents. When we call contains(), Java List by default calls ==, while ArrayList calls equals(). It doesn't make sense to create a new arraylist every time we want to compare, which is why I created a member directly. 

Also, this helps in writing the cache back to the file. We load the cache into this list, add any new entries to the list if needed, write the list back to the file.

core/os/device/device.proto
```diff
@@ -269,3 +271,20 @@ message VulkanPhysicalDevice {
...
+
+// DeviceValidationCacheEntry describes the properties of the device that can
+// change validation results if updated.
+message DeviceValidationCacheEntry {
+  OS os = 1;
+  // driverVersion from the VulkanPhysicalDevice. (Per VulkanPhysicalDevice)
+  repeated uint32 driver_version = 2;
+  // The pre-release driver APK version code
+  int64 version_code = 3;
+}
```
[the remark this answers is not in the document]
silence-do-good: The current idea is that we cache only the devices that passed validation. If we cache even the failed ones, then we need to enable a way to re-run validation, which I feel is going to be an additional button or something and complicates stuff.
Given that if validation fails on one driver version, it is never going to pass on that driver again. But if validation failed because the user mistakenly presses the home button or does something with the device, then the only way to re-run validation in the current state would be to delete the cache and try again. To avoid this situation, I feel its better to just not cache failed devices. 
What do you think @lpy and @pmuetschard ?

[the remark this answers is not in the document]
pmuetschard: OK. Let's design for caching success and failure (i.e. the proto message), but for v1 only actually cache successes.

gapic/src/main/com/google/gapid/widgets/LoadingIndicator.java
silence-do-good: Done.

gapic/src/main/com/google/gapid/widgets/LoadingIndicator.java
```diff
@@ -90,6 +90,12 @@ private static int paint(Image image, GC g, int x, int y, int w, int h) {
     return x + (w - s.width) / 2 + s.width;
   }
 
+  public void paintImage(Image image, GC g, int x, int y, Point size) {
```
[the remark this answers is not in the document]
silence-do-good: Done.

gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -630,6 +706,9 @@ private void updateOnApiChange(
         durationLabel.setText(isPerfetto ? DURATION_LABEL : FRAMES_LABEL);
         updateDurationSpinner(dev, config, dur.getType());
         perfettoConfig.setVisible(isPerfetto);
+        resetValidationStatus();
```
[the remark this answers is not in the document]
silence-do-good: Had it as a default state to show in between devices change. But the loader is fast enough to hide this, so removing it.

gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -582,6 +606,58 @@ private void updateOnDeviceChange(Settings settings, DeviceCaptureInfo dev) {
...
+      private void runValidationCheck(DeviceCaptureInfo dev, TraceTypeCapabilities config, Shell shell) {
+        if (dev == null || Devices.skipDeviceValidation.get()) {
+          return;
+        }
+        setValidationStatus(models.devices.getValidationStatus(dev));
+        if (isPerfetto(config) && !models.devices.getValidationStatus(dev)) {
+          validationStatusLoader.startLoading();
+          validationStatusText.setText("Device is being validated");
+          rpcController.start().listen(models.devices.validateDevice(dev),
+              new UiErrorCallback<DeviceValidationResult, DeviceValidationResult, DeviceValidationResult>(shell, LOG) {
```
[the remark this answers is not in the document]
silence-do-good: Done.

gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -582,6 +606,58 @@ private void updateOnDeviceChange(Settings settings, DeviceCaptureInfo dev) {
         updatePerfettoConfigLabel(settings);
       }
 
+      private void runValidationCheck(DeviceCaptureInfo dev, TraceTypeCapabilities config, Shell shell) {
+        if (dev == null || Devices.skipDeviceValidation.get()) {
+          return;
+        }
+        setValidationStatus(models.devices.getValidationStatus(dev));
+        if (isPerfetto(config) && !models.devices.getValidationStatus(dev)) {
```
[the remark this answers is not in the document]
silence-do-good: Done.

gapic/src/main/com/google/gapid/models/Devices.java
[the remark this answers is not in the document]
silence-do-good: Done.

gapic/src/main/com/google/gapid/models/Devices.java
```diff
@@ -295,4 +334,104 @@ public boolean isStadia() {
       return device.getConfiguration().getOS().getKind() == Device.OSKind.Stadia;
     }
   }
+
+  /*
+   *  Class that handles the gapis communication for device validation and the load/store
+   *  of the validation cache.
+   */
+  private static class DeviceValidationInfo {
+    private static final String VALIDATION_CACHE = ".gapic-device-validation-cache";
```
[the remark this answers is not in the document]
silence-do-good: Done.

gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -237,6 +252,14 @@ protected void createButtonsForButtonBar(Composite parent) {
       traceInput.addModifyListener(modifyListener);
 
       modifyListener.handleEvent(null); // Set initial state of widgets.
+
+      traceInput.validationStatus.addListener(new PropertyChangeListener() {
```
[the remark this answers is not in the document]
silence-do-good: Done.

gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -333,6 +363,19 @@ public TraceInput(Composite parent, Models models, Widgets widgets, Runnable ref
         apiLabel = createLabel(mainGroup, "Type*:");
         api = createApiDropDown(mainGroup);
         api.getCombo().setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, false));
+        // dummy label to fill the 3rd column
+        createLabel(mainGroup, "");
+
+        validationStatusLoader = widgets.loading.createWithImage(mainGroup, widgets.theme.smile(), widgets.theme.error());
```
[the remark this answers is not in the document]
silence-do-good: I couldn't find any images with the green check in it. Didn't want to download a random one from the internet due to potential copyright issues. Please suggest a good alternate.

gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -582,6 +606,58 @@ private void updateOnDeviceChange(Settings settings, DeviceCaptureInfo dev) {
         updatePerfettoConfigLabel(settings);
       }
 
+      private void runValidationCheck(DeviceCaptureInfo dev, TraceTypeCapabilities config, Shell shell) {
+        if (dev == null || Devices.skipDeviceValidation.get()) {
```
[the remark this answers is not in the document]
silence-do-good: Done.

gapic/src/main/com/google/gapid/models/Devices.java
```diff
@@ -57,11 +71,19 @@
   private List<Device.Instance> replayDevices;
   private Device.Instance selectedReplayDevice;
   private List<DeviceCaptureInfo> devices;
+  private DeviceValidationInfo deviceValidationInfo;
+  private List<DeviceCaptureInfo> validatedDevices;
+
+  public static final Flag<Boolean> skipDeviceValidation = Flags.value("skip_device_validation", false,
```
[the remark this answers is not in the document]
silence-do-good: Done.

core/os/device/device.proto
```diff
@@ -214,6 +214,8 @@ message Drivers {
   OpenGLDriver opengl = 1;
   // The Vulkan driver support.
   VulkanDriver vulkan = 2;
+  // Driver APK version
+  int64 prerelease_driver_apk_version_code = 3;
```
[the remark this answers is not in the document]
silence-do-good: Done.

core/os/android/adb/device.go
```diff
@@ -241,6 +241,11 @@ func newDevice(ctx context.Context, serial string, status bind.Status) (*binding
 		i.GenID()
 	}
 
+	c, _ := d.DriverVersionCode(ctx)
+	if c != -1 {
+		d.To.Configuration.Drivers.PrereleaseDriverApkVersionCode = int64(c)
+	}
+
```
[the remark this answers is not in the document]
silence-do-good: Done.

core/os/android/adb/commands.go
```diff
@@ -381,6 +381,19 @@ func extrasFlags(extras []android.ActionExtra) []string {
...
 
+// Returns the package version code of the graphics driver
+func (b *binding) DriverVersionCode(ctx context.Context) (int, error) {
+	driver, err := b.SystemProperty(ctx, driverProperty)
+	if err != nil {
+		return -1, err
+	}
+	ip, err := b.InstalledPackage(ctx, driver)
+	if err != nil {
+		return -1, err
```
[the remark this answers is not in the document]
silence-do-good: Done.

silence-do-good:
@pmuetschard 
I've updated the PR with your requested changes. Will create a new PR in AGI once its all good to go. Easier to track here because of the existing comments.

gapic/src/main/BUILD.bazel
```diff
@@ -76,6 +76,9 @@ java_library(
 proto_library(
     name = "settings_proto",
     srcs = ["com/google/gapid/settings.proto"],
+    deps = [
+      "//core/os/device:device_proto",
```
pmuetschard: indent 4 spaces (this is why presubmit failed).

gapic/src/main/com/google/gapid/models/Devices.java
```diff
@@ -142,6 +152,16 @@ public void selectReplayDevice(Device.Instance dev) {
     listeners.fire().onReplayDeviceChanged(dev);
   }
 
+  public ListenableFuture<DeviceValidationResult> validateDevice(DeviceCaptureInfo device) {
+    return MoreFutures.transform(deviceValidationInfo.doValidation(device), e -> {
```
pmuetschard: Why the transform, if you are not transforming the result?

gapic/src/main/com/google/gapid/models/Devices.java
```diff
@@ -295,4 +315,89 @@ public boolean isStadia() {
...
+      DeviceValidation.ValidationEntry entry  = buildValidationEntry(device);
+      if (deviceValidation.getValidationEntriesList().contains(entry)) {
+        return DeviceValidationResult.getPassedResult();
+      }
+      return DeviceValidationResult.getFailedResult();
+    }
+
+    public ListenableFuture<DeviceValidationResult> doValidation(DeviceCaptureInfo device) {
+      if (device == null) {
+        return immediateFuture(DeviceValidationResult.getPassedResult());
```
pmuetschard: shouldn't this be failed?

gapic/src/main/com/google/gapid/models/Devices.java
```diff
@@ -295,4 +315,89 @@ public boolean isStadia() {
...
+      }
+
+      return MoreFutures.transform(client.validateDevice(device.path), e -> {
+        updateValidationStatus(currentEntry, e);
+        return new DeviceValidationResult(e.getError(), !e.hasError(), false);
+      });
+    }
+
+    protected void updateValidationStatus(DeviceValidation.ValidationEntry entry, Service.ValidateDeviceResponse response) {
```
pmuetschard: this needs to happen on the UI thread.

gapic/src/main/com/google/gapid/models/Settings.java
```diff
@@ -220,6 +220,10 @@ public void save() {
     return proto.getPerfettoBuilder();
   }
 
+  public SettingsProto.DeviceValidation.Builder deviceValidation() {
```
pmuetschard: follow the pattern of using `writeFoo` if you are returning a builder, to indicate that this can mutate the proto.

gapic/src/main/com/google/gapid/settings.proto
```diff
@@ -164,4 +166,25 @@ message Settings {
...
+  message Result {
+    bool passed = 1;
+  }
+
+  message ValidationEntry {
+    Device device = 1;
+    Result result = 2;
+  }
+  repeated ValidationEntry validation_entries = 1;
+}
```
pmuetschard: missing new line

gapic/src/main/com/google/gapid/util/URLs.java
```diff
@@ -18,4 +18,7 @@
 public interface URLs {
   public static final String FILE_BUG_URL =
       "https://github.com/google/gapid/issues/new?template=standard-bug-report-for-gapid.md";
+  // TODO(b/148700785): Create the device compatibility doc and update here
+  public static final String DEVICE_COMPATIBILITY_URL =
+      "https://github.com/google/gapid";
```
pmuetschard: point to agi instead

gapic/src/main/com/google/gapid/views/TracerDialog.java
```diff
@@ -582,6 +589,56 @@ private void updateOnDeviceChange(Settings settings, DeviceCaptureInfo dev) {
...
+          validationStatusLoader.stopLoading();
+          validationStatusText.setText("Validation skipped.");
+          validationStatus = true;
+        } else {
+          validationStatusLoader.updateStatus(result.passed);
+          validationStatusText.setText("Validation " + (result.passed ? "Passed." : "Failed. " + VALIDATION_FAILED_LANDING_PAGE));
+          validationStatusLoader.stopLoading();
+          validationStatus = result.passed;
+        }
+        (TraceInput.this).notifyListeners(SWT.Modify, new Event());
```
pmuetschard: Remove the `(TraceInput.this)` part. I know it's from my previous comment, I just wanted to point out what class to to call `notifyListeners` on.

gapic/src/main/com/google/gapid/widgets/LoadingIndicator.java
```diff
@@ -164,6 +180,27 @@ public Widget(Composite parent, boolean showRefresh) {
...
+          paint(successImage, e.gc, 0, 0, getSize());
+        }
+        else {
+          paint(failureImage, e.gc, 0, 0, getSize());
+        }
+      });
+    }
+
+    public void updateStatus(boolean status) {
+      this.status = status;
```
pmuetschard: add `scheduleForRedraw(this)` since the state changed

gapic/src/main/com/google/gapid/widgets/LoadingIndicator.java
```diff
@@ -148,12 +156,20 @@ public Widget createWidgetWithRefresh(Composite parent) {
...
-    private boolean loading = false;
+    private Image successImage;
+    private Image failureImage;
+
+    protected boolean loading = false;
+    protected boolean status = false;
 
     public Widget(Composite parent, boolean showRefresh) {
       super(parent, SWT.DOUBLE_BUFFERED);
+      successImage = null;
```
pmuetschard: Get rid of this constructor and simply call your new one with either `null, null`, or `refresh, refresh` above.

gapic/src/main/com/google/gapid/models/Devices.java
silence-do-good: This is just for updating the cache that a new entry is going to be added. It will reach this step only if the device wasn't in the cache already. The Devices model will handle the proto response message while constructing a new ValidationResult object and pass that to the UI.

silence-do-good:
Cleaned up the code after the review comments. @pmuetschard

gapic/src/main/com/google/gapid/widgets/LoadingIndicator.java
```diff
@@ -148,22 +148,43 @@ public Widget createWidgetWithRefresh(Composite parent) {
 
   /**
    * Widget that shows the loading indicator while loading and is blank once done.
+   * Can optionally show an image when done.
    */
   public class Widget extends Canvas implements Loadable, Repaintable {
-    private boolean loading = false;
+    private Image successImage;
+    private Image failureImage;
```
pmuetschard: Make these final.

gapic/src/main/com/google/gapid/widgets/LoadingIndicator.java
```diff
@@ -148,22 +148,43 @@ public Widget createWidgetWithRefresh(Composite parent) {
...
+
+    protected boolean loading = false;
+    protected boolean status = false;
 
-    public Widget(Composite parent, boolean showRefresh) {
+    public boolean hasImage = false;
```
pmuetschard: This should be private and final, however, I don't think the field is necessary.

silence-do-good: The only loading widget currently in use is the refresh devices, and the validation status. But if someone uses just the loader without refresh or the image, it will cause a null pointer derefencing, which is why I thought this preliminary check might come in handy

pmuetschard: sorry, i think GH ate one of my comments. The reason I said it was not necessary is because of the comment that got eaten. Below, in the paint handler I was saying to to check each image for null, because `hasImage` is true if either image is not null, so if only a success image was given, but a failure is to be shown, it would fail with an NPE.

I think the code below should be:
```
if (loading) {
  ..
} else {
  Image image = status ? sucessImage : failureImage;
  if (image != null) {
    paint(image, ...);
  }
}
```
Making that change would make `hasImage` obsolete, while supporting various cases of image present/not-present.

silence-do-good: Done. Updated this in the AGI PR.

PR #3726, opened 2020-01-31
Allow marking and zooming into multi-selections.

pmuetschard:
(and counter selections)

Bug: http://b/146328747

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/models/CounterTrack.java
+++ b/gapic/src/main/com/google/gapid/perfetto/models/CounterTrack.java
@@ -35,7 +35,6 @@ import org.eclipse.swt.widgets.Composite;
 
 import java.util.Arrays;
 import java.util.Set;
-import java.util.stream.LongStream;
 
 public class CounterTrack extends Track.WithQueryEngine<CounterTrack.Data> {
   private static final String VIEW_SQL_DELTA =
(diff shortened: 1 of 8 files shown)


gapic/src/main/com/google/gapid/perfetto/models/Selection.java
```diff
@@ -133,17 +145,19 @@ public void addSelection(MultiSelection other) {
...
 
     public void zoom(State state) {
-      // TODO: handle zooming into multiple selections.
-      if (selections.size() == 1) {
-        firstSelection().zoom(state);
+      getRange().ifNotEmpty(state::setVisibleTime);
+    }
+
+    private TimeSpan getRange() {
+      TimeSpan[] range = new TimeSpan[] { TimeSpan.ZERO };
```
stellama0208: Can we use `new TimeSpan(0, 0)` here instead of an array?

gapic/src/main/com/google/gapid/perfetto/models/CounterTrack.java
```diff
@@ -120,21 +120,29 @@ private String counterSQL() {
 
   public ListenableFuture<Data> getValue(long t) {
     return transform(expectOneRow(qe.query(valueSql(t))), row -> {
-      Data data = new Data(null, new long[1], new double[1]);
+      Data data = new Data(null, new long[2], new double[2]);
       data.ts[0] = row.getLong(0);
+      data.ts[1] = row.getLong(1);
       data.values[0] = row.getDouble(2);
+      data.values[0] = data.values[1];
```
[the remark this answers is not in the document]
pmuetschard: Good catch, thanks!

gapic/src/main/com/google/gapid/perfetto/models/Selection.java
pmuetschard: no, because it has to be final. Putting it into an array allows us to modify it.

stellama0208: Thanks for explaining!

PR #3730, opened 2020-02-03
Add vkCreateDevice to deviceinfo.

lpy:
In order to trigger the driver to do more works on some platforms, for example
registering their data producers, deviceinfo needs to create a Vulkan device.

Bug: b/148598399


lpy:
Thanks for the feedback, ptal :)

core/os/device/deviceinfo/cc/vk.cpp
```diff
@@ -163,6 +166,60 @@ bool vkPhysicalDevices(
         std::string(prop.deviceName));
   }
 
+  uint32_t queue_family_index = static_cast<uint32_t>(-1);
+  VkPhysicalDevice physical_device{};
```
AWoloszyn: My other comment got eaten.
We probably want to create a VkDevice for EVERY physical device, instead of simple "one physical device at random" that way if there is also something like swiftshader on the system, we don't actually create the wrong one.

lpy:
Thanks, ptal!

PR #3731, opened 2020-02-04
Update shallow_since fields in workspace.bzl.

lpy:
Update shallow_since fields in workspace.bzl to silence build DEBUG message.

Bug: N/A

### DIFF
--- a/tools/build/workspace.bzl
+++ b/tools/build/workspace.bzl
@@ -147,6 +147,7 @@ def gapid_dependencies(android = True, mingw = True, locals = {}):
         remote = "https://github.com/nothings/stb",
         commit = "f67165c2bb2af3060ecae7d20d6f731173485ad0",
         build_file = "@gapid//tools/build/third_party:stb.BUILD",
+        shallow_since = "1572280202 -0700",
     )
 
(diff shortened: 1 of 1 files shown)


tools/build/workspace.bzl
```diff
@@ -329,6 +331,7 @@ def _grpc_deps(locals):
         locals = locals,
         # on the master-with-bazel branch
         url = "https://boringssl.googlesource.com/boringssl/+archive/afc30d43eef92979b05776ec0963c9cede5fb80f.tar.gz",
+        sha256 = "54ff586a2bafbf3b7f9a9ca037f3f3490281061670aba2962a8431e9492f4b27"
```
hevrard: Builds are failing, apparently this should be:
1c5c3e596b286414d70cc97f3e9d4e0831d73ca33b6969ec57bf56cf35cd8a61
Can you double check?

pmuetschard: remove this, it unfortunately doesn't work, because GOB puts some sort of timestamps into the tar file, which means it changes every time it's downloaded.

lpy:
ptal

Issue #3728, opened 2020-02-03
Be graceful if `libgpudataproducer.so` does not output to `stderr`


mattyclarkson:
We've been cleaning up our code. As part of that we removed all debugging from our `libgpudataproducer.so`.

I've been trying to figure out why our `gpu.counters` doesn't work with GAPID/`gapit` anymore.

It was because we don't output anything on `stderr` so the [StartFunc](https://github.com/google/gapid/blob/571246c744f5f3c2d148434cfcd6390d684f3aa5/gapidapk/deviceinfo.go#L214-L216) never gets invoked. This makes GAPID/`gapis`/`gapit` hang.

Reverse engineering that took a bit of time 🤦‍♂️

Is there documentation for `libgpudataproducer.so`? Make sure to add that you must print something to `stderr` before sleeping!

Might be worth putting a bit of defensive programming in there. If nothing is received in 5(?) seconds, it errors out and puts a helpful hint explaining that GAPID needs to receive something on `stderr` from `libgpudataproducer.so`.

lpy:
Fixed in #3732

PR #3732, opened 2020-02-04
Log to stderr in launch_producer

lpy:
The data producer is not required to log anything, previously we check the
output to determine the start of data producer, this patch modifies the
launch_producer to output to stderr to make sure there is at least one
output to be captured.

Bug: #3728

### DIFF
--- a/cmd/launch_producer/main.cpp
+++ b/cmd/launch_producer/main.cpp
@@ -24,9 +24,10 @@
 
 #include <string>
 
+// TODO(b/148950543): Figure out why stdout is not captured.
 #define _LOG(lvl, name, msg, ...)                          \
   do {                                                     \
-    printf(name ": " msg "\n", ##__VA_ARGS__);             \
+    fprintf(stderr, name ": " msg "\n", ##__VA_ARGS__);    \
     __android_log_print(lvl, "GAPID", msg, ##__VA_ARGS__); \
   } while (false)
 


gapidapk/deviceinfo.go
```diff
@@ -211,9 +211,9 @@ func launchPerfettoProducerFromApk(ctx context.Context, d adb.Device, startFunc
 	crash.Go(func() {
 		buf := bufio.NewReader(reader)
 		for {
+			startFuncOnce := task.Once(task.Delay(startFunc, 3*time.Second))
```
mattyclarkson: I'm not sure this is a great solution as it will add a 3 second delay when determining `gpu.counter` support.

I think the original solution of writing to `stderr` is correct. It just could do with being documented somewhere.

However, I think it *should* error out and dump and error message if the producer doesn't write to `stderr` within 3 seconds.

As this in in the `for` loop, won't it keep creating delayed functions if the producer *does* write to `stderr`?

A possible solution would be:

```
startFuncOnce = task.Once(startFunc, 3*time.Second)
startFuncDelayed = task.Delay(func() {
  log.W(ctx, "libgpudataproducer.so did not write to stderr to signify it has started")
  startFuncOnce(ctx)
})
startFuncDelayed()
for {
  line, e := buf.ReadString('\n')
  startFuncOnce(ctx)
  startFuncDelayed.cancel()  // <-- no idea if this exists...

**Comment by mattyclarkson** [comment]
I just thought that there is a nice solution for this. If the `start(void)` was changed to `start(started_t started)` the `libgpudataproducer.so` can do:

```
[[noreturn]] void start(started_t started) {
  register_pefetto_data_source();
  started();
  while (1) { sleep(100000000) };
}
```

Then `launch_producer` can do:

```
  LOG_INFO("Calling start at %p", startFunc);
  (*startFunc)([]() {
    std::cerr << "<this-is-the-started-search-string>" << std::endl;
  });
  LOG_WARN("Producer has exited.");
```

Then `deviceinfo.go` can check that `<this-is-the-started-search-string>` is in `stderr` and we *know* that the `libgpudataproducer.so` has started.

This is much nicer for the `libgpudataproducer.so` developer because it becomes obvious that they need to invoke the callback.

This will probably break all existing `libgpudataproducer.so` libraries though due to the ABI of `start` changing.

I mentioned this a while ago: it would be much better if the `libgpudataproducer.so` had a plugin interface virtual function table to fill out as that would be much more strongly typed:

```
typedef void(*started_t)();
typedef void(*start_t)(started_t);
struct gapid_plugin {
  start_t start;
};
```

Then `libgpudataproducer.so` can then fulfil that plugin interface. It would be easier to be forwards/backwards compatible then. I can raise that as another tracking issue if necessary.

mattyclarkson:
That being said. I think having the `libgpudataproducer.so` output to `stderr` is a decent solution, it just needs to be documented or defensively programmed against.

lpy:
Our launch_producer binary loads the libgpudataproducer.so, we can output something from there. We actually already output something, I am checking why it didn't work

lpy:
I added an stderr output from launch_producer. ptal

We actually capture stdout and stderr, however, it seems to me stdout is not captured. :/

lpy:
done, ptal!

PR #3736, opened 2020-02-04
Fix and document capture-time tweak of vkCreateImage

hevrard:
Bug: b/148783351

### DIFF
--- a/gapii/README.md
+++ b/gapii/README.md
@@ -59,3 +59,13 @@ GAPII needs to support all of these cases.
 
 Each encoded command holds an identifier of the thread that made the call.
 Commands from all threads are encoded into a single chronological stream.
+
+## Command modifications during capture
+
+Some commands may have to be modified for the capture to be possible.
+The following lists are maintained as a best-effort, they may not be exhaustive nor up-to-date.
+
+### Vulkan
+
+- `vkCreateImage`: the `VK_IMAGE_USAGE_TRANSFER_SRC_BIT` is always set, in order to be able to read the image contents.
+  The `VK_IMAGE_USAGE_TRANSIENT_ATTACHMENT_BIT` is cleared accordingly, because it is incompatible with the transfer bit.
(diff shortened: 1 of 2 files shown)


gapii/README.md
```diff
@@ -58,4 +58,14 @@ Some graphics APIs permit simultaneous sharing of contexts by multiple threads.
 GAPII needs to support all of these cases.
 
 Each encoded command holds an identifier of the thread that made the call.
-Commands from all threads are encoded into a single chronological stream.
\ No newline at end of file
+Commands from all threads are encoded into a single chronological stream.
+
+## Command modifications at capture time
```
johnplate: "at capture time" implies a point in time instead of a range, so for me it would sound better to use "Command modifications during capture".

gapii/README.md
```diff
@@ -58,4 +58,14 @@ Some graphics APIs permit simultaneous sharing of contexts by multiple threads.
...
+
+Some commands may be need to be modified for the capture to be possible.
```
johnplate: "may be need to be" sounds weird, might be wrong. I would use "may have to be".

gapii/README.md
```diff
@@ -58,4 +58,14 @@ Some graphics APIs permit simultaneous sharing of contexts by multiple threads.
...
+The following lists are maintained as a best-effort, they may not be exhaustive nor up-to-date.
+
+### Vulkan
+
+- `vkCreateImage`: the `VK_IMAGE_USAGE_TRANSFER_SRC_BIT` is always set, in order to be able to read the image contents.
+  The `VK_IMAGE_USAGE_TRANSIENT_ATTACHMENT_BIT` is accordingly un-set, because it is incompatible with the transfer bit.
```
johnplate: "cleared" instead of "un-set"

gapii/README.md
hevrard: done

gapii/README.md
hevrard: ok done

gapii/README.md
hevrard: typo indeed, done

AWoloszyn:
Your removal of VK_IMAGE_USAGE_TRANSIENT_ATTACHMENT_BIT is not sufficient.

PR #3740, opened 2020-02-04
Do not try to read from TRANSIENT images

hevrard:
Bug: related to first fix for b/148783351


hevrard:
After #3750 , we might as well just aim for the proper fix, see b/148857112

PR #3743, opened 2020-02-04
Allocate a new destination binding if it does not exist

yalcinmelihyasin:
during the copy of the descriptor set. Previously copy was
incorrectly dropped if the destination binding had not been
allocated


AWoloszyn:
Probably want a Bug: label.

PR #3744, opened 2020-02-04
Change quantized slice color in GPU track to be based on the first slice name

rschiu:
Bug: b/148808741

### DIFF
--- a/gapic/src/main/com/google/gapid/perfetto/views/TrackPanel.java
+++ b/gapic/src/main/com/google/gapid/perfetto/views/TrackPanel.java
@@ -142,6 +142,18 @@ public abstract class TrackPanel<T extends TrackPanel<T>> extends Panel.Base
     };
   }
 
+  // Helper function to determine the color of a slice.
+  protected StyleConstants.Gradient getSliceColor(String title, int depth) {
+    int commaIndex = title.indexOf(',');
+    int colorCode = (commaIndex == -1) ? title.hashCode() :
+        title.substring(0, commaIndex).hashCode();
+    return StyleConstants.gradient(colorCode ^ depth);
+  }
+
+  protected StyleConstants.Gradient getSliceColor(String title) {
+    return getSliceColor(title, 0);
+  }
+
   private static class Tooltip {
     private static final Splitter LINE_SPLITTER =
         Splitter.on(CharMatcher.anyOf("\r\n")).omitEmptyStrings().trimResults();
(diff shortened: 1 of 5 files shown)


rschiu:
@pmuetschard Let me know what you think of this change.  An alternative way I can think of to do this would be to add a field to store the color hash in SliceTrack.Data.

rschiu:
It doesn't fix all the problems with quantized slice, but it does fix the bug I am referencing.  I did file another bug b/148896447 relating to quantized slice, maybe that's what you thinking about?

I am not seeing additional growing/shrinking slice problem that we didn't already have.  It shouldn't be introducing new problem since I am just changing the color.  This just make zooming in and out visually a lot more pleasing.  Could you maybe give it a try?

Please keep me up to date on your discussion.

pmuetschard:
This is with your patch applied.
Zoomed out:
Screen Shot 2020-02-05 at 12 34 00 PM
Zoomed in:
Screen Shot 2020-02-05 at 12 34 28 PM
(the zoom-in happened on the center of the zoomed out image).

Notice how the multi-color slices seen in the zoomed-in version are all absorbed into the blue slice? This is exactly the same as without your change, except that before it could be a completely different color based on the titles of the aggregated slices. You are just trading the representation color from a usually different color to the color of one of the aggregated slices. This is why I'm saying it's both marginally better (color represents an actual slice found in the aggregation) and marginally worse (it's misleading on the weight of that slice) than before.

pmuetschard:
If you like it better, I'm fine with merging this in. Can you please make the same change in the other panels that render slices (i know, copy-pasted code, sigh).

rschiu:
Did you make any changes other than applying the patch?  I only change the GPU Panel because I want your opinion first.  So the patch wouldn't affect the thread slices.  Here is before:

Screenshot from 2020-02-05 13-47-34
Here is after:
Screenshot from 2020-02-05 13-46-39

Notice in the after shot, there are still gaps in between the slices.  So it doesn't mislead on the weight of the slice.  I am aware that the problem you are talking about is not solve with this patch.  But I don't think it made it worse.  The multiple slices being absorb is not due to this patch.

pmuetschard:
I applied the patch to the thread slices for the images I posted, because I didn't have a GPU trace handy, but it behaves exactly the same on the GPU slices. As I mentioned, it's trading one slightly wrong color for another slightly, in-a-different-way, wrong color. However, I'm OK to make the change if you prefer the new wrongness over the old wrongness. We should, however, be consistent and so make the change to all slice renderers.

rschiu:
PTAL

gapic/src/main/com/google/gapid/perfetto/views/TrackPanel.java
```diff
@@ -142,6 +142,17 @@ public void stop() {
     };
   }
 
+  // Helper function to determine the color of a slice.
+  protected StyleConstants.Gradient getSliceColor(String title, int depth) {
+    int commaIndex = title.indexOf(',');
+    int colorCode = (commaIndex == -1) ? title.hashCode() :
+      title.substring(0, commaIndex).hashCode();
```
pmuetschard: indent two more spaces.

gapic/src/main/com/google/gapid/perfetto/views/TrackPanel.java
```diff
@@ -142,6 +142,17 @@ public void stop() {
...
+    return StyleConstants.gradient(colorCode ^ depth);
+  }
+  protected StyleConstants.Gradient getSliceColor(String title) {
```
pmuetschard: Add a new line between methods.

PR #3748, opened 2020-02-05
Allow easier time highlight modification in Pan/Zoom mode.

stellama0208:
- Before this change: in Pan/Zoom mode, to do a time highlight range
 modification, you have to Ctrl-drag the edge in time ruler panel. And
 there's no yellow bold edge rendering searving as a hint.
 - After this change: in Selection/Pan/Zoom mode, modification can be
 done in a same manner by a plain dragging on edge in the time ruler panel.
 - The idea behind this commit is to have at least one place, where time
 highlight dragging always works, no matter in which mouse mode.
 - Related ticket: http://b/148789881.


stellama0208:
> Should we allow dragging the full edge in every mode?

After some thoughts, my personal preference would say I'd vote for the full edge solution. 
So I've updated the PR to allow full edge dragging in all the modes. Hey Pascal would you mind check this code change and UX experience again? If it's good, let's check in the code this way! ^_^

Issue #3729, opened 2020-02-03
System Trace with vkreplay


mattyclarkson:
**Environment information:**
 - GAPID version:1.7.0-dev-20200203
 - Host OS: Linux/Windows
 If tracing on Android:
 - Device model: Makalu GED
 - Android Version: 10

**Bug description**
Additional arguments do not get passed to `vkreplay`

**Reproduction steps**
Install `vkreplay` on device, attempt to perform a system replay with the following settings

Application: `android.intent.action.MAIN:com.example.vkreplay/android.app.NativeActivity`
Additional Arguments: `--EnablePortabilityTableSupport true --SelfManageMemoryAllocation true -o /data/local/tmp/lineage2.vktrace`

`adb logcat '*.S' 'vkreplay:V'` will output
```
01-03 10:16:19.492 15604 15627 I vkreplay: argc = 1
01-03 10:16:19.492 15604 15627 I vkreplay: argv[0] = vkreplay
01-03 10:16:19.492 15604 15627 V vkreplay: Replaying with v1.11
01-03 10:16:19.492 15604 15627 E vkreplay: No trace file specified.
01-03 10:16:19.492 15604 15627 D vkreplay: vkreplay_main returned -1
```

The arguments do not seem to get passed through the executable. The following does work:

```
adb shell am start \
  -a android.intent.action.MAIN \
  -n com.example.vkreplay/android.app.NativeActivity \
  --es args "'--EnablePortabilityTableSupport true
              --SelfManageMemoryAllocation    true
               -o                              /data/local/tmp/lineage2.vktrace'"
```

```
01-03 10:18:10.951 15826 15849 I vkreplay: argc = 7
01-03 10:18:10.951 15826 15849 I vkreplay: argv[0] = vkreplay
01-03 10:18:10.951 15826 15849 I vkreplay: argv[1] = --EnablePortabilityTableSupport
01-03 10:18:10.951 15826 15849 I vkreplay: argv[2] = true
01-03 10:18:10.951 15826 15849 I vkreplay: argv[3] = --SelfManageMemoryAllocation
01-03 10:18:10.951 15826 15849 I vkreplay: argv[4] = true
01-03 10:18:10.951 15826 15849 I vkreplay: argv[5] = -o
01-03 10:18:10.951 15826 15849 I vkreplay: argv[6] = /data/local/tmp/lineage2.vktrace
01-03 10:18:10.951 15826 15849 V vkreplay: Replaying with v1.11
```

[...]

PR #3751, opened 2020-02-05
Pass on "additional args" in system profile mode

hevrard:
Bug: #3729


hevrard:
This change is basically replicating the code here:
https://github.com/google/gapid/blob/0827598aadce1249ed1272dc48a01f82c8f065a5/gapii/client/adb.go#L138

PR #3755, opened 2020-02-05
Revert "Fix thread slices due to column name change (#3741)"

rschiu:
This reverts commit 51ce17e5ab307aa4f816e5ec0726e542483bd452.


pmuetschard:
why?

lpy:
I have verified that this patch caused the thread slices keep loading, after I reverted the original patch everything works fine.

I have also captured a new trace with the latest qpr2 release user build, and the `sched` table still has `row_id` instead of `id`.

pmuetschard:
Please update the PR description with a rollback reason. It should have been part of the commit message, which I asked for, but that can't be helped now. Commit messages should be more about the "why", not so much the "what".

PR #3758, opened 2020-02-06
Fix the mismatch between issue command id vs trace command id

yalcinmelihyasin:
due to the initial commands

b/148923131

### DIFF
--- a/gapis/api/vulkan/find_issues.go
+++ b/gapis/api/vulkan/find_issues.go
@@ -295,7 +295,7 @@ func (t *findIssues) Flush(ctx context.Context, out transform.Writer) error {
 				return
 			}
 			// The debug report is issued for a trace command
-			issue.Command = api.CmdID(label - uint64(t.numInitialCmds))
+			issue.Command = api.CmdID(label)
 			issue.Error = fmt.Errorf("%s", msg)
 			issue.Severity = service.Severity(uint32(eMsg.GetSeverity()))
 			t.issues = append(t.issues, issue)
--- a/gapis/resolve/report.go
+++ b/gapis/resolve/report.go
@@ -24,6 +24,7 @@ import (
 	"github.com/google/gapid/gapis/database"
 	"github.com/google/gapid/gapis/messages"
 	"github.com/google/gapid/gapis/replay"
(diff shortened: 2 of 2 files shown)


gapis/api/vulkan/find_issues.go
```diff
@@ -295,7 +295,7 @@ func (t *findIssues) Flush(ctx context.Context, out transform.Writer) error {
 				return
 			}
 			// The debug report is issued for a trace command
-			issue.Command = api.CmdID(label - uint64(t.numInitialCmds))
+			issue.Command = api.CmdID(label)
```
pmuetschard: I don't think this change is a good idea. `issue.Command` should remain a valid command ID. To me it looks like this change just adds an offset of `t.numInitialCmds` here, just to remove it again when resolving the report.

gapis/resolve/report.go
```diff
@@ -117,10 +118,13 @@ func (r *ReportResolvable) Resolve(ctx context.Context) (interface{}, error) {
 		}
 	}
 
+	InitialCommands, _, _ := initialcmds.InitialCommands(ctx, r.Path.Capture)
```
pmuetschard: lowercase variable names
I'm not a fan of this new dependency. Especially since it looks like it's just here to undo the change in `find_issues.go`.

gapis/api/vulkan/find_issues.go
yalcinmelihyasin: Well, it is actually the other way around if you look at the rest of the code. 

The label here is a cmdID counted with the initial commands. We used to convert it by substracting the number of the initial commands. 

If we do the conversion here, we have no way to understand that if it's an issue coming find_issues.go with a cmdID without initial commands  or if it's an issue coming from externs.go with a cmdID including initial commands. We cannot resolve that at that point.

gapis/resolve/report.go
yalcinmelihyasin: No, it is not what is happening here. 

This has to be done because the issues coming from externs.go counted with the initial commands.

We have two different way to create issues and they have been counted differently before, we have to unify them in a single place. And this seems to be the best place for it,

yalcinmelihyasin:
This is updated and merged in new repo

PR #3763, opened 2020-02-06
Fix taking graphics trace on non-perfetto enabled devices.

AWoloszyn:


lpy:
Is this a dup of https://github.com/google/gapid/pull/3762?

AWoloszyn:
> Is this a dup of #3762?

Yes, and like Pascal's solution better so use that one I will close this one

PR #3764, opened 2020-02-06
Allow users to customize their Perfetto tracing config.

pmuetschard:


gapic/src/main/com/google/gapid/perfetto/views/TraceConfigDialog.java
```diff
@@ -742,4 +808,52 @@ protected void okPressed() {
...
+      input.setText(TextFormat.printToString(settings.perfetto().getCustomConfig()));
+    }
+
+    @Override
+    public void update(Settings settings) {
+      try {
+        TextFormat.merge(input.getText(), settings.writePerfetto()
+            .getCustomConfigBuilder()
+            .clear());
+        settings.writePerfetto().setUseCustom(true);
```
stellama0208: It seems that the settings are not actually parsed and remembered into seperate fields here, neither is there a `onSwitchedTo` method for BasicInputArea, so there comes some surprise that the change get lost when switching back and forth between these two modes. 
Should we implement this part in this PR as well? Or is it intended that this part will be done in some future PR?

pmuetschard: The idea is to populate the text box with the current version of the proto. That is, when switching from basic to advanced, the advanced text is reset to what would be used from the settings in the basic mode. It's not worth trying to be too smart and infer the basic mode selections (especially since it's impossible to maintain fidelity) when going from advanced mode to basic mode. And since the advanced mode input gets reset when switching back to advanced, there is no point in remembering the advanced mode text.

I have adjusted the link text to make it a bit more clear that this is happening.

stellama0208: I see the point that it's hard to maintain fidelity, then it's good to go!

PR #3765, opened 2020-02-06
Create Pipeline Strip in Pipeline View

hliatis:
Also handles hovering and arrows over disabled stages

Use stack layout in Pipeline View

### DIFF
--- a/gapic/src/main/com/google/gapid/views/PipelineView.java
+++ b/gapic/src/main/com/google/gapid/views/PipelineView.java
@@ -15,6 +15,7 @@
  */
 package com.google.gapid.views;
 
+import static com.google.gapid.util.Colors.lerp;
 import static com.google.gapid.util.Loadable.MessageType.Error;
 import static com.google.gapid.util.Loadable.MessageType.Info;
 import static com.google.gapid.util.Logging.throttleLogRpcError;
@@ -62,20 +63,26 @@ import org.eclipse.swt.custom.ST;
 import org.eclipse.swt.custom.ScrolledComposite;
 import org.eclipse.swt.custom.StyleRange;
 import org.eclipse.swt.custom.StyledText;
+import org.eclipse.swt.custom.StackLayout;
 import org.eclipse.swt.graphics.Point;
 import org.eclipse.swt.graphics.Rectangle;
+import org.eclipse.swt.graphics.RGB;
+import org.eclipse.swt.graphics.Color;
 import org.eclipse.swt.layout.FillLayout;
 import org.eclipse.swt.layout.GridData;
 import org.eclipse.swt.layout.GridLayout;
+import org.eclipse.swt.layout.RowLayout;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Group;
 import org.eclipse.swt.widgets.Label;
 import org.eclipse.swt.widgets.TabFolder;
 import org.eclipse.swt.widgets.TabItem;
+import org.eclipse.swt.widgets.Button;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.HashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.logging.Logger;
 
@@ -90,6 +97,10 @@ public class PipelineView extends Composite
   protected final LoadablePanel<Composite> loading;
   protected final Theme theme;
   protected final Composite stagesContainer;
+  protected final Color hoverColor;
+
+  protected String selectedStage = null;
+  protected Button hoveredButton = null;
 
   public PipelineView(Composite parent, Models models, Widgets widgets) {
     super(parent, SWT.NONE);
@@ -101,12 +112,17 @@ public class PipelineView extends Composite
     loading = LoadablePanel.create(this, widgets,
         panel -> createComposite(panel, new FillLayout(SWT.VERTICAL)));
 
-    stagesContainer = createComposite(loading.getContents(), new FillLayout());
+    stagesContainer = createComposite(loading.getContents(), new GridLayout(1, false));
+
+    RGB selectRGB = getDisplay().getSystemColor(SWT.COLOR_BLUE).getRGB();
+    RGB unselectRGB = getDisplay().getSystemColor(SWT.COLOR_LIST_BACKGROUND).getRGB();
+    hoverColor = new Color(getDisplay(), lerp(selectRGB, unselectRGB, 0.5f), (int)(0.5 * 255));
 
     models.capture.addListener(this);
     models.commands.addListener(this);
     models.resources.addListener(this);
     addListener(SWT.Dispose, e -> {
+      hoverColor.dispose();
       models.capture.removeListener(this);
       models.commands.removeListener(this);
       models.resources.removeListener(this);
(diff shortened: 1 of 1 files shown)


hliatis:
Late amend makes the labels in the buttons bigger and has the size of the arrow be percentages of the box sizes.

hliatis:
Tooltips

gapic/src/main/com/google/gapid/views/PipelineView.java
```diff
@@ -184,202 +192,293 @@ protected void onUiThreadError(Loadable.Message error) {
...
+          stageButton.addListener(SWT.Selection, e -> {
+            stageStack.topControl = stageMap.get(stage.getDebugName());
+            stageComposite.layout();
+            selectedStage = stage.getDebugName();
+            stripComposite.redraw();
+          });
+
+          stageButton.setBackground(getDisplay().getSystemColor(SWT.COLOR_LIST_BACKGROUND));
+          stageButton.setFont(theme.bigBoldFont());
+          stageButton.setToolTipText(stage.getStageName() + " (" + stage.getDebugName() + ")");
```
johnplate: No need for DebugName here, it is already in the button. Change to:
```stageButton.setToolTipText(stage.getStageName());```

File: gapic/src/main/com/google/gapid/views/PipelineView.java [hunk 2]
```diff
@@ -184,202 +192,293 @@ protected void onUiThreadError(Loadable.Message error) {
...
   }
 
-  private void createPipelineTabs(TabFolder folder, List<API.Pipeline> pipelines) {
+  private void createPipelineTabs(Composite folder, List<API.Pipeline> pipelines) {
+    HashMap<String, Composite> stageMap = new HashMap<String, Composite>();
+
     if (!pipelines.isEmpty()) {
-      for (API.Pipeline result : pipelines) {
-        List<API.Stage> stages = result.getStagesList();
+      Composite stripComposite = withLayoutData( createComposite(folder, new FillLayout(SWT.HORIZONTAL)),
```
johnplate: The strip looks weird when stretched too much on a big screen. Let's not fill the width. Please replace the line with:
```
      RowLayout stripLayout = new RowLayout();
      stripLayout.fill = true;
      stripLayout.spacing = 0;
      stripLayout.wrap = true;
      Composite stripComposite = withLayoutData( createComposite(folder, stripLayout),
```

gapic/src/main/com/google/gapid/views/PipelineView.java
```diff
@@ -184,202 +192,293 @@ protected void onUiThreadError(Loadable.Message error) {
...
+              stripComposite.redraw();
+            });
+          }
 
-          FillLayout nameLayout = new FillLayout(SWT.VERTICAL);
-          nameLayout.marginHeight = 5;
-          Composite nameComposite = withLayoutData( createComposite(stageGroup, nameLayout),
-              new GridData(SWT.BEGINNING, SWT.BEGINNING, false, false));
+          if (stageIndex != stages.size()-1) {
+            Composite arrowSpace = createComposite(stripComposite, new FillLayout());
```
johnplate: This doesn't work anymore with the non-filled strip. Replace with:
```
            Label arrowSpace = new Label(stripComposite, SWT.FILL);
            arrowSpace.setText("     ");
            arrowSpace.setFont(theme.bigBoldFont());
```

gapic/src/main/com/google/gapid/views/PipelineView.java
```diff
@@ -184,202 +192,293 @@ protected void onUiThreadError(Loadable.Message error) {
...
+            API.Stage nextStage = stages.get(stageIndex+1);
 
-          Label stageName = createLabel(nameComposite, stage.getStageName() + " (" + stage.getDebugName() + ")");
-          stageName.setFont(theme.bigBoldFont());
+            arrowSpace.addListener(SWT.Paint, e -> {
+              Rectangle areaSize = arrowSpace.getClientArea();
```
johnplate: needs getBounds() with the above change

gapic/src/main/com/google/gapid/views/PipelineView.java
```diff
@@ -184,202 +192,293 @@ protected void onUiThreadError(Loadable.Message error) {
...
+              int lineLength = (int)(areaSize.width * 0.9);
+              int triangleBaseLength = (int)(areaSize.height * 0.15);
```
johnplate: Bigger arrowhead looks better, please change 0.9 to 0.8 and 0.15 to 0.2.

gapic/src/main/com/google/gapid/views/PipelineView.java
```diff
@@ -184,202 +192,293 @@ protected void onUiThreadError(Loadable.Message error) {
...
-        for (API.Stage stage : stages) {
-          TabItem item = createStandardTabItem(folder, stage.getDebugName());
+          if (!stage.getEnabled()) {
+            stageButton.setEnabled(false);
+            stageButton.addListener(SWT.Paint, e -> {
+              Rectangle areaSize = stageButton.getBounds();
+              e.gc.setBackground(getDisplay().getSystemColor(SWT.COLOR_LIST_FOREGROUND));
+              e.gc.setLineWidth((int)(areaSize.height * 0.1));
+              e.gc.drawLine(0, areaSize.height / 2, areaSize.width, areaSize.height / 2);
+              e.gc.drawPolygon(new int[] {0, 0, areaSize.width - 1, 0, areaSize.width - 1, areaSize.height - 1, 0, areaSize.height - 1});
```
johnplate: Remove this line, don't draw a custom frame for an inactive stage.

gapic/src/main/com/google/gapid/views/PipelineView.java
```diff
@@ -91,6 +96,9 @@
   protected final Theme theme;
   protected final Composite stagesContainer;
 
+  protected String selectedStage = null;
```
johnplate: add the following line after this line:
```protected Control selectedButton = null;```

File: gapic/src/main/com/google/gapid/views/PipelineView.java [hunk 8]
```diff
@@ -184,202 +192,293 @@ protected void onUiThreadError(Loadable.Message error) {
...
+          API.Stage stage = stages.get(stageIndex);
+          stageMap.put(stage.getDebugName(), createStage(stageComposite, stage));
+
+          Button stageButton = new Button(stripComposite, SWT.FLAT);
+          stageButton.setText(stage.getDebugName());
+          stageButton.addListener(SWT.Selection, e -> {
+            stageStack.topControl = stageMap.get(stage.getDebugName());
+            stageComposite.layout();
+            selectedStage = stage.getDebugName();
+            stripComposite.redraw();
```
johnplate: To fix the redraw on Windows, replace these two lines with:
```
            if (selectedButton != null) {
              selectedButton.redraw();
            }
            selectedStage = stage.getDebugName();
            selectedButton = stageButton;
            selectedButton.redraw();
```

gapic/src/main/com/google/gapid/views/PipelineView.java
```diff
@@ -184,202 +192,293 @@ protected void onUiThreadError(Loadable.Message error) {
...
+            stageButton.addListener(SWT.Paint, e -> {
+              Rectangle areaSize = stageButton.getBounds();
+              e.gc.setLineWidth((int)(areaSize.height * 0.1));
+
+              if (hoveredButton == stageButton) {
+                RGB hoverColor = getDisplay().getSystemColor(SWT.COLOR_BLUE).getRGB();
+                stageButton.setBackground(new Color(stageButton.getDisplay(), hoverColor.red+30, hoverColor.green+30, hoverColor.blue, 63));
+              } else {
+                stageButton.setBackground(getDisplay().getSystemColor(SWT.COLOR_LIST_BACKGROUND));
+              }
```
johnplate: Delete these six lines here. It is too late to change the button background color in the paint event (which happens after the widget has been drawn).
Further it is not good to create a new system resource (color) each time, and not to dispose them.

gapic/src/main/com/google/gapid/views/PipelineView.java
```diff
@@ -184,202 +192,293 @@ protected void onUiThreadError(Loadable.Message error) {
...
 
-          Composite stageGroup = createComposite(folder, new GridLayout());
+              if (hoveredButton == stageButton || (selectedStage != null && selectedStage.equals(stageButton.getText()))) {
```
johnplate: This works shorter and faster now with:
```if (hoveredButton == stageButton || selectedButton == stageButton) {```

File: gapic/src/main/com/google/gapid/views/PipelineView.java [hunk 11]
```diff
@@ -184,202 +192,293 @@ protected void onUiThreadError(Loadable.Message error) {
...
+                e.gc.setForeground(getDisplay().getSystemColor(SWT.COLOR_BLUE));
+              } else {
+                e.gc.setBackground(getDisplay().getSystemColor(SWT.COLOR_LIST_FOREGROUND));
+              }
+              e.gc.drawPolygon(new int[] {0, 0, areaSize.width - 1, 0, areaSize.width - 1, areaSize.height - 1, 0, areaSize.height - 1});
+            });
+
+            stageButton.addListener(SWT.MouseEnter, e -> {
+              hoveredButton = stageButton;
+              stripComposite.redraw();
```
johnplate: This is the right place to change the button background color. No need to redraw the strip, which doesn't work anyway, because it does not redraw all children.
So replace the line with:
```stageButton.setBackground(hoverColor);```
You have to create the hoverColor when initializing the pipeline view, and dispose it properly.
Also please interpolate the hoverColor to be halfway between blue and SWT.COLOR_LIST_BACKGROUND.

File: gapic/src/main/com/google/gapid/views/PipelineView.java [hunk 12]
```diff
@@ -184,202 +192,293 @@ protected void onUiThreadError(Loadable.Message error) {
...
+            });
+
+            stageButton.addListener(SWT.MouseExit, e -> {
+              hoveredButton = null;
+              stripComposite.redraw();
```
johnplate: Replace this line with resetting the background color:
```stageButton.setBackground(getDisplay().getSystemColor(SWT.COLOR_LIST_BACKGROUND));```

File: gapic/src/main/com/google/gapid/views/PipelineView.java [hunk 13]
```diff
@@ -184,202 +192,293 @@ protected void onUiThreadError(Loadable.Message error) {
...
+            });
+          }
 
-          FillLayout nameLayout = new FillLayout(SWT.VERTICAL);
-          nameLayout.marginHeight = 5;
-          Composite nameComposite = withLayoutData( createComposite(stageGroup, nameLayout),
-              new GridData(SWT.BEGINNING, SWT.BEGINNING, false, false));
+          if (stageIndex != stages.size()-1) {
```
johnplate: Insert the following selection of the initial button before this line:
```
          if (selectedButton == null) {
            selectedButton = stageButton;
            selectedStage = stage.getDebugName();
            stageStack.topControl = stageMap.get(selectedStage);
          }
```

gapic/src/main/com/google/gapid/views/PipelineView.java
```diff
@@ -184,202 +192,293 @@ protected void onUiThreadError(Loadable.Message error) {
...
+      }
 
-          FillLayout dataLayout = new FillLayout(SWT.VERTICAL);
-          dataLayout.spacing = 5;
-          Composite dataComposite = withLayoutData( createComposite(stageGroup, dataLayout),
+      if (selectedStage == null || stageMap.get(selectedStage) == null) {
+        selectedStage = pipelines.get(0).getStagesList().get(0).getDebugName();
+      }
+
+      stageStack.topControl = stageMap.get(selectedStage);
```
johnplate: Delete this here, moved.

gapic/src/main/com/google/gapid/views/PipelineView.java
hliatis: Ok, but just letting you know that the document and mockups include the debug name

gapic/src/main/com/google/gapid/views/PipelineView.java
hliatis: Why?

gapic/src/main/com/google/gapid/views/PipelineView.java
hliatis: Little worried about wrap being true. Check with Pau about this. I think he may have wanted it to shrink instead. The doc isn't explicit about this.

By the way, unless I implemented one of your changes wrong, this causes the strip to be left justified instead of centered. Also, the boxes are now different sizes, which the mockup does not imply. In fact, these were the reasons I used Fill Layout in the first place. I wasn't entirely pleased with it either, especially since BLEND never quite looked right.

Should we use a grid layout like we did for the KVP grids?

gapic/src/main/com/google/gapid/views/PipelineView.java
hliatis: This is causing an error when I change draw calls. Wouldn't selected button already be disposed?

gapic/src/main/com/google/gapid/views/PipelineView.java
hliatis: Except now every time I change commands, I go straight back to IA/CS instead of the stage I was at.

That was the whole point of using the StageMap. The buttons are disposed each time you switch commands because you may have different pipelines bound, meaning the stage you were on might not exist anymore. The one constant are the names. By using the stage map, I could detect if a stage at the newly selected command no longer existed (the selectedStage string would no longer be in the map) or not, and act accordingly (go the first stage available, whether it is IA or CS).

gapic/src/main/com/google/gapid/views/PipelineView.java
hliatis: See above comment

gapic/src/main/com/google/gapid/views/PipelineView.java
hliatis: Mind you, this is trivially fixed by resetting selectedButton to null every time we change pipelines

gapic/src/main/com/google/gapid/views/PipelineView.java
johnplate: The mockups don't address that the layout can be resized, so it is ok to deviate if we have a reason. Left justified is also intended. I will check everything with Pau. No need for grid layout.

gapic/src/main/com/google/gapid/views/PipelineView.java
johnplate: The thicker custom black line did not looked like grayed out, so I wanted to have it gray anyway. Thinner frame is also better to be a contrast to the arrow through it. Both is achieved with the original frame, so no need for a custom frame.

gapic/src/main/com/google/gapid/views/PipelineView.java
johnplate: Of course you can change/fix/improve anything I propose. My code is mostly a hint how to move forward.

gapic/src/main/com/google/gapid/views/PipelineView.java
johnplate: You can change it back and implement another solution to remember and redraw the previously selected button when you press another button. This is what caused the bug on Windows.

hliatis: Okay. After reviewing the code a bit, the best change is to probably get rid of selectedButton and have stageMap map stage name's to an plain object that contains both the composite and the button. The reason for this, is that I can't check if the selected stage from the last pipeline still exists in the new one until after I create all the buttons and stage composites.

gapic/src/main/com/google/gapid/views/PipelineView.java
```diff
@@ -101,12 +112,17 @@ public PipelineView(Composite parent, Models models, Widgets widgets) {
     loading = LoadablePanel.create(this, widgets,
         panel -> createComposite(panel, new FillLayout(SWT.VERTICAL)));
 
-    stagesContainer = createComposite(loading.getContents(), new FillLayout());
+    stagesContainer = createComposite(loading.getContents(), new GridLayout(1, false));
+
+    RGB selectRGB = getDisplay().getSystemColor(SWT.COLOR_BLUE).getRGB();
+    RGB unselectRGB = getDisplay().getSystemColor(SWT.COLOR_LIST_BACKGROUND).getRGB();
+    hoverColor = new Color(getDisplay(), lerp(selectRGB, unselectRGB, 0.5f), (int)(0.5 * 255));
```
johnplate: Please use:
```hoverColor = new Color(getDisplay(), lerp(selectRGB, unselectRGB, 0.5f));```
We want 100% opacity here, because if the color is 50% transparent and it would work, we would see 50% of the desktop shining through. This is not intended.
If the design doc says 50% opacity, it means you should apply 50% of the foreground color over the background color. This is achieved with the 50% interpolation.

File: gapic/src/main/com/google/gapid/views/PipelineView.java [hunk 16]
```diff
@@ -184,202 +200,298 @@ protected void onUiThreadError(Loadable.Message error) {
...
+          stageButton.addListener(SWT.Selection, e -> {
+            stageStack.topControl = stageMap.get(stage.getDebugName()).stageComposite;
+            stageComposite.layout();
+            Button selectedButton = stageMap.get(selectedStage).stageButton;
+            if (selectedButton != null) {
+              selectedButton.redraw();
+            }
+            selectedStage = stage.getDebugName();
+            selectedButton = stageButton;
+            selectedButton.redraw();
```
johnplate: With your solution there is no need to check the validity of selectedButton anymore.
It is more likely that selectedStage is not found in stageMap, which would cause get() to return null and a null reference exception. So better use something like:
```
            StageUI stageUI = stageMap.get(selectedStage);
            if (stageUI != null) {
              stageUI.stageButton.redraw();
            }
            selectedStage = stage.getDebugName();
            stageButton.redraw();
```

johnplate:
Good solution with the StageUI class. Almost done.

PR #3766, opened 2020-02-07
Add tooltips for Enums in Pipeline View

hliatis:
Bug: 148873707

### DIFF
--- a/gapic/src/main/com/google/gapid/views/PipelineView.java
+++ b/gapic/src/main/com/google/gapid/views/PipelineView.java
@@ -53,6 +53,7 @@ import com.google.gapid.widgets.Widgets;
 
 import org.eclipse.jface.text.source.SourceViewer;
 import org.eclipse.jface.viewers.ArrayContentProvider;
+import org.eclipse.jface.viewers.ColumnViewerToolTipSupport;
 import org.eclipse.jface.viewers.StyledCellLabelProvider;
 import org.eclipse.jface.viewers.TableViewer;
 import org.eclipse.jface.viewers.TableViewerColumn;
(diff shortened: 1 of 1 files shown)


johnplate:
Looks good overall, but a few comments/questions:
- Is there an associated buganizer task?
- Which enums are supposed to be covered by this PR? During my test only most KVP enums worked, but none in tables (on Windows, didn't test on Linux yet).
- The Cull Mode enum in Rasterization State did not work.

hliatis:
Cull Mode is actually not a Vulkan enum but a bitfield (for reasons). I'll look into it.

johnplate:
VkCullModeFlagBits is an enum that is used as bit field. As the existing four enums cover all possible bit combinations, we can simply see it as enum.

hliatis:
It's sent to the client as a bitfield. The plumbing for bitfields is different from simple enums. I am looking into a solution now so that it can handle other, more "pure" bitfields as well

hliatis:
For bitfields, the UI displays a truncated name separated by '|' (except for colors, where there is no separator). 

Would you be okay with a long tooltip where it's the longer names separated by '|'?

johnplate:
As the client receives it as bitfield, I think we can leave the cull mode as is for now.

hliatis:
I mean, I have the most basic solution where I show all the bit flags used separated by a |. The tip can be long for something like using RGBA bits in a color bitfield. Shall I push it?

johnplate:
Please don't forget to answer this:
> * Which enums are supposed to be covered by this PR? During my test only most KVP enums worked, but none in tables (on Windows, didn't test on Linux yet).

hliatis:
All enums, including tables. I am not sure what is going on there. I am using SWT's standard way of doing this.

PR #3767, opened 2020-02-07
Fix vulkan event linking issue when zoomed too out.

stellama0208:
Related ticket: http://b/148952158


stellama0208:
> This doesn't work as is as soon as aggregation happens across slices with different submission ids.
> 
> The inner-most query aggregates over (quantum_ts, depth), but you are not using an aggregation function for the submission_id column. Same is true for the next level query, which groups by (depth, label, i). A more stringent SQL parser than SQLite wouldn't even allow this.

Hi Pascal, I've updated the SQL query. The inner-most query will aggregate by `first_value(submission_id) over win1`, the outer-most query will aggregate by `first_value(submission_id) over (partition by depth, label, i)`, the middle one is kept as `submission_id` for consistency and the reason of no newer grouping.

Please take another look, thanks!

stellama0208:
This PR is moved to AGI repo (https://github.com/google/agi/pull/20), will close it here.

PR #3768, opened 2020-02-07
Make the vsync marker contrast higher.

lpy:
Bug: b/149043217


lpy:
Before
image

After
image

pmuetschard:
IMHO: it's now too dark.
It's definitely too close to the title background color. The original color was chosen to be at a reasonable point between the background (0xff) and the title background (0xf1). It should stay between the two. Please work with Carl to get a color, since he and I worked on the current colors.

lpy:
Based on Carl's feedback, change the gray

After:
image

PR #3769, opened 2020-02-07
Change API function tables from maps to flat arrays

purvisa-at-google-com:
Many look-ups are performed on these tables and they are only 64k of void* entries max.

### DIFF
--- a/gapir/cc/function_table.h
+++ b/gapir/cc/function_table.h
@@ -49,20 +49,25 @@ class FunctionTable {
   inline Function* lookup(Id id);
 
  private:
-  // Map of the supported function ids to the actual function implementations
-  std::unordered_map<Id, Function> mFunctions;
+  // Array of the actual function implementations by function ID
+  // This is stored as an array rather than a map because many lookups are
+  // performed at replay time and this becomes a bottleneck when stored as a
+  // map. The 64k entries (limit mandated elsewhere in the code by the vm
+  // bytecode instruction packing) are small enough that storing them as an
+  // array isn't a problem.
+  Function mFunctions[65536] = {};
 };
 
 inline FunctionTable::Function* FunctionTable::lookup(Id id) {
-  auto func = mFunctions.find(id);
-  if (func == mFunctions.end()) {
+  Function& ret = mFunctions[id];
+  if (ret == nullptr) {
     return nullptr;
   }
-  return &func->second;
+  return &ret;
 }
 
 inline void FunctionTable::insert(Id id, Function func) {
-  if (mFunctions.count(id) != 0) {
+  if (mFunctions[id] != NULL) {
     GAPID_FATAL("Duplicate functions inserted into table");
   }
   mFunctions[id] = func;
--- a/gapir/cc/interpreter.cpp
+++ b/gapir/cc/interpreter.cpp
@@ -94,11 +94,7 @@ void Interpreter::registerBuiltin(uint8_t api, FunctionTable::Id id,
 
 void Interpreter::setRendererFunctions(uint8_t api,
                                        FunctionTable* functionTable) {
-  if (functionTable != nullptr) {
-    mRendererFunctions[api] = functionTable;
-  } else {
-    mRendererFunctions.erase(api);
-  }
+  mRendererFunctions[api] = functionTable;
 }
 
 void Interpreter::resetInstructions() {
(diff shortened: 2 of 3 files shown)


gapir/cc/function_table.h
```diff
@@ -49,23 +55,23 @@ class FunctionTable {
   inline Function* lookup(Id id);
 
  private:
-  // Map of the supported function ids to the actual function implementations
-  std::unordered_map<Id, Function> mFunctions;
+  // Array of the actual function implementations by function ID
+  Function mFunctions[65536];
```
AWoloszyn: Function mFunctions[65536] = {};
Then you shouldnt have to zero-initailize in the constructor by hand.

gapir/cc/function_table.h
```diff
@@ -49,23 +55,23 @@ class FunctionTable {
...
+  if (ret == nullptr) {
     return nullptr;
   }
-  return &func->second;
+  return &ret;
 }
 
 inline void FunctionTable::insert(Id id, Function func) {
-  if (mFunctions.count(id) != 0) {
+  if (mFunctions[(uint16_t)id] != NULL) {
```
AWoloszyn: What happens if id > 655536? Probably deserves an assert, as this will be a tricky bug to track down in the future./
Or make Id a 16-bit number.

gapir/cc/interpreter.cpp
```diff
@@ -76,8 +76,15 @@ Interpreter::Interpreter(core::CrashHandler& crash_handler,
                     stack->printStack();
                     return true;
                   });
+
+  mRendererFunctions = new FunctionTable*[16];
```
AWoloszyn: Why allocate this on the heap if its always going to be the same size?
FunctionTable* mRendererFunctions[16] = {};

gapir/cc/function_table.h
purvisa-at-google-com: typedef uint16_t Id;

AWoloszyn: Then the cast should not be necessary?

AWoloszyn: And to make things more clear, we can change the constant 65536 to 
std::numeric_limits<Id>::max. That makes it perfectly clear that we cannot overflow.

purvisa-at-google-com: Fair.

gapir/cc/function_table.h
purvisa-at-google-com: Good idea, I forgot we're using c++11

gapir/cc/interpreter.cpp
purvisa-at-google-com: Done.

gapir/cc/function_table.h
AWoloszyn: Can we also remove the cast (Also c-style casts are forbidden in the style guide)

purvisa-at-google-com:
Comments addressed. Can I get another review?

gapir/cc/function_table.h
```diff
@@ -49,20 +49,20 @@ class FunctionTable {
   inline Function* lookup(Id id);
 
  private:
-  // Map of the supported function ids to the actual function implementations
-  std::unordered_map<Id, Function> mFunctions;
+  // Array of the actual function implementations by function ID
```
hevrard: Maybe add in comment *why* it's good to use an array here, rather than e.g. a map.

gapir/cc/interpreter.h
```diff
@@ -181,10 +181,10 @@ class Interpreter {
   const MemoryManager* mMemoryManager;
 
   // The builtin functions.
-  std::unordered_map<uint8_t, FunctionTable> mBuiltins;
+  FunctionTable mBuiltins[16];
```
hevrard: Add a short comment about why it's `16` here.

gapir/cc/interpreter.h
```diff
@@ -181,10 +181,10 @@ class Interpreter {
...
 
   // The current renderer functions.
-  std::unordered_map<uint8_t, FunctionTable*> mRendererFunctions;
+  FunctionTable* mRendererFunctions[16] = {};
```
hevrard: Same here, maybe have a single comment for both `16` values, if you find a nice way to do this.

gapir/cc/function_table.h
purvisa-at-google-com: Done

gapir/cc/interpreter.h
purvisa-at-google-com: Done

gapir/cc/interpreter.h
purvisa-at-google-com: Done

PR #3774, opened 2020-02-07
Add Vulkan version annotation to the api files

zakerinasab:
This change adds Vulkan version annotation to the api
files in gapis/api/vulkan/api. This allows to distinguish
between Vulkan releases if a third party user had to use
an older API version than GAPID.

### DIFF
--- a/gapis/api/vulkan/api/command_buffer_control.api
+++ b/gapis/api/vulkan/api/command_buffer_control.api
@@ -49,7 +49,6 @@
   @unused ref!VulkanDebugMarkerInfo                      DebugInfo
 }
 
-@since("1.0")
 @indirect("VkDevice")
 cmd VkResult vkCreateCommandPool(
     VkDevice                       device,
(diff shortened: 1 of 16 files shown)


pmuetschard:
Should we call this `@since`?

zakerinasab:
> Should we call this `@since`?

Done.

gapis/api/vulkan/api/command_buffer_control.api
```diff
@@ -643,6 +651,7 @@ sub void dovkCmdExecuteCommands(ref!vkCmdExecuteCommandsArgs cmds) {
   leaveSubcontext()
 }
 
+@since("1.0")
```
AWoloszyn: I would be happy to not include @since(1.0) and just assume that everything without an annotation is EITHER an extension or core functionality. (Just to trim down the number of things we have to add to EVERY single entry-point)

PR #3777, opened 2020-02-07
Force stop already running gapidapk before starting

silence-do-good:
The current instance of gapidapk is kept alive in the device and resumed
during the loadDevices phase. This causes a prolonged wait and eventual
failure to find the android device, if the driver is updated and gapid is
launched without restarting the device.

Bug: 149109076

### DIFF
--- a/gapidapk/deviceinfo.go
+++ b/gapidapk/deviceinfo.go
@@ -108,6 +108,9 @@ func fetchDeviceInfo(ctx context.Context, d adb.Device) error {
 		return nil
 	}
 
+	// Close any previous runs of the apk
+	apk.Stop(ctx)
+
 	driver, err := d.GraphicsDriver(ctx)
 	if err != nil {
 		return err


lpy:
It seems to me if there are two instances launched there will be an issue.

rschiu:
I don't think it is possible to have two instances running.  This is an Android service.  So if the service was already launched, the OS will not relaunched it.
In the normal case, the GAPID will just reconnected with the existing deviceinfo and retrieve the data correctly.  But if the prerelease driver was changed, then the deviceinfo service will get into some kind of zombie state because the original driver it loaded no longer exist.

rschiu:
Right, maybe if just add a timeout, that will be good enough.  We really should add a timeout anyway.
This should only run on device discovery, which is done once per session.  It's uncommon for two GAPID doing device discovery at the same time.  And if that does happens, user just need to click refresh, as long as we implemented the timeout.

silence-do-good:
> I mean two gapid instances, now after EnsureInstalled, we kill the service, which I think may interrupt the existing socket because one instance could be waiting for it.
> 
> The service will be killed by the OS eventually. Per android dev doc, the service code can't kill itself though... :(

Yes. I've noticed the process gets killed if the device is left idle for some time. But how long it takes to do that, I'm not sure. If we can force the user to restart the device after driver updates, then there is no need for this fix. But if not, it is very visible.

silence-do-good:
> It seems to me if there are two instances launched there will be an issue.

Is this a very common use case? I haven't done this so far so I did not think in this direction.

pmuetschard:
The problem we are trying to fix with this is a corner case, but it is annoying that the service is not restarted when the driver changes. Is there a `android:configChanges` event we could specify in the manifest to get restarted by the system if the driver changes? If not, I'm OK with this change. The possibility that the force-stop will interfere with another device info is mostly just theoretical.

`apk.Stop()` causes an `am force-stop`. Will that stop services, too?

lpy:
We can listen to package change event and react to that if a specific package we are interested in is changed.

PR #3779, opened 2020-02-10
Make checks for API function pointers and handle mappings debug only

purvisa-at-google-com:


gapis/api/templates/specific_gfx_api.cpp.tmpl
```diff
@@ -145,7 +145,13 @@
         {{$args := (GetAnnotation $ "indirect").Arguments}}
         {{$elem := (index $.CallParameters 0).Name}}
         {{$func := Macro "CmdName" $}}
-        if ({{Template "GetIndirectedFunction" "Annotations" $args "Element" $elem "Function" $func}}) {
+        if (
+        #ifdef DEBUG
```
hevrard: The only `#ifdef DEBUG` that I could find in the codebase is actually a `#ifndef NDEBUG`:
https://github.com/google/gapid/blob/2efbe29f4d25dfa279927dc1d84534713884e24c/core/memory_tracker/cc/posix/memory_tracker.inc#L66

I personally prefer `#ifdef DEBUG`, but maybe there's a good reason it hasn't been used in the other case?

gapis/api/templates/specific_gfx_api.cpp.tmpl
```diff
@@ -145,7 +145,13 @@
...
+        {{Template "GetIndirectedFunction" "Annotations" $args "Element" $elem "Function" $func}}
+        #else
+        true
```
hevrard: Should this be `false`, to make the `if` body dead-code when not in DEBUG mode?

purvisa-at-google-com: We still want to call the API function, we just don't want to check all the pointers and handles for validity before doing it.

gapis/api/templates/specific_gfx_api.cpp.tmpl
AWoloszyn: We use NDEBUG because that is what "assert.h" uses, and is the "standard" way to do things. Not every compiler defined "DEBUG" and NDEBUG I think catches some weird cases that you see in bazel. (What happens when you are building fastbuild vs/ release vs/ debug)

gapis/api/templates/specific_gfx_api.cpp.tmpl
hevrard: OK, thanks for the clarification.

AWoloszyn: Having a preprocessor macro inside of a condition is ... awkward.

```
if (
#ifdef
.....
#else
true
#endif
) {
```
I would prefer we generate something slightly nicer.

```
#ifndef NDEBUG
if (!(...)) {
   GAPID_ERROR("Error during calling function {{$name}}");
      return false;
}
#endif
.....
```

gapis/api/templates/specific_gfx_api.cpp.tmpl
purvisa-at-google-com: I also prefer DEBUG, but changed

gapis/api/templates/specific_gfx_api.cpp.tmpl
purvisa-at-google-com: If we did that we'd call ATrace_beginSection() without calling ATrace_endSection() and any other post-call logic would need to be duplicated. All of the API file code is "awkward" :)

PR #3781, opened 2020-02-10
Rewrite GAPIR stack to use C++ type system based stack

purvisa-at-google-com:
Also updates Interpreter to use the new stack


gapir/cc/context.cpp
```diff
@@ -123,13 +123,11 @@ bool Context::interpret(bool cleanup, bool isPrewarm) {
...
       // There is only one vulkan "renderer" so we create it when requested.
       mVulkanRenderer = VulkanRenderer::create();
-      if (mVulkanRenderer->isValid()) {
-        mVulkanRenderer->setListener(this);
-        Api* api = mVulkanRenderer->api();
-        interpreter->setRendererFunctions(api->index(), &api->mFunctions);
-        GAPID_INFO("Bound Vulkan renderer");
-        return true;
-      }
+      mVulkanRenderer->setListener(this);
```
AWoloszyn: What happens if the vulkan renderer is invalid?
Do we want GAPIR to crash in this case?

gapir/cc/interpreter.cpp
```diff
@@ -236,27 +337,82 @@ Interpreter::Result Interpreter::pushI(uint32_t opcode) {
...
+      mStack.push<uint32_t>(data);
+      break;
+
+    case BaseType::Uint64:
+      mStack.push<uint64_t>(data);
+      break;
+
     case BaseType::Float:
-      data <<= 23;
+      data <<=
```
AWoloszyn: The alignment here is very odd.

gapir/cc/interpreter.cpp
```diff
@@ -236,27 +337,82 @@ Interpreter::Result Interpreter::pushI(uint32_t opcode) {
...
+    case BaseType::ConstantPointer:
+      mStack.push(Stack::ConstantPointer(data));
+      break;
+
+    case BaseType::VolatilePointer:
+      mStack.push(Stack::VolatilePointer(data));
       break;
+
     default:
+      assert(false);  // Unhandled type in pushI
```
AWoloszyn: GAPID_ERROR please, otherwise this will get completely ignored in release.

AWoloszyn: GAPID_FATAL also is fine.

gapir/cc/interpreter.cpp
```diff
@@ -287,8 +505,70 @@ Interpreter::Result Interpreter::loadV(uint32_t opcode) {
...
+    case BaseType::ConstantPointer:
+      mStack.push(Stack::ConstantPointer(*(int32_t*)address));
+      break;
+
+    case BaseType::VolatilePointer:
+      mStack.push(Stack::VolatilePointer(*(int32_t*)address));
+      break;
+
+    default:
+      assert(false);  // Unhandled type in loadC
```
AWoloszyn: GAPID_FATAL

gapir/cc/interpreter.cpp
```diff
@@ -302,46 +582,104 @@ Interpreter::Result Interpreter::load(uint32_t opcode) {
...
 
 Interpreter::Result Interpreter::resource(uint32_t opcode) {
-  mStack.push<uint32_t>(extract26bitData(opcode));
+  auto id = extract26bitData(opcode);
+  mStack.push<uint32_t>(id);
   return this->call(Interpreter::RESOURCE_FUNCTION_ID);
 }
 
-Interpreter::Result Interpreter::post() {
+Interpreter::Result Interpreter::post(uint32_t) {
```
AWoloszyn: Why do these take uint32_t values now?

gapir/cc/interpreter.cpp
```diff
@@ -411,99 +749,135 @@ Interpreter::Result Interpreter::strcpy(uint32_t opcode) {
...
+  } else if (type == typeid(void*)) {
+    auto val = mStack.pop<void*>();
+    if (sizeof(void*) == 8) {
+      uint64_t* ptr = (uint64_t*)&val;
+      *ptr = (*ptr << 26) | data;
+    } else if (sizeof(void*) == 4) {
+      uint32_t* ptr = (uint32_t*)&val;
+      *ptr = (*ptr << 26) | data;
+    } else {
+      assert(false);  // This system doesn't have 8 or 4 byte pointers?
```
AWoloszyn: GAPID_FATAL.

gapir/cc/interpreter.cpp
```diff
@@ -411,99 +749,135 @@ Interpreter::Result Interpreter::strcpy(uint32_t opcode) {
...
-      value |= data;
-      value &= 0x000fffffffffffffULL;
-      value |= exponent;
-      break;
-    }
-    // Extending the value with 26 new LSB
-    default: {
-      value = (value << 26) | data;
-      break;
+  const std::type_index& type = mStack.typeAtTop();
```
AWoloszyn: Are we compiling with RTTI turned on?
I would hope that we are not .

gapir/cc/interpreter.h
```diff
@@ -129,6 +129,9 @@ class Interpreter {
   // 7th to 13th MSB (both inclusive) of the opcode
   BaseType extractType(uint32_t opcode) const;
 
+  // Get 6 bit data out from an opcode located in the 26-20 LSB of the opcode.
+  uint32_t extract6bitData(uint32_t opcode) const;
```
AWoloszyn: This function does not behave the same as the rest. It should have a different name.
Every other ExtractXBitData extracts from the LSB, this extracts from somewhere in the middle/

What do these 6 bits represent? Is is just arbitrary data? or is it something more./

gapir/cc/stack.h
```diff
@@ -26,457 +26,172 @@
...
-    }
-    BaseType type = getTopType();
-    if (type != BaseType::VolatilePointer) {
-      GAPID_WARNING("popVolatile called for type %s", baseTypeName(type));
-      mValid = false;
-      return nullptr;
-    }
-    return pop<T*>();
+  void push(T value) {
+    assert(mNumEntries < mMaxEntries);
```
AWoloszyn: We do not use assert. We have our own.

gapir/cc/stack.h
```diff
@@ -26,457 +26,172 @@
...
-      mValue.u32 = uint32_t(val);
+    template <class T>
+    const T& get() {
+      assertCanStore<T>();
+#ifndef NDEBUG
+      // We only check the typeid for get() in debug builds. The rest of the
+      // time we just make sure the number of bytes is correct.
+      assert(sizeof(T) == mDataSize && mDataType == typeid(T));
+#else
+      assert(sizeof(T) == mDataSize);
```
AWoloszyn: This is going to get optimized out ANYWAY.

assert() is disabled #ifdef NDEBUG.

gapis/api/templates/vulkan_gfx_api_extras.tmpl
```diff
@@ -209,158 +209,123 @@ bool Vulkan::replayCreateVkDeviceImpl(Stack* stack, size_val physicalDevice,
...
-      // Get all queues for this device and bind them.
-      for (auto queue : getVkQueues(stubs.vkGetDeviceQueue, device, createInfo)) {
-        mIndirectMaps.VkQueuesToVkDevices[queue] = device;
-      }
-      return true;
-    } else {
-      GAPID_WARNING("Error during calling function replayRegisterVkDevice");
+    GAPID_DEBUG("replayRegisterVkDevice(%zu, %zu, %p)", physical_device, device, createInfo);
+    if (mVkDeviceFunctionStubs.count(device) > 0) {
+      // TODO(antiagainst): The same device handle returned from the driver. Is this possible?
```
AWoloszyn: This comment can be removed, this is not actually possible.

gapir/cc/interpreter.cpp
purvisa-at-google-com: Why would you hope that we are not?

gapir/cc/stack.h
purvisa-at-google-com: Of course we do :)

gapir/cc/stack.h
purvisa-at-google-com: Fair point. Whoops!

gapir/cc/interpreter.h
purvisa-at-google-com: Sorry, this slipped through from another optimisation that is yet to be put up for code review.

gapir/cc/interpreter.cpp
purvisa-at-google-com: clang-format

gapir/cc/context.cpp
purvisa-at-google-com: Oh crap. Good catch. Knocked this out by accident!!!

gapir/cc/interpreter.cpp
AWoloszyn: RTTI has historically been poorly supported on Android.
Also it has a non-trivial performance cost for all objects regardless of if RTTI is used in that object or not.

The second reason is that it is good practice for the developers of this tool to apply similar constraints as our users have. If only to have more practice with how our users work. (Most game developers disable RTTI & Exceptions)

gapir/cc/interpreter.cpp
purvisa-at-google-com: For symmetry with the other functions. I was playing with replacing the switch function with an array of function pointers and it was required then to get the same signature on each function. I can remove the unused int if you prefer it.

gapir/cc/interpreter.cpp
purvisa-at-google-com: Like all things it, the appropriateness of RTTI depends on when and where you use it. Much of the usage here is compile time evaluated and has zero overhead. I changed the code because vtune was telling me that we were spending more time in our own RTTI replacement than we would spend using C++'s built in type system. Slower and non-standard loses to faster and standard in my books.

purvisa-at-google-com:
> This change does a **LOT** more than just change the gapir stack to use the C++ type system. In fact by the looks of it, most of the performance gained here is due to the other changes, not the change to the stack itself.

Vtune was showing the old stack functions as some of the most expensive functions by cumulative self-time. With these changes, the stack barely appears anymore,

AWoloszyn:
> > This change does a **LOT** more than just change the gapir stack to use the C++ type system. In fact by the looks of it, most of the performance gained here is due to the other changes, not the change to the stack itself.
> 
> Vtune was showing the old stack functions as some of the most expensive functions by cumulative self-time. With these changes, the stack barely appears anymore,

This largely depends on if you simply moved the time out of those functions and smeared them into the outer functions. I am not saying this is NOT a valid change, but we can't validate that unless we have it without the rest of the changes that also have a performance impact.

gapir/cc/stack.h
```diff
@@ -26,457 +26,172 @@
...
-    }
-    BaseType type = getTopType();
-    if (type != BaseType::ConstantPointer) {
-      GAPID_WARNING("popConstant called for type %s", baseTypeName(type));
-      mValid = false;
-      return nullptr;
-    }
-    return pop<const T*>();
-  }
+  void discard(uint32_t count) { assert(mNumEntries-- > 0); }
```
AWoloszyn: This is ignoringthe (count) parameter.

gapir/cc/stack.h
```diff
@@ -26,457 +26,172 @@
...
+  uint32_t mMaxEntries;
+  Entry* mStack;
+
   const MemoryManager* mMemoryManager;
 };
 
+template <class T>
+void Stack::Entry::set(const T& val) {
+  assertCanStore<T>();
+  *((T*)mData) = val;
```
AWoloszyn: This is illegal. (And the compiler should warn/error about this).
If the alignment of mData < alignment of val, this may segfault.

gapir/cc/stack.h
```diff
@@ -26,457 +26,172 @@
...
+    const T& get() {
+      assertCanStore<T>();
+#ifndef NDEBUG
+      // We only check the typeid for get() in debug builds. The rest of the
+      // time we just make sure the number of bytes is correct.
+      assert(sizeof(T) == mDataSize && mDataType == typeid(T));
+#else
+      assert(sizeof(T) == mDataSize);
+#endif
+      return *((T*)mData);
```
AWoloszyn: This may segfault, depending on the data type.

gapir/cc/interpreter.cpp
AWoloszyn: Might be good to at least name the parameter to prevent confusion?

gapir/cc/context.cpp
```diff
@@ -231,101 +231,77 @@ void Context::registerCallbacks(Interpreter* interpreter) {
...
-          GAPID_INFO("[%u]replayCreateRenderer(%u)", label, id);
-          auto existing = mGlesRenderers.find(id);
-          if (existing != mGlesRenderers.end()) {
-            delete existing->second;
-          }
-          // Share objects with the root GLES context.
-          // This will essentially make all objects shared between all contexts.
-          // It is ok since correct replay will only reference what it is
-          // supposed to.
+        GAPID_INFO("[%u]replayCreateRenderer(%u)", label, id);
```
AWoloszyn: So what happens now if the stack is not valid?

gapir/cc/interpreter.cpp
AWoloszyn: Can you move the comment to another line, so that it doesn't have this alignment issue

gapir/cc/interpreter.cpp
```diff
@@ -236,27 +337,82 @@ Interpreter::Result Interpreter::pushI(uint32_t opcode) {
...
+
+    case BaseType::Int16:
+      mStack.push<int16_t>(data);
+      break;
+
     case BaseType::Int32:
+      data |= (data & 0x80000) != 0
+                  ? 0xfff00000ULL
+                  : 0x0;  // Sign extend the 20 bit immediate value
+      mStack.push(*((int32_t*)&data));
```
AWoloszyn: This breaks strict aliasing rules.

gapir/cc/stack.h
```diff
@@ -26,457 +26,176 @@
...
+  uint32_t mMaxEntries;
+  Entry* mStack;
+
   const MemoryManager* mMemoryManager;
 };
 
+template <class T>
+void Stack::Entry::set(const T& val) {
+  assertCanStore<T>();
+  *((T*)&mData) = val;
```
AWoloszyn: I think this is still not spec-valid, typically for this case you would memcpy in and out, rather than try to assign.

gapir/cc/stack.h
```diff
@@ -26,457 +26,176 @@
...
 
-    bool getTo(void** p) const {
-      if (mType != BaseType::AbsolutePointer) {
-        return false;
-      }
-      *p = mValue.p;
-      return true;
+    Entry& operator=(const Entry& rhs) {
+      this->mDataSize = rhs.mDataSize;
+      memcpy((void*)&mData, (void*)&rhs.mData, maxEntrySize);
```
AWoloszyn: I am assuming the only thing that can ever end up in here is a POD type?

gapir/cc/interpreter.cpp
purvisa-at-google-com: Some people leave unused arguments nameless to make it clear they're not used. I picked this up from past software I worked on. TBH I prefer always naming. Changed.

gapis/api/templates/vulkan_gfx_api_extras.tmpl
purvisa-at-google-com: Can the if block be removed as well as the comment then? This isn't my code.

gapir/cc/stack.h
purvisa-at-google-com: Fixed. Good catch! I've been sick this week :(

gapir/cc/stack.h
purvisa-at-google-com: Changed to use std::aligned_storage.

gapir/cc/stack.h
purvisa-at-google-com: Changed to use std::aligned_storage.

gapir/cc/context.cpp
purvisa-at-google-com: Undefined. Almost certainly a crash. The checking was slowing down replay in a way that was impeding the functional usefulness of the code as a profiler. If you are interested I could leave the checking in for debug builds only. As long as we make sure to generate valid VM bytecode the stack should never become invalid. Processors and operating systems don't typically check their stacks for validity for the same reason. Performance trumps security in a profiling tool like the one we're making.

gapir/cc/interpreter.cpp
purvisa-at-google-com: Do we not break aliasing rules all over GAPID with all the void* trickery we're performing? The whole conceptual operation of this function implies type aliasing.

gapir/cc/stack.h
purvisa-at-google-com: Do you still object here since the change to std::aligned_storage?

gapir/cc/stack.h
purvisa-at-google-com: The code will work with anything that can be shallow copied.

gapir/cc/context.cpp
AWoloszyn: We should at least keep these in in a debug build. It is too easy to make a mistake when generating the instructions, that having the check is valuable (this is why it exists in the first place).

AWoloszyn: We should definitely leave these in at least in debug. It helps a lot when we run into the "Generating invalid VM bytecode" case, which can certainly happen if we are in an invalid case.

gapir/cc/interpreter.cpp
AWoloszyn: We do it in a slightly different way for the old system, but I guess we run into similar problems with the union.
The spec-compliant way to solve this is with a memcpy
```
int32_t v;
memcpy(&v, &data, sizeof(v));
mStack.push(v);
```

A memcpy is probably also the right thing to do in the set functions. As long as the compile optimizes memcpy (which many do) then it should remain valid.

gapis/api/templates/vulkan_gfx_api_extras.tmpl
AWoloszyn: Yes, it is illegal for a handle to be duplicated.

gapir/cc/stack.h
AWoloszyn: memcpy(&mData, &val, sizeof(val))

Would remain spec-valid, and prevent this issue. (Also fix your failing builds)

core/cc/assert.h
```diff
@@ -17,8 +17,10 @@
 #ifndef CORE_ASSERT_H
 #define CORE_ASSERT_H
 
+#include <cassert>
```
AWoloszyn: I don't think we use assert() anywhere here?

core/cc/assert.h
```diff
@@ -17,8 +17,10 @@
...
 #include "log.h"
 
+#ifndef NDEBUG
```
AWoloszyn: You can't change the semantics of GAPID_ASSERT unless you are 100% sure no-one relies on it NOT getting elided in release.

I would instead create a GAPID_DEBUG_ASSERT which **very clearly** denotes that it only exists in debug.
(this breaks arena::allocate_aligned for example).

gapir/cc/interpreter.cpp
```diff
@@ -302,46 +591,104 @@ Interpreter::Result Interpreter::load(uint32_t opcode) {
...
+      break;
+  }
+
+  return SUCCESS;
 }
 
 Interpreter::Result Interpreter::pop(uint32_t opcode) {
   mStack.discard(extract26bitData(opcode));
-  return mStack.isValid() ? SUCCESS : ERROR;
+  return SUCCESS;
```
AWoloszyn: If we are just going to ALWAYS return SUCCESS, why even bother with a return?
Same below

gapir/cc/stack.h
```diff
@@ -26,457 +26,188 @@
...
+        mStack(new Entry[size]),
+        mMemoryManager(memoryManager) {}
+
+  ~Stack() { delete[] mStack; }
+
   template <typename T>
-  T pop() {
-    if (!popCheck("pop")) {
-      return T();
+  T pop(typename std::enable_if<std::is_pointer<T>::value>::type* = 0) {
```
AWoloszyn: Please add method comments here (Looks like most of the code-removal was comments)

gapir/cc/stack_test.cpp
```diff
@@ -49,127 +49,62 @@ class StackTest : public ::testing::Test {
...
 
 TEST_F(StackTest, PushValue) {
   uint32_t x = 123456789;
-  mStack->pushValue(BaseType::Uint32, x);
+  mStack->push(x);
   EXPECT_EQ(123456789, mStack->pop<uint32_t>());
-  EXPECT_TRUE(mStack->isValid());
-}
-
-TEST_F(StackTest, PushValueErrorStackOverflow) {
```
AWoloszyn: Can we create death tests in debug for these. (Since I think we agreed to have the stack crash when it becomes invalid at least in debug).

PR #3784, opened 2020-02-12
Add "InlineResource" VLIW instruction to interpreter bytecode

purvisa-at-google-com:
This stores small resources "inline" in the instruction bytecode, which is conceptually kind of nasty but highly performant. Additionally, this new instruction supports some inlined "resource patch ups" which are basically push/store and load/store instructions that can also be inlined into the compound instruction, to allow common patch-ups of the inlined resource to all be performed in a single operation.

### DIFF
--- a/gapir/cc/interpreter.cpp
+++ b/gapir/cc/interpreter.cpp
@@ -182,8 +182,8 @@ void Interpreter::exec() {
   mExecResult.set_value(SUCCESS);
 }
 
-BaseType Interpreter::extractType(uint32_t opcode) const {
-  return BaseType((opcode & TYPE_MASK) >> TYPE_BIT_SHIFT);
+uint32_t Interpreter::extract6bitData(uint32_t opcode) const {
+  return (opcode & TYPE_MASK) >> TYPE_BIT_SHIFT;
 }
 
 uint32_t Interpreter::extract20bitData(uint32_t opcode) const {
@@ -192,6 +192,10 @@ uint32_t Interpreter::extract20bitData(uint32_t opcode) const {
 
 uint32_t Interpreter::extract26bitData(uint32_t opcode) const {
   return opcode & DATA_MASK26;
+}
+
+BaseType Interpreter::extractType(uint32_t opcode) const {
+  return BaseType(extract6bitData(opcode));
 }
 
 bool Interpreter::registerApi(uint8_t api) {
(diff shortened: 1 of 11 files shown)


AWoloszyn:
What kind of effect does this have on the average replay? (I assume this only really have much of an effect in Vulkan)
   1) How many inline resources are typical
   2) How large does this make the instruction stream? (This could be a problem on more memory-constrained devices, as well as places with important transfer speed requirements, as these will no longer end up in the cache)
   3) How many resources do we end up now duplicating into the instruction stream rather than having in the cache.

gapir/cc/interpreter.cpp
```diff
@@ -337,6 +341,49 @@ Interpreter::Result Interpreter::resource(uint32_t opcode) {
...
+
+  auto roundedDataSize = (dataSize / 4) + (dataSize % 4 != 0 ? 1 : 0);
+
+  for (unsigned int i = 0; i < numValuePatchUps; ++i) {
+    uint32_t destination =
+        mInstructions[mCurrentInstruction + 1 + roundedDataSize + i * 2];
+    uint32_t value =
+        mInstructions[mCurrentInstruction + 1 + roundedDataSize + i * 2 + 1];
+
+    *((void**)mMemoryManager->volatileToAbsolute(destination)) =
```
AWoloszyn: This makes some assumptions about the size of the patchups?
What happens if the patched value is not sizeof(void*) (This is quite normal in Vulkan for example), and DEFINITELY normal in GL.

gapis/replay/builder/builder.go
```diff
@@ -316,6 +358,79 @@ func (b *Builder) CommitCommand() {
...
+				newInstructions = append(newInstructions, b.instructions[index])
+			}
+		} else {
+			newInstructions = append(newInstructions, b.instructions[index])
+		}
+	}
+
+	b.instructions = b.instructions[:b.cmdStart]
+	for _, newInstruction := range newInstructions {
+		b.instructions = append(b.instructions, newInstruction)
```
AWoloszyn: b.instructions = append(b.instructions, newInstructions...)

gapir/cc/interpreter.cpp
purvisa-at-google-com: The "patch ups" offered by this instruction are only applicable to void* sized data. Maybe this should be commented to a greater degree. Unless I misunderstand the memory manager, the way we're using this in builder.go should not insert "patch ups" for things that aren't pointer sized? If there is a problem we should address it there. I did not want to reimplement the entire type system inside this instruction.

gapis/replay/builder/builder.go
purvisa-at-google-com: That's not the same? We want to drop all the "b.instructions" beyond b.cmdStart

purvisa-at-google-com: done

purvisa-at-google-com:
> What kind of effect does this have on the average replay? (I assume this only really have much of an effect in Vulkan)
> 
> 1. How many inline resources are typical
> 2. How large does this make the instruction stream? (This could be a problem on more memory-constrained devices, as well as places with important transfer speed requirements, as these will no longer end up in the cache)
> 3. How many resources do we end up now duplicating into the instruction stream rather than having in the cache.

https://docs.google.com/spreadsheets/d/1j9WfjuLtBClYrUjtp0nEdnxu9EsqmC0CoKZ1rIEAUz8/edit#gid=0

https://docs.google.com/spreadsheets/d/1LQG4Q1swcUB56hFMjt5EAxX8hIR0CgKtF5h4PkC0un0/edit#gid=1037122485

gapir/cc/interpreter.cpp
AWoloszyn: So, looking more carefully it looks like we are only patching pointers?
Is there a particular reason for that? Because in Vulkan at least, there are a TON of handles that also get patched, that this could benefit.

gapis/replay/protocol/opcode.go
```diff
@@ -20,28 +20,29 @@ import "fmt"
...
+	OpExtend         = Opcode(13)
+	OpAdd            = Opcode(14)
+	OpLabel          = Opcode(15)
+	OpSwitchThread   = Opcode(16)
+	OpJumpLabel      = Opcode(17)
+	OpJumpNZ         = Opcode(18)
+	OpJumpZ          = Opcode(19)
+	OpNotification   = Opcode(20)
+	OpWait           = Opcode(21)
+	OpInlineResource = Opcode(22)
```
AWoloszyn: You are missing the other changes needed in this file.

purvisa-at-google-com: done

gapis/replay/asm/instructions.go
```diff
@@ -305,6 +306,92 @@ func (a Resource) Encode(r value.PointerResolver, w binary.Writer) error {
...
+		}
+
+		if i*4+3 < len(a.Data) {
+			v3 = uint32(a.Data[i*4+3]) * 256 * 256 * 256
+		}
+
+		data[i] = v0 + v1 + v2 + v3
+	}
+
+	return opcode.InlineResource{ // ALAN: Inline
```
AWoloszyn: // What does this comment mean?

purvisa-at-google-com: // ALAN is a note to myself that I forgot to delete before creating the PR :)

gapir/cc/interpreter.cpp
purvisa-at-google-com: Will address in follow up commit that extends the functionality.

PR #3790, opened 2020-02-21
Serialize buffer contents on transfer from external queue.

bjoeris:
This PR is work toward supporting external memory (https://github.com/bjoeris/gapid/tree/extern-mem).

When a `VkBufferMemoryBarrier` (in a `vkCmdPipelineBarrier` or `vkCmdWaitEvents` call) has `srcQueueFamilyIndex = VK_QUEUE_FAMILY_EXTERNAL`, this indicates that the buffer contents may have been changed by an external process. This PR saves the contents when tracing, and restores the contents when replaying.

Capture:
  - Command buffer recording: maintain a map (`mExternalMemoryBarriers`) from command buffers to the `VkBufferMemoryBarrier`s recorded in the command buffers that acquire from `VK_QUEUE_FAMILY_EXTERNAL`.
  - Queue submission:
    1. Insert new command buffers to:
        1. acquire the buffers from the external queue family
        2. copy the contents into a staging buffer, and
        3. transfer the buffer back to the external queue family (so that the original acquire operation is valid).
      These new command buffers are inserted *before* the command buffers containing the corresponding `VkBufferMemoryBarrier`s.
    2. When the queue submission is complete, map the staging memory, and record the contents as an `ExternalMemoryData` extra.

Replay:
  - Command buffer recording: remove all queue family transfers to/from `VK_QUEUE_FAMILY_EXTERNAL`; for queue family transfers from `VK_QUEUE_FAMILY_EXTERNAL`, add `VK_ACCESS_TRANSFER_WRITE_BIT` to `srcAccessMask` and add `VK_PIPELINE_STAGE_TRANSFER_BIT` to the `srcStageMask`.
  - Queue submission:
    1. Load the contents into a staging buffer
    2. Insert new command buffers to:
        1. copy the contents from the staging buffer into the corresponding buffers.

### DIFF
--- a/gapii/cc/vulkan_external_memory.cpp
+++ b/gapii/cc/vulkan_external_memory.cpp
@@ -1,0 +1,484 @@
+/*
+ * Copyright (C) 2020 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "gapii/cc/vulkan_external_memory.h"
+#include "gapii/cc/vulkan_layer_extras.h"
+#include "gapii/cc/vulkan_spy.h"
+#include "gapis/api/vulkan/vulkan_pb/extras.pb.h"
+
+namespace gapii {
+
+void VulkanSpy::recordExternalBarriers(
+    VkCommandBuffer commandBuffer, uint32_t bufferMemoryBarrierCount,
+    const VkBufferMemoryBarrier* pBufferMemoryBarriers,
+    uint32_t imageMemoryBarrierCount,
+    const VkImageMemoryBarrier* pImageMemoryBarriers) {
+  static const uint32_t VK_QUEUE_FAMILY_EXTERNAL = ~0U - 1;
+
+  size_t externalBufferBarrierCount = 0;
+  for (uint32_t i = 0; i < bufferMemoryBarrierCount; ++i) {
+    if (pBufferMemoryBarriers[i].msrcQueueFamilyIndex ==
+        VK_QUEUE_FAMILY_EXTERNAL) {
+      ++externalBufferBarrierCount;
+    }
+  }
+
+  size_t externalImageBarrierCount = 0;
+  for (uint32_t i = 0; i < imageMemoryBarrierCount; ++i) {
+    if (pImageMemoryBarriers[i].msrcQueueFamilyIndex ==
+        VK_QUEUE_FAMILY_EXTERNAL) {
+      ++externalImageBarrierCount;
+    }
+  }
+
+  if (externalBufferBarrierCount == 0 && externalImageBarrierCount == 0) {
+    return;
+  }
+
+  std::vector<VkBufferMemoryBarrier>& bufBarriers =
+      mExternalBufferBarriers[commandBuffer];
+  bufBarriers.reserve(bufBarriers.size() + externalBufferBarrierCount);
+  for (uint32_t i = 0; i < bufferMemoryBarrierCount; ++i) {
+    if (pBufferMemoryBarriers[i].msrcQueueFamilyIndex ==
+        VK_QUEUE_FAMILY_EXTERNAL) {
+      bufBarriers.push_back(pBufferMemoryBarriers[i]);
+    }
+  }
+}
+
+ExternalMemoryStaging::ExternalMemoryStaging(
+    VulkanSpy* spy, CallObserver* observer, VkQueue queue, uint32_t submitCount,
+    const VkSubmitInfo* pSubmits, VkFence fence)
+    : spy(spy), observer(observer), queue(queue), origFence(fence) {
+  QueueObject& queueObj = *spy->mState.Queues[queue];
+  queueFamilyIndex = queueObj.mFamily;
+  device = queueObj.mDevice;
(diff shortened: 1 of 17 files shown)


gapis/api/vulkan/api/queue.api
```diff
@@ -85,6 +85,8 @@ cmd void vkGetDeviceQueue(
 @threadSafety("app")
 @indirect("VkQueue", "VkDevice")
 @submission
+@custom
+@override_observer
```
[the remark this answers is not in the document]
bjoeris: The issue is that I need to both modify the arguments to vkQueueSubmit (not possible in an extern, as far as I can tell), and have access to the CallObserver, to write the Extra (not possible with an spy_override). These two pieces also need to share information, which makes it messy to split this into a separate extern and spy_override (but probably not impossible)

gapii/cc/vulkan_external_memory.cpp
```diff
@@ -0,0 +1,497 @@
...
+      (uint32_t)commandBuffers.size() + 1,  // commandBufferCount
+  };
+  res = fn->vkAllocateCommandBuffers(device, &commandBufferAllocInfo,
+                                     commandBuffers.data());
+  if (res != VkResult::VK_SUCCESS) {
+    GAPID_ERROR(
+        "Error allocating command buffer for external memory observations");
+    return res;
+  }
+  stagingCommandBuffer = commandBuffers.back();
```
AWoloszyn: I think you may have to fix up the dispatch table here. 
see   `set_dispatch_from_parent((void*)command_buffer, (void*)device);`

That way other layers will play nice with while GAPID is enabled.

gapii/cc/vulkan_external_memory.cpp
```diff
@@ -0,0 +1,497 @@
...
+      memReqs.msize,  // allocationSize
+      0,              // memoryTypeIndex
+  };
+
+  VkPhysicalDevice physDevice = spy->mState.Devices[device]->mPhysicalDevice;
+  VkPhysicalDeviceMemoryProperties memProps =
+      spy->mState.PhysicalDevices[physDevice]->mMemoryProperties;
+  res = VkResult::VK_SUCCESS;
+  stagingMemory = 0;
+  for (size_t i = 0; i < 32; ++i) {
```
AWoloszyn: Hardcoded 32? Max memory types?

gapii/cc/vulkan_external_memory.cpp
```diff
@@ -0,0 +1,497 @@
...
+    return res;
+  }
+
+  std::vector<VkBufferMemoryBarrier> acquireBufferBarriers;
+  acquireBufferBarriers.reserve(cmdBuf.buffers.size());
+  std::vector<VkBufferMemoryBarrier> releaseBufferBarriers;
+  releaseBufferBarriers.reserve(cmdBuf.buffers.size());
+
+  std::vector<VkImageMemoryBarrier> acquireImageBarriers;
+  // acquireImageBarriers.reserve(cmdBuf.images.size());
```
AWoloszyn: Please remove commented out code.

gapii/cc/vulkan_extras.inc
```diff
@@ -161,3 +161,47 @@ void walkImageSubRng(
 void parseShaderModule(
     StageData* stage,
     gapil::Ref<DescriptorInfo>& descriptors);
+
+// TODO: does this need a mutex?
```
AWoloszyn: Shouldn't spy is globally synchronized

gapii/cc/vulkan_extras.inc
```diff
@@ -161,3 +161,47 @@ void walkImageSubRng(
...
+std::map<VkCommandBuffer, std::vector<VkBufferMemoryBarrier>> mExternalBufferBarriers;
```
AWoloszyn: Please std::unordered_map, unless we REALLY need command-buffer handles to be ordered.

gapii/cc/vulkan_external_memory.cpp
bjoeris: Done.

gapii/cc/vulkan_extras.inc
bjoeris: Comment removed.

gapii/cc/vulkan_extras.inc
bjoeris: Done.

gapii/cc/vulkan_extras.cpp
```diff
@@ -876,16 +875,11 @@ void VulkanSpy::SpyOverride_vkDestroyInstance(
...
-    VkBufferCreateInfo override_create_info = *pCreateInfo;
-    override_create_info.musage |=
-        VkBufferUsageFlagBits::VK_BUFFER_USAGE_TRANSFER_SRC_BIT;
-    return mImports.mVkDeviceFunctions[device].vkCreateBuffer(
-        device, &override_create_info, pAllocator, pBuffer);
-  } else {
-    return mImports.mVkDeviceFunctions[device].vkCreateBuffer(
-        device, pCreateInfo, pAllocator, pBuffer);
-  }
+  VkBufferCreateInfo override_create_info = *pCreateInfo;
```
bjoeris: @AWoloszyn I changed the behaviour here to always add TRANSFER_SRC usage, whether or not we are suspended. I need this bit to be set at least for all external buffers, but it seemed simpler to just enable it all the time. This mirrors the behaviour in `SpyOverride_vkCreateImage` (where there is no `is_suspended` check). If you think it is important to maintain the check, I can change this to only enable TRANSFER_SRC for external buffers.

AWoloszyn: I think that is fine. (Also more consistent). This change was originally to be less invasive, but I expect this to matter not at all for buffers, so feel free to make the change.

gapii/cc/vulkan_external_memory.cpp
AWoloszyn: That is a good place.

gapii/cc/vulkan_external_memory.cpp
AWoloszyn: I think that is a perfectly fine place

PR #3791, opened 2020-02-25
VK_KHR_timeline_semaphore support.

AWoloszyn:
This adds VK_KHR_timeline_semaphore support to GAPID.
This handles the simple cases (always being able to make forward
progress) as well as the more complex cases (blocking on
host or other commands).


gapis/api/vulkan/api/synchronization.api
```diff
@@ -214,36 +227,75 @@ cmd void vkDestroySemaphore(
...
+    VkSemaphore                                 semaphore,
+    u64                                         value) {
+  sem := Semaphores[semaphore]
+  if sem == null {
+    vkErrorInvalidSemaphore(semaphore)
+  } else if sem.TimelineSemaphoreInfo == null {
+    vkErrorInvalidSemaphore(semaphore)
+  } else {
+    sem.TimelineSemaphoreInfo.value = value
+    recordSemaphoreSignal(semaphore, value)
```
bjoeris: Should this call queueForwardProgress()?

gapis/api/vulkan/api/queued_command_tracking.api
```diff
@@ -40,146 +40,226 @@
...
+  } else if dont_queue && !are_signaled {
+    did_run.b = false
+  } else {
+    if dont_queue {
+      executeDeferredSubmit(sub)
+    }
+    for i in (0 .. len(sub.WaitSemaphores)) {
+      sem := sub.WaitSemaphores[as!u32(i)]
+      s := Semaphores[sem]
+      s.Signaled = false
```
bjoeris: I think this is for loop is redundant after calling `semaphoresAreSignaled`

gapii/cc/vulkan_extras.cpp
```diff
@@ -995,4 +1003,28 @@ void VulkanSpy::walkImageSubRng(
     }
   }
 }
+
+void VulkanSpy::recordWaitedSemaphores(CallObserver* observer, VkDevice device,
+                                       const VkSemaphoreWaitInfo* wait_info,
+                                       bool isKHR) {
```
bjoeris: I don't think we actually need to save the semaphore values here. I think replay already has all the information we need, even if only a subset of the semaphores are signaled (either because flags contains `VK_SEMAPHORE_WAIT_ANY_BIT`, or because `vkWaitSemaphores` returned `VK_TIMEOUT`).

At replay, we know the semaphore values after all the unblocked commands, so, if we wanted to be conservative, we could replay a `WAIT_ANY` as a `WAIT_ALL`, where the semaphores we know are unsignaled are filtered out.

gapis/api/vulkan/vulkan.go
```diff
@@ -358,6 +360,36 @@ func (API) ResolveSynchronization(ctx context.Context, d *sync.Data, c *path.Cap
...
+				if q.PendingSubmissions().Len() > 0 {
+					blocked = true
+					break
+				}
+			}
+		}
+		if !blocked {
+			for j := 0; j < len(blockedCommands); j++ {
+				d.UnblockingCommands[blockedCommands[j]] = id
+			}
```
bjoeris: should `blockedCommands` be cleared here?

gapis/api/vulkan/extensions/khr_timeline_semaphore.api
```diff
@@ -74,25 +98,33 @@ class VkSemaphoreWaitInfoKHR {
...
 }
 
 // Partial implementation
 @extension("VK_KHR_timeline_semaphore")
 @indirect("VkDevice")
+@custom
 cmd VkResult vkWaitSemaphoresKHR(
     VkDevice                                    device,
     const VkSemaphoreWaitInfoKHR*               pWaitInfo,
     u64                                         timeout) {
```
bjoeris: Missing device validity check in `vkSignalSemaphore(KHR)` and `vkWaitSemaphore(KHR)`

gapis/api/templates/vulkan_gfx_api_extras.tmpl
```diff
@@ -577,6 +577,111 @@ bool Vulkan::replayGetEventStatus(Stack* stack, bool pushReturn) {
...
+   auto expected = stack->pop<uint64_t>();
+   auto pValue = stack->pop<uint64_t*>();
+   auto semaphore = stack->pop<uint64_t>();
+   auto device = static_cast<size_val>(stack->pop<size_val>());
+   if (stack->isValid()) {
+        GAPID_DEBUG("vkGetSemaphoreCountValue(%zu, %" PRIu64 ")", device, semaphore);
+        uint32_t return_value;
+        do {
+          return_value = mVkDeviceFunctionStubs[device].vkGetSemaphoreCounterValue(device, semaphore, pValue);
+        } while (*pValue < expected && (return_value == VK_SUCCESS));
```
bjoeris: Why this spin, instead of just using `vkWaitSemaphores`?

gapis/api/sync/data.go
```diff
@@ -91,6 +91,11 @@ type Data struct {
 	SyncNodes        []SyncNode
 	CmdSyncNodes     map[api.CmdID]SyncNodeIdx
 	SubcommandLookup *api.SubCmdIdxTrie
+	// UnblockingCommands is a map of what command fully unblocks
+	// some other command. Or to put it another way, the command that answers
+	// the question: If we want to play up to command N, what command do we
+	// have to replay until, in order to prevent a deadlock.
+	UnblockingCommands map[api.CmdID]api.CmdID
```
bjoeris: It looks like `UnblockingCommands` handles the case where a submission blocks on a timeline semaphore that is host-signaled later *in the trace*. But what if the trace ended with the command still blocked? I think we need to make sure those commands never get submitted.

AWoloszyn: Chatted offline, but a re-cap here.
This is a correct analysis. However since we only stop at frame boundaries, this should not be a problem in practice. We have a lot of other problems to fix if we ever end up in a case where an EOF is not a safe place to block the GPU.

gapis/api/templates/vulkan_gfx_api_extras.tmpl
AWoloszyn: We could do that here. I wanted to be consistent with our VkGetFenceStatus. (Which spins instead of VkWaitForFences, because of hilarious driver bugs).
But if we think this won't actually break I am happy to switch.

gapis/api/vulkan/api/device.api
```diff
@@ -233,6 +236,19 @@ sub ref!DeviceObject createDeviceObject(const VkDeviceCreateInfo* data) {
             ShaderInt8: ext.shaderInt8,
           )
         }
+        case VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SHADER_ATOMIC_INT64_FEATURES_KHR: {
```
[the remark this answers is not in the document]
AWoloszyn: That is not intentional, I will figure out how/fix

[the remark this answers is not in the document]
AWoloszyn: Oh wait, yes this is intentional, I noticed it was missing when I was doing TIMELINE_SEMAPHORE support. I didnt want to have a merge-conflict later.

gapis/api/vulkan/api/synchronization.api
AWoloszyn: Great catch, yes it should.

gapis/api/vulkan/vulkan.go
AWoloszyn: Yep, fixed.

PR #3843, opened 2020-09-01
Fix for destroying the last instance on windows.

AWoloszyn:
If a game destroys that last instance, some static data
will fail to get cleaned up causing a crash. Prevent
libgapii from being unloaded on windows in this case.

### DIFF
--- a/gapii/cc/spy.cpp
+++ b/gapii/cc/spy.cpp
@@ -16,15 +16,14 @@
 
 #include "spy.h"
 
+#include <cstdlib>
+#include <memory>
+#include <sstream>
+#include <thread>
+#include <vector>
+
 #include "connection_header.h"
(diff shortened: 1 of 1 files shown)


gapii/cc/spy.cpp
```diff
@@ -98,11 +96,14 @@ const int32_t kSuspendIndefinitely = -1;
 thread_local gapii::CallObserver* gContext = nullptr;
 
 }  // anonymous namespace
-
+#include <windows.h>
```
aellem, line 99: Shouldn't this be inside a 
#if TARGET_OS == GAPID_OS_WINDOWS?

AWoloszyn: Thanks! I actually had moved it up, but forgot to remove this one. Fixed now.

